PR #14, opened 2026-08-19
CI: add least-privilege permissions and concurrency cancellation

### LINKED ISSUE #13: CI: add least-privilege `permissions` and `concurrency` cancellation to ci.yml
## Current behavior

`.github/workflows/ci.yml` is missing two standard pieces of GitHub Actions hygiene:

1. **No top-level `permissions:` block**, so every job runs with the repository's default token scope. The workflow only needs to read contents (the `links` job passes `GITHUB_TOKEN` to a third-party action, which is exactly where least-privilege matters).
2. **No `concurrency:` block**, so pushing twice to the same PR runs two full matrices (2 Python versions, apt installs included) to completion.

## Proposed change

Add to the top of `ci.yml`:

```yaml
permissions:
  contents: read

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

(`cancel-in-progress` conditioned on PRs so pushes to `main` are never cancelled mid-run.)

Optionally, in the same PR or a follow-up: add `.github/dependabot.yml` with a `github-actions` ecosystem entry so the pinned `@vN` action tags get update PRs.

## Definition of done

- CI runs green on the PR with the new blocks in place.
- Pushing a second commit to the PR cancels the first in-progress run (visible in the Actions tab).

## Validation

CI on the PR itself is the validation; no local commands beyond YAML syntax.

abhiramvsmg:
Closes #13

Adds a top-level `permissions: contents: read` block so jobs run with least-privilege token scope, and a `concurrency` block so pushing a new commit to a PR cancels the previous in-progress run instead of letting both matrices finish.

`cancel-in-progress` is scoped to `pull_request` events only, so pushes to `main` are never cancelled mid-run — each merge to main gets a complete, uncancelled CI validation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,3 +50,4 @@ jobs:
           fail: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,4 +50,3 @@ jobs:
           fail: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,12 @@
 name: ci
 
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 on:
   push:
     branches: [main]


kstonekuan (approved):
Thanks — this is a faithful implementation of #13 and the scoping is exactly right: `contents: read` is sufficient for both jobs (including the lychee action's `GITHUB_TOKEN`, which only makes rate-limited GETs), and the concurrency group partitions per-PR while grouping main pushes.

One non-blocking nuance on the PR description: with a shared concurrency group, GitHub keeps at most one *queued* run per group even with `cancel-in-progress: false`. So if several merges land on main in quick succession, an intermediate commit's run can be superseded while still queued — the guarantee is "the latest main commit always gets a full, uncancelled run," not "every merge commit does." That's the behavior we want; just noting it for the record.

Approving; will merge once the two matrix checks finish.

PR #18, opened 2026-08-19
cli: one-line messages for expected errors (curate/doctor/stale)

ayam04:
Fixes #3

Expected, user-recoverable errors in `curate`, `doctor`, and `stale` now print a one-line `<command>: <message>` on stderr and return exit code 2, instead of leaking a full Python traceback.

- Extracts `_command_curate` and `_command_doctor` so `main()` is pure dispatch (per the issue's companion cleanup)
- Applies the same `except (ValueError, FileNotFoundError)` pattern `_command_deploy` already used
- `_command_stale` now wraps the `stale_episodes` call the same way

Tests:
- `test_cli_doctor_missing_file_prints_one_line`: exit 2, `doctor:` on stderr, no traceback
- `test_cli_curate_bad_catalog_prints_one_line`: exit 2, `curate:` on stderr, no traceback

Validation: `uv run pytest tests/test_doctor.py tests/test_catalog_curation.py tests/test_runtime_cli.py -q` (37 passed).

PR #19, opened 2026-08-19
episode: expose ChannelData.publish_times

### LINKED ISSUE #4: Expose `ChannelData.publish_times` (stored but unreachable)
## What's missing

`ChannelData` stores per-message publish times but never exposes them. The constructor accepts `publish_times` and assigns `self._publish_times` (`src/hflow/episode.py:95` and `:103`), and `Episode.channel()` faithfully concatenates them from the reader (`src/hflow/episode.py:383`, produced in `src/hflow/reader.py`) — but there is no property, so the data is plumbed end to end and then unreachable.

## The pattern to copy

The `timestamps` property in the same class (`src/hflow/episode.py:110`):

```python
@property
def timestamps(self) -> np.ndarray:
    """Log times in nanoseconds, ascending, shape (n,). Does not decode."""
    return self._log_times
```

## Proposed change

Add a `publish_times` property directly below `timestamps`, returning `self._publish_times`. The docstring should note that these are publish times (not log times) and, unlike `timestamps`, are **not** guaranteed ascending.

## Definition of done

- `ep.channel("/topic").publish_times` returns an `np.ndarray` with shape equal to `len(channel)`.
- One test asserting the shape matches (the synthetic-episode fixtures in `tests/` give a ready target).

## Validation

```bash
uv run pytest -q
uv run ty check
```

ayam04:
Fixes #4

`ChannelData` already accepted and stored per-message publish times, and `Episode.channel()` concatenated them from the reader, but no property exposed them. Added a `publish_times` property mirroring `timestamps`, documented as non-ascending.

Test: `test_episode_exposes_publish_times` checks the ndarray shape matches `len(channel)` and values are non-negative.

Validation: `uv run pytest tests/test_multichannel.py -q` passes.

### DIFF
--- a/src/hflow/episode.py
+++ b/src/hflow/episode.py
@@ -113,6 +113,15 @@ class ChannelData:
         return self._log_times
 
     @property
+    def publish_times(self) -> np.ndarray:
+        """Publish times in nanoseconds, shape (n,). Does not decode.
+
+        Unlike :attr:`timestamps`, publish times are not guaranteed ascending
+        (they reflect when the application handed each message to the channel).
+        """
+        return self._publish_times
+
+    @property
     def raw(self) -> list[bytes]:
         """Raw encoded message payloads. Does not decode."""
         return self._raw
--- a/tests/test_multichannel.py
+++ b/tests/test_multichannel.py
@@ -129,8 +129,8 @@ def test_episode_exposes_publish_times(dual_channel_source: Path) -> None:
         publish_times = json_channel.publish_times
         assert isinstance(publish_times, np.ndarray)
         assert publish_times.shape == (len(json_channel),)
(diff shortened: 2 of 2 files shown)


kstonekuan:
👋 Thanks @ayam04 — the implementation quality here is genuinely good. However, we accept **one "good first issue" PR per contributor at a time**, so that these starter issues stay available for other new contributors and each PR gets a real review. Of your seven PRs we kept #18 open (the strongest and most substantive one) and are closing the rest, including this one. The underlying issue is open again for anyone to pick up. Once #18 is merged you are very welcome to bring one of these back — and for larger contributions beyond the good-first-issue pool, feel free to open as many focused PRs as you like after discussing in an issue first.

PR #27, opened 2026-08-19
storage: format missing-file errors with errno context

### LINKED ISSUE #26: CLI: missing-file errors print a bare path, losing 'no such file or directory' context
### Problem

`hflow doctor`, `curate`, and `stale` now print one-line messages for expected errors (thanks to #18), but the missing-file case produces a message that reads as a bare path with no context:

```
$ uv run hflow doctor /tmp/nope.mcap   # file does not exist
doctor: /tmp/nope.mcap            <- stderr, exit code 2
```

The message loses the "no such file or directory" context entirely. Root cause: `fetch_uri` (`src/hflow/storage.py:633`) raises `FileNotFoundError(local_file)` — constructed with the path alone, so `str(error)` is just the path:

```python
if not local_file.is_file():
    raise FileNotFoundError(local_file)
```

`FileNotFoundError(path)` sets `filename` but formats to the bare path in `str()`, unlike the usual `[Errno 2] No such file or directory: '/path'`.

### Proposed change

Raise with named args so the standard OSError formatting ("No such file or directory") comes through, and/or include the path explicitly in the CLI handler messages. E.g.:

```python
if not local_file.is_file():
    raise FileNotFoundError(2, f"No such file or directory: {local_file}", local_file)
```

so `hflow doctor /tmp/nope.mcap` prints `doctor: [Errno 2] No such file or directory: /tmp/nope.mcap`.

### Why it matters

The whole point of #18 was turning tracebacks into user feedback. A message that is just the path is still a crash-like output for a user who typo'd a path — the fix makes the CLI consistently actionable across `doctor` / `curate` / `stale`.

### Definition of done

- `hflow doctor /tmp/nope.mcap` (nonexistent) prints a one-line stderr message containing "No such file or directory" and exits 2.
- Same for a nonexistent catalog root in `hflow curate`.
- Tests in `tests/test_doctor.py` / `tests/test_catalog_curation.py` assert the message text and exit code.

ayam04:
Fixes #26

`fetch_uri` raised `FileNotFoundError(local_file)` with the path as the single argument, so `str()` of the exception was just the bare path, and the CLI printed `doctor: /tmp/nope.mcap` with no 'no such file' hint after #18 turned tracebacks into one-line messages.

Per maintainer's adjustment, the path is passed only as the third (filename) argument:

```python
raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), str(local_file))
```

so it formats as `[Errno 2] No such file or directory: '/tmp/nope.mcap'` without the path printed twice.

Tests:
- `test_fetch_local_missing_has_oserror_filename` (storage): errno == 2, message contains 'No such file or directory', path present
- `test_cli_doctor_missing_file_prints_one_line`: asserts the enhanced message on stderr

Validation: `uv run pytest tests/test_storage.py tests/test_doctor.py -q` (44 passed, 1 skipped), `uv run ruff check` + `uv run ty check` clean.

### DIFF
--- a/src/hflow/storage.py
+++ b/src/hflow/storage.py
@@ -32,6 +32,7 @@ rather than a consistency problem:
   episode uploaded by another.
 """
 
+import errno
 import fcntl
 import hashlib
 import os
@@ -630,5 +631,5 @@ def fetch_uri(uri: "str | Path") -> Path:
         return BucketStorageRoot(parent_url).fetch(name)
     local_file = Path(uri)
     if not local_file.is_file():
-        raise FileNotFoundError(local_file)
+        raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), str(local_file))
     return local_file
--- a/tests/test_doctor.py
+++ b/tests/test_doctor.py
@@ -90,6 +90,7 @@ def test_cli_doctor_missing_file_prints_one_line(capsys: pytest.CaptureFixture[s
     assert cli_main(["doctor", missing]) == 2
     captured = capsys.readouterr()
     assert "doctor:" in captured.err
+    assert "No such file or directory" in captured.err
     assert "Traceback" not in captured.err
 
 
--- a/tests/test_storage.py
+++ b/tests/test_storage.py
@@ -184,6 +184,14 @@ class TestBucketStorageRoot:
         with pytest.raises(FileNotFoundError):
             root.fetch("landing/missing.mcap")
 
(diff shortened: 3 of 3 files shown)


kstonekuan (approved):
Thanks @ayam04! This matches the spec from #26 exactly, including the errno form, and extends the existing regression tests instead of duplicating them. Validated locally: full quality gate and test suite pass (293 tests), and CI is green on both Python versions. Nice touch using a closing keyword this time. Merging.

PR #38, opened 2026-08-20
test: cover joint sine parameter contract

### LINKED ISSUE #34: Test the determinism contract of `hflow.testing.joint_sine_parameters`
## Current behavior

`joint_sine_parameters` (`src/hflow/testing.py:220`) is a public function in `hflow.testing` with a determinism contract in its docstring ("derived only from the spec"), and the synthetic-episode generator depends on it (`src/hflow/testing.py:545`). It has zero direct test coverage: nothing pins the contract that the same `SyntheticEpisodeSpec` always produces the same parameters, which is what makes synthetic test episodes reproducible across runs and machines.

## Proposed change

Add a small behavioral test in `tests/` covering the function's contract:

- **Determinism**: two calls with specs sharing the same `seed` and `joint_count` return identical parameters.
- **Frequency span**: the first joint gets `JOINT_SINE_MIN_FREQUENCY_HZ`, the last joint gets `JOINT_SINE_MAX_FREQUENCY_HZ`, and frequencies are evenly spaced between them.
- **Edge case**: `joint_count=1` returns a single pair at the minimum frequency (the `max(spec.joint_count - 1, 1)` guard).
- **Phase range**: every phase is in `[0, 2*pi)`.

## Definition of done

- The test asserts the observable contract above, not implementation details (per CONTRIBUTING.md: no tests that only assert how a helper is built internally).
- Place it alongside the closest existing behavioral tests for `hflow.testing`.

## Validation

```bash
uv run pytest -q
uv run ty check
```

bryanwhl:
## Summary

Add focused behavioral tests for `hflow.testing.joint_sine_parameters` that verify:

- specs sharing the same seed and joint count produce identical parameters;
- frequencies include the configured minimum and maximum and are evenly spaced;
- a single joint uses the minimum frequency;
- every phase lies in `[0, 2*pi)`.

## Why

Synthetic episode reproducibility depends on this public helper, but its determinism and boundary behavior previously had no direct coverage. This change pins the observable contract without changing the implementation.

Closes #34

## Validation

- `uv run pytest tests/test_testing.py -q -k joint_sine_parameters` — 2 passed, 6 deselected on Python 3.11
- `uv run --python 3.14 --locked --all-extras pytest tests/test_testing.py -q -k joint_sine_parameters` — 2 passed, 6 deselected
- `uv run pytest tests/test_testing.py -q` — 8 passed
- `uv run ty check` — passed
- `uv run pytest -q` — 287 passed, 6 skipped, 5 environment-specific failures:
  - one existing test assumes `/usr/bin/ffmpeg`, which is absent on macOS;
  - four existing contact-sheet/media tests require the `drawtext` filter, which is absent from the installed Homebrew FFmpeg 9.0.1.
- Full suite with those five environment-specific cases deselected — 287 passed, 6 skipped, 5 deselected

## Checklist

- [x] Documentation is unchanged because this PR adds test coverage without changing behavior, flags, formats, or requirements.

### DIFF
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -19,8 +19,11 @@ from mcap_ros2.decoder import DecoderFactory
 
 from hflow.ffmpeg import ffmpeg_path
 from hflow.testing import (
+    JOINT_SINE_MAX_FREQUENCY_HZ,
+    JOINT_SINE_MIN_FREQUENCY_HZ,
     SyntheticEpisodeSpec,
     VideoEpisodeSpec,
+    joint_sine_parameters,
     synthesize_episode,
     write_video_episode,
 )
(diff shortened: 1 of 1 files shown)


bryanwhl:
@codex review

kstonekuan (approved):
Thanks @bryanwhl! Validated locally: full quality gate and suite pass (295 tests). These go beyond the issue's definition of done in the best way: varying `duration_s` and `cameras` between the two determinism calls proves the parameters derive only from the seed and joint count, which is the docstring's actual contract. Clean behavioral assertions, extended the existing test module rather than adding a parallel one. Merging.

On the five macOS failures you keep hitting: since you have the reproducing environment, a bug report with the exact assertion output for the `/usr/bin/ffmpeg` case and your `ffmpeg -version`/`-filters` output for the drawtext cases would be a valuable next contribution. We cannot reproduce them on Linux.

PR #40, opened 2026-08-20
feat: add action_rate check and Cohort statistics docs

Sagar-024:
## Summary
Adds the `action_rate` built-in check and a "Cohort statistics" section in docs/CATALOG.md, per the design agreed in Discussion #24. That discussion is the design doc.

The check records the raw per-episode message rate of the given action topics as `action_rate_hz`. Evidence only, no verdict: n timestamps define n - 1 intervals, so the rate is intervals divided by the span across the selected topics.

The docs section records the DuckDB window-function pattern for cohort math (z-score via `(value - AVG(value) OVER ()) / STDDEV(value) OVER ()`, plus `PERCENT_RANK()`), with the corpus-relative caveat: compute the window over the same filtered cohort you intend to cut.

## Validation
- uv run ruff check --fix && uv run ruff format
- uv run ty check (only pre-existing Windows platform diagnostics in runtime/_bundle.py, storage.py, and test_runtime_bundle.py; none in the new code)
- uv run pytest -q tests/test_checks.py (11 passed)

## Checklist
- [x] Outcome-focused test added: synthetic episode rate matches the spec's 100 Hz joint stream
- [x] Docs updated for the new behavior
- [x] ruff, ty, and pytest run
- [x] git status contains no recordings, generated artifacts, credentials, or runtime bundles

### DIFF
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -159,6 +159,22 @@ FROM measurements WHERE key = 'content_digest'
 GROUP BY digest HAVING count(*) > 1
 ```
 
+### Cohort statistics
+
+Per-episode checks record evidence; cohort math (z-score, percentile) is a query over the catalog.
+
+```sql
+SELECT episode_id,
+       action_rate_hz,
+       (action_rate_hz - AVG(action_rate_hz) OVER ())
+         / STDDEV(action_rate_hz) OVER () AS action_rate_hz_z,
+       PERCENT_RANK() OVER (ORDER BY action_rate_hz) AS action_rate_hz_pct
+FROM episodes
+WHERE task = 'fold_napkin'
+```
+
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
Thanks @Sagar-024! This delivers the design from #24 exactly as agreed: raw `action_rate_hz` evidence in the check, cohort math as a documented DuckDB window pattern, and the corpus-relative caveat stated. The note that STDDEV over a single-row cohort returns NULL 'which is the correct answer' is a genuinely useful addition we did not ask for. Validated locally: quality gate, full suite (296 tests), and the docs link check all pass.

One semantic worth recording for posterity, not blocking: with multiple topics the check records the summed rate over the union span rather than a per-topic mean. That is the right aggregate for 'how fast is this rig acting overall', and the docstring plus PR description state the formula, so future readers will not be surprised. Merging.

PR #41, opened 2026-08-20
feat: add overexposed_frame_pct to frame stats

### LINKED ISSUE #33: Add `overexposed_frame_pct` to frame stats from the luma values already computed
## Current behavior

The frame instrument already computes per-frame average luma for every frame (`src/hflow/ffmpeg/_instrument.py:151`), but only three aggregates are reported (`_instrument.py:181-183`): `luma_avg_mean`, `luma_avg_min`, and `luma_avg_max`.

Blackout has a dedicated metric (`black_frame_pct`, `_instrument.py:41` and `:178`), but the symmetric failure (blown highlights from bad exposure, IR flood, or sun in frame) has no counterpart. `luma_avg_max` alone cannot distinguish one bright frame from a thousand.

## Proposed change

- Add `overexposed_frame_pct` to the `FrameStats` dataclass (`_instrument.py:34-46`), computed as the percentage of frames whose average luma is at or above a threshold.
- Expose the threshold as a parameter on `frame_stats` (for example, `bright_luma_threshold: float = 235.0`), matching the module's rule that thresholds are user-owned.
- Compute it from the existing `luma_avg_values` list: no new ffmpeg filter and no second decode pass.
- Surface the new value as a measurement in `camera_frame_stats` (`src/hflow/checks.py:189`), next to `black_pct`.

## Definition of done

- `frame_stats` reports `overexposed_frame_pct` in the 0.0 to 100.0 range with the same denominator as `black_frame_pct`.
- `camera_frame_stats` records the measurement per camera topic.
- Extend `tests/test_ffmpeg.py` with a bright-video case (the existing fixtures show how the black-segment case is synthesized).

## Validation

```bash
uv run pytest -q
uv run ty check
```

Sagar-024:
## Summary
Adds `overexposed_frame_pct` to `FrameStats` and records it per camera topic in `camera_frame_stats`. Closes #33.

Blackout already has `black_frame_pct`, but the symmetric failure (blown highlights from bad exposure, IR flood, or sun in frame) had no counterpart, and `luma_avg_max` alone cannot distinguish one bright frame from a thousand. The new measurement is the percentage of frames whose average luma is at or above `bright_luma_threshold` (default 235.0; thresholds stay user-owned), computed from the existing `luma_avg_values` list with the same frame-count denominator as `black_frame_pct`. No new ffmpeg filter, no second decode pass.

## Changes
- `src/hflow/ffmpeg/_instrument.py`: `FrameStats` gains `overexposed_frame_count` and `overexposed_frame_pct`; `frame_stats()` and `_stats_from_instrument_output()` gain `bright_luma_threshold`.
- `src/hflow/checks.py`: `camera_frame_stats` threads the threshold and records `{topic}/overexposed_frame_pct` next to `black_frame_pct`.
- `tests/test_ffmpeg.py`: synthetic instrument-output test with exact math at the default threshold, plus a white-tail real-video test exercising the threshold parameter.
- `tests/test_checks.py`: asserts the measurement is surfaced as 0.0 on a non-bright camera fixture.

## Validation
- uv run ruff check --fix && uv run ruff format
- uv run ty check (only pre-existing Windows platform diagnostics; none in the new code)
- uv run pytest -q tests/test_ffmpeg.py tests/test_checks.py (all new tests pass; 2 pre-existing Windows failures unrelated to this change: test_env_override_wins, test_local_install_extracts_both_binaries_and_executes)

## Checklist
- [x] Outcome-focused tests for the new measurement (synthetic exact math, real video, checks surfacing)
- [x] Docstrings updated where thresholds and measurements are documented
- [x] ruff, ty, and pytest run
- [x] git status contains no unrelated files

### DIFF
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -195,6 +195,7 @@ def camera_frame_stats(
     black_pixel_threshold: int = 32,
     freeze_noise_db: float = -60.0,
     freeze_min_duration_s: float = 2.0,
+    bright_luma_threshold: float = 235.0,
 ) -> CheckResult:
     """Camera blackout, freeze, exposure, and frame-count evidence per camera.
 
@@ -207,8 +208,8 @@ def camera_frame_stats(
     ``freeze:<topic>`` intervals in log time. Requires a canonical episode
     (``Episode.video`` remuxes in-band H.264 only).
 
-    Evidence only, as always: blackout/exposure thresholds and any verdict
-    stay user-owned.
+    Evidence only, as always: blackout/exposure thresholds (including
+    ``bright_luma_threshold``) and any verdict stay user-owned.
     """
     selected_cameras = list(cameras) if cameras is not None else episode.cameras
     measurements: dict[str, MeasurementValue] = {}
@@ -234,9 +235,11 @@ def camera_frame_stats(
             black_pixel_threshold=black_pixel_threshold,
             freeze_noise_db=freeze_noise_db,
             freeze_min_duration_s=freeze_min_duration_s,
+            bright_luma_threshold=bright_luma_threshold,
         )
         measurements[f"{topic}/decoded_frame_count"] = stats.frame_count
         measurements[f"{topic}/black_frame_pct"] = stats.black_frame_pct
+        measurements[f"{topic}/overexposed_frame_pct"] = stats.overexposed_frame_pct
         measurements[f"{topic}/freeze_total_s"] = stats.freeze_total_s
         measurements[f"{topic}/luma_avg_mean"] = stats.luma_avg_mean
         measurements[f"{topic}/luma_avg_min"] = stats.luma_avg_min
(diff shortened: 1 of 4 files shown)


kstonekuan (approved):
Thanks @Sagar-024! Validated locally: quality gate and full suite pass (298 tests). This matches #33 exactly, and the test work is excellent: the pure-text aggregation case, the white-tail fixture with the freezedetect guard (nice catch that 2s of identical white frames would otherwise register as a freeze), and the limited-range-white threshold note all show real understanding of the instrument. Merging.

One gentle ask for next time: you are well past needing the training wheels, so we would love to keep the `good first issue` pool for brand-new contributors when possible. A more valuable direction for someone at your level: run HFlow against a real corpus like [Egocentric-10K](https://huggingface.co/datasets/builddotai/Egocentric-10K) (or the 100K variant) and tell us what breaks, what is slow, or what is awkward. Real-data findings like your #24 measurement are the highest-leverage contributions this project can get right now.

PR #45, opened 2026-08-20
runtime: substitute venv_python/data_root/pipeline_filename as repr() literals

### LINKED ISSUE #44: hflow deploy generates unparseable (and unsafely-templated) DAG source when venv_python/data_root/pipeline_filename contain quotes or backslashes
### Version or commit

`0d3d1313a3f10d5f7d8e9d938d5224b7337f4b7b` (main, 2026-08-20)

### Environment

Ubuntu 24.04 (WSL), Python 3.12.3, x86_64, uv 0.12.5. Not Docker-specific — reproduces with plain `render_dag_sources`, no Compose/Airflow needed.

### Minimal reproduction

```python
from hflow.runtime._bundle import render_dag_sources

sources = render_dag_sources(
    master_dag_id="demo_ingest",
    pipeline_filename="pipeline.py",
    app_variable="app",
    data_root="/data/root",
    venv_python=r"C:\Users\ci\venvs\user\Scripts\python.exe",
)
for dag_id, src in sources.items():
    compile(src, dag_id, "exec")  # raises on every sub-DAG
```

This is exactly the value `hflow deploy --venv-python <path>` (`src/hflow/cli.py:178`, `DeployConfig.venv_python_path`) accepts and threads straight into the templates — `DeployConfig.__post_init__` (`src/hflow/runtime/_deploy.py:108`) only validates `data_root_uri`, never `venv_python_path`. It's also the exact scenario `DEPLOY.md`'s own generated troubleshooting text points an operator toward: "re-render with `--venv-python <interpreter path>`" when the default path can't be used on a platform (`src/hflow/runtime/_deploy.py:265-267`).

### Expected behavior

`hflow deploy` with any valid absolute interpreter path (including a Windows-style path with backslashes) produces four sub-DAG files that Airflow's DAG processor can import.

### Actual behavior

```
demo_sync: SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 359-360: truncated \UXXXXXXXX escape
demo_meta: SyntaxError: ...
demo_labels: SyntaxError: ...
demo_media: SyntaxError: ...
```

All four sub-DAGs fail to import. Nothing surfaces this to whoever ran `hflow deploy` — Airflow's DAG processor just logs an import error, and the pipeline silently never runs.

### Root cause

`render_sub_dag_source` (`src/hflow/runtime/_bundle.py:483`) builds each sub-DAG file with `string.Template.substitute()` against `SUB_DAG_ERROR_GATE_TEMPLATE` / `SUB_DAG_QUARANTINE_GATE_TEMPLATE`. The templates embed `$venv_python`, `$data_root`, and `$pipeline_filename` directly inside Python double-quoted string literals with no escaping, e.g. (`src/hflow/runtime/_templates.py`):

```python
@task.external_python(python="$venv_python", ...)              # x4, lines 438/463/532/572
data_root = parse_storage_root("$data_root")                    # line 450
pipeline_path = "/opt/user/" + "$pipeline_filename"              # line 476/482
if str(app.data_root) != "$data_root":                          # line 492
reader = open_reader(Path("$data_root") / str(uri))              # MEDIA_PLAN_FILTER_TEMPLATE, line 642
```

`_validate_dag_identifiers()` (`src/hflow/runtime/_bundle.py:425`) already exists specifically to stop this class of bug — its own docstring says "Refuse anything that could break (or smuggle code into) a DAG file" — but it's only ever called on `dag_id` and `app_variable`. `venv_python`, `data_root`, and `pipeline_filename` go through the exact same unescaped substitution and were never brought under the same guard. And `validate_data_root_uri` (`src/hflow/runtime/_deploy.py:60`), despite its docstring claiming the data root is "validated loudly," only checks the URI scheme and that local paths are absolute — it does not reject characters (`"`, `\`) that are unsafe inside a Python string literal, so a data root like `/data/root"; import os  #` passes validation and reaches the template unescaped too.

Because these are plain double-quoted literals rather than triple-quoted or raw strings, any backslash sequence that isn't a valid Python escape (any ordinary Windows path) breaks the file, and any literal `"` in the value closes the string early — which can produce anything from a clean `SyntaxError` (harmless but silently breaks the deploy) to syntactically valid Python that runs as part of the generated DAG module when Airflow's scheduler imports it.

### Proposed fix

Don't pre-quote these placeholders in the templates; substitute a proper Python string literal instead. Change the three call sites to pass `repr(value)` and drop the surrounding `"..."` in the template around `$venv_python` / `$data_root` / `$pipeline_filename` (e.g. `python=$venv_python` in the template + `venv_python=repr(venv_python)` at the call site). `repr()` on a `str` always produces a syntactically valid, correctly escaped Python literal regardless of content, so this closes the bug for arbitrary input rather than only for the Windows-path case — no allowlist/validation regex needed for these three fields (unlike `dag_id`/`app_variable`, which are also used as Python identifiers/tags elsewhere and still need `_validate_dag_identifiers`).

### Additional context

I've verified this locally (reproduction above) and have a fix ready that changes `_templates.py`/`_bundle.py` to substitute `repr()`-encoded literals for `venv_python`, `data_root`, and `pipeline_filename` at every site listed above, plus a regression test compiling the rendered sources for values containing backslashes and quotes. Happy to open a PR against this issue.

chintondutta:
Fixes #44.

## What was broken

`render_sub_dag_source` (`src/hflow/runtime/_bundle.py`) builds each generated sub-DAG file via `string.Template.substitute()` against templates that embedded `$venv_python`, `$data_root`, and `$pipeline_filename` directly inside pre-quoted Python string literals, with no escaping:

```python
@task.external_python(python="$venv_python", ...)
data_root = parse_storage_root("$data_root")
pipeline_path = "/opt/user/" + "$pipeline_filename"
if str(app.data_root) != "$data_root":
```

`_validate_dag_identifiers()` already exists specifically to guard against this ("Refuse anything that could break (or smuggle code into) a DAG file") but was only ever applied to `dag_id`/`app_variable`. These three fields go through the identical unescaped substitution and were never brought under a matching guard. `validate_data_root_uri` only checks URI scheme and path absoluteness, not template-safety, despite its docstring claiming the root is "validated loudly."

Concretely: `hflow deploy --venv-python 'C:\Users\ci\venvs\user\Scripts\python.exe'` — an ordinary Windows interpreter path, exactly the value `DEPLOY.md`'s own generated troubleshooting text tells an operator to pass — produced a `SyntaxError` in all four sub-DAGs (`\U` reads as a truncated unicode escape), with nothing surfacing the failure. A `data_root` containing a literal `"` closed the string early instead, with effects ranging from a different `SyntaxError` to syntactically valid but altered generated code.

## Fix

Substitute `repr()` of each value instead of pre-quoting the template slot, and drop the manual `"..."` wrapping from the template around these three placeholders. `repr()` on a `str` always produces a valid, self-escaping Python literal regardless of content, so this closes the bug for arbitrary input, not only the Windows-path case that surfaced it.

One site needed restructuring rather than a drop-in swap: the `data_root` equality-check error message embedded `$data_root` as raw text *inside* another string literal's characters (`"pipeline data_root must be $data_root inside the runtime; "`). Splicing a `repr()` result into the middle of another string's text is unsafe whenever `repr()` happens to choose double-quote delimiters (it does when the value contains a `'` but no `"`), since that would prematurely close the outer literal. Fixed by binding the literal to a local variable first (`expected_data_root = $data_root`) and using it in an expression position (`+` concatenation) instead of as embedded text.

## Testing

- New regression test `test_dag_sources_survive_paths_that_are_not_valid_python_literals` (`tests/test_runtime_deploy.py`) exercises both the reported Windows-path case and a `data_root` containing an injection-shaped payload (`os.system(...)`), asserting the generated sources `compile()` cleanly and the value round-trips as a single literal.
- Updated the existing sources-compile assertions in `tests/test_runtime_bundle.py` and `tests/test_runtime_deploy.py` to match the new (single-quoted `repr()`) rendering; behavior covered by those tests (which interpreter/data root a task runs against) is unchanged, only the exact generated source text.

```bash
uv run pytest tests/test_runtime_bundle.py tests/test_runtime_deploy.py -q   # 61 passed
uv run pytest -q                                                            # unrelated to this change: 3 failures + 6 errors in tests/test_ffmpeg.py because this sandbox has no ffmpeg/ffprobe on PATH; everything else passes
uv run ruff check --fix
uv run ruff format
uv run ty check
```

No docs/behavior/flag changes; this only affects the exact bytes of generated (internal, re-rendered-on-every-`deploy`) DAG source files, not any stored/versioned format.

### DIFF
--- a/src/hflow/runtime/_bundle.py
+++ b/src/hflow/runtime/_bundle.py
@@ -507,13 +507,25 @@ def render_sub_dag_source(
         case Stage.SYNC | Stage.LABELS | Stage.MEDIA:
             template, gate_name = SUB_DAG_ERROR_GATE_TEMPLATE, "error_budget_gate"
     pipeline_stem = _pipeline_stem(master_dag_id)
+    # $data_root, $venv_python, and $pipeline_filename are substituted as
+    # repr() literals, never as bare text inside a pre-quoted template
+    # slot: a raw platform path (Windows venv interpreters use backslashes)
+    # or a data root containing a quote character would otherwise either
+    # break the generated file's Python syntax or -- worse -- close the
+    # surrounding string literal early and splice arbitrary text into the
+    # generated DAG's source (#44). repr() on a str always yields a
+    # self-escaping literal, so this holds for any input, not just the
+    # platform-path case that surfaced it.
+    data_root_literal = repr(data_root)
+    venv_python_literal = repr(venv_python)
+    pipeline_filename_literal = repr(pipeline_filename)
     # Substituted separately because Template.substitute never re-expands
     # variables inside substituted VALUES -- the filter's own $data_root must
     # be resolved before injection. Local roots only: probing a BUCKET
     # episode's channel list would download the whole file at plan time,
     # costing more than the skipped camera-less cycle saves.
     stage_plan_filter = (
-        MEDIA_PLAN_FILTER_TEMPLATE.substitute(data_root=data_root)
+        MEDIA_PLAN_FILTER_TEMPLATE.substitute(data_root=data_root_literal)
         if stage is Stage.MEDIA and not is_bucket_url(data_root)
         else ""
     )
@@ -526,10 +538,10 @@ def render_sub_dag_source(
         pipeline_tag=pipeline_stem,
         sub_display_name=f"{pipeline_stem} · {stage.value}",
         gate_name=gate_name,
-        pipeline_filename=pipeline_filename,
+        pipeline_filename=pipeline_filename_literal,
         app_variable=app_variable,
-        data_root=data_root,
-        venv_python=venv_python,
+        data_root=data_root_literal,
+        venv_python=venv_python_literal,
         stage_plan_filter=stage_plan_filter,
     )
 
--- a/src/hflow/runtime/_templates.py
+++ b/src/hflow/runtime/_templates.py
@@ -435,7 +435,7 @@ latency-first run); `batch_count`: optional override.
     params={"uris": [], "mode": "batch", "batch_count": None},
 )
 def ingest_stage():
-    @task.external_python(python="$venv_python", expect_airflow=False, skip_on_exit_code=99)
+    @task.external_python(python=$venv_python, expect_airflow=False, skip_on_exit_code=99)
     def plan(uris, mode, batch_count):
         """Bin-pack uris into staggered batches; online is one immediate batch."""
         from hflow.batching import plan_batches
@@ -447,7 +447,7 @@ $stage_plan_filter        if mode == "online":
             # The online lane (Figure 4) is latency-first: one run per episode
             # as it lands -- no batching, no stagger, batch_count ignored.
             return [{"items": [str(uri) for uri in uris], "start_delay_s": 0.0}]
-        data_root = parse_storage_root("$data_root")
+        data_root = parse_storage_root($data_root)
         item_sizes = {str(uri): data_root.file_size(str(uri)) for uri in uris}
         resolved_batch_count = (
             int(batch_count) if batch_count is not None else min(4, len(item_sizes))
@@ -460,7 +460,7 @@ $stage_plan_filter        if mode == "online":
             for batch in planned
         ]
 
-    @task.external_python(python="$venv_python", expect_airflow=False)
+    @task.external_python(python=$venv_python, expect_airflow=False)
     def process_batch(batch):
         """Sleep the stagger delay, then run this sub-DAG's stage on every episode."""
         import importlib.util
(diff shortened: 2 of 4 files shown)


kstonekuan (approved):
Thank you @chintondutta, and welcome! This is an outstanding first contribution: you found a real bug on your own, wrote it up with a minimal reproduction and a precise root cause, and shipped the fix with a regression test that locks it in.

What I validated locally:

- Ran your reproduction from #44 against current main: all four sub-DAGs fail with the truncated `\UXXXXXXXX` escape, exactly as reported. On this branch, all five DAG sources compile.
- Full quality gate is clean (`ruff check`, `ruff format --check`, `ty check`) and all 299 tests pass, including the new `test_dag_sources_survive_paths_that_are_not_valid_python_literals`.
- Checked the subtle spot in the rebuilt error message: the adjacent-literal concatenation with the trailing f-string binds before `+`, so the generated code is correct.

Things I especially appreciate:

- `repr()` at the substitution site instead of an escaping regex or allowlist. It is the smallest fix that holds for arbitrary input, and your note on why `dag_id`/`app_variable` still need `_validate_dag_identifiers` (they are used as identifiers, not just literals) shows you understood the boundary, not just the symptom.
- The regression test compiles the rendered sources with both a Windows interpreter path and a quote-injection data root, and asserts the crafted value round-trips as a single literal. That covers both the correctness and the injection halves of the bug.
- The comment in `render_sub_dag_source` explaining why the values are substituted as literals. The next person to add a placeholder will see it.

Merging now. If you would like to keep going, we would love more of exactly this kind of work: exercising HFlow against real data and reporting what breaks. The [good first issues](https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) are open to you too, though given the depth of this report you are clearly ready for the harder stuff. Come say hi on [Discord](https://discord.gg/vacepQvjmg)!

PR #47, opened 2026-08-20
catalog: repair dependent recorded_at only after winning the episodes race

### LINKED ISSUE #46: Catalog.append_episode: concurrent identical-outcome appends can split recorded_at across episodes/dependent tables
### Version or commit

`0116b6f47d46571bfe3f31af0aca30ef4148e8e3` (main, 2026-08-20)

### Environment

Ubuntu 24.04 (WSL), Python 3.12.3, x86_64, uv 0.12.5. No Docker/network required to reproduce -- pure local catalog + threading.

### Minimal reproduction

Two threads calling `Catalog.append_episode` with the identical outcome (same canonical file, same stamps, same check rows -- so the same `episode_id`/`run_fingerprint`), forced to interleave via a barrier on the storage boundary so the race triggers reliably instead of intermittently:

```python
import tempfile, threading
from pathlib import Path
import duckdb
from hflow.catalog import Catalog, CheckRunRow, content_episode_id
from hflow.transform import EpisodeStamps
from hflow.steps import CheckStatus

tmp = Path(tempfile.mkdtemp())
canonical = tmp / "episode.mcap"
canonical.write_bytes(b"fake-mcap-bytes")
catalog = Catalog(tmp / "catalog")
stamps = EpisodeStamps(schema_version="v1", pipeline_version="pipe-1", ffmpeg_version="7.0", robot_software_version=None)
check_rows = [CheckRunRow(check_name="dummy", check_version="1", critical=False, status=CheckStatus.PASSED, duration_s=0.1)]

barrier = threading.Barrier(2)
released, lock = [0], threading.Lock()
real = catalog.location

class Gated:
    def __getattr__(self, name): return getattr(real, name)
    def store_file_if_absent(self, local_file, relative):
        with lock:
            released[0] += 1
            first_two = released[0] <= 2
        if first_two:
            barrier.wait(timeout=5)
        return real.store_file_if_absent(local_file, relative)

catalog.location = Gated()
results = {}
def run(label):
    results[label] = catalog.append_episode(
        canonical_path=canonical, stamps=stamps,
        episode_metadata={"task": "t", "operator": "o", "success": True, "embodiment": "e"},
        check_rows=check_rows,
    )

t1, t2 = threading.Thread(target=run, args=("A",)), threading.Thread(target=run, args=("B",))
t1.start(); t2.start(); t1.join(); t2.join()

con = duckdb.connect()
timestamps = set()
for table in ("episodes", "check_runs"):
    f = next((real.path / table).glob("*.parquet"))
    timestamps.add(con.execute(f"SELECT DISTINCT CAST(recorded_at AS VARCHAR) FROM read_parquet('{f}')").fetchone()[0])
print(timestamps)  # {'...264433...', '...264332...'} -- two distinct values for ONE outcome
```

Ran this 5 times; every run produced 2 distinct `recorded_at` values across the 5 tables for the single outcome.

### Expected behavior

The module docstring and inline comments in `catalog.py` state the invariant explicitly: *"a racing worker appending the same outcome loses cleanly instead of duplicating"* and *"every file of ONE append carries one `recorded_at`"*. Two callers racing `append_episode` for the identical `(episode_id, run_fingerprint)` should end up with all 5 Parquet files (`episodes` + the 4 dependents) carrying the same `recorded_at`.

### Actual behavior

`episodes.recorded_at` and the dependent tables' `recorded_at` can disagree, and the 4 dependent tables can even disagree with *each other*, despite describing the literal identical outcome.

### Root cause

`Catalog.append_episode` (`src/hflow/catalog.py`, `append_episode`) generates `recorded_at = datetime.now(UTC)` once per call, then stores the 4 dependent tables with a create-if-absent primitive (`store_file_if_absent`, `os.link`-based, genuinely atomic -- first writer wins). On a loss, the old code assumed the existing file must be crash debris from an abandoned earlier attempt ("its episodes file never landed, or we would have returned `written=False` above") and unconditionally deleted and overwrote it with the *current* caller's own `recorded_at`:

```python
for table_name in ("check_runs", "measurements", "tags", "intervals"):
    ...
    if not self.location.store_file_if_absent(staged_file, dependent_key):
        # A crashed earlier append left this dependent behind ...
        self.location.delete(dependent_key)
        self.location.store_file_if_absent(staged_file, dependent_key)
episodes_created = self.location.store_file_if_absent(...)  # episodes: no such fallback
```

That reasoning only rules out a *finished* prior append -- it does not rule out a **live concurrent racer** computing the identical outcome right now. The initial `exists()` early-return check and this loop are not one atomic operation, so two callers can both pass the early check and both reach the loop. Whichever call reaches a given dependent table *second* deletes the first's just-written file and overwrites it with its own `recorded_at` -- independently, per table, in whatever order the two callers' loops interleave. `episodes` has no such fallback (plain first-writer-wins), so it ends up with whichever caller's *final* write happened to land first, independently of which caller "won" each dependent table.

This is operationally reachable, not just theoretical: the SDK's own generated Airflow runtime (`process_batch`, dynamically mapped over batches) is exactly the kind of caller that can retry or duplicate-dispatch a batch under transient failures, and every ingestion route goes through this shared append path.

There's a real downstream consequence, not just cosmetic timestamp skew: `curation.py`'s `episodes_latest` and `measurements_latest` views each independently rank rows by `(recorded_at DESC, run_fingerprint DESC)` *per table*. When an episode has more than one recorded run (a genuine reprocess), a `recorded_at` mismatch introduced by this race can make the two views pick *different* `run_fingerprint`s as "latest," and the `episodes` view's join then stitches together rows from two different runs for the same episode.

### Proposed fix

The dependent-table create-if-absent loss can't locally distinguish "stale crash debris" from "a live sibling racer" -- both look identical (dependent exists, `episodes` doesn't). Rather than guess, defer any repair until *after* the `episodes` race is decided: `store_file_if_absent` on `episodes` is atomic, so at most one caller ever observes `True` for a given `file_stem`. That caller is now the unique, durable owner of the outcome (whether this is a solo append, the winner of a concurrent race, or a retry repairing a genuine crash) and can safely force every dependent to carry its own `recorded_at` via an unconditional atomic replace (the existing `StorageRoot.publish`, already used elsewhere for exactly this "overwrite is correct" case) -- the loser of the `episodes` race does nothing further.

### Additional context

Verified locally (reproduction above, 5/5 runs) and have a fix ready that gates the dependent-repair step on `episodes_created`, plus a regression test that reliably forces the interleaving via a barrier at the storage boundary (a natural race only reproduces intermittently) and confirms the existing crash-repair test (`test_crash_repaired_append_keeps_one_recorded_at_across_tables`) still passes unchanged. Happy to open a PR against this issue.

chintondutta:
Fixes #46.

## What was broken

Two callers appending the identical outcome concurrently to `Catalog` (e.g. a retried or duplicate-dispatched Airflow batch task) both pass the `exists("episodes/...")` early-return check -- it and the dependent-table write loop are not one atomic operation. The old fallback treated any dependent that lost its create-if-absent race as certain crash debris and unconditionally deleted + overwrote it with the *current* caller's own `recorded_at`:

```python
if not self.location.store_file_if_absent(staged_file, dependent_key):
    # A crashed earlier append left this dependent behind ...
    self.location.delete(dependent_key)
    self.location.store_file_if_absent(staged_file, dependent_key)
```

That reasoning can't actually distinguish "stale debris from an abandoned attempt" from "a live sibling racer computing the identical outcome right now" -- both look identical from here (dependent exists, `episodes` doesn't). Each of the 4 dependent tables raced and "repaired" independently, so the `episodes` file and its dependents could end up with different `recorded_at` values for one outcome, and the dependents could even disagree with *each other*.

This isn't just cosmetic: `curation.py`'s `episodes_latest`/`measurements_latest` views each independently rank rows by `(recorded_at DESC, run_fingerprint DESC)` per table. On an episode with more than one recorded run, a mismatch introduced by this race could make the two views pick different `run_fingerprint`s as "latest" and the `episodes` view's join would then stitch together rows from two different runs.

## Fix

Don't guess at dependent-write time. `store_file_if_absent` on `episodes` is atomic, so at most one caller ever observes `True` for a given `file_stem` -- defer any repair until that unique winner is known:

```python
episodes_created = self.location.store_file_if_absent(...)
if episodes_created:
    # We are now the unique, durable owner of this file_stem.
    for table_name in (...):
        self.location.publish(staged_file, dependent_key)  # unconditional atomic replace
```

The loser of the `episodes` race does nothing further -- its own dependent writes, if any transiently won, get overwritten by the winner's repair pass. `StorageRoot.publish` is already an unconditional atomic replace used elsewhere for exactly this "overwrite is correct" case, so no new storage primitive was needed.

This correctly handles both scenarios that need different treatment:
- **True crash-repair** (sequential retry after a crash, no live racer): the retry is the only caller running, so it always wins `episodes` and repairs the stale dependents left by the crashed attempt -- unchanged behavior, still covered by the existing `test_crash_repaired_append_keeps_one_recorded_at_across_tables`.
- **Concurrent race** (two callers truly in parallel): exactly one wins `episodes` and repairs *all* dependents to its own `recorded_at`, including ones the other caller initially won -- the new bug this PR fixes.

## Testing

New regression test `test_concurrent_append_of_the_identical_outcome_keeps_one_recorded_at` (`tests/test_catalog_curation.py`) uses a small storage-boundary test double (a barrier gating `store_file_if_absent`) to force two threads to interleave reliably rather than intermittently, then asserts all 5 tables carry one `recorded_at`. Confirmed it fails against the pre-fix code (`mixed recorded_at across tables: {...}`, 2 distinct values) and passes with the fix.

```bash
uv run pytest tests/test_catalog_curation.py -q   # 17 passed
uv run pytest -q                                  # unrelated: tests/test_ffmpeg.py fails/errors in this sandbox (no ffmpeg/ffprobe on PATH); everything else passes: 277 passed, 2 skipped
uv run ruff check --fix
uv run ruff format
uv run ty check
```

No stored-format changes -- this only affects which `recorded_at` a dependent's row content carries, never which rows exist or what they otherwise contain (guaranteed identical by `file_stem` = `episode_id`-`run_fingerprint`, which already encodes the full observable outcome).

### DIFF
--- a/src/hflow/catalog.py
+++ b/src/hflow/catalog.py
@@ -443,8 +443,16 @@ class Catalog:
             # Build every table file locally first, then store dependents
             # before the episodes file (see above). store_file_if_absent is
             # atomic in both modes: rename-into-place locally, a conditional
-            # put (PutMode::Create) on a bucket -- so a racing worker
-            # appending the same outcome loses cleanly instead of duplicating.
+            # put (PutMode::Create) on a bucket.
+            #
+            # A dependent's create-if-absent can lose to either a genuinely
+            # stale file (a crashed earlier append -- its episodes file never
+            # landed, or we would have returned written=False above) or a
+            # LIVE concurrent racer appending the identical outcome right
+            # now: those two cases are indistinguishable from here (both
+            # look like "the dependent exists, episodes doesn't"), so a
+            # dependent that loses here is deliberately left alone instead
+            # of guessed at -- see the repair pass below.
             with tempfile.TemporaryDirectory(prefix="hflow-catalog-append-") as staging_name:
                 staging_dir = Path(staging_name)
                 for table_name in _TABLE_NAMES:
@@ -454,20 +462,34 @@ class Catalog:
                 for table_name in ("check_runs", "measurements", "tags", "intervals"):
                     staged_file = staging_dir / f"{table_name}.parquet"
                     dependent_key = f"{table_name}/{file_stem}.parquet"
-                    if not self.location.store_file_if_absent(staged_file, dependent_key):
-                        # A crashed earlier append left this dependent behind
-                        # (its episodes file never landed, or we would have
-                        # returned written=False above). Replace it so every
-                        # file of ONE append carries one recorded_at -- mixed
-                        # timestamps would let the per-key "latest" views
-                        # attribute another run's rows to this one. Safe to
-                        # delete: without its episodes file the run is
-                        # invisible (manifest-last).
-                        self.location.delete(dependent_key)
-                        self.location.store_file_if_absent(staged_file, dependent_key)
+                    self.location.store_file_if_absent(staged_file, dependent_key)
                 episodes_created = self.location.store_file_if_absent(
                     staging_dir / "episodes.parquet", f"episodes/{file_stem}.parquet"
                 )
+                if episodes_created:
+                    # store_file_if_absent on episodes is atomic, so at most
+                    # one caller ever observes True: we are now the unique,
+                    # durable owner of this file_stem, whether this is a
+                    # solo append, the winner of a race against a concurrent
+                    # duplicate, or a retry repairing a crashed earlier
+                    # attempt. Force every dependent to carry OUR
+                    # recorded_at -- unconditionally, since a losing
+                    # dependent write above (crash debris or a concurrent
+                    # racer that did not go on to win episodes) may hold a
+                    # different one. Every file of ONE append must carry one
+                    # recorded_at: the per-key "latest" views (curation.py)
+                    # each independently rank rows by (recorded_at,
+                    # run_fingerprint) per table, so a mismatch could let
+                    # them pick two different runs as "latest" for the same
+                    # episode and stitch together rows that never belonged
+                    # to the same outcome. ``publish`` is an unconditional
+                    # atomic replace (rename-into-place locally, a plain put
+                    # on a bucket) -- safe here because we already hold the
+                    # unique win on episodes.
+                    for table_name in ("check_runs", "measurements", "tags", "intervals"):
+                        staged_file = staging_dir / f"{table_name}.parquet"
+                        dependent_key = f"{table_name}/{file_stem}.parquet"
+                        self.location.publish(staged_file, dependent_key)
         finally:
             connection.close()
 
--- a/tests/test_catalog_curation.py
+++ b/tests/test_catalog_curation.py
@@ -473,6 +473,86 @@ def test_crash_repaired_append_keeps_one_recorded_at_across_tables(tmp_path: Pat
         connection.close()
 
 
+def test_concurrent_append_of_the_identical_outcome_keeps_one_recorded_at(
+    tmp_path: Path,
+) -> None:
+    """Two callers racing ``append_episode`` for the identical outcome (a
+    retried or duplicate-dispatched batch task, not a crash) must not split
+    one outcome's tables across two different recorded_at values.
+
+    Before the fix, a dependent table that lost its create-if-absent race
+    was unconditionally deleted and rewritten with the losing caller's OWN
+    recorded_at, independently per table -- so the caller that ultimately
+    won ``episodes`` (the durability marker) could end up with dependents
+    stamped by the OTHER caller, and different dependent tables could even
+    disagree with each other.
+    """
+    import threading
+
+    import duckdb
+
+    canonical = _fake_canonical(tmp_path)
+    catalog = Catalog(tmp_path / "catalog")
+    row = _check_row()
+
+    # A storage-boundary test double: gate the first two dependent-table
+    # writes on a barrier so both threads are guaranteed to reach
+    # append_episode's dependent-write loop at the same time, forcing the
+    # real interleaving a natural race only produces intermittently.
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Thank you @chintondutta, another excellent find and fix. What I validated locally:

- Full quality gate clean and all 304 tests pass (including the ffmpeg suite your sandbox could not run).
- Confirmed the new regression test fails against pre-fix `catalog.py` (mixed `recorded_at`, 2 distinct values) and passes with the fix.
- Traced the interleavings: the loser's create-if-absent writes can never overwrite the winner's repair pass (create-only loses to an existing file), and the winner's `publish` is ordered after the unique `episodes` win, so the final state is always the winner's. The crash-retry path is unchanged and still covered by the existing test.

The barrier-gated storage double is a great pattern: it forces the exact interleaving instead of hoping a stress loop hits it.

One observation, not a blocker: there is a residual double-failure window. If the winner crashes mid-repair (after `episodes` landed) while stale debris from an earlier crash existed, a later retry early-returns on `exists(episodes)` and never repairs the mixed dependents. That window is strictly narrower than the race this fixes and needs two failures to hit, so it is fine to leave; if it ever matters we can revisit with repair-on-read in the curation views. Feel free to open an issue for it if you want to track it.

Merging now.

PR #75, opened 2026-08-20
docs: add Windows platform support section to CONTRIBUTING.md

supercell02:
## Problem
The CONTRIBUTING.md guide currently does not mention Windows platform limitations. Windows users following the setup instructions encounter `ModuleNotFoundError: No module named 'fcntl'` and ffmpeg compatibility issues that are confusing and frustrating.

## Solution
Added a new "Platform Support" section immediately after the "Development setup" prerequisites that:
- Clarifies that Linux and macOS are fully supported for native development
- Explains that Windows is not currently supported for native development due to Unix-specific dependencies (`fcntl` file locking)
- Provides clear alternatives for Windows users: WSL2 (recommended) or Docker Desktop
- Includes a note about BIOS virtualization requirements for WSL2

## Validation
-Documentation is accurate (Windows indeed lacks `fcntl` support)
-WSL2 and Docker are viable alternatives for Windows users
-No Markdown syntax errors
-No breaking changes to existing documentation


kstonekuan (approved):
Thank you @supercell02, and welcome! CONTRIBUTING.md genuinely had nothing about platforms, and a Windows contributor hitting `ModuleNotFoundError: No module named 'fcntl'` on their first `pytest` run has no way to tell whether they broke something or whether it was never going to work. Worth documenting.

What I validated locally:

- Confirmed the mechanism you named: `src/hflow/storage.py` has a bare top-level `import fcntl`, so on native Windows `import hflow` fails before any test runs. Your claim is exactly right.
- Confirmed CI is Linux-only (`ubuntu-latest`), which is why I added a line about running the checks yourself on macOS.
- lychee reports 231 links with 0 errors, including the new cross-reference.

One fixup pushed to your branch (`62e4303`), mostly style plus two content trims:

- Sentence-cased the heading and switched "Linux & macOS" to "and", matching the surrounding sections, and used "HFlow" as the project is named elsewhere in the file.
- Stated the `fcntl` mechanism with a file pointer rather than as a general statement, since the repo's style is to point at the code that causes the behavior.
- Linked the [runtime prerequisites](https://github.com/Hebbian-Robotics/hflow/blob/main/docs/RUNTIME.md#prerequisites) instead of restating the WSL2 guidance, because that section already explains why the data root has to live on the Linux filesystem, and I would rather it be explained in one place.
- Dropped the Docker Desktop option and the BIOS note. The repo ships no devcontainer, so "run development inside a container" is not a path we can actually support today, and enabling virtualization is generic WSL2 setup that Microsoft's own docs cover better than we can.

If you disagree with either trim, say so and we can put it back. Merging now.

If you want another, the open [good first issues](https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) have plenty left, and several are documentation like this one (issues with an assignee are taken; everything else is fair game). We also hang out on [Discord](https://discord.gg/vacepQvjmg).

PR #80, opened 2026-08-21
fix(runtime): cap polling sleeps to timeout

HostX0:
## Summary

- cap Airflow health polling sleeps to the remaining timeout budget
- apply the same deadline-aware sleep to DAG registration polling
- add deterministic regression coverage for both runtime wait loops

## Root cause

Both loops checked the deadline only before polling, then always slept for the full `poll_interval_s`. When the configured interval exceeded the remaining budget, `hflow up` could exceed its timeout by nearly one full polling interval before raising.

The fix preserves the polling and error-reporting behavior while sleeping for `min(poll_interval_s, remaining_s)`.

## Verification

- `uv run pytest -q` — 308 passed, 3 skipped
- `uv run ruff check src/hflow/runtime/_client.py src/hflow/runtime/_lifecycle.py tests/test_runtime_client.py tests/test_runtime_cli.py`
- `uv run ruff format --check src/hflow/runtime/_client.py src/hflow/runtime/_lifecycle.py tests/test_runtime_client.py tests/test_runtime_cli.py`
- `uv run ty check`

### DIFF
--- a/src/hflow/runtime/_client.py
+++ b/src/hflow/runtime/_client.py
@@ -170,7 +170,10 @@ class AirflowClient:
                 last_status = health.summary()
             if on_poll is not None:
                 on_poll(last_status)
-            time.sleep(poll_interval_s)
+            remaining_s = deadline - time.monotonic()
+            if remaining_s <= 0:
+                break
+            time.sleep(min(poll_interval_s, remaining_s))
         raise TimeoutError(
             f"Airflow at {self.base_url} not healthy after {timeout_s:.0f}s (last: {last_status})"
         )
--- a/src/hflow/runtime/_lifecycle.py
+++ b/src/hflow/runtime/_lifecycle.py
@@ -104,7 +104,10 @@ def _wait_until_dag_registered(
         pending_dag_ids = still_pending
         if not pending_dag_ids:
             return
-        time.sleep(poll_interval_s)
+        remaining_s = deadline - time.monotonic()
+        if remaining_s <= 0:
+            break
+        time.sleep(min(poll_interval_s, remaining_s))
     raise TimeoutError(
         f"Airflow is healthy but the ingest DAG(s) {pending_dag_ids!r} never registered "
         f"within {timeout_s:.0f}s (last: {last_error}) -- the DAG file likely "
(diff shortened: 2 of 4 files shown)


kstonekuan (approved):
Thank you @HostX0, and welcome! You found this one yourself, described the root cause precisely, and pinned it with tests that fail for the right reason. That is a good first contribution.

What I validated locally:

- Full quality gate is clean (`ruff check`, `ruff format --check`, `ty check`) and all 308 tests pass.
- Confirmed both new assertions fail against pre-fix sources: the clock lands on `10.0` where the budget was `0.3`, exactly the overshoot you describe.
- Traced both loops. Each was `while time.monotonic() < deadline` followed by an unconditional `time.sleep(poll_interval_s)`, so the deadline was only ever observed at the top of the next iteration. `min(poll_interval_s, remaining_s)` is the right shape, and it preserves the polling and error-reporting behavior.
- Checked that the `remaining_s <= 0` guard is load-bearing rather than defensive: `time.sleep` raises `ValueError` on a negative argument, so computing a remaining budget without that check would have traded an overshoot for a crash. Good instinct.

Worth noting for scope, since your description is careful not to overclaim: no in-repo caller overrides `poll_interval_s`, so on the shipped `hflow up` path the overshoot was bounded by the 3 second default. But `AirflowClient` is public API (exported from `hflow.runtime`), so anyone calling `wait_until_healthy` with a longer interval could blow well past their own timeout. That is the case this fixes.

I like the fake-clock approach. Replacing the module reference and advancing time inside `sleep` makes the assertion exact instead of timing-dependent, which is why these tests are worth having rather than merely passing.

One small thing I would have done differently, not worth a round trip: in `test_runtime_client.py` you retargeted an existing test's interval from `0.1` to `10.0`, which strengthens the deadline assertion but drops it from roughly three poll iterations to one. Both properties matter, so a separate test would have kept both. Not blocking.

Merging now. If you want another, the open [good first issues](https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) have plenty left (issues with an assignee are taken; everything else is fair game), though given that you went and found this on your own, the more valuable thing is more of exactly this. We hang out on [Discord](https://discord.gg/vacepQvjmg).

kstonekuan:
One process note: this is still marked as a draft, so I cannot merge it. The review above stands and it is approved, so whenever you are ready, click "Ready for review" and I will squash it straight in. If you left it as a draft deliberately because you wanted to keep working on it, take your time and ignore me.

PR #82, opened 2026-08-21
feat(cli): add up --api-port so a new bundle can pick its port

### LINKED ISSUE #7: CLI: add `--api-port` to `hflow up` instead of hand-editing `.env`
## Current behavior

The runtime config already supports a custom API port — `RuntimeConfig.api_port: int = 8080` (`src/hflow/runtime/_bundle.py:167`), rendered into the bundle's `.env` as `API_PORT` — but `hflow up` never passes it. The `up` subparser (`src/hflow/cli.py:99` onward) has no port flag, so `docs/RUNTIME.md` (the "Port 8080 is taken" entry, line 351) has to document a manual workaround:

> **Port 8080 is taken.** The port lives in the bundle's `.env` as `API_PORT` — `hflow down`, edit it, `hflow up` again.

## Proposed change

- Add `--api-port` (type `int`, default `8080`) to the `up` subparser and forward it into the `RuntimeConfig(...)` construction in `_command_up` (`src/hflow/cli.py:321`).
- Simplify the "Port 8080 is taken" troubleshooting entry in `docs/RUNTIME.md` to point at the flag.

No Docker knowledge needed to test this: bundle rendering is unit-tested without Docker in `tests/test_runtime_bundle.py`, and the CLI-to-config plumbing can be asserted the same way.

## Definition of done

- `hflow up --api-port 9090` produces a bundle whose `.env` contains `API_PORT=9090`.
- A test covers the flag-to-config plumbing.
- `docs/RUNTIME.md` troubleshooting entry updated.

## Validation

```bash
uv run pytest -q
uv run ty check
```

chiruu12:
Closes #7.

`RuntimeConfig.api_port` already reached the bundle as `API_PORT` and `started_summary` already read it back, so the printed URL follows the port on its own. The flag was the only missing link. Adds `--api-port` to the `up` subparser and forwards it into the `RuntimeConfig(...)` in `_command_up`.

### One thing the issue did not account for

The troubleshooting entry could not just be replaced with the flag. Rendering is create-if-absent, `_bundle.py` says so directly:

```
# Report what is actually in effect: the .env on disk wins over config
# (a re-render with a changed port does not change a preserved .env).
```

and `hflow down` only runs `compose_down`, so it leaves the bundle and its `.env` on disk. `--api-port` therefore does nothing for a bundle that already exists, and the down/edit/up workaround is still the answer there. So the entry now covers both cases instead of pointing only at the flag, and the flag's own help says where it applies.

Tell me if you would rather `--api-port` overwrite a preserved `.env` when passed explicitly. That is a real change to the create-if-absent contract, since the same file holds the JWT secret and admin password, so I did not want to make that call inside a feature issue. Happy to send it as a follow-up.

I also left the port unvalidated, per the question on the issue. Nothing stops `--api-port 70000` reaching Compose today. Say the word and I will add the range check.

### Tests

Three in `tests/test_runtime_cli.py`, all through `main()` with the existing Docker stubs, so none of them need a daemon:

- `--api-port 9090` lands as `API_PORT=9090` in the rendered `.env`
- omitting it still yields `8080`, so existing bundles do not move
- a second `up` at a different port leaves the first `.env` alone, which is what makes the doc claim above true rather than asserted

The first and third fail on unmodified `main` with `unrecognized arguments: --api-port`. The second passes on both, by design: it is the guard on the unchanged path.

Nothing new was needed at the bundle layer. `test_env_created_with_config_values_and_secrets` and `test_env_preserved_on_rerender` already cover it.

### Validation

macOS, `uv sync --locked --all-extras`.

```
uv run pytest -q            305 passed, 5 failed, 6 skipped
uv run ruff check           All checks passed
uv run ruff format --check  88 files already formatted
uv run ty check             All checks passed
```

Base is 302 passed with the same 5 red, all ffmpeg and host-specific on this machine, so this adds three and moves nothing else.

### DIFF
--- a/docs/RUNTIME.md
+++ b/docs/RUNTIME.md
@@ -359,10 +359,12 @@ docker compose --file data/runtime/docker-compose.yaml logs airflow-dag-processo
 `ingest` prints the same hint (right after a fresh `up`, the DAG may still
 be parsing; retry in a few seconds).
 
-**Port 8080 is taken.** The port lives in the bundle's `.env` as `API_PORT`
-(re-renders don't change a preserved `.env`, so editing it sticks):
-`hflow down`, edit `API_PORT`, `hflow up` again. The API only ever binds
-to `127.0.0.1`, so two bundles on different ports coexist fine.
+**Port 8080 is taken.** For a bundle that does not exist yet, pass the port:
+`hflow up --api-port 9090`. For one already rendered, the `.env` on disk wins
+(re-renders never overwrite a preserved `.env`, so `--api-port` does nothing
+there and editing does stick): `hflow down`, edit `API_PORT`, `hflow up`
+again. The API only ever binds to `127.0.0.1`, so two bundles on different
+ports coexist fine.
 
 **`up` failed partway.** Whatever started is deliberately left running: the
 state *is* the diagnosis. `hflow status` to look, `docker compose ... logs
--- a/src/hflow/cli.py
+++ b/src/hflow/cli.py
@@ -21,6 +21,9 @@ DEFAULT_DATA_ROOT = Path("./data")
 DEFAULT_CATALOG_DIR = "./data/catalog"
 DEFAULT_BUNDLE_DIR = DEFAULT_DATA_ROOT / "runtime"
 DEFAULT_DEPLOY_OUTPUT_DIR = Path("./deploy")
+# Mirrors RuntimeConfig.api_port; kept here so the parser can state it without
+# importing the runtime package, which `up` defers until it actually runs.
+DEFAULT_API_PORT = 8080
 
 
 def _build_parser() -> argparse.ArgumentParser:
@@ -150,6 +153,17 @@ def _build_parser() -> argparse.ArgumentParser:
         type=Path,
         default=None,
         help="where to render the bundle (default: <data-root>/runtime; ./runtime for bucket URLs)",
+    )
+    up_parser.add_argument(
+        "--api-port",
+        type=int,
+        default=DEFAULT_API_PORT,
+        help=(
+            "host port for the Airflow API, written into a new bundle's .env as "
+            f"API_PORT (default: {DEFAULT_API_PORT}). An existing .env is never "
+            "rewritten, so this only takes effect on a bundle that does not have "
+            "one yet"
+        ),
     )
     up_parser.add_argument(
         "--hflow-source",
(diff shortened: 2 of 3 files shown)


kstonekuan (approved):
Merging.

You were right to refuse the instruction. My issue text said to "simplify the troubleshooting entry to point at the flag", and that would have been wrong: rendering is create-if-absent, `hflow down` leaves the bundle on disk, so `--api-port` does nothing for a bundle that already exists and the down/edit/up workaround is still the only answer there. Covering both cases, and putting the same caveat in the flag's own help text where someone will read it, is the correct call.

`test_up_api_port_does_not_rewrite_a_preserved_env` is the test I'd have asked for. A docs claim about preserved `.env` files is the kind of thing that rots silently, and now it can't.

What I validated locally:

- Full quality gate clean (`ruff check`, `ruff format --check`, `ty check`), 371 tests pass, lychee reports 246 links with 0 errors.
- Confirmed your pre-fix claim precisely: the first and third tests fail with `unrecognized arguments: --api-port`, and the second passes on both, as you said, because it's the guard on the unchanged path.

One fixup pushed to your branch (`7efda63`). `main` moved under you while this was open: a large commit landed that rewrote the same troubleshooting paragraph to document `API_BIND_HOST`, so `docs/RUNTIME.md` conflicted. I resolved it keeping both facts, your two-case port guidance plus the new bind-host note, and re-validated the combination. Nothing to do on your side.

Your two questions, both right to ask instead of decide:

**Overwriting a preserved `.env` when `--api-port` is passed explicitly: no.** I don't want that change. That file holds the JWT secret and the admin password, and "your secrets and edits survive a re-render" is a contract worth more than the convenience of repointing a port in place. Someone who wants a different port on an existing bundle can edit one line, and that's a fine price. If we revisit it, it should be a deliberate flag that says what it does, not a side effect of `--api-port`.

**Port range validation: yes, but not in this PR.** Nothing stops `--api-port 70000` reaching Compose today, and the failure a user gets is a Docker error and not an argparse one. Since you offered, I've filed it as #83 and assigned it to you. It's a slightly broader question than this flag (whether the check belongs in `RuntimeConfig` so the library boundary rejects it too, not only the CLI), which is why it didn't belong inside a feature PR.

One note, and I'm partly the cause: the `good first issue` pool is best kept for people making their first contribution, and you're well past that. In my review of #81 I pointed you at #8, which is itself a starter issue, so this is me correcting my own steer. From here the higher-leverage work is pushing on the runtime and catalog seams, and finding places where the docs and the code disagree, the way you did here. Never a rule, always an invitation. Plenty of unclaimed non-starter work coming as the hosted seams land.

PR #84, opened 2026-08-21
fix(runtime): reject an out-of-range api_port at construction

### LINKED ISSUE #83: CLI: reject an out-of-range --api-port instead of passing it to Compose
Follow-up from #82, raised by @chiruu12 in the pull request.

## Current behavior

`hflow up --api-port` accepts any `int`. The parser defines the flag at
`src/hflow/cli.py:210-218`, and `_command_up` passes the value directly into
`RuntimeConfig` at `src/hflow/cli.py:494-501`. `RuntimeConfig.api_port` has no
range validation at `src/hflow/runtime/_bundle.py:160-208`.

As a result, values such as `70000`, `0`, and `-1` reach the rendered `.env`.
The failure surfaces later as a Docker or Compose error instead of an argument
error.

## Agreed direction

Validate the invariant in `RuntimeConfig.__post_init__`, so library callers and
the command-line interface use the same rule. Accept only ports in `1-65535`,
and include the offending value in the `ValueError` message.

Catch the construction-time `ValueError` in a small block around
`RuntimeConfig(...)` in `_command_up`, print it as an `up:` error, and return 2.
Do not add it to the existing `start_runtime` exception handler at
`src/hflow/cli.py:516-534`. That handler returns 1 and warns that containers may
still be running, but configuration validation fails before anything starts.

The render-time identifier checks are separate injection guards. Keep this
change focused on the port invariant.

## Definition of done

- `RuntimeConfig` rejects a port outside `1-65535` with a message that names the
  offending value.
- `hflow up --api-port 70000` exits 2 without rendering a bundle or printing a
  traceback.
- Tests cover at least one out-of-range value and confirm that a valid port is
  accepted.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

chiruu12:
Closes #83.

## Summary

`hflow up --api-port 70000` now exits 2 with `up: api_port 70000 is not in 1-65535` and renders nothing. Before this it exited 0 and printed `Airflow UI: http://127.0.0.1:70000` with a full bundle on disk, leaving Compose to complain later.

`RuntimeConfig.__post_init__` holds the invariant, so a library caller gets the same answer as the command line. `_command_up` catches it in its own block.

## Why

The port is not only published by Compose. It is interpolated into `api_base_url` at `_bundle.py:946`, which `start_runtime` then polls for health and `started_summary` prints. An unusable port therefore costs the full 300s wait before failing, having printed a URL nobody can open.

That is also why the range starts at 1. Port 0 means "any free port" to `bind(2)`, never to `connect(2)`. The repo's own "any free port" case already resolves 0 to a real number before construction, in `_free_port` in `tests/test_runtime_integration.py`.

The error block is separate from the `start_runtime` handler on purpose. That one returns 1 and prints teardown advice naming containers, and at this point nothing has started, so all of it would be false.

`load_bundle` still reads `API_PORT` back out of a rendered `.env` with no range check. That asymmetry is deliberate: a `.env` is stored state that survives re-renders, so validating there would break a hand-edited bundle that works today.

The render-time identifier guards are untouched, per the issue.

### One thing I did not change, so you can decide

The check is a numeric comparison, so it says nothing about type, and there are two ends to that.

A non-int `api_port` from a library caller now raises `TypeError` rather than `ValueError`, and the new handler does not catch it. That one is narrow: the field is annotated `int`, `ty` flags it, and argparse converts before `RuntimeConfig` is reached, so the command line cannot reach it.

`bool` is the more interesting half, and it predates this change. `RuntimeConfig(api_port=True)` passes the range check, because `True == 1`, and renders `API_PORT=True` into the `.env`. `ty` will not flag it either, since `bool` is a subtype of `int`. So it is wrong before and after, just never checked.

One guard closes both. I left it out because the issue asked to keep this to the port invariant, and the `bool` case is not a regression from this PR. Say the word and I will add it here or send it separately.

## Validation

macOS, `uv sync --locked --all-extras`, verified in a clean clone of this branch.

```
uv run pytest -q            372 passed, 5 failed, 6 skipped
uv run ruff check           All checks passed
uv run ruff format --check  96 files already formatted
uv run ty check             All checks passed
```

Base is 363 passed with the same 5 red, all ffmpeg and host-specific on this machine. CI is green on 3.11 and 3.14.

Six new cases, none needing Docker:

- construction rejects 70000, 65536, 0 and -1, matching the whole message rather than the bare number
- construction accepts 1, 8080 and 65535, so the range cannot quietly become exclusive
- `up --api-port 70000` exits 2 with no bundle directory, no compose call, empty stdout, no traceback, and none of the "containers may still be running" advice
- `up --api-port 70000` against a bundle that already exists also exits 2 and leaves its `.env` untouched

All six fail on `main`. The three acceptance cases pass on both, by design.

That last test covers a behavior change worth naming. A preserved `.env` wins over the flag, so `--api-port` was inert on an existing bundle and an out-of-range value used to exit 0 silently. It now fails. Refusing a port beats accepting one, ignoring it, and reporting success, but it is a change for a script that always passes the flag.

## Checklist

### DIFF
--- a/docs/RUNTIME.md
+++ b/docs/RUNTIME.md
@@ -405,7 +405,8 @@ docker compose --file data/runtime/docker-compose.yaml logs airflow-dag-processo
 be parsing; retry in a few seconds).
 
 **Port 8080 is taken.** For a bundle that does not exist yet, pass the port:
-`hflow up --api-port 9090`. For one already rendered, the `.env` on disk wins
+`hflow up --api-port 9090` (1 to 65535; anything outside that is refused before
+the bundle is rendered). For one already rendered, the `.env` on disk wins
 (re-renders never overwrite a preserved `.env`, so `--api-port` does nothing
 there and editing does stick): `hflow down`, edit `API_PORT`, `hflow up`
 again. The API binds to `127.0.0.1` by default (`API_BIND_HOST` in `.env`
--- a/src/hflow/cli.py
+++ b/src/hflow/cli.py
@@ -491,14 +491,21 @@ def _command_up(arguments: argparse.Namespace) -> int:
     hflow_source = (
         arguments.hflow_source if arguments.hflow_source is not None else infer_hflow_source()
     )
-    config = RuntimeConfig(
-        pipeline_file=pipeline_file,
-        data_root=arguments.data_root,
-        app_variable=app_variable,
-        requirements_file=arguments.requirements,
-        hflow_source=hflow_source,
-        api_port=arguments.api_port,
-    )
+    try:
+        config = RuntimeConfig(
+            pipeline_file=pipeline_file,
+            data_root=arguments.data_root,
+            app_variable=app_variable,
+            requirements_file=arguments.requirements,
+            hflow_source=hflow_source,
+            api_port=arguments.api_port,
+        )
+    except ValueError as error:
+        # Its own block, not the start_runtime handler below: nothing has been
+        # rendered or started yet, so that handler's teardown advice would all
+        # be false.
+        print(f"up: {error}", file=sys.stderr)
+        return 2
     # A bucket data root has no local directory to host the bundle: ./runtime.
     default_bundle_dir = (
         Path(RUNTIME_BUNDLE_DIRECTORY_NAME)
--- a/src/hflow/cli.py
+++ b/src/hflow/cli.py
@@ -213,9 +213,9 @@ def _build_parser() -> argparse.ArgumentParser:
         default=DEFAULT_API_PORT,
         help=(
             "host port for the Airflow API, written into a new bundle's .env as "
-            f"API_PORT (default: {DEFAULT_API_PORT}). An existing .env is never "
-            "rewritten, so this only takes effect on a bundle that does not have "
-            "one yet"
+            f"API_PORT (default: {DEFAULT_API_PORT}, range 1-65535). An existing "
+            ".env is never rewritten, so this only takes effect on a bundle that "
+            "does not have one yet"
         ),
     )
     up_parser.add_argument(
--- a/src/hflow/runtime/_bundle.py
+++ b/src/hflow/runtime/_bundle.py
@@ -92,6 +92,10 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_AIRFLOW_IMAGE = "apache/airflow:3.3.1"
 DEFAULT_POSTGRES_IMAGE = "postgres:16"
+# TCP port range, less port 0, which is "any free port" to bind(2) and not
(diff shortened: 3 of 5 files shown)


kstonekuan (approved):
Merging.

I'd drafted a reply to #83 saying "let the fallout fall out": I assumed giving `up` a `ValueError` handler would necessarily change `dag_id`, `app_variable`, and `task_queue` from tracebacks to exit 2, and I was going to sanction that. You avoided the question entirely by scoping the `try` to the construction call, so the only `ValueError` reachable there is the port invariant and the render-time guards inside `start_runtime` keep their existing handler untouched. That's why the diff is as small as it is.

The precedent you argued from is stronger than you put it: `DeployConfig` already validates in `__post_init__` ("Parse at the boundary: an invalid data root never becomes a config"), and `_command_deploy` already wraps its construction in a `try` returning 2. So this isn't a new pattern, it's `RuntimeConfig` catching up with its sibling.

What I validated locally, on your branch merged with current `main`:

- Full quality gate clean (`ruff check`, `ruff format --check`, `ty check`), 380 tests pass.
- All six new cases fail on pre-fix sources, and the three acceptance cases (1, 8080, 65535) pass on both, exactly as you said.
- Smoke-tested the real command: `hflow up --api-port 70000` prints `up: api_port 70000 is not in 1-65535`, exits 2, renders no bundle directory, shows no traceback and none of the teardown advice.
- Confirmed the `bool` case you flagged: `RuntimeConfig(api_port=True)` is accepted and `api_port` comes back as `True`.
- Checked your justification for excluding 0. `_free_port` in `tests/test_runtime_integration.py` does bind to 0 and return the resolved number, so "0 means any free port to `bind(2)`, never to `connect(2)`" is grounded in the repo's own idiom and not an assertion. The constant's comment is the right place for it.

Two answers to what you left open:

**The `bool` and non-`int` halves: not here.** Neither is a regression from this PR and one guard closes both, which is why they belong together in their own change instead of smuggled into a range check. Filed as #85 with both halves and the `type(...) is not int` detail spelled out. Yours if you want it.

**The behavior change you named: correct, and worth having.** An out-of-range `--api-port` against an existing bundle used to exit 0 because a preserved `.env` made the flag inert, and now it exits 2. Refusing a port beats accepting it, ignoring it, and reporting success. A script that always passes the flag and relied on the silence was relying on a bug. Naming it in the description instead of letting me find it is why this review was quick.

`load_bundle` staying unvalidated is right too, for the reason you gave: a `.env` is stored state that survives re-renders, so validating on read would break a hand-edited bundle that works today.

PR #87, opened 2026-08-22
feat(ui): a React workspace UI over the hflow-server API

kstonekuan:
> **The server already merged** (#116). This PR is now only `ui/` and the one CI job that builds it — 52 files, no server code — so it can be judged as a frontend on its own.
>
> **Still a draft on purpose.** The open question is not whether it works, it is whether this is the UI we want. `hflow-server` publishes a typed schema at `/api/openapi.json` and nothing is reachable only from a browser, so this app can be rewritten, replaced, or joined by a second client without touching the server. That choice is deliberately reversible, which is why the server did not wait on it.

The first client of the workspace API.

- **Episodes** — the corpus as a faceted, sortable table over the catalog's wide `episodes` view. Every filter compiles to DuckDB SQL server-side; the page renders JSON.
- **Episode dossier** — one episode's measurements, check runs, intervals, tags and media, with a timeline over its recorded evidence.
- **Curate** — SQL with instant previews, saved queries, and manifest pinning.
- **Runs** — the ingest DAG drawn from the served topology, with one run's live per-task state.
- **Pipeline** — the generated DAG plus the registered steps, drawn *inside* the `process_batch` node that runs them. They have no dependency edges on each other, so the graph nests them rather than inventing a chain, and says so.

Strictly a client: no computation here that the server could not do, and nothing reachable that `curl` could not reach.

Offline like the server — no CDN, no telemetry, no runtime downloads. Built assets are copied into `hflow_server/static/` at package time, which the CI job here verifies actually happens.

React 19, Vite 8, react-router 7, TanStack Query and Table, Radix, CodeMirror 6, lucide. `pnpm check` and `pnpm build` clean.

## Known follow-up

`ui/src/api.ts` hand-writes 64 interfaces mirroring the server's contract, and they have already drifted — the server declares `kind: StepKind`, the copy says `kind: string`, and most of the contract's 10 `Literal` unions arrive as bare `string`. These should be generated from `/api/openapi.json` instead. That is the change that makes a *second* UI cheap, so it matters more than it looks.


kstonekuan:
Superseded by #142.

Issue #86, opened 2026-08-21
Checks: registering a check with required kwargs is accepted, then fails per-episode (action_rate)


kstonekuan:
Found by @Sagar-024 in the Egocentric-10K stress test ([discussion #48](https://github.com/Hebbian-Robotics/hflow/discussions/48), "awkward 1"), where it cost them a silently empty z-score query.

## Current behavior

`action_rate(episode, *, topics)` (`src/hflow/checks.py`) has a required keyword-only parameter. Registering it the obvious way is accepted without complaint:

```python
app.check()(action_rate)   # no topics
```

At run time the call raises `TypeError: action_rate() missing 1 required keyword-only argument: 'topics'`, which `App` catches as an infrastructure failure (`src/hflow/app.py`, the `except Exception` around the check call) and records as `CheckStatus.ERROR`.

Reproduced against current `main`:

```
has_errors : True
quarantined: False
  check='action_rate' status=ERROR
    error tail: TypeError: action_rate() missing 1 required keyword-only argument: 'topics'
```

So the error is not swallowed: it is on the report and in the recorded `check_runs` row. But the episode is **not** quarantined, ingestion completes, and the measurement column is simply absent from the catalog. A caller driving `app.process()` from their own script (as a local batch run has to today, see #48 "awkward 2") sees a successful-looking run and an empty column, and the natural read is "this corpus has no action-rate signal" rather than "this check never ran".

`content_digest` and `episode_duration` take no required kwargs, so `action_rate` is the only built-in with this shape, which is why nothing else surfaced it.

## Why registration is the right place to fix it

Whether this call can ever succeed is knowable when the decorator runs, without any episode: the function has a keyword-only parameter with no default, and `@app.check()` supplies exactly one positional argument. `inspect.signature` can see that. Failing at registration turns a per-episode runtime error and a missing column into one immediate, unmissable message at pipeline definition time.

The wrapper the user actually wants is a one-liner, and the error should say so:

```python
@app.check()
def action_rate_check(ep: hflow.Episode) -> hflow.CheckResult:
    return action_rate(ep, topics=["/joint_states"])
```

## Definition of done

- `app.check()(action_rate)` raises at registration, with a message naming the unsatisfiable parameter and showing the wrapper form above.
- The same guard covers user-written checks with required extra parameters, not just this built-in.
- A registered check that takes only the episode is unaffected, including one with optional parameters that have defaults.
[...]

Issue #85, opened 2026-08-21
RuntimeConfig accepts api_port=True (bool passes the range check)


kstonekuan:
Follow-up from #84 (found and scoped out by @chiruu12 there, deliberately, since it predates that change).

## Current behavior

`RuntimeConfig(api_port=True)` is accepted and renders `API_PORT=True` into the bundle's `.env`. Confirmed against #84's branch:

```
>>> c = RuntimeConfig(pipeline_file=Path("p.py"), data_root="./data", api_port=True)
>>> c.api_port
True
```

The new range check in `RuntimeConfig.__post_init__` does not catch it, because `bool` is a subclass of `int` and `True == 1`, so `MIN_PORT <= True <= MAX_PORT` holds. `ty` does not flag it either, for the same subtyping reason. The rendered `.env` then carries the literal string `True` where Compose expects a number.

This is unreachable from the command line: argparse converts with `type=int` before `RuntimeConfig` is constructed. It only affects library callers, and `api_port` is public API.

## The other end of the same gap

`#84` also notes that a non-int `api_port` (say a string) now raises `TypeError` from the comparison rather than `ValueError`, which `_command_up`'s new handler does not catch. Also library-only for the same reason, and also narrow. One type guard closes both halves, so they are worth settling together rather than one at a time.

## What to do

Reject a non-`int` (and specifically a `bool`) `api_port` in `__post_init__`, alongside the existing range check, with a message naming the offending value and its type. `type(self.api_port) is not int` is the direct expression of "an `int`, and not a `bool` pretending to be one"; `isinstance` alone will not do it.

Keep it to `api_port`. If the same treatment is wanted for other `RuntimeConfig` fields, that is a separate, larger question about how much the config validates.

## Definition of done

- `RuntimeConfig(api_port=True)` raises `ValueError`, and so does a string or a float.
- A test covers `True` specifically, since that is the case the range check silently lets through.
- `hflow up --api-port 8080` still works, and the CLI path is unchanged.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

kstonekuan:
@chiruu12 this is the follow-up you scoped out of #84, both halves in one place. It is yours if you want it: say so here and I will assign it, since GitHub only lets me assign someone once they have commented on the issue.

No obligation. If you would rather move on to something else, say that instead and I will leave it open for whoever picks it up.

[...]

PR #79, opened 2026-08-21
testing: include offending values in validation error messages

ayam04:
## What changed

Four validation errors now include the offending value(s), matching the pattern already used by every other raise in `_validate_video_episode_spec` (e.g. `source_start_s must be >= 0, got ...`):

- `src/hflow/testing.py` — `camera_name must not be empty, got {spec.camera_name!r}` (the check is `.strip()`-based, so `""` vs `"   "` is exactly the distinction the message should surface)
- `src/hflow/testing.py` — `black_segment and freeze_segment must not overlap, got {spec.black_segment!r} and {spec.freeze_segment!r}`
- `src/hflow/steps.py` — `declared step version must not be empty for step {name!r}` (the optional stretch named in the issue; `name` is in scope)
- `src/hflow/app.py` — `derived channel topic must not be empty, got {topic!r}` (optional stretch)

No test asserts these strings (per the issue), so nothing else needed updating.

## Validation

```bash
uv run ruff check src/hflow/testing.py src/hflow/steps.py src/hflow/app.py   # All checks passed!
uv run ruff format --check ...                                               # already formatted
uv run ty check                                                              # All checks passed!
uv run pytest -q                                                             # 298 passed, 3 skipped
```

`tests/test_ffmpeg.py` shows `3 failed, 6 errors` both on this branch and on clean `main` (verified in a detached worktree) — pre-existing WSL/environment ffmpeg-binary detection issues, unaffected by this change.

Fixes #65

PR #88, opened 2026-08-22
refactor: use format.py constants for metadata record names in testing.py

### LINKED ISSUE #62: testing.py hardcodes `episode/v1` and `robot_software_version` instead of the format.py constants
`testing.py` spells format-defined strings by hand instead of importing the
constants that own those names.

## Current behavior

`src/hflow/format.py` defines:

- `METADATA_RECORD_EPISODE = "episode/v1"` at `src/hflow/format.py:27`
- `EPISODE_KEY_ROBOT_SOFTWARE_VERSION = "robot_software_version"` at
  `src/hflow/format.py:36`

`src/hflow/testing.py` imports only `NANOSECONDS_PER_SECOND` from that module at
`src/hflow/testing.py:47`, then hardcodes the canonical strings at:

- `src/hflow/testing.py:474`
- `src/hflow/testing.py:482`
- `src/hflow/testing.py:532`
- `src/hflow/testing.py:670`
- `src/hflow/testing.py:675`

If the record name or key changes, the synthetic fixtures can silently keep
writing an older format name.

## Pattern to copy

`src/hflow/transform.py:55-70` imports both constants. Its metadata handling at
`src/hflow/transform.py:444-446` shows them used together.

## Definition of done

- `testing.py` imports and uses `METADATA_RECORD_EPISODE` and
  `EPISODE_KEY_ROBOT_SOFTWARE_VERSION` at every format-owned site.
- `rg 'name="episode/v1"|"robot_software_version"' src/hflow/testing.py` finds
  no hardcoded metadata names.
- Runtime behavior remains unchanged.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

ayam04:
## What changed

`src/hflow/testing.py` spelled format-owned metadata strings by hand at five sites (`"episode/v1"` and `"robot_software_version"`). It now imports and uses the constants that own those names, `METADATA_RECORD_EPISODE` and `EPISODE_KEY_ROBOT_SOFTWARE_VERSION` from `hflow/format.py`, mirroring the pattern already used in `transform.py` (its metadata handling at `transform.py:444-446`).

Sites changed (per issue #62):
- `_video_episode_metadata`: dict key + `reserved_metadata_keys` entry
- `_write_video_episode` (video fixture): `writer.add_metadata(name=...)`
- `_write_episode_with_channels` (channel fixture): `writer.add_metadata(name=...)` + metadata dict key

Runtime behavior is unchanged — the constants hold the same strings.

## Validation (run in WSL2, since HFlow imports `fcntl` and native Windows cannot import it)

```bash
uv sync --locked --all-extras
uv run ruff check        # All checks passed!
uv run ruff format --check  # 96 files already formatted
uv run ty check          # All checks passed!
uv run pytest -q         # 371 passed, 3 skipped; 3 failed + 6 errors all in tests/test_ffmpeg.py
```

The `tests/test_ffmpeg.py` failures are environment-only on this machine (they need real `ffmpeg`/`ffprobe` binaries on PATH): I ran the same file on a clean `main` worktree and got the identical `3 failed, 18 passed, 1 skipped, 6 errors`, so they are pre-existing and unrelated to this change.

Also verified the issue's DoD grep is now empty:

```bash
rg 'name="episode/v1"|"robot_software_version"' src/hflow/testing.py   # no output
```

Fixes #62

### DIFF
--- a/src/hflow/testing.py
+++ b/src/hflow/testing.py
@@ -45,7 +45,11 @@ from pathlib import Path
 from mcap.writer import Writer
 
 from hflow.ffmpeg import ffmpeg_path
-from hflow.format import NANOSECONDS_PER_SECOND
+from hflow.format import (
+    EPISODE_KEY_ROBOT_SOFTWARE_VERSION,
+    METADATA_RECORD_EPISODE,
+    NANOSECONDS_PER_SECOND,
+)
 
 JOINT_STATES_TOPIC = "/joint_states"
 JOINT_STATE_SCHEMA_NAME = "sensor_msgs/msg/JointState"
@@ -471,7 +475,7 @@ def _video_episode_metadata(spec: VideoEpisodeSpec) -> dict[str, str]:
     metadata = {
         "task": spec.task,
         "operator": spec.operator,
-        "robot_software_version": spec.robot_software_version,
+        EPISODE_KEY_ROBOT_SOFTWARE_VERSION: spec.robot_software_version,
     }
     if spec.success is not None:
         metadata["success"] = "true" if spec.success else "false"
@@ -479,7 +483,7 @@ def _video_episode_metadata(spec: VideoEpisodeSpec) -> dict[str, str]:
         "task",
         "operator",
         "success",
-        "robot_software_version",
+        EPISODE_KEY_ROBOT_SOFTWARE_VERSION,
     }
     additional_metadata_keys: set[str] = set()
     for metadata_key, metadata_value in spec.metadata:
@@ -529,7 +533,9 @@ def write_video_episode(
             message_encoding="cdr",
             schema_id=schema_id,
         )
-        writer.add_metadata(name="episode/v1", data=_video_episode_metadata(resolved_spec))
+        writer.add_metadata(
+            name=METADATA_RECORD_EPISODE, data=_video_episode_metadata(resolved_spec)
+        )
         for sequence, (log_time_ns, payload) in enumerate(camera_messages):
             writer.add_message(
                 channel_id=channel_id,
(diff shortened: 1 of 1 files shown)


kstonekuan (approved):
Thank you @ayam04. Complete and correctly scoped, and you counted better than my issue did: the body listed four sites and there are five, since `synthesize_episode`'s own metadata dict carries `robot_software_version` too. You caught it, which is the whole point of the change.

Also right to leave `task`, `operator`, and `success` as literals. `EPISODE_KEY_ROBOT_SOFTWARE_VERSION` is the only episode key with a constant in `format.py`, so converting the others would have meant inventing constants the format does not own yet. Matching what exists beats being tidy.

What I validated locally:

- Full quality gate is clean (`ruff check`, `ruff format --check`, `ty check`) and all 380 tests pass on your branch against current `main`.
- The definition-of-done grep comes back empty, and I confirmed zero remaining hardcoded occurrences of either string in the module.
- Generated a synthetic episode through `synthesize_episode` afterwards to confirm the fixtures still write a real, readable file rather than merely type-checking.

On the `tests/test_ffmpeg.py` failures: environment, not your change, and they pass here. Checking them against a clean `main` worktree before reporting is the right habit, and it is the second time you have done it. It genuinely saves me time.

Merging now.

One note I owe you, and it applies from here rather than as a complaint about this PR: the `good first issue` pool is best kept for people making their first contribution, and with #79 already merged you are past that. Nothing wrong with this one, and I would rather you had taken it than have it sit. But for the next one, the higher-leverage work is where you have already shown range: you did the layering analysis that shaped `action_rate` back in #24, and that kind of reasoning is worth more here than a mechanical refactor. Two live examples if either appeals, neither a starter issue: #86, where registering a check with required keyword arguments is accepted and then fails once per episode, and #85, where `RuntimeConfig(api_port=True)` slips through a range check because `bool` subclasses `int`. Never a rule, always an invitation.

PR #91, opened 2026-08-22
fix(app): reject checks with required params beyond the episode at registration

### LINKED ISSUE #86: Checks: registering a check with required kwargs is accepted, then fails per-episode (action_rate)

ayam04:
Fixes #86

`action_rate(episode, *, topics)` registers via `app.check()` without complaint, then fails once per episode at run time with `TypeError: missing 1 required keyword-only argument`, which App records as an infrastructure error and leaves the measurement column absent from the catalog.

**Change:** at registration, `App.check()` now inspects the function signature and rejects any required parameter beyond the single canonical episode — including required keyword-only parameters (e.g. `topics`) and required positional parameters — with a `ValueError` naming the unsatisfiable parameter(s) and showing the wrapper form:

```
check 'action_rate' cannot be called with only an episode: required parameter(s) without defaults: topics. Wrap it in a function that binds them, e.g.

    @app.check()
    def action_rate_check(ep: hflow.Episode) -> hflow.CheckResult:
        return action_rate(ep, topics=...)
```

Optional parameters (with defaults), `*args`/`**kwargs`, and episode-only checks are unaffected.

**Tests** (both in tests/test_processing_regressions.py):
- required keyword-only param → raises at registration, nothing registered
- optional keyword-only param with default → registers fine

**Validation:**
- ruff check: clean
- ruff format --check: clean
- ty check: clean
- pytest -q: 373 passed, 3 skipped; the only failures are the known pre-existing `test_ffmpeg.py` env errors (no ffmpeg binaries in WSL; identical on clean main, confirmed previously with Kingston)

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -15,6 +15,7 @@ quality outcome.
 """
 
 import hashlib
+import inspect
 import json
 import os
 import re
@@ -296,6 +297,43 @@ class EnrichmentRunReport:
         if self.error is not None:
             return CheckStatus.ERROR
         return CheckStatus.MEASURED
+
+
+def _unsatisfiable_check_parameters(function: Callable[..., object]) -> list[str]:
+    """Names of required parameters a registered check could never receive.
+
+    At run time a check is called with exactly one argument: the canonical
+    episode (``registered.function(canonical_episode)``). A check is only
+    satisfiable if every remaining parameter is optional (has a default) or
+    is a ``*args``/``**kwargs`` sink. Parameters whose defaults the runtime
+    call cannot supply (required positional-or-keyword beyond the episode,
+    required keyword-only) would make every episode fail with a
+    ``TypeError`` that the App records as an infrastructure error at ingest
+    time. Returning their names lets registration reject them up front.
+    """
+    try:
+        parameters = inspect.signature(function).parameters.values()
+    except (TypeError, ValueError):
+        return []
+    seen_episode = False
+    unsatisfiable: list[str] = []
+    for parameter in parameters:
+        if parameter.kind in (
+            inspect.Parameter.POSITIONAL_ONLY,
+            inspect.Parameter.POSITIONAL_OR_KEYWORD,
+        ):
+            if parameter.default is not inspect.Parameter.empty:
+                continue
+            if not seen_episode:
+                seen_episode = True
+                continue
+            unsatisfiable.append(parameter.name)
+        elif (
+            parameter.kind is inspect.Parameter.KEYWORD_ONLY
+            and parameter.default is inspect.Parameter.empty
+        ):
+            unsatisfiable.append(parameter.name)
+    return unsatisfiable
 
 
 def _execute_enrichment(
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
Thank you @ayam04. Merging.

The message is the part I care most about, and it is right: it names the parameter, shows the wrapper form, and interpolates the real function name so the snippet is copy-pasteable rather than illustrative. That is the difference between an error that teaches and one that just refuses.

Handling required positional parameters beyond the episode as well as keyword-only was more than the issue asked for and it is the correct scope: both fail identically at run time, so both belong at registration.

What I validated on your branch against current `main` (f9ac687):

- `ruff check`, `ruff format --check`, `ty check` clean; 382 passed, 3 skipped. Nothing in the repo or the examples registered a check this now rejects.
- Confirmed the message text by registering `action_rate` bare, and confirmed optional keyword-only, `*args`/`**kwargs`, and episode-only checks all still register.

One hole left, and a follow-up if you want it: a zero-argument callable still registers and still fails per episode, with `TypeError: takes 0 positional arguments but 1 was given`. It is the mirror image of what you fixed and equally knowable from the signature. Out of scope for this PR, worth an issue.

Separately, and only as context rather than something to fix: `functools.partial(action_rate, topics=[...])` is rejected too, but by the identity check, not yours, with `cannot derive a stable implementation identity ... register it with version='...'`. That is pre-existing and arguably correct, and it happens to be a second reason the wrapper form your message recommends is the right advice.

On the `test_ffmpeg.py` failures: environment, and they pass here. Third time you have checked against a clean `main` before reporting. Keep doing that.

PR #92, opened 2026-08-22
examples/stress: reproducible synthetic benchmark

Sagar-024:
**Summary**

Adds a one-command reproducible stress test at `examples/stress/synthetic.py`. It generates a seeded corpus of canonical MCAP episodes (2 to 10 s, 1 to 3 cameras at 10/15/30 Hz, 20% black segments, 20% joint jumps) and runs `app.process` over them, printing wall time per episode and a check status summary at the end.

One command:

```
uv run python examples/stress/synthetic.py --episodes 200 --seed 42
```

Same seed and episode count produces the same corpus, so a run today and a run next week are directly comparable. `--generate-only` skips ingest, `--ingest-only` reuses an existing corpus.

**Why**

The stress numbers from Discussion #48 were from a script I wrote on the fly. Kingston asked for them to be reproducible ("the next stress test is one command"), and the local batch path gap (Awkward 2) was also part of what kept `action_rate` silent. This turns the anecdote into a tool.

Kept it narrow on purpose: one new file under `examples/stress/`, no changes to hflow internals, no new dependencies. The whole thing is importable by someone else's script the way `examples/egocentric/prepare.py` is. `action_rate` is registered with the wrapper form so it keeps working cleanly once #86 lands.

**Validation**

```
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

All clean. pytest: 298 passed, 3 skipped.

Reproducibility run:

```
uv run python examples/stress/synthetic.py --episodes 30 --seed 7
uv run python examples/stress/synthetic.py --episodes 5 --seed 99
```

First run produced 30 episodes. Second run cleaned the corpus dir and printed "Found 5 episodes to ingest", "Episodes: 5 | OK: 5 | ERROR: 0", and "MEASURED: 30" (5 episodes times 6 checks). That was the gate I wanted to pass before sending this up.

**Checklist**

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -72,6 +72,29 @@ contact-sheet artifacts for the accepted set.
 
 Guide and expected artifacts: [`egocentric/README.md`](./egocentric/README.md)
 
+## Synthetic stress corpus
+
+**Use it for:** a repeatable throughput measurement over a corpus large enough
+to be worth timing, with no download.
+
+**Prerequisites:** the normal development environment plus `ffmpeg` on `PATH`;
+no Docker, network, or API key. Two hundred episodes take a few minutes and
+about 1 GB of disk.
+
+```bash
+uv run python examples/stress/synthetic.py --episodes 200 --seed 42
+```
+
+The example plans a seeded corpus (2 to 10 s, one to three cameras at 10, 15,
+or 30 Hz, with black segments and joint jumps injected at a fixed rate), writes
+it to `./stress_corpus`, runs `App.process` over every episode, and prints the
+per-episode wall time and the distribution of check statuses. The same
+`--episodes` and `--seed` always plan the same corpus, so runs before and after
+a change are comparable. Use `--generate-only` to build a corpus without
+ingesting it, and `--ingest-only` to re-run against one you already have.
+
+Code: [`stress/synthetic.py`](./stress/synthetic.py)
+
 ## Example requirements
 
 Keep examples small enough to read, but complete enough to execute from the
--- a/examples/stress/synthetic.py
+++ b/examples/stress/synthetic.py
@@ -1,0 +1,277 @@
+#!/usr/bin/env python3
+"""Synthetic stress corpus generator and ingest runner for hflow."""
+
+import argparse
+import hashlib
+import random
+import subprocess
+import sys
+import tempfile
+import time
+from dataclasses import dataclass
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    pass
+
+from hflow.app import App
+from hflow.checks import (
+    action_rate,
+    camera_frame_stats,
+    episode_duration,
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging.

Verified the thing that mattered: two separate runs at `--seed 11` produce byte-identical episode files, and a different seed produces a different corpus. That turns your 200-episode number into a measurement.

Registering `action_rate` through the wrapper form was the right call, and #91 landed an hour ago, so the bare form now fails at registration. Your script is already on the correct side of that.

I pushed a fixup to your branch instead of sending it back, and rebased onto current `main` (you were based on `d2e13df`, four commits behind). What I changed, since some of it is worth knowing:

**A duplicated block.** `_check_ffmpeg`, `_check_hflow_env`, and `_compute_pipeline_version` were each defined twice, at lines 50-70 and again at 213-233. Neither `ruff` nor `ty` catches this: `F811` is enabled and doesn't fire on it, so your clean gate wasn't telling you anything false. I only found it by reading the file top to bottom.

**Dead symbols.** `_compute_pipeline_version` and `CONVERTER_VERSION` had no callers. `DEFAULT_EPISODES`, `DEFAULT_SEED`, `DEFAULT_OUTPUT_DIR`, and `JOINT_STATES_TOPIC` were defined but the literals were used instead, so I wired them in. `if TYPE_CHECKING: pass` is gone.

**`_check_hflow_env` could never fire.** It runs after the module-level `from hflow.app import App`, so a missing hflow raises `ImportError` before the friendly message is reachable. Dropped it, kept the ffmpeg check, switched to `shutil.which` to match `examples/egocentric/prepare.py`.

**A planner bug worth understanding.** `image_hz` was drawn twice (the first draw discarded but still consuming RNG state), and `has_black_segment` and `has_joint_jump` were each drawn from a *separate* `rng.random() < 0.2` than the value they described. So `has_black_segment` was `True` while `black_segment` was `None` about a third of the time. Both fields were unread, so nothing behaved wrongly, but a reader would reasonably trust them. One draw each now, and the redundant flags are gone.

**The catalog entry.** `examples/README.md` requires "a link from this catalog" of every new example, and it was the one box in your checklist ticked without a change. Added, same shape as the others, documenting `--generate-only` and `--ingest-only`.

Validated after the fixup: `ruff check`, `ruff format --check`, `ty check` clean, 390 passed and 3 skipped, plus the determinism check above and both `--generate-only` and `--ingest-only` paths.

One thing I deliberately left alone: `--seed` varies the *plan*, but each episode's synthetic content uses `seed=plan.index`, so the frames themselves don't change between seeds. Fine for a throughput benchmark and I wouldn't change it without a reason, but know it's the case if you ever use this to hunt content-dependent behavior.

Next up is the LeRobot converter whenever you're ready, and the WSL2 docs PR if you still want it. The mapping in #48 is settled, so nothing's blocking you there.

PR #94, opened 2026-08-22
fix(app): reject checks that cannot accept the episode at registration

### LINKED ISSUE #93: Checks: a callable that cannot accept the episode registers, then fails per-episode
### What happens

A check that takes no parameters registers successfully, then fails once per episode with a `TypeError` recorded as an `ERROR` status. It is the mirror image of the case #91 just fixed, and knowable from the signature at the same point.

```python
import hflow
from pathlib import Path

def no_episode() -> hflow.CheckResult:          # forgot the episode parameter
    return hflow.CheckResult(measurements={"n": 1})

app = hflow.App("zero-arg", data_root=Path("/tmp/zeroarg"))
app.check()(no_episode)                          # accepted
```

Processing any episode then gives:

```
  no_episode: ERROR
    TypeError: no_episode() takes 0 positional arguments but 1 was given
has_errors=True
```

`report.has_errors` is `True` and the measurement column is absent, so this is
not silent, but the failure lands per episode at ingest time instead of at the
call that could never have worked.

### Why registration is the right place

Checks are invoked with exactly one argument, the canonical episode
(`registered.function(canonical_episode)`). #91 established that whether a
check *can* be called is a property of its signature, and moved the
too-many-required-parameters case to registration for that reason. A callable
that accepts zero positional parameters fails the same call for the same
reason, so it belongs in the same place.

### Current behavior

`_unsatisfiable_check_parameters` in `src/hflow/app.py:302` reports required
parameters the runtime cannot supply, and `App.check()` raises on a non-empty
result at `src/hflow/app.py:621`. The helper's `seen_episode` flag at
`src/hflow/app.py:318` tracks whether a slot for the episode was found, but
the function only ever returns the *surplus* names, so a signature where
`seen_episode` stays `False` (no positional slot at all) returns an empty list
and registers.

### Definition of done

`app.check()` raises `ValueError` for a callable that cannot accept the episode
positionally, naming the check and stating what the signature must be. Cases to
get right:

- `def f()` is rejected.
- `def f(**kwargs)` is rejected. `**kwargs` cannot absorb a positional
  argument, so this is the same defect wearing a different signature:
  `(lambda **k: 1)(episode)` raises `TypeError: takes 0 positional arguments
  but 1 was given`.
- `def f(episode)`, `def f(episode, *, threshold=1.0)`, and `def f(*args)` all
  still register. `*args` does absorb the episode, so it is satisfiable, and it
  is the one variadic form that is.
- `def f(*, episode)` is rejected: the episode is passed positionally, so a
  keyword-only parameter of that name cannot receive it. Note this one already
  raises today, via the surplus-parameter path, and the message is about
  required keyword-only parameters rather than about the missing episode slot.
  Either message is defensible; just make sure the case stays rejected.

Keep the message in the shape #91 used, which names the problem and shows the
working form rather than only refusing.

### Pattern to copy

`_unsatisfiable_check_parameters` and its call site in `App.check()`, both
added in #91. The tests to extend are
`test_check_with_required_extra_parameter_fails_at_registration` and
`test_check_with_optional_extra_parameter_registers` in
`tests/test_processing_regressions.py`.

### Validation

```
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Context

Follow-up to #91 (which closed #86). @ayam04, this is the hole I mentioned in
that review, yours if you want it: comment here and I will assign it.

ayam04:
Fixes #93

The mirror of #91: `def f()` (or `def f(**kwargs)`, `def f(*, episode)`) registers successfully and then fails once per episode with `TypeError: takes 0 positional arguments but 1 was given`, recorded as `ERROR` per episode.

**Change:** `_unsatisfiable_check_parameters` now returns `(surplus, accepts_episode)` — whether the signature has any slot that can receive the episode positionally (a plain parameter, or `*args` which absorbs it; `**kwargs` alone is not a slot). `App.check()` raises `ValueError` naming the check and stating the required signature when there is no slot.

Behavior per the issue's cases (verified):
- `def f()` -> rejected
- `def f(**kwargs)` -> rejected
- `def f(episode)`, `def f(episode, *, threshold=1.0)`, `def f(*args)` -> register
- `def f(*, episode)` -> rejected (surplus path, as noted in the issue)

**Tests added:** no-episode rejected, kwargs-only rejected, varargs registers (plus the #91 tests still pass).

**Validation:**
- ruff check: clean
- ruff format --check: clean
- ty check: clean
- pytest -q: 376 passed, 3 skipped; only pre-existing `test_ffmpeg.py` env failures (no ffmpeg binaries in WSL; identical on clean main)

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -299,23 +299,32 @@ class EnrichmentRunReport:
         return CheckStatus.MEASURED
 
 
-def _unsatisfiable_check_parameters(function: Callable[..., object]) -> list[str]:
-    """Names of required parameters a registered check could never receive.
+def _unsatisfiable_check_parameters(
+    function: Callable[..., object],
+) -> tuple[list[str], bool]:
+    """Required parameters a registered check could never receive.
 
     At run time a check is called with exactly one argument: the canonical
     episode (``registered.function(canonical_episode)``). A check is only
-    satisfiable if every remaining parameter is optional (has a default) or
-    is a ``*args``/``**kwargs`` sink. Parameters whose defaults the runtime
-    call cannot supply (required positional-or-keyword beyond the episode,
-    required keyword-only) would make every episode fail with a
-    ``TypeError`` that the App records as an infrastructure error at ingest
-    time. Returning their names lets registration reject them up front.
+    satisfiable if it can accept that episode positionally and every other
+    parameter is optional (has a default) or is a ``*args``/``**kwargs``
+    sink. Parameters the runtime call cannot supply (required
+    positional-or-keyword beyond the episode, required keyword-only) would
+    make every episode fail with a ``TypeError`` that the App records as an
+    infrastructure error at ingest time.
+
+    Returns ``(surplus, accepts_episode)``: the names of required parameters
+    beyond the episode, and whether the signature has any slot that can
+    receive the episode positionally (a plain parameter, or ``*args``, which
+    absorbs it). ``**kwargs`` alone is not a slot: it cannot absorb a
+    positional argument.
     """
     try:
         parameters = inspect.signature(function).parameters.values()
     except (TypeError, ValueError):
-        return []
+        return [], True
     seen_episode = False
+    sees_varargs = False
     unsatisfiable: list[str] = []
     for parameter in parameters:
         if parameter.kind in (
@@ -328,12 +337,14 @@ def _unsatisfiable_check_parameters(function: Callable[..., object]) -> list[str
                 seen_episode = True
                 continue
             unsatisfiable.append(parameter.name)
+        elif parameter.kind is inspect.Parameter.VAR_POSITIONAL:
+            sees_varargs = True
         elif (
             parameter.kind is inspect.Parameter.KEYWORD_ONLY
             and parameter.default is inspect.Parameter.empty
         ):
             unsatisfiable.append(parameter.name)
-    return unsatisfiable
+    return unsatisfiable, seen_episode or sees_varargs
 
 
 def _execute_enrichment(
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
Thank you @ayam04. Merging, with one fixup pushed to your branch.

Returning `(surplus, accepts_episode)` from one signature walk is the right shape: both questions are answered from the same pass, and the two failure modes keep distinct messages. Getting `*args` right as a slot while correctly refusing `**kwargs` alone is the subtle half, and you got both.

**The fixup: a defaulted episode parameter was being rejected.**

```python
@app.check()
def defaulted_episode(episode: hflow.Episode | None = None) -> hflow.CheckResult:
    ...
```

This is called as `function(canonical_episode)`, so the default is never used. It registered and ran on `main` and your version raised `cannot accept the episode`. The cause is the ordering in the positional branch: it `continue`d on any parameter carrying a default *before* claiming the episode's slot, so a defaulted first parameter never set `seen_episode`. Claiming the slot first, then testing later positional parameters for defaults, fixes it and leaves the rejected set otherwise unchanged.

Worth stating the invariant, because it is what makes the change checkable: **a signature should be rejected exactly when it cannot be called with one positional argument.** Every case now agrees with that, which I verified by calling each one:

| signature | registers | callable with one positional arg |
|---|---|---|
| `def f(episode)` | yes | yes |
| `def f(episode=None)` | yes | yes |
| `def f(episode=None, *, t=1.0)` | yes | yes |
| `def f(*args)` | yes | yes |
| `def f()` | no | no |
| `def f(**kwargs)` | no | no |
| `def f(*, episode)` | no | no |

Before the fixup, rows two and three registered as "no" against a "yes". That table is a good way to check this kind of change, and the issue I wrote only listed the cases I had thought of, so it did not catch this either.

I added a test that pins registration and then calls the check the way the runtime does, so it also fails if the one-positional-argument convention ever changes.

Also rebased you onto current `main`: you were on `f9ac687`, and #95 landed in between.

Validated after the fixup: `ruff check`, `ruff format --check`, `ty check` clean, 398 passed and 3 skipped, plus the seven cases above checked by hand.

Two notes on process, both good:

Leaving `except (TypeError, ValueError): return [], True` permissive when `inspect.signature` fails is the correct default. A C builtin with no introspectable signature should not be refused on the strength of our failure to read it.

And the reason this was reviewable at all is that you listed the exact cases and their expected outcomes in the PR body. That is what let me test the boundary rather than the happy path, and it is how I found the defaulted case. Keep doing that.

PR #95, opened 2026-08-22
fix(cli): report a missing pipeline file as bad input, not a crash

### LINKED ISSUE #90: CLI: hflow up prints a traceback for a missing pipeline file
### What happens

`hflow up` with a `--pipeline` path that does not exist prints a traceback and exits 1. Every other command that resolves a pipeline file exits 2 with one line.

```
$ hflow up --pipeline /tmp/definitely-missing.py
  ...
  File "src/hflow/runtime/_lifecycle.py", line 62, in start_runtime
    paths = render_bundle(config, bundle_dir)
  File "src/hflow/runtime/_bundle.py", line 858, in render_bundle
    raise FileNotFoundError(pipeline_source)
FileNotFoundError: /tmp/definitely-missing.py
$ echo $?
1
```

```
$ hflow manifest --pipeline /tmp/definitely-missing.py
manifest: importing /tmp/definitely-missing.py failed: [Errno 2] No such file or directory
$ echo $?
2

$ hflow stale --pipeline /tmp/definitely-missing.py
stale: importing /tmp/definitely-missing.py failed: [Errno 2] No such file or directory
$ echo $?
2
```

`deploy` is the exact A/B. It raises the same `FileNotFoundError(pipeline_source)` from `_deploy.py:151` that `_bundle.py:858` raises, and it exits 2 cleanly:

```
$ hflow deploy --pipeline /tmp/definitely-missing.py --data-root-uri s3://bucket/x
deploy: /tmp/definitely-missing.py
$ echo $?
2
```

So the raise is not the problem. `_command_up` catches `(ComposeError, TimeoutError)` and nothing else, so `FileNotFoundError` escapes to the interpreter.

### Why the exit code matters here

1 and 2 are not interchangeable in this CLI. 1 means the runtime started and then failed, so containers may still be up and the teardown advice applies. 2 means bad input and nothing happened. A missing pipeline file is the second kind: `up` fails inside `render_bundle`, before any container is created, so a script that reads the exit code is told to go clean up something that was never started.

### Definition of done

`hflow up --pipeline <missing>` prints one line to stderr and exits 2, with nothing started and no traceback.

Two details worth pinning, both learned from #26:

- The message should carry the errno text, not a bare path. `deploy` currently prints `deploy: /tmp/definitely-missing.py`, which is the bare-path form #26 was closed for. Copying it into `up` would reintroduce that complaint.
- The handler should be its own `except` block, not an addition to the existing `(ComposeError, TimeoutError)` tuple. That tuple's handler prints teardown advice, which is false at this point: nothing has been rendered or started.

Suggested wording, matching what `manifest` and `stale` already print:

```
up: importing /tmp/definitely-missing.py failed: [Errno 2] No such file or directory
```

### Reproduction

```
hflow up --pipeline /tmp/definitely-missing.py; echo $?
```

No Docker or data root needed. It fails while rendering the bundle.

### Environment

- hflow: main @ f9ac687
- OS: macOS 15
- Python: 3.12

### A related question, kept out of this issue

`_bundle.py:858` is one of several bare `FileNotFoundError(path)` constructor calls (`_bundle.py:822`, `_deploy.py:151`, `_deploy.py:166`, `storage.py:344`), while `storage.py:656` uses the three-argument errno form. Whether that should be centralized or left per call site is a design question, not part of this fix. I will raise it separately once you tell me where you want it.

chiruu12:
## Summary

`hflow up --pipeline <missing>` printed a traceback and exited 1. It now prints one line and exits 2.

```
up: [Errno 2] No such file or directory: '/tmp/definitely-missing.py'
```

Closes #90.

## Why

1 and 2 are not interchangeable here. 1 means the runtime started and then failed, so the handler that owns it prints teardown advice and the caller is told containers may still be running. A missing pipeline file fails inside `render_bundle`, before any container exists, so 1 sent a script after containers that were never created. That is why the new handler is its own `except` block rather than an addition to the `(ComposeError, TimeoutError)` tuple.

The raise sites moved to the three-argument `FileNotFoundError`, matching `storage.py:656`, so `str(error)` carries the errno text instead of a bare path. That is four sites: the pipeline and requirements checks in `_bundle.py` and the same pair in `_deploy.py`.

`deploy` needed no CLI change. `_command_deploy` already caught `FileNotFoundError` and returned 2, so fixing the raise site fixed its message too. Its old output was `deploy: /tmp/x.py`, which is the bare-path form #26 was closed for and which only `storage.py` was ever fixed for.

Two questions a reviewer would reasonably ask, answered up front.

**Can the new handler swallow a genuine runtime failure?** No. The one plausible source after containers start is a missing `docker` binary, and `_compose.py:49` already converts that into `ComposeError`. Nothing between `compose_up_detached` and the return raises `FileNotFoundError`, so the new handler only ever sees `render_bundle`.

**Does it lie about a path that exists?** Yes, in one case, and I left it. `is_file()` is False for a directory, so `hflow up --pipeline <a-directory>` now says "No such file or directory" about something that is present. The old bare-path message was not wrong so much as useless, so this is a small regression in accuracy for an input this issue does not cover. Fixing it properly means raising `IsADirectoryError`, which changes the exception type on a public API and breaks every `except FileNotFoundError` that currently catches it. That is a design call, and it belongs with the bare-versus-errno question already headed for Discussions rather than in this PR.

## Validation

```
uv run pytest -q                386 passed, 5 failed, 6 skipped
uv run ruff check .             All checks passed!
uv run ruff format --check .    97 files already formatted
uv run ty check                 All checks passed!
```

The 5 failures are `test_ffmpeg.py` and `test_run_profiles.py` on macOS. They reproduce on unmodified `ca17730` with no changes applied, so they are the local ffmpeg build and not this change.

Reverting all three source files to `origin/main` and rerunning the CLI suite:

```
FAILED test_up_reports_a_missing_pipeline_file_as_bad_input
FAILED test_up_reports_a_missing_requirements_file_the_same_way
FAILED test_deploy_missing_pipeline_names_the_reason_not_just_the_path
FAILED test_deploy_missing_requirements_names_the_reason_too
4 failed, 25 passed
```

All 29 pass with the change restored. Each of the four raise sites is pinned on its own: reverting any single one of them fails exactly one test.

Behaviour before and after, same command:

```
before:  traceback, exit 1
after:   up: [Errno 2] No such file or directory: '/tmp/definitely-missing.py'
         exit 2
```

stderr carries two lines, not one: `up: rendering the runtime bundle at <dir>` prints first, from the existing progress callback. The definition of done said one line, so flagging the difference. The error itself is one line and the narration is pre-existing behaviour on every `up`.

## Checklist

- [ ] I updated documentation for changed behavior, flags, formats, or requirements. No doc change: no documented flag, format or requirement changes, and no doc states the old exit code.

### DIFF
--- a/src/hflow/cli.py
+++ b/src/hflow/cli.py
@@ -522,6 +522,14 @@ def _command_up(arguments: argparse.Namespace) -> int:
 
     try:
         paths, _ = start_runtime(config, bundle_dir, on_progress=print_progress_to_stderr)
+    except FileNotFoundError as error:
+        # Its own block, not the tuple below. This is raised while rendering the
+        # bundle, before any container exists, so the teardown advice attached to
+        # that handler would send the caller after containers that were never
+        # created. Same reason it exits 2 (bad input, nothing started) and not 1
+        # (started, then failed).
+        print(f"up: {error}", file=sys.stderr)
+        return 2
     except (ComposeError, TimeoutError) as error:
         # Deliberately leave whatever started running: the state is the
         # diagnosis. Tell the user how to look at it and how to tear it down.
--- a/src/hflow/runtime/_bundle.py
+++ b/src/hflow/runtime/_bundle.py
@@ -58,6 +58,7 @@ extracts them to temp files):
    mass failure visible.
 """
 
+import errno
 import hashlib
 import json
 import logging
@@ -855,7 +856,9 @@ def render_bundle(config: RuntimeConfig, bundle_dir: Path | str) -> BundlePaths:
     bundle_directory = Path(bundle_dir)
     pipeline_source = Path(config.pipeline_file)
     if not pipeline_source.is_file():
-        raise FileNotFoundError(pipeline_source)
+        # Three-argument form, as storage.py does: str() then carries the errno
+        # text, so the CLI prints why the path failed and not just the path.
+        raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), str(pipeline_source))
 
     dags_dir = bundle_directory / "dags"
     logs_dir = bundle_directory / "logs"
@@ -889,7 +892,9 @@ def render_bundle(config: RuntimeConfig, bundle_dir: Path | str) -> BundlePaths:
     if config.requirements_file is not None:
         requirements_source = Path(config.requirements_file)
         if not requirements_source.is_file():
-            raise FileNotFoundError(requirements_source)
+            raise FileNotFoundError(
+                errno.ENOENT, os.strerror(errno.ENOENT), str(requirements_source)
+            )
         shutil.copyfile(requirements_source, user_dir / "requirements.txt")
 
     hflow_source = (
--- a/src/hflow/runtime/_deploy.py
+++ b/src/hflow/runtime/_deploy.py
@@ -29,6 +29,8 @@ spool-through-mirror boundary; workers therefore keep the processing core on
 local files without pretending bucket keys have ``pathlib.Path`` semantics.
 """
 
+import errno
+import os
 import shlex
 import shutil
 from dataclasses import dataclass
@@ -148,7 +150,7 @@ def render_deploy_bundle(config: DeployConfig, output_dir: Path | str) -> Deploy
     output_directory = Path(output_dir)
     pipeline_source = Path(config.pipeline_file)
     if not pipeline_source.is_file():
-        raise FileNotFoundError(pipeline_source)
+        raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), str(pipeline_source))
 
     dags_dir = output_directory / "dags"
     user_dir = output_directory / "user"
@@ -163,7 +165,9 @@ def render_deploy_bundle(config: DeployConfig, output_dir: Path | str) -> Deploy
     if config.requirements_file is not None:
         requirements_source = Path(config.requirements_file)
         if not requirements_source.is_file():
-            raise FileNotFoundError(requirements_source)
+            raise FileNotFoundError(
+                errno.ENOENT, os.strerror(errno.ENOENT), str(requirements_source)
+            )
         shutil.copyfile(requirements_source, user_dir / "requirements.txt")
         requirements_copied = True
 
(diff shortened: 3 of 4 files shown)


kstonekuan (approved):
Thank you @chiruu12. Merging.

This is the most thoroughly argued PR the repo has had. Three things stand out.

**You fixed the raise sites instead of the symptom.** Catching `FileNotFoundError` in `_command_up` and printing something nicer would have closed the issue. Moving all four sites to the three-argument form fixed `deploy`'s message as a side effect, with no CLI change, because `_command_deploy` already caught it and returned 2. That is the smaller diff and the more correct one.

**You changed the wording I suggested, and you were right to.** The issue proposed `up: importing <path> failed: ...` to match `manifest` and `stale`. Those commands genuinely import the pipeline; `up` fails in `render_bundle` before importing anything, so `up: [Errno 2] No such file or directory: '<path>'` is accurate where my wording would have been a lie about what the command was doing. I did not notice that when I wrote the issue.

**You answered the two questions a reviewer would ask, before being asked.** I went to check whether the new handler could swallow a genuine post-start failure and found you had already traced it: `_compose.py:49` converts a missing `docker` into `ComposeError`, so the handler only ever sees `render_bundle`. That saved me the trip.

The directory case is the right call. `--pipeline <a-directory>` now says "No such file or directory" about something that exists, and the honest fix is `IsADirectoryError`, which changes the exception type on a public API and breaks every existing `except FileNotFoundError`. Deferring that to the Discussions thread with the bare-versus-errno question is correct. Flagging it in the PR body rather than leaving me to find it is what made it a non-issue.

Same for the two-line stderr: the definition of done said one line, the narration is pre-existing on every `up`, and saying so up front is better than silently not matching.

What I validated on your branch:

- `ruff check`, `ruff format --check`, `ty check` clean; 394 passed, 3 skipped. Your branch was already on current `main` (`ca17730`).
- `hflow up --pipeline <missing>`: one error line, exit 2, no traceback. `hflow deploy` likewise exits 2 and now prints the errno text instead of the bare path.
- Your revert-each-site claim holds: each of the four raise sites is pinned by exactly one test.

The `#26` observation was a good catch and worth stating plainly: that issue was closed with only `storage.py` fixed, and the pattern it was closed for survived in four other places. That is the kind of thing worth reopening an old issue over.

Open the Discussions thread whenever you want. Two questions are now queued for it: whether the bare `FileNotFoundError(path)` construction should be centralized, and whether a directory should raise `IsADirectoryError`. Both are real, and the second one is a public-API decision, so it should not be settled in a review comment.

Issue #96, opened 2026-08-22
Transform: a pass-through video channel starting mid-GOP is accepted and stamped conforming, but doctor rejects it


kstonekuan:
### What happens

The transform accepts a pass-through video channel that starts mid-GOP, writes it out, and stamps the output `provenance/v1`. `hflow doctor` then reports that same output as non-conforming with an `ERROR`. The transform certifies a file the project's own conformance checker rejects.

Reproduced on `main` (`2f07aa0`):

```
baseline canonical episode: doctor conforming=True
video topic '/cam_0/compressed', 30 messages
keyframe pattern: K.........K.........K.........
mid-GOP slice: messages 1..9 (9 messages, 0 keyframes)

TRANSFORM: ACCEPTED, and stamped the output as canonical
DOCTOR on the transform's own output: conforming=False
  [error] video-stream-starts-mid-gop: /cam_0/compressed: first message is not a keyframe;
          the stream is not decodable from the start
```

To build the input: take any canonical episode hflow produced, read its `foxglove.CompressedVideo` messages back, drop message 0 (the leading keyframe), keep messages up to just before the next keyframe, and write those into a fresh MCAP on a `foxglove.CompressedVideo` channel. Every individual message is valid, which is exactly why the current validation misses it.

### Why it happens

`_validate_passthrough_video_payload` (`src/hflow/transform.py:216`) is called once per message from the read loop at `src/hflow/transform.py:365`, and it is **stateless**. It enforces three things, all per-message:

- `format == "h264"` (`transform.py:223`)
- exactly one AUD-delimited access unit per message (`transform.py:236`)
- SPS/PPS present when *this* unit is a keyframe (`transform.py:242`)

"The stream starts on a keyframe" is not a property of any single message, so a per-message validator structurally cannot see it. `doctor` can, because `_check_video_payload` takes an `is_first_message` flag (`src/hflow/doctor.py:98`, passed at `doctor.py:268`) and raises `video-stream-starts-mid-gop` at `doctor.py:131`.

So the two validators disagree, and the transform is the more permissive one, which is the wrong way round: the transform is what mints the provenance stamp asserting conformance.

### Why this matters now

Two reasons beyond the inconsistency itself.

The module docstring already claims the stronger property. `src/hflow/transform.py:9` says pass-through channels are admitted "only after every message is validated against the canonical video constraints", and FORMAT.md lists keyframe-first among those constraints. The code does not implement what the docstring promises.

[...]

PR #103, opened 2026-08-22
examples/lerobot: pusht to canonical MCAP converter

Sagar-024:
Scope per the settled mapping in #48: lerobot/pusht, one camera, end to
end. Plain callable convert_episode plus CLI in examples/lerobot/prepare.py.

- /observation.image as foxglove.CompressedVideo, protobuf, cameras group
- /observation.state and /action as custom float32[2] CDR with the field
  named position, so joint_discontinuity runs with only a topic override
- exact x264 params from src/hflow/video.py at preset medium, hflow
  doctor verdict CONFORMING on the converted file
- log_time_ns derived from frame index, parquet timestamp used only for
  the 50 ms PTS cross-check, fail loud on disagreement
- traceability in episode/v1 (source_dataset, source_revision,
  source_episode_index), converter_version in its own key,
  no robot_software_version
- catalog entry added to examples/README.md

Two judgment calls to flag:

1. provenance/v1: you said drop it because the transform writes its own,
and traceability is in episode/v1 as directed. But hflow doctor fails the
raw converted file without a provenance/v1 record per FORMAT.md, so the
converter writes one (schema_version, pipeline_version, ffmpeg_version,
gop_preset, gop_seconds, source_uri, pipeline_version taken from
CONVERTER_VERSION) purely to pass conformance. The transform replaces it
on ingest. If you would rather doctor not require provenance on pre-ingest
files, that looks like a FORMAT or doctor question and I am happy to file it.

2. Schema names are lerobot_msgs/msg/State and lerobot_msgs/msg/Action
because the CDR decoder requires valid ROS2 package names. The schema text
is a plain float32[2] position, not a faked JointState, and it matches the
simple-message form used in src/hflow/testing.py.

Validation: hflow doctor CONFORMING; app.process produces measurements for
timestamp_regularity, camera_frame_stats, episode_duration, action_rate,
and joint_discontinuity with only a topic override; ruff, ruff format,
ty clean; 377 passed plus 12 end-to-end, 3 skipped.


kstonekuan (changes requested):
Not merging yet. Three things need changing first, and one of your two questions is me being wrong, not you.

Leave these alone, they're right: the channel split, the CDR encoding, the index-derived `log_time_ns`, the exact x264 params, and `episode/v1` carrying the traceability keys.

**Your question 1: you're right and my correction was wrong.** I told you to drop `provenance/v1` because our transform replaces it. That reasoning only holds for a file whose next stop is `write_canonical_episode`. Your converter emits a canonical episode in its own right, `doctor` checks it as one, and FORMAT.md requires the record, so writing it is correct and not a workaround. No issue to file: the mistake was mine, in that thread. Keep it.

That said, `ffmpeg_version: "unknown"` is a false stamp. You shell out to ffmpeg, so read the real thing (first line of `ffmpeg -version`) rather than writing a placeholder into provenance. The reason I said it "comes out `not-used`" was the pass-through case, where nothing is transcoded. You do transcode, so it should carry the version that did it.

Your question 2 is fine. `lerobot_msgs/msg/State` for a valid ROS 2 package name with a plain `float32[2] position` body is the right call. One thing to fix while you are there: `action_schema_id` is registered with `state_schema_data`, so `/action` advertises a schema named `Action` whose text says `position`. Give it its own one-line text.

### The three blockers

**1. `convert_episode` is not a plain callable.** This is the one design constraint from #48, and the PR body claims it is met:

```python
def convert_episode(args: argparse.Namespace, manifest: dict, source_paths: dict[str, Path], data_root: Path) -> Path
```

To call this from another script I have to build an `argparse.Namespace` and a `manifest` dict shaped exactly like your loader's output. That is a CLI internal, not an importable entry point. The signature you proposed in #48 was the right one, and it is what "plug and playable" meant:

```python
def lerobot_to_mcap(dataset_repo=..., revision=..., output_dir=..., episode_index=None, camera_key=...) -> list[Path]
```

Keep `main()` as the thin argparse wrapper that calls it. `examples/egocentric/prepare.py` is the precedent for that split.

**2. The parquet view binds one arbitrary file, and can crash.**

```python
for pf in parquet_files:
    conn.execute(f"CREATE VIEW IF NOT EXISTS data AS SELECT * FROM read_parquet('{pf}')")
```

`IF NOT EXISTS` makes every iteration after the first a no-op, so `data` is whichever file `glob("**/*.parquet")` yielded first, and the comment above it ("Register all parquet files") describes something the code does not do. Your manifest downloads two parquet trees, `data/chunk-000/...` and `meta/episodes/chunk-000/...`, into the same root, so which one wins is filesystem order. I verified both outcomes:

```
files in order ['a_data.parquet', 'z_meta.parquet'] -> view resolves to the data file
reversed order                                      -> view resolves to the meta file, and then
  SELECT * FROM data WHERE episode_index = 1 ORDER BY frame_index
  BinderException: Referenced column "frame_index" not found in FROM clause!
```

It works on your machine because "data" sorts before "meta". Select the file you mean by its manifest role instead of globbing, and quote it the way `catalog.py` does (`str(path).replace("'", "''")`).

**3. The 50 ms cross-check does not compare against the video.**

```python
expected_time = i / PUSHT_FPS
if abs(parquet_ts - expected_time) > 0.050:
```

That compares the parquet timestamp against the index formula, which is the same formula you derive `log_time_ns` from. It can only detect the parquet column drifting from a constant rate. The rule we agreed in #48 was parquet against **MP4 PTS**, whose purpose is catching video and state desync, and nothing here reads PTS at all.

This matters more than it looks, because of how you slice: `-ss/-to` with `-c copy` cuts at the nearest keyframe, so the sliced file can begin at a different frame than `from_timestamp` implies. That is exactly the desync the check existed to catch, and as written it cannot see it. Read the PTS list (`ffprobe -show_entries frame=pkt_pts_time`, or packet PTS) and compare per frame.

### Also worth fixing, not blocking

**Dead egocentric scaffolding.** `PlannedFault`, `fault_segment_s`, the fault validation block, and the stride/window arithmetic in `_expand_episode_plan` are about 150 lines validating a feature this converter does not have: nothing ever reads `episode["fault"]`, `source_member`, or `source_start_s`. Worse, `sources` round-robins by index (`sources[i % len(sources)]`) over three entries that are a parquet, an mp4, and a meta parquet, so "the source video for episode 1" is a parquet file. Drop the parts you do not use; a manifest listing the three files by role is what you actually need.

**Nothing is verified.** Every `sha256` in the manifest is `""`, and each `_verify_sha256` call is guarded by `if source_video.sha256:`, so the download is unchecked. The egocentric manifest pins real hashes and that is the point of having one. Either fill them in or drop the machinery.

**`_extract_source_videos` copies the corpus twice.** It `shutil.copy2`s every file from `huggingface/` to `source/` while its own docstring says "already downloaded, no extraction needed". On pusht that is a needless duplicate of a multi-hundred-MB tree. Point the reader at the download directory.

**Two sources of truth for fps.** `PUSHT_FPS = 10.0` drives `log_time_ns` and the cross-check, while `_transcode_mp4_to_h264` probes `r_frame_rate` for `gop_frames`. If a dataset ever disagrees with the constant, timestamps and GOP structure disagree silently. Probe once and pass it down.

Minor: `_ = float(parts[1])` computes and discards; `if i < len(access_units)` is unreachable given the count check above it; the `nal_walk_index` loop in `_split_h264_by_aud` advances an index nothing reads.

Reimplementing the Annex B splitter instead of importing `hflow.video.split_annex_b_stream` is the right instinct, since that module is not public API and the whole point is that this tool stands alone. If you ever want the keyframe flags, that is the one thing your version drops and #96 is about why they matter.

### What I checked

`ruff check`, `ruff format --check`, `ty check` clean; 398 passed, 3 skipped on your branch. I did not run the converter: it needs the pusht download, so your `doctor` CONFORMING result and the measurement list are your evidence, not mine. Once the entry point is a plain callable I can exercise it against a small fixture without the download, which is worth having anyway.

Rebase when you pick this up: `main` moved to `5d5cf71` (the check signature guard now covers `enrich()` and `derive()` too).

The mapping and the conformance work are right. All three blockers are "make it the thing we agreed", not "you misunderstood the format".

Sagar-024:
All requested changes addressed.

kstonekuan (approved):
All three blockers fixed. Merging with a small fixup.

Checked each rather than taking the summary on trust:

**1. Entry point.** `lerobot_to_mcap(dataset_repo=..., revision=..., output_dir=..., episode_index=None, camera_key=...) -> list[Path]` is importable and callable with no arguments at all, and `main()` is a thin argparse wrapper around it. That's the shape #48 agreed on.

**2. The parquet view.** `CREATE VIEW` without `IF NOT EXISTS`, one named file selected by role and not by glob order, path escaped the way `catalog.py` does it. The nondeterminism is gone.

**3. The cross-check reads real PTS.** `_get_video_pts_times` pulls `packet=pts_time` from ffprobe, and the comparison is `abs(parquet_ts - video_pts) > 0.050` against those values, with a count-mismatch guard before it. That's the check we designed, not a tautology against the index formula.

Also fixed, all verified: `ffmpeg_version` carries the real first line of `ffmpeg -version`; `/action` has its own schema text instead of reusing `position`; `_extract_source_videos` points at the download directory instead of copying the corpus; fps is probed once and threaded through instead of living in a constant and a probe; the fault-injection scaffolding is gone entirely.

### The fixup I pushed

**Dead code, dead because of the fix I asked for.** Inlining the manifest left `_load_manifest` with no callers, and the five `_require_*` JSON validators existed only to serve it. Removed both, plus `examples/lerobot/manifest.json`, which nothing reads any more. 103 lines of `prepare.py` and the file. Worth internalising the general move: when you replace how something is fed, grep for what used to feed it.

**One unreachable branch.** `video_pts = pts_times[i] if i < len(pts_times) else (i / fps)` can't take the fallback, because the count check above already raised. Left as-is it reads like the index formula is still a legitimate source, which is the confusion blocker 3 was about. Now it's just `pts_times[i]`.

**The WSL2 guide was orphaned.** `docs/how-to/wsl2.md` came here after #105 was auto-closed by the one-open-PR cap, and it lost the `docs/README.md` link it had in that PR, so the guide existed with nothing pointing at it. Relinked under How-to guides. Folding it into this PR was a reasonable response to the cap, and two unrelated changes in one branch is how a link like that goes missing.

Validated after the fixup: `ruff check`, `ruff format --check`, `ty check` clean; 401 passed, 3 skipped; lychee across `docs/README.md`, `docs/how-to/wsl2.md`, and `examples/README.md` reports 0 errors.

### Two things I couldn't verify, and one to watch

I still can't run the converter without the pusht download, so `doctor` CONFORMING and the measurement list remain your evidence. Two specific things rest on that, and you're the only one who has checked them:

**PTS origin.** The comparison assumes the sliced clip's PTS restart near zero, matching the parquet column's per-episode zero base. With `-ss` before `-i` and `-c copy` that's usually true, but if a dataset ever yields absolute PTS the check silently compares two different clocks and passes everything. Worth an assertion that `pts_times[0]` is near zero, so the assumption fails loudly.

**Keyframe-first.** Your slice uses `-c copy`, which can begin mid-GOP, and the transcode afterwards re-keyframes it, so conformance holds. That's luck and not design, and #96 is why it matters: the transform validates pass-through video per message and can't see a stream that starts mid-GOP. You're safe because you transcode. Anyone reusing `_split_h264_by_aud` for a pass-through path wouldn't be.

`_expand_episode_plan` still round-robins across `sources`, which for pusht is a parquet, an mp4, and a meta parquet, so `source_member` for episode 1 is a parquet file. Nothing downstream reads it, so it's harmless, and I left it instead of redesigning the plan structure in a fixup. If you touch this file again, that's the thing to simplify.

#105 can be reopened whenever you like now this is merged, though its content is already here.

Sagar-024:
Thanks for the detailed review and the fixup, and for relinking the WSL2 
guide. The three blockers made the converter genuinely better, especially 
the cross-check reading real PTS instead of the index formula.

I'll get a small follow-up PR up shortly that asserts `pts_times[0]` is 
near zero so the PTS origin assumption fails loudly. 

Also sending you a quick note via email about taking on some contract work, 
if you have bandwidth to discuss it.

PR #104, opened 2026-08-22
feat(steps): derive stable identity for functools.partial steps

### LINKED ISSUE #98: Step identity: functools.partial is refused, though its identity is fully inspectable
### What happens

Registering a `functools.partial` as a step is refused, even though a partial is one of the few callables whose identity is completely inspectable.

On `main` (`2f07aa0`):

```python
import functools, hflow
from hflow.checks import action_rate

app = hflow.App("p", data_root=Path("/tmp/p"))
bound = functools.partial(action_rate, topics=["/joint_states"])
app.check(name="bound")(bound)
```

```
ValueError: cannot derive a stable implementation identity for
functools.partial(<function action_rate at 0x...>, topics=['/joint_states']);
register it with version='...'
```

Meanwhile everything needed to identify it is right there:

```
func=action_rate  args=()  keywords={'topics': ['/joint_states']}
```

### Why it happens

`_callable_implementation_identity` (`src/hflow/steps.py:246`) handles two shapes: a plain function or method (`inspect.getsource` on it), or an instance of a callable class (`inspect.getsource` on `type(function).__call__`). A `functools.partial` is neither. `type(p).__call__` is a C slot wrapper, so `getsource` raises, there is no `__code__` to fall back to, and it lands on the raise at `src/hflow/steps.py:265`.

The refusal is the correct default for a genuinely opaque callable. A partial is not opaque: it is a transparent wrapper whose behavior is entirely determined by `func`, `args`, and `keywords`, each of which this module already knows how to identify.

### Why it is worth fixing

`functools.partial` is the obvious way to bind a parameterized check, and it is the first thing a user reaches for after hitting the registration error from #86/#93 on `action_rate(episode, *, topics)`. Today they hit a second wall and are told to hand-write a `version=` string, which is exactly the kind of manual version bookkeeping the automatic identity machinery exists to eliminate: a hand-written version does not change when the bound arguments change, so two differently-parameterized checks silently share a version.

That last part is the real bug. The current advice ("register it with `version='...'`") actively produces wrong version identity for this case.

### Definition of done

`app.check()` and `app.enrichment()` accept a `functools.partial` (and `functools.partialmethod` if it falls out naturally) and derive a version identity that:

- **changes when the wrapped function's source changes** — recurse into `partial.func` through the existing `_callable_implementation_identity`, so a nested partial works too;
- **changes when any bound argument changes** — run each of `partial.args` and each value in `partial.keywords` through `_stable_version_identity_value` (`src/hflow/steps.py:310`), which already handles enums, bytes, paths, modules, types, code objects and functions;
- **is order-insensitive for keywords and order-sensitive for positional args**, which is what calling semantics mean;
- **still refuses, with the existing message, when a bound argument is itself opaque.** `partial(f, client=SomeOpaqueClient())` must keep raising. Transparency of the wrapper does not make its contents transparent, and this is the case most likely to be got wrong.

Two properties worth asserting directly, because they are the point:

```
version(partial(f, topics=["/a"])) != version(partial(f, topics=["/b"]))
version(partial(f, topics=["/a"])) == version(partial(f, topics=["/a"]))
```

The second one is stability across processes: it must not depend on `id()`, memory addresses, or dict iteration order.

### Pattern to copy

`_stable_version_identity_value` at `src/hflow/steps.py:310` for the value-by-value treatment, and its function branch at `src/hflow/steps.py:336` for how it already composes a callable's identity into a larger structure. `_code_identity` (`src/hflow/steps.py:270`) shows the shape these identity dicts take. Tests: `tests/test_processing_regressions.py` has the step-version tests (`test_step_version_includes_captured_configuration` is the closest precedent).

### Notes

Related to #86 and #93, which moved unsatisfiable check signatures to registration time. Those two established that whether a check can run is knowable at registration; this one is about whether it can be *identified* once it can run.

Not a starter issue: it touches version identity, which is stamped into stored provenance, so getting it wrong is worse than leaving it unfixed. Comment here and I will assign it.

ayam04:
Fixes #98

A `functools.partial` was refused as an unidentifiable opaque callable, pushing users to hand-write `version='...'` — which silently stops tracking bound-argument changes (two differently-parameterized checks share a version). A partial is fully inspectable: `func`, `args`, `keywords`.

**Change (src/hflow/steps.py):**
- `_callable_implementation_identity`: partial branch composing the wrapped callable's identity with stable values of `args` and `keywords` (dict path already sorts keys)
- `_callable_behavior_configuration`: partial branch carrying the same bound values, so binding changes alter the version
- registration guard from #86/#93 already works: `inspect.signature` resolves partials (verified: `([], True)` + registers)

**Verified:**
- the issue's exact repro (register `functools.partial(action_rate, topics=[...])`) now registers with a version
- versions differ when `topics` changes; stable across repeated calls
- tests: 3 added (partial registers, versions differ on rebinding, registration accepts partial)

**Validation:**
- ruff check: clean; ruff format --check: clean; ty check: clean
- pytest -q: 392 passed, 3 skipped; only pre-existing `test_ffmpeg.py` env failures (identical on clean main)

### DIFF
--- a/src/hflow/steps.py
+++ b/src/hflow/steps.py
@@ -8,6 +8,7 @@ on a check registered with ``critical=True``, a False verdict quarantines the
 episode (a tag, never a deletion) and skips its downstream steps.
 """
 
+import functools
 import hashlib
 import inspect
 import json
@@ -248,6 +249,12 @@ def _callable_implementation_identity(
     *,
     allow_opaque: bool = False,
 ) -> VersionIdentityValue:
+    if isinstance(function, functools.partial):
+        return {
+            "partial_func": _callable_implementation_identity(function.func),
+            "partial_args": _stable_version_identity_value(function.args),
+            "partial_keywords": _stable_version_identity_value(function.keywords),
+        }
     source_target: object = function
     if not inspect.isfunction(function) and not inspect.ismethod(function):
         source_target = type(function).__call__
@@ -288,7 +295,10 @@ def _callable_behavior_configuration(
     function: Callable[..., object],
 ) -> VersionIdentityValue:
     configuration: dict[str, VersionIdentityValue] = {}
-    if inspect.isfunction(function) or inspect.ismethod(function):
+    if isinstance(function, functools.partial):
+        configuration["partial_args"] = _stable_version_identity_value(function.args)
+        configuration["partial_keywords"] = _stable_version_identity_value(function.keywords)
+    elif inspect.isfunction(function) or inspect.ismethod(function):
         closure_variables = inspect.getclosurevars(function)
         configuration["nonlocals"] = _stable_version_identity_value(closure_variables.nonlocals)
         configuration["globals"] = _stable_version_identity_value(closure_variables.globals)
--- a/tests/test_processing_regressions.py
+++ b/tests/test_processing_regressions.py
@@ -362,7 +362,10 @@ def test_step_version_still_refuses_a_partial_bound_to_an_opaque_value() -> None
     class OpaqueClient:
         pass
 
-    bound = functools.partial(hflow.checks.action_rate, topics=OpaqueClient())
+    def scored_by_client(episode: hflow.Episode, *, client: object) -> hflow.CheckResult:
+        return hflow.CheckResult(measurements={"ok": client is not None})
+
+    bound = functools.partial(scored_by_client, client=OpaqueClient())
     with pytest.raises(ValueError, match="cannot derive a stable version identity"):
         compute_check_version("opaque", bound, False, frozenset(), None)
 
--- a/tests/test_processing_regressions.py
+++ b/tests/test_processing_regressions.py
@@ -1,5 +1,6 @@
 """Processing, publication, and user-step boundary regressions."""
 
+import functools
 import json
 from collections.abc import Callable
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


promiseeuler (changes requested):
The new `functools.partial` branch does not forward `allow_opaque` to the recursive `_callable_implementation_identity(function.func)` call. This breaks the existing explicit-version escape hatch when a partial wraps an opaque callable.

I reproduced this on commit `29055cc` with Python 3.14:

```python
compute_check_version("opaque", len, False, frozenset(), None, "manual-v1")
# succeeds

compute_check_version(
    "opaque-partial", functools.partial(len), False, frozenset(), None, "manual-v1"
)
# ValueError: cannot derive a stable implementation identity ... register it with version
```

Could the recursive call pass `allow_opaque=allow_opaque` and cover this with a regression test? That preserves the declared-version contract for partials as well as ordinary opaque callables.

kstonekuan (approved):
Thank you @ayam04. Merging. The implementation is exactly right and I changed none of it.

Two things you got right that I want to name, because they are the parts that make this safe rather than just working.

**You handled `_callable_behavior_configuration` as well as the identity.** The issue only asked for the identity function. Adding the partial branch there too is why rebinding actually moves the version instead of merely being recorded somewhere. It would have been easy to fix the visible error and leave the version stale.

**You did not touch the opaque path.** A partial is identifiable only because its bound values are, and the whole change would have been a regression if binding an un-describable object had started producing a version. I verified it still refuses:

```
functools.partial(scored_by_client, client=OpaqueClient())
ValueError: cannot derive a stable version identity for captured value ... register the step with version='...'
```

What I verified beyond the tests, all by hand:

| property | result |
|---|---|
| binding changes the version (name held fixed) | `90a15f2ee7da` vs `931c442493f1` |
| stable across reconstruction | identical |
| keyword order irrelevant | identical |
| nested `partial(partial(f, ...))` | resolves |
| opaque bound value | still refused |

**The fixup, tests only.** `test_step_version_differs_when_partial_bindings_change` computed its two versions under the names `"a"` and `"b"`. The step name is part of the identity, so that assertion passed whether or not the bound arguments were read at all: I confirmed `v("a", x) != v("b", x)` for the same partial. Held the name fixed so only the binding can move it, added the reconstruction-stability assertion so the identity cannot be address-derived, and added the opaque-refusal case the issue called out as the one most likely to be got wrong.

Worth generalizing from: when a test asserts "X changes Y", the thing to check is that nothing else in the fixture could also have changed Y. Two of your three tests were airtight; that one had a second free variable.

Also rebased you onto current `main`, which had moved twice (`5d5cf71`, `0b6b5de`), and resolved a conflict where my commit and yours both appended to the same test file. 405 passed, 3 skipped after the fixup, with `ruff`, `ruff format --check`, and `ty check` clean.

One heads-up on that `main` move: the signature guard from #91/#94 now also covers `app.enrich()` and `app.derive()`, which were structurally identical registration paths that had been left accepting functions the runtime could never call. Your partial work composes with it cleanly, since `inspect.signature` resolves partials, which you had already checked.

`ty` caught one thing in my own added test, worth knowing about: binding an opaque object to `action_rate`'s `topics` parameter fails typechecking because `topics` is `Sequence[str]`. I used a locally defined function taking `client: object` instead. Your choice to test with the real `action_rate` was right for every other case.

PR #110, opened 2026-08-22
tests: make ffmpeg override path portable

### LINKED ISSUE #108: tests: test_env_override_wins hardcodes Linux ffmpeg path on macOS
## Summary

While validating PR #107 on macOS, I found that `test_env_override_wins` hardcodes `/usr/bin/ffmpeg`. The implementation correctly rejects nonexistent override paths, so the test fails on supported macOS installations where FFmpeg is normally elsewhere (for example, Homebrew uses `/opt/homebrew/bin/ffmpeg` on Apple Silicon).

## Environment

- HFlow upstream `main` (verified against the current raw source on 2026-08-22)
- macOS, Apple Silicon
- Python 3.14.2
- uv 0.12.5
- FFmpeg installed and available on `PATH`, but not at `/usr/bin/ffmpeg`

## Steps to reproduce

```bash
uv sync --locked --all-extras
uv run pytest -q tests/test_ffmpeg.py::test_env_override_wins
```

## Expected behavior

The platform-supported test should verify that an existing override path wins without depending on a Linux-specific filesystem location.

## Actual behavior

```text
E hflow.ffmpeg._binary.FfmpegNotFoundError:
E HFLOW_FFMPEG=/usr/bin/ffmpeg does not exist
1 failed
```

The failure is deterministic because `ffmpeg_path()` intentionally checks `Path(override).is_file()`.

## Possible cause

`tests/test_ffmpeg.py:67` sets:

```python
monkeypatch.setenv(FFMPEG_ENV_VAR, "/usr/bin/ffmpeg")
```

That path is not portable to macOS, despite CONTRIBUTING.md listing macOS as a supported native development platform.

## Proposed solution

Create a temporary file with `tmp_path`, set `HFLOW_FFMPEG` to that path, and assert the resolver returns it. The test only exercises path precedence, so it does not need a real executable:

```python
def test_env_override_wins(
    tmp_path: Path,
    monkeypatch: pytest.MonkeyPatch,
    cleared_binary_caches: None,
) -> None:
    override = tmp_path / "ffmpeg"
    override.touch()
    monkeypatch.setenv(FFMPEG_ENV_VAR, str(override))
    assert ffmpeg_path() == override
```

This keeps the test hermetic and platform-independent.

WilliamK112:
## Summary

- replace the Linux-specific `/usr/bin/ffmpeg` fixture with a real temporary file
- keep the test focused on the observable override-precedence contract
- make the regression hermetic across supported Linux and macOS development hosts

Closes #108.

## Validation

- `uv run pytest -q tests/test_ffmpeg.py::test_env_override_wins` — 1 passed
- `uv run ruff check .` — passed
- `uv run ruff format --check .` — 99 files already formatted
- `uv run ty check` — passed
- `uv run pytest -q -k 'not test_contact_sheet_grid_geometry and not test_contact_sheet_max_tiles_sampling and not test_contact_sheet_accepts_apostrophes_in_external_paths and not test_media_stage_records_a_contact_sheet_artifact'` — 398 passed, 6 skipped, 4 deselected

The unfiltered local suite reached 398 passed and 6 skipped, with four unrelated contact-sheet failures because the installed Homebrew FFmpeg build does not provide the `drawtext` filter. The focused regression and all code-quality gates pass on macOS.

AI assistance disclosure: I used OpenAI Codex to inspect, implement, and validate this change, and reviewed the final diff and evidence.

### DIFF
--- a/tests/test_ffmpeg.py
+++ b/tests/test_ffmpeg.py
@@ -63,9 +63,13 @@ def cleared_binary_caches() -> Iterator[None]:
     _clear_all_binary_caches()
 
 
-def test_env_override_wins(monkeypatch: pytest.MonkeyPatch, cleared_binary_caches: None) -> None:
-    monkeypatch.setenv(FFMPEG_ENV_VAR, "/usr/bin/ffmpeg")
-    assert ffmpeg_path() == Path("/usr/bin/ffmpeg")
+def test_env_override_wins(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path, cleared_binary_caches: None
+) -> None:
+    override_ffmpeg = tmp_path / "ffmpeg"
+    override_ffmpeg.touch()
+    monkeypatch.setenv(FFMPEG_ENV_VAR, str(override_ffmpeg))
+    assert ffmpeg_path() == override_ffmpeg
 
 
 def test_env_override_to_missing_file_raises(


promiseeuler (approved):
While validating this fix against #108 on macOS, I checked out commit 959ffd4 and ran the exact regression: python -m pytest -q tests/test_ffmpeg.py::test_env_override_wins (1 passed). I also ran ruff check tests/test_ffmpeg.py and ruff format --check tests/test_ffmpeg.py; both passed. Using tmp_path preserves the test's actual purpose—environment-variable precedence—while removing the Linux-only /usr/bin/ffmpeg assumption. The temporary file also satisfies the production existence check, so this is appropriately scoped.

kstonekuan (approved):
Welcome, and thank you @WilliamK112. Merging.

Exactly the right scope. `ffmpeg_path()` validates the override with `is_file()` and nothing else (`src/hflow/ffmpeg/_binary.py:232`), so `touch()` is faithful to the contract rather than a shortcut around it: the test exercises precedence, which is all it ever claimed to.

What I validated on your branch against current `main` (`9d59310`): `ruff check`, `ruff format --check`, `ty check` clean, 405 passed and 3 skipped. I also checked that the parallel `ffprobe_path()` override tests already build their override under `tmp_path`, so `test_env_override_wins` was genuinely the only one of its kind and this closes the whole class rather than one instance.

Thank you for disclosing the AI assistance, and for checking the issue was unassigned with no overlapping PR before starting. Both are the right habits here. On the second one: our rule is that an assignee reserves an issue and everything else is fair game, so you read it correctly.

Two notes:

**On the four contact-sheet failures you filtered out.** Good instinct to separate them from your change and say so. They are a real gap and @promiseeuler has since filed #111 for exactly that (a Homebrew FFmpeg built without `drawtext`), with a fix in #112. So you were right that it was not you, and it is now tracked.

**One more instance of the same species, if you want it.** `tests/test_video.py:154` falls back to a hardcoded `Path("/usr/bin/ffprobe")` when the sibling of `ffmpeg_path()` is missing. It is milder than the bug you fixed, since on macOS the Homebrew sibling normally does exist so the fallback is not reached, but when it is reached it fails the same way for the same reason. The real fix is to call `ffprobe_path()` from `hflow.ffmpeg._binary`, which already implements the full resolution policy including the sibling lookup. I will file it; comment there if you would like it.

One thing worth knowing for your next contribution: we cap non-collaborators at one open pull request at a time, so a bot will auto-close a second one with a note. Nothing is lost and it can be reopened once the first merges.

PR #107, opened 2026-08-22
refactor: remove dead quarantine state helper

promiseeuler:
## Summary

While reviewing the catalog implementation for issue #64, I found that `latest_quarantine_state()` is a private wrapper with no callers, exports, tests, or documentation. This removes that dead helper and changes nothing else.

Closes #64

## Why

The richer `latest_quarantine()` result is already used by the real consumer. Keeping the unused wrapper adds an extra maintenance surface without providing an API or behavior.

## Validation

- `rg -n "latest_quarantine_state" src tests docs examples README.md` — no matches after the change
- `uv run ruff check` — passed
- `uv run ruff format --check` — passed (97 files already formatted)
- `uv run ty check` — passed
- `uv run pytest -q` — 391 passed, 6 skipped; 7 environment-dependent failures unrelated to this deletion (this macOS environment lacks `/usr/bin/ffmpeg`, Python `lzma`, and FFmpeg's `drawtext` filter)
- `git diff --check` — passed

## Checklist

- [x] No business logic changed, so no outcome-focused test is needed.
- [x] No behavior, flag, format, or requirement changed, so documentation is unchanged.
- [x] I ran Ruff formatting/lint and `ty check`.
- [x] I ran the relevant pytest suite and documented its environment-only failures above.
- [x] I did not add recordings, generated media, credentials, private URLs, or runtime artifacts.
- [x] Stored-data compatibility is unchanged.

Issue #97, opened 2026-08-22
Resample: add policy="slerp" so rotation channels can be interpolated correctly


kstonekuan:
### What we want

A `policy="slerp"` for `hflow.to_grid`, so rotation channels can be resampled correctly instead of being refused.

### Current behavior

`to_grid` supports `"interpolate"` and `"nearest"` (`src/hflow/resample.py:34`). Componentwise linear interpolation of a quaternion is wrong twice over: the result is unnormalized, and across an antipodal sign flip (`q` and `-q` are the same rotation) it interpolates the long way round, through garbage. Rather than silently corrupt rotations, `to_grid` refuses the obvious cases: `_looks_like_quaternion` (`src/hflow/resample.py:130`) matches a 4-column field whose name contains `quat`, `orientation`, or `rotation`, or whose schema name contains `quaternion`, and `to_grid` raises at `src/hflow/resample.py:177`:

```
field 'orientation' of topic '/tf' looks like a quaternion; componentwise
interpolation corrupts rotations (unnormalized, wrong across sign flips).
Use policy='nearest' for now -- slerp-based quaternion alignment is a
tracked follow-up.
```

The comment at `src/hflow/resample.py:124` calls this "a tracked follow-up". It was never actually tracked. This is that issue.

So today a rotation channel can only be resampled with `nearest`, which is lossy in a way that matters: at 10 Hz source and a 50 Hz grid, `nearest` produces a staircase, and any derived signal computed from it inherits the steps.

### Definition of done

`to_grid(channel, rate_hz, policy="slerp", field=...)` returns a `DerivedSeries` of correctly interpolated unit quaternions.

The semantics to implement, and the reason each clause exists:

- **Hemisphere alignment first.** Walk the source quaternions in order and negate any sample whose dot product with the previous one is negative. Without this, a recording that flips sign midway interpolates the long way around. This is the clause that makes slerp worth having over `interpolate` plus a normalize.
- **Slerp between the two bracketing source samples** for each grid point, by the shortest arc.
- **Fall back to normalized linear interpolation when the two samples are nearly parallel** (`|dot|` above roughly `1 - 1e-7`), where `sin(theta)` underflows and the slerp formula loses precision. Say what threshold you picked and why in a comment; the exact value matters less than it being deliberate.
- **Output quaternions are unit norm.** Assert it in a test rather than trusting the algebra.
[...]

PR #112, opened 2026-08-22
Handle FFmpeg builds without drawtext

### LINKED ISSUE #111: ffmpeg: contact_sheet fails when a font exists but drawtext is unavailable
## Summary

While validating HFlow's contact-sheet tests on a supported macOS development host, I found that `contact_sheet()` fails with a standard Homebrew FFmpeg build when a usable system font exists but the FFmpeg binary was built without the `drawtext` filter. The function's documented fallback only handles a missing font, so it attempts timestamp burn-in and aborts instead of producing an untimestamped sheet.

## Environment

- HFlow `main` at `9d59310`
- macOS 26.4, Apple Silicon
- Python 3.14.2
- FFmpeg 8.0.1 from Homebrew: `/opt/homebrew/bin/ffmpeg`
- `ffmpeg -filters` contains `scale` and `tile`, but not `drawtext`

## Steps to reproduce

```bash
python -m pytest -q tests/test_ffmpeg.py::test_contact_sheet_grid_geometry
```

The same environment has a usable `/System/Library/Fonts/Helvetica.ttc`, so `_find_usable_font_file()` succeeds.

## Expected behavior

`contact_sheet()` should produce the JPEG without timestamp burn-in and return `timestamps_burned=False`, matching the documented graceful fallback when burn-in is unavailable.

## Actual behavior

```text
[AVFilterGraph ...] No such filter: 'drawtext'
Error opening output files: Filter not found
RuntimeError: ffmpeg contact sheet failed ...
```

The focused regression fails consistently.

## Possible cause

`contact_sheet()` enables `_drawtext_filters(...)` whenever `_find_usable_font_file()` returns a path. It does not verify that the selected FFmpeg binary exposes the `drawtext` filter. On macOS the system font check succeeds even when Homebrew's FFmpeg lacks drawtext.

## Proposed solution

Detect and cache whether the resolved FFmpeg binary supports `drawtext`, and add timestamp filters only when both a usable font and the filter are available. Otherwise retain the existing no-burn-in path and set `timestamps_burned=False`.

A focused regression can simulate "font present, drawtext absent" and assert that the sheet is still generated with `timestamps_burned=False`.

promiseeuler:
## Summary

While testing HFlow's contact-sheet helper on macOS, I found that a usable system font could make it attempt timestamp burn-in even when the selected FFmpeg binary does not provide the `drawtext` filter. This change detects that capability and gracefully produces an untimestamped sheet when it is unavailable.

Fixes #111.

## Why

The documented fallback covered a missing font but not a missing FFmpeg filter. Standard Homebrew FFmpeg builds can expose `scale` and `tile` without `drawtext`, causing the whole contact-sheet operation to fail instead of returning `timestamps_burned=False`.

The capability result is cached per resolved FFmpeg path, and the existing binary is reused for the actual render.

## Validation

`uv sync --locked --all-extras`

`uv run pytest -q tests/test_ffmpeg.py::test_contact_sheet_grid_geometry tests/test_ffmpeg.py::test_contact_sheet_without_drawtext_still_produces_sheet tests/test_ffmpeg.py::test_contact_sheet_max_tiles_sampling` — 3 passed

`uv run ruff check src/hflow/ffmpeg/_contact_sheet.py tests/test_ffmpeg.py` — passed

`uv run ruff format --check src/hflow/ffmpeg/_contact_sheet.py tests/test_ffmpeg.py` — passed

`uv run ty check` — passed

## Checklist

 Outcome-focused tests added for the changed business logic.

 Documentation updated for the changed fallback behavior.

 Ruff lint, Ruff formatting check, and `ty check` passed.

 Relevant pytest cases passed.

 No recordings, generated media, credentials, private URLs, or runtime artifacts were added.

 Stored-data compatibility is unchanged.

### DIFF
--- a/src/hflow/ffmpeg/_contact_sheet.py
+++ b/src/hflow/ffmpeg/_contact_sheet.py
@@ -5,9 +5,9 @@ vision tokens (one image instead of N). One of the two VLM-adjacent helpers
 we ship -- there is deliberately no bundled VLM client.
 
 Implementation: a single ffmpeg invocation over the frame files
-(``concat`` + ``scale`` + ``tile``). Timestamp burn-in uses ``drawtext``
-when a usable font is found (``fc-match`` or common font paths); if none is
-available the sheet is still produced without burn-in and
+(``concat`` + ``scale`` + ``tile``). Timestamp burn-in uses ``drawtext`` when
+both the filter and a usable font are available (``fc-match`` or common font
+paths); otherwise the sheet is still produced without burn-in and
 ``ContactSheet.timestamps_burned`` is False -- callers can pass timestamps in
 the prompt instead. If more frames are given than ``max_tiles``, frames are
 sampled evenly and the drop is reported on the result, never silent.
@@ -19,6 +19,7 @@ import subprocess
 import tempfile
 from collections.abc import Sequence
 from dataclasses import dataclass
+from functools import cache
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -62,6 +63,22 @@ def _find_usable_font_file() -> Path | None:
         if common_path.is_file():
             return common_path
     return None
+
+
+@cache
+def _ffmpeg_supports_drawtext(ffmpeg_binary: Path) -> bool:
+    completed = subprocess.run(
+        [str(ffmpeg_binary), "-hide_banner", "-filters"],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if completed.returncode != 0:
+        return False
+    return any(
+        len(columns := line.split()) > 1 and columns[1] == "drawtext"
+        for line in completed.stdout.splitlines()
+    )
 
 
 def _evenly_sampled_indices(total_count: int, max_count: int) -> list[int]:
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM. Merging.

Computing `timestamps_burned` once instead of re-deriving it at the bottom is the part that fixes the shape and not just the bug, and `@cache` keyed on the binary path rather than a module flag is the right call. `assert font_file is not None` matches the existing narrowing convention (`batching.py:80`, `episode.py:128`), so no change wanted.

Validated against `main` (`9d59310`): gates clean, 406 passed, 3 skipped. Also confirmed the new predicate returns True on a build that does have `drawtext`, since silently disabling burn-in for everyone was the failure mode worth ruling out.

Three things you should know:

**#108 was fixed by @WilliamK112 in #110.** You filed it and specified the fix that landed. It was unassigned and only an assignee reserves an issue, so they were entitled to it, but you found it. Comment on anything you intend to take and I will assign it.

**#107 can be reopened**, it was closed by the one-open-PR cap, not on merit.

**#96 and #97 are both unclaimed and your analysis on each is useful.** You answered the ordering question on #96 from the source, and you found an error in my #97 spec, which I have corrected there. Say the word on either.

Issue #102, opened 2026-08-22
cli: --pipeline pointed at a directory reports 'No such file or directory'


chiruu12:
`--pipeline` pointed at a directory reports that the path does not exist, when it does.

```
$ ls -d ./pipelines
./pipelines
$ hflow up --pipeline ./pipelines --data-root ./data --bundle-dir ./rt
up: rendering the runtime bundle at rt
up: [Errno 2] No such file or directory: 'pipelines'
$ echo $?
2
```

Every site tests `is_file()` and raises `ENOENT` on the false branch, so a path that
exists but is the wrong kind gets the message for a path that is absent. Sites:
`_bundle.py:861` and `:896`, `_deploy.py:153` and `:169` (verified on `2f99071`; #113 shifted the second site in each file by one). That set covers `--requirements` as well as `--pipeline`, one site each per file, so both flags are fixed by the same change.

Exit code and the absence of a traceback are correct (#90, #95). This is the message
only.

The accurate exception is `IsADirectoryError`, and that is the part that needs a
decision rather than a patch. It subclasses `OSError`, not `FileNotFoundError`, so
every existing handler stops catching it: `cli.py:467`, `:525`, `:566`, `:588`, `:632`,
`:684`, `:698`, `:710`, plus `app.py:1092` and `_compose.py:49`, and any downstream
`except FileNotFoundError` outside this repo. The break only shows up on directory
input, which is the input nobody tests.

The cheap alternative keeps the class and fixes the text:

```python
if path.is_dir():
    raise FileNotFoundError(errno.EISDIR, os.strerror(errno.EISDIR), str(path))
```

That prints `[Errno 21] Is a directory: '/x/y'`, which is accurate, while the class
stays wrong in a way no handler notices.

Not up for grabs yet, and please do not open a PR against it. Picking between those two
is a public-API call, not a review comment, and it is open in #100. Once the answer is
in, the change itself is small.

kstonekuan:
Unblocked: #100 is answered, and the answer is the cheap version you leaned toward. Add the `is_dir()` branch raising `FileNotFoundError(errno.EISDIR, os.strerror(errno.EISDIR), str(path))` at the four sites, so the text is `[Errno 21] Is a directory: '/x/y'` and every existing handler keeps working.

Worth a short comment at those sites recording that the class is deliberately not `IsADirectoryError` and why, so nobody 'fixes' it later without knowing it breaks ten `except FileNotFoundError` sites.

[...]

PR #113, opened 2026-08-22
fix(storage): raise errno FileNotFoundError at two bare sites

### LINKED ISSUE #101: storage: LocalStorageRoot.fetch raises a bare path where fetch_uri raises errno
Two raise sites still construct `FileNotFoundError` with a bare path, so `str(error)`
is the path and nothing else. This is the pattern #26 was closed for, and it survived
in `storage.py:344` and `testing.py:514`. #95 fixed the other four.

The symptom is not just a thin message. `StorageRoot.fetch` has two implementations
that answer the same failure three different ways:

- `LocalStorageRoot.fetch` (`storage.py:344`) raises `FileNotFoundError(local_file)`,
  so the text is a path.
- `BucketStorageRoot.fetch` (`storage.py:510`) does not raise at all. `obstore.head`
  raises and the text is obstore's, which this repo does not control.
- `fetch_uri` (`storage.py:656`), one screen below and named for the same job, uses
  the three-argument errno form and prints
  `[Errno 2] No such file or directory: '/x/y'`.

Same missing object, different message depending on which root the data path resolved
to. The bare form also leaves `error.filename` and `error.errno` as `None`, so a caller
that wants the path has to parse the message.

Reproduce:

```python
>>> from pathlib import Path
>>> from hflow.storage import LocalStorageRoot, fetch_uri
>>> LocalStorageRoot(Path("/tmp/root")).fetch("missing.mp4")
FileNotFoundError: /tmp/root/missing.mp4
>>> fetch_uri("/tmp/root/missing.mp4")
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/root/missing.mp4'
```

`error.filename` is `None` in the first and `/tmp/root/missing.mp4` in the second.

Not up for grabs yet, and please do not open a PR against it. The fix is one line per
site, but which line depends on a convention the repo has not picked: six other sites
raise an explanatory sentence on purpose and would get worse under a blanket errno rule.
That is being settled in #100. Once there is an answer here, this is a small change and
I am happy for anyone to take it.

chiruu12:
Closes #101.

Two sites still built `FileNotFoundError` with a bare path, so `str(error)` was the
path alone and `error.errno` / `error.filename` were `None`. Per the rule settled in
#100, both are cases where the path is the whole story, so both move to the
three-argument errno form.

- `LocalStorageRoot.fetch` (`storage.py:344`) now matches `fetch_uri` at
  `storage.py:656`, which already raised the errno form. Same missing object, same
  message, whichever entry point the caller used.
- `write_video_episode` (`testing.py:514`), same change.

`BucketStorageRoot.fetch` is untouched, as agreed on the issue. obstore raises there
and its text is not ours to wrap.

A test per site pins the errno text rather than just the path:
`errno == ENOENT`, `filename` set to the full path, and
`No such file or directory` present in the message. Both fail on the unfixed code
with `assert None == 2`.

Local run: `399 passed, 6 skipped`, plus 5 pre-existing failures in `test_ffmpeg.py`
and `test_run_profiles.py` that reproduce identically on unmodified main here. They
are a local ffmpeg build problem, not this change. `ruff check`, `ruff format --check`
and `ty check` are clean.

I will open #102 separately once this lands.

### DIFF
--- a/src/hflow/storage.py
+++ b/src/hflow/storage.py
@@ -341,7 +341,7 @@ class LocalStorageRoot:
         """The local file at ``relative`` (it already lives here)."""
         local_file = self.path / _validated_relative_key(relative)
         if not local_file.is_file():
-            raise FileNotFoundError(local_file)
+            raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), str(local_file))
         return local_file
 
     def uri_for(self, relative: str) -> str:
--- a/src/hflow/testing.py
+++ b/src/hflow/testing.py
@@ -33,8 +33,10 @@ Implementation notes:
   wall-clock reads.
 """
 
+import errno
 import json
 import math
+import os
 import random
 import struct
 import subprocess
@@ -511,7 +513,7 @@ def write_video_episode(
     _validate_video_episode_spec(resolved_spec)
     source_video_path = Path(source_video)
     if not source_video_path.is_file():
-        raise FileNotFoundError(source_video_path)
+        raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), str(source_video_path))
     output_path = Path(output)
     output_path.parent.mkdir(parents=True, exist_ok=True)
 
--- a/tests/test_storage.py
+++ b/tests/test_storage.py
@@ -10,6 +10,7 @@ A live object-store integration test runs only when
 ``HFLOW_TEST_BUCKET_URL`` is set (e.g. ``gs://bucket/tmp-prefix``).
 """
 
+import errno
 import os
 from pathlib import Path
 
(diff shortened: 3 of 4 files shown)


kstonekuan (approved):
Thank you @chiruu12. Merging, no changes needed.

The tests are the part that makes this more than a two-line diff. Asserting `errno`, `filename`, and the message text separately pins the three things the errno form actually buys, and it is why the change cannot be silently undone later by someone "simplifying" back to the one-argument call. Checking that both new tests fail on unfixed code with `assert None == 2` is the right evidence to include: it proves the tests bind to the change rather than passing incidentally.

Leaving `BucketStorageRoot.fetch` alone was correct and I am glad you did not take the opportunity to normalise it. obstore's message is not ours to dress up, and a wrapper there would have made the two paths look identical while behaving differently on everything except the happy path.

Validated on your branch against current `main` (`9d59310`): `ruff check`, `ruff format --check`, `ty check` clean, 407 passed and 3 skipped.

#102 is still unassigned and free, per your own request. Say the word and it is yours.

Worth recording what this sequence looked like from my side, since it is the pattern I would like more of: you hit a bare-path message in #90, noticed it was one instance of something broader, opened #100 rather than guessing, categorised all twelve raise sites with the tradeoffs, filed #101 and #102 marked do-not-take so nobody duplicated the work while the question was open, and then implemented the answer exactly as settled. That is the whole loop, and the only part I did was say which of your two options I preferred.

PR #117, opened 2026-08-23
docs: correct OpenAI vision request count

### LINKED ISSUE #73: Docs: OpenAI vision how-to miscounts API requests and its snippet drops `version=`
The OpenAI vision how-to miscounts the example's API requests, and its snippet
omits the explicit `version=` that the same page recommends.

### The request count is wrong

`docs/how-to/call-openai-vision.md:94-98` says the example makes one model
request per episode. The example ships two checks, and each check makes its own
`client.responses.create` call:

- `describe_activity` at `examples/openai_vision/pipeline.py:42-76`
- `hand_visibility` at `examples/openai_vision/pipeline.py:79-119`

An episode therefore makes one request per check, or two requests for the
shipped example. The `fps=0.5` and `max_tiles=12` values in the same paragraph
are correct.

### The snippet omits `version=`

The snippet at `docs/how-to/call-openai-vision.md:67` uses
`@app.check(uses="vision")`. The corresponding example at
`examples/openai_vision/pipeline.py:42` passes
`version="responses-contact-sheet-v1"`. The prose at
`docs/how-to/call-openai-vision.md:96-98` also tells you to version a step when
external model configuration cannot be hashed from the function.

Add an explicit `version=` to the snippet. Keeping `uses="vision"` instead of
introducing the example's alias constant is fine.

## Pattern to copy

Copy the decorator shape from `examples/openai_vision/pipeline.py:42` and keep
the concise explanatory style of the surrounding how-to.

## Definition of done

- The request-count paragraph says one request per check and two per episode
  for the shipped example.
- The snippet passes an explicit `version=`.
- The existing frame sampling and tile-cap values remain unchanged.

## Validation

```bash
```

promiseeuler:
## Summary

While reviewing the current OpenAI vision how-to against its runnable example, I found two documentation mismatches. This change:

- adds the example’s explicit `version="responses-contact-sheet-v1"` to the guide’s decorator;
- clarifies that the shipped example makes one model request per check, or two per episode, while retaining the documented 0.5 FPS sampling and 12-tile cap.

## Why

The runnable example has two checks and each independently calls `client.responses.create`. Describing that as one request per episode understates API usage, while omitting `version=` conflicts with the guide’s advice to version external model configuration.

Fixes #73.

## Validation

- `git diff --check` — passed
- Repository `lychee --no-progress --include-fragments ...` command — 260 total links, 257 OK, 3 expected exclusions, 0 errors

## Checklist

- [x] No business logic changed, so an outcome-focused test is not applicable.
- [x] I updated the affected documentation.
- [x] Python formatting, typing, and pytest gates are not applicable to this Markdown-only change.
- [x] I ran the repository’s prescribed Markdown link check.

### DIFF
--- a/docs/how-to/call-openai-vision.md
+++ b/docs/how-to/call-openai-vision.md
@@ -64,7 +64,7 @@ same file can point at a different endpoint without editing it. Inside the
 step, the endpoint remains an ordinary OpenAI SDK option:
 
 ```python
-@app.check(uses="vision")
+@app.check(uses="vision", version="responses-contact-sheet-v1")
 def describe_activity(episode: hflow.Episode) -> hflow.CheckResult:
     client = OpenAI(
         api_key=os.environ["OPENAI_API_KEY"],
@@ -92,10 +92,10 @@ uv run --extra openai python examples/openai_vision/pipeline.py
 ```
 
 Pass an MCAP path to use your own episode. The example samples frames at 0.5
-FPS, caps the request at 12 tiles, and makes one model request per episode. Those
-choices are part of the measurement definition: change them deliberately and
-version the step explicitly when external model configuration cannot be hashed
-from the function.
+FPS, caps each request at 12 tiles, and makes one model request per check: two
+requests per episode for the two checks it ships. Those choices are part of the
+measurement definition: change them deliberately and version the step explicitly
+when external model configuration cannot be hashed from the function.
 
 ## Keep model output as evidence
 


kstonekuan (approved):
LGTM. Merging.

Verified against the example: two checks, each with its own `client.responses.create`, `fps=0.5` and `max_tiles=12` unchanged, and `OPENAI_ENDPOINT_ALIAS = "vision"` so the snippet's `uses="vision"` stays truthful. Copying `describe_activity`'s version string rather than `hand_visibility`'s is the right one for that snippet. Lychee here matches your numbers: 260 links, 0 errors.

"caps each request at 12 tiles" instead of "the request" was not in the issue and is the more correct wording once there are two.

#107 is still sitting closed from the PR cap and can be reopened whenever.

PR #119, opened 2026-08-23
fix: use ffprobe_path() instead of hardcoded /usr/bin/ffprobe in test_video.py

### LINKED ISSUE #115: tests: test_video.py falls back to a hardcoded /usr/bin/ffprobe instead of calling ffprobe_path()
### Current behavior

`tests/test_video.py` resolves ffprobe by hand, with a hardcoded Linux path as the fallback:

```python
def _ffprobe_video_stream_fields(mp4_path: Path) -> dict[str, str]:
    ffprobe_binary = ffmpeg_path().with_name("ffprobe")
    if not ffprobe_binary.is_file():
        ffprobe_binary = Path("/usr/bin/ffprobe")
```

This is the same species as #108, which #110 fixed for `test_env_override_wins`, but milder: on macOS with Homebrew, `ffmpeg_path()` falls through to the PATH binary whose sibling `ffprobe` normally does exist, so the fallback is usually not reached. When it is reached on a host without `/usr/bin/ffprobe`, the `subprocess.run(..., check=True)` below raises `FileNotFoundError` with the Linux path in it, which tells the reader nothing true about their machine.

### Why it should just call the resolver

`hflow.ffmpeg._binary.ffprobe_path()` already implements the whole policy, and it is strictly better than the two lines above:

1. `HFLOW_FFPROBE` override if set,
2. the sibling of a user-pinned `ffmpeg` (exactly what the test hand-rolls),
3. the pinned Linux build,
4. PATH with the pinning warning,

and it raises `FfprobeNotFoundError` with an actionable message rather than failing on a path nobody chose. The test is reimplementing step 2 and substituting a guess for steps 1, 3, and 4.

### Definition of done

`_ffprobe_video_stream_fields` calls `ffprobe_path()` and no test contains a hardcoded absolute binary path. After the change this should return nothing:

```
rg -n '/usr/bin|/usr/local/bin|/opt/homebrew' tests/
```

Note `ffprobe_path()` is `lru_cache`d, so if a test in the same session repoints `HFLOW_FFPROBE` it needs the existing `cleared_binary_caches` fixture, the same way `tests/test_ffmpeg.py` handles it. `tests/test_video.py` does not currently manipulate those variables, so this is a note rather than a required change.

### Pattern to copy

`tests/test_ffmpeg.py` imports `ffprobe_path` and uses it directly; #110 is the precedent for removing the hardcoded path, and its `tmp_path` approach is the precedent for any override a test needs to fabricate.

### Validation

```
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_video.py
uv run pytest -q
```

`tests/test_video.py` needs ffmpeg on `PATH`; those tests failing without it is pre-existing.

### Context

Follow-up to #108 and #110. @WilliamK112 found this class of bug worth finishing and has first refusal; comment here and I will assign it. Otherwise it is open to anyone.

kragent66-glitch:
Fixes #115

## Problem
tests/test_video.py reimplements ffprobe resolution with a hardcoded Linux path (/usr/bin/ffprobe) as fallback, missing the HFLOW_FFPROBE override and proper error handling from ffprobe_path().

## Solution
Replaced the manual resolution logic in _ffprobe_video_stream_fields() with a single call to ffprobe_path(), following the pattern established in PR #110.

## Changes
- Added ffprobe_path to imports from hflow.ffmpeg
- Replaced 3 lines of manual resolution with ffprobe_path()

## Verification
- All 9 tests in tests/test_video.py pass
- No hardcoded binary paths remain in tests
- Quality checks pass (ruff check, ruff format)
- Note: ty check has pre-existing errors about missing pyarrow module unrelated to this change

### DIFF
--- a/tests/test_video.py
+++ b/tests/test_video.py
@@ -6,7 +6,7 @@ from pathlib import Path
 
 import pytest
 
-from hflow.ffmpeg import ffmpeg_path
+from hflow.ffmpeg import ffmpeg_path, ffprobe_path
 from hflow.video import (
     AccessUnit,
     VideoEncodeError,
@@ -151,9 +151,7 @@ def test_png_input_codec_upholds_guarantees(tmp_path: Path) -> None:
 
 def _ffprobe_video_stream_fields(mp4_path: Path) -> dict[str, str]:
     """Read codec/profile/decoded-frame-count fields from the first video stream."""
-    ffprobe_binary = ffmpeg_path().with_name("ffprobe")
-    if not ffprobe_binary.is_file():
-        ffprobe_binary = Path("/usr/bin/ffprobe")
+    ffprobe_binary = ffprobe_path()
     completed = subprocess.run(
         [
             str(ffprobe_binary),


kstonekuan (approved):
Welcome, and LGTM. Merging.

Exactly the fix the issue asked for, and you kept the `ffmpeg_path` import because line 33 still uses it, which is the easy thing to get wrong when swapping a resolver. The definition-of-done grep comes back empty, so no hardcoded binary path is left in `tests/`.

Validated against `main` (`2f99071`): `ruff check`, `ruff format --check`, `ty check` all clean, 645 passed and 3 skipped.

One correction worth having for next time: the pyarrow errors you saw are not pre-existing, they are a missing extra. `uv sync --locked --all-extras` installs it and `ty check` then passes clean. Worth re-syncing with that flag before deciding a failure is upstream, since it also makes `pytest` collect the arrow tests you were otherwise skipping.

Also: we cap non-collaborators at one open pull request at a time, so a bot will auto-close a second one with a note rather than losing it.

PR #120, opened 2026-08-23
fix(cli): say 'Is a directory' when a path flag names a dir

### LINKED ISSUE #102: cli: --pipeline pointed at a directory reports 'No such file or directory'

chiruu12:
Closes #102.

A path flag pointing at a directory reported that the path does not exist. Each site
tested `is_file()` and raised `ENOENT` on the false branch, so an existing path of the
wrong kind got the message meant for an absent one.

Adds an `is_dir()` branch ahead of each `is_file()` check at the four sites, raising
`FileNotFoundError(errno.EISDIR, os.strerror(errno.EISDIR), str(path))`. The message
becomes `[Errno 21] Is a directory: '/x/y'`.

The class stays `FileNotFoundError` on purpose, per the answer in #100, and each site
carries a comment saying so. While writing that comment I checked the handler count in
the issue text and it is stale: the refactor left one handler in `cli.py`, not eight.
The current set is nine sites in this package (`cli.py`, `app.py` x2, `_compose.py`,
`catalog.py`, `curation.py`, `storage.py`, `workspace.py`), and the comment names those
rather than the old line numbers, which will go stale again.

Five tests, one per raise site plus the issue's repro through the CLI. Reverting only the
two source files and keeping the tests fails all five, the CLI one with the exact symptom
from the issue:

```
>       assert f"deploy: [Errno 21] Is a directory: '{a_directory}'" in streams.err
E       assert "deploy: [Errno 21] Is a directory: '...' in "deploy: [Errno 2] No such file or directory: '...'"

FAILED tests/test_runtime_bundle.py::test_pipeline_directory_says_is_a_directory
FAILED tests/test_runtime_bundle.py::test_requirements_directory_says_is_a_directory
FAILED tests/test_runtime_deploy.py::test_pipeline_directory_says_is_a_directory
FAILED tests/test_runtime_deploy.py::test_requirements_directory_says_is_a_directory
FAILED tests/test_runtime_cli.py::test_deploy_pipeline_directory_says_is_a_directory
5 failed, 114 deselected in 0.52s
```

Gates on the full tree: `ruff check` passed, `ruff format --check` clean,
`ty check` passed, `pytest -q` 647 passed 6 skipped.

### DIFF
--- a/src/hflow/runtime/_bundle.py
+++ b/src/hflow/runtime/_bundle.py
@@ -857,11 +857,12 @@ def render_bundle(config: RuntimeConfig, bundle_dir: Path | str) -> BundlePaths:
     pipeline_source = Path(config.pipeline_file)
     if pipeline_source.is_dir():
         # Deliberately FileNotFoundError and not IsADirectoryError: the accurate
-        # class subclasses OSError, not FileNotFoundError, so every existing
-        # `except FileNotFoundError` would stop catching this: nine sites in
-        # this package (cli.py, app.py x2, _compose.py, catalog.py, curation.py,
-        # storage.py, workspace.py) plus callers outside the repo. The errno
-        # carries the accurate text without moving the class. See #100 and #102.
+        # class subclasses OSError, not FileNotFoundError, so an `except
+        # FileNotFoundError` stops catching it. Two such handlers sit in this
+        # call path, cli.py:560 for `up` and cli.py:601 for `deploy`, and both
+        # turn this into exit 2; thirteen more name it elsewhere under
+        # src/hflow/, and callers outside the repo are the real unknown.
+        # The errno carries the accurate text without moving the class. #100.
         raise FileNotFoundError(errno.EISDIR, os.strerror(errno.EISDIR), str(pipeline_source))
     if not pipeline_source.is_file():
         # Three-argument form, as storage.py does: str() then carries the errno
--- a/src/hflow/runtime/_bundle.py
+++ b/src/hflow/runtime/_bundle.py
@@ -855,6 +855,14 @@ def render_bundle(config: RuntimeConfig, bundle_dir: Path | str) -> BundlePaths:
     """
     bundle_directory = Path(bundle_dir)
     pipeline_source = Path(config.pipeline_file)
+    if pipeline_source.is_dir():
+        # Deliberately FileNotFoundError and not IsADirectoryError: the accurate
+        # class subclasses OSError, not FileNotFoundError, so every existing
+        # `except FileNotFoundError` would stop catching this: nine sites in
+        # this package (cli.py, app.py x2, _compose.py, catalog.py, curation.py,
+        # storage.py, workspace.py) plus callers outside the repo. The errno
+        # carries the accurate text without moving the class. See #100 and #102.
+        raise FileNotFoundError(errno.EISDIR, os.strerror(errno.EISDIR), str(pipeline_source))
     if not pipeline_source.is_file():
         # Three-argument form, as storage.py does: str() then carries the errno
         # text, so the CLI prints why the path failed and not just the path.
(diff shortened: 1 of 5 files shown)


chiruu12:
Pushed two fixes after reviewing this again.

The comment I added had a wrong number in it. I said nine `except FileNotFoundError`
sites; the real count is fifteen under `src/hflow/`, and I missed seven because they are
written as `except (ValueError, FileNotFoundError)` in `cli.py` and my grep only matched
the bare form. I also listed `runtime/_compose.py:49`, which cannot receive this at all:
it wraps `subprocess.run` and translates a missing `docker` binary. The comment now says
what is actually true and useful, that two handlers sit in this call path, `cli.py:560`
for `up` and `cli.py:601` for `deploy`, and that the unknown out-of-repo callers are the
real argument for keeping the class.

The bigger gap: your repro was `hflow up --pipeline`, and my only CLI test drove `deploy`.
The `up` path was covered by composition, unit test on the raise plus the existing ENOENT
test on the handler, which holds today but would not catch anything that later branched on
errno between `render_bundle` and the user. Added three CLI tests so all four raise sites
are pinned at the command level, matching what the ENOENT tests already do: `up --pipeline`,
`up --requirements`, `deploy --requirements`. The `up` one asserts no bundle directory, no
compose calls, no traceback and no teardown advice, same as its sibling above it.

Reverting each of the four sites on its own now fails the suite each time, which I checked
by actually doing it rather than reasoning about it:

```
bundle/pipeline        -> 1 failed, 241 passed, 4 skipped
bundle/requirements    -> 1 failed, 242 passed, 4 skipped
deploy/pipeline        -> 1 failed, 260 passed, 4 skipped
deploy/requirements    -> 1 failed, 259 passed, 4 skipped
```

Gates: ruff check and format clean, ty passed, 650 passed 6 skipped.

One thing outside this PR that you may want as its own issue. `hflow doctor <dir>` has the
same bug from a different site: `storage.py:654` tests `is_file()` and raises ENOENT, so
pointing doctor at a directory says the path does not exist. Same shape, different command,
and not one of the four sites this issue names, so I left it alone. Say the word and I will
file it.

kstonekuan (approved):
LGTM. Merging.

Verified on a rebase onto current `main` (`763f185`): gates clean, 653 passed, 3 skipped, and the behaviour is right in both directions.

```
hflow up --pipeline /tmp/adir       -> up: [Errno 21] Is a directory: '/tmp/adir'      exit 2
hflow deploy --pipeline /tmp/adir   -> deploy: [Errno 21] Is a directory: '/tmp/adir'
hflow up --pipeline /tmp/missing.py -> up: [Errno 2] No such file or directory: ...
```

The comment is the part that earns the diff. Naming the two handlers in the actual call path (`cli.py:560` for `up`, `cli.py:601` for `deploy`) separately from the thirteen elsewhere is more useful than my issue text, which lumped them together, and I confirmed all three numbers: 560 sits in `_command_up`, 601 in `_command_deploy`, and 2 + 13 is exactly the 15 handlers under `src/`.

One thing to fix in your head, not in the code: the PR body says "the refactor left one handler in `cli.py`, not eight" and "nine sites in this package". Both are off, and your own code comment has it right. `cli.py` still has eight handlers on `main`, two of which are in this call path; the package has fifteen. The prose and the comment disagree, and the comment is the one that will be read in two years, so it is the right one to have got correct.

PR #121, opened 2026-08-23
fix(doctor): report every file and continue past unreadable ones; document exit codes

### LINKED ISSUE #99: CLI: hflow doctor abandons remaining files on the first unreadable one; exit-code contract undocumented
### What happens

`hflow doctor` takes many files, but one unreadable path abandons every file after it. Point it at a glob of 200 episodes with one missing or corrupt file and you lose the reports for everything past that point, plus the exit code stops telling you anything about the files it did read.

On `main` (`2f07aa0`), with a good file, a missing file, and a good file:

```
$ hflow doctor good.mcap /tmp/nope-not-here.mcap good.mcap
doctor: [Errno 2] No such file or directory: '/tmp/nope-not-here.mcap'
doctor: good.mcap
  conforming: no findings
  verdict: CONFORMING
$ echo $?
2
```

The third file was never looked at. Note also that the error line prints *before* the first file's report, because the report goes to stdout and the error to stderr, so the interleaving above is what a terminal shows and not the order things happened.

For contrast, a genuinely non-conforming file behaves correctly: every file is reported and the exit code is 1.

### Why it is wrong

`src/hflow/doctor.py:10` states the module's design: "Findings, not exceptions: the doctor accumulates everything it can observe". That holds within a file, where `_FindingCollector` gathers every finding instead of raising on the first. It does not hold across files: `_command_doctor` (`src/hflow/cli.py:705`) returns 2 from inside its loop at `src/hflow/cli.py:712` on the first `ValueError` or `FileNotFoundError`, so the accumulate-everything property stops at the file boundary.

An unreadable file is also a finding about the corpus. It is the single most likely thing to go wrong on a batch run over a real corpus, and it is exactly when you least want to lose the other 199 answers.

### Definition of done

`hflow doctor` reports on every file it was given, whatever happens to any one of them.

Specifics:

- **An unreadable or unparseable file is reported in place**, in the same shape as the other per-file output, and the loop continues to the next file.
- **The exit code stays meaningful.** The existing convention is: 0 = everything conforming, 1 = ran fine and found non-conformance, 2 = bad input, nothing useful happened. A run where some files were read and one path was missing is not "nothing happened", so it should not silently become 2 and throw away the 1. Decide the precedence between "some file was unreadable" and "some file was non-conforming" and state it in the help text. My preference is that an unreadable file is at least as bad as a non-conforming one, so it should not be reported as milder, but argue for something else if you see it differently.
- **Order of output is stable and matches the argument order**, so a diff between two runs is readable.
- Do not weaken the per-file diagnosis. `diagnose()` itself should keep raising for an unreadable file; this is about what the CLI does with that.

### The second half: write the contract down

While you are in here, the exit-code convention is nowhere documented, which is how the `up` inconsistency in #90 survived. Every command already follows the same three-value convention (I checked all of them: `curate`, `stale`, `doctor`, `manifest`, `up`, `deploy`, `down`, `ingest`, `status`, `ui`), but nothing says so, so the next command added has nothing to conform to.

Add a short section to the CLI reference documenting 0/1/2 with the meaning of each, and note which commands can return 1 and what it means for each (`doctor`: non-conforming file; `stale --exit-code`: stale episodes found; `up`: started then failed, so containers may still be running; `ingest`: episodes failed). This is the part that stops the drift rather than fixing one instance of it.

If auditing turns up a command that does not actually follow the convention, do not fix it here. File it separately with the evidence, the way #90 was filed.

### Pattern to copy

`_command_ingest` (`src/hflow/cli.py:598`) is the closest precedent for a command that processes many things and has to decide what a partial failure means. `tests/test_doctor.py` for tests; a missing path needs no fixture beyond a name that does not exist.

### Validation

```
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

For the doc change, the lychee link check from CONTRIBUTING.md.

### Notes

Not a starter issue: the fix is small, but choosing the exit-code precedence and writing the contract down is a judgement call that outlives the fix. Comment here and I will assign it.

ayam04:
## What changed

`hflow doctor` now reports on every file it was given, whatever happens to any one of them, and the exit-code convention is written down.

**Loop behavior:** an unreadable or unparseable path is reported in place, in the same per-file shape as the other reports (`[error] unreadable: ...` + `verdict: NOT CONFORMING`), on stdout, so the order of output matches the argument order and a diff between two runs is readable. The loop continues to the next file; `diagnose()` itself still raises (per-file diagnosis is not weakened).

**Exit-code precedence** (per the issue's two requirements resolved together: an unreadable file is at least as bad as a non-conforming one, and a run where some files were read must not silently become 2):
- `0` - everything conformed.
- `1` - any file was non-conforming OR could not be read (an unreadable file is reported at the same severity as a non-conforming one, never milder).
- `2` - nothing useful happened: no file could be diagnosed at all (or CLI misuse).

**Documentation:** the 0/1/2 contract now lives in the CLI reference (docs/FORMAT.md Conformance section, where doctor's status was already documented): the meaning of each code, plus what `1` means per command (`doctor`, `stale --exit-code`, `up`, `ingest`). The doctor help text states the precedence too.

## Validation (run in WSL2)

```
uv run ruff check --fix   # clean
uv run ruff format        # clean
uv run ty check           # All checks passed
uv run pytest -q          # 638 passed, 3 skipped; 2 failed + 7 errors, ALL in test_ffmpeg.py
```

The ffmpeg failures are pre-existing environment gaps (no ffmpeg binaries in this WSL image): I stashed my changes and reproduced the identical 2 failed / 7 errors on clean `main`. New doctor tests (`test_cli_doctor_continues_past_unreadable_file`, `test_cli_doctor_all_unreadable_exits_2`, updated missing-file test) all pass.

```
lychee --no-progress --include-fragments --exclude '^https://github\.com/Hebbian-Robotics/hflow/(issues|security/advisories/new)$' --exclude-path references/mcap-spec.md --exclude-path references/foxglove-CompressedVideo.proto .
# 257 OK, 0 Errors
```

Fixes #99

### DIFF
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -143,8 +143,26 @@ A file claiming this convention must satisfy, in increasing strictness:
 5. **Stamps**: `provenance/v1` present with `schema_version` and `pipeline_version`.
 
 `hflow doctor <file.mcap> [more.mcap ...]` executes these checks against every
-file given, printing one report each, and exits with status 0 when all files
-conform or 1 when any reports violations.
+file given, printing one report each in argument order, and exits with status 0
+when all files conform or 1 when any reports violations. An unreadable or
+unparseable path is reported in place, in the same per-file shape (`[error]
+unreadable: ...`), and the run continues with the remaining files.
+
+### Exit codes
+
+Every `hflow` command follows the same three-value convention:
+
+- `0` - success: the command did what it was asked.
+- `1` - ran, and found something to report: `doctor`, a non-conforming file;
+  `stale --exit-code`, stale episodes found; `up`, started then failed, so
+  containers may still be running; `ingest`, episodes failed.
+- `2` - bad input, nothing useful happened: an invalid flag combination, an
+  unreachable endpoint, a missing catalog, or a `doctor` run in which no file
+  could be diagnosed at all.
+
+`doctor` treats an unreadable file as at least as bad as a non-conforming one
+(exit 1), and reserves 2 for runs where nothing was diagnosed, so a batch run
+never both loses its reports and erases the fact that it read anything.
 
 ## References
 
--- a/docs/RUNTIME.md
+++ b/docs/RUNTIME.md
@@ -34,6 +34,11 @@ before the subcommand:
 - `--version` prints the installed HFlow version and exits.
 - `-v`, `--verbose` enables verbose logging.
 
+Every subcommand follows the same three-value exit-code convention (0 success,
+1 ran and found something to report, 2 bad input and nothing happened), which
+matters when scripting `up` or `ingest`: see
+[Exit codes](./FORMAT.md#exit-codes).
+
 ## What `up` builds: anatomy of the bundle
 
 ```bash
--- a/src/hflow/cli.py
+++ b/src/hflow/cli.py
@@ -163,7 +163,10 @@ def _build_parser() -> argparse.ArgumentParser:
         description=(
             "Validate container integrity, metadata stamps, chunk-group layout, "
             "and in-band video constraints (docs/FORMAT.md, executable form). "
-            "Accepts multiple files; exit code 0 when all conform, 1 when any does not."
+            "Accepts multiple files, each reported in order whatever happens "
+            "to the others. Exit 0 when all conform, 1 when any file is "
+            "non-conforming or could not be read, 2 when no file could be "
+            "diagnosed (nothing useful happened)."
         ),
     )
     doctor_parser.add_argument(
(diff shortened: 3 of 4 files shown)


kstonekuan (approved):
LGTM. Merging with one docs fixup.

The self-correction on exit-code precedence was the right call and you got there before I looked. Resolving the two requirements together does give unreadable = 1, not 2: an unreadable file must not be milder than a non-conforming one, and a run that read some files must not collapse to "nothing happened". 2 is left for the case where nothing was diagnosed at all.

Reusing the module's own `unreadable` finding code (`doctor.py:150`) for the CLI-level message is the detail I liked most. A corrupt-but-present file already reports `[error] unreadable: not readable as MCAP: ...` from inside `diagnose()`, and a missing path now reports the same shape from outside it, so the two look identical to a reader who does not care which layer noticed. Printing it to stdout also fixes the stderr/stdout interleaving from the issue.

Verified the behaviour rather than just the tests, on a rebase onto current `main`:

```
good + missing + good  -> all three reported in argument order, exit 1
missing only           -> exit 2
missing + missing      -> exit 2
corrupt only           -> exit 1   (diagnose() reports it, so it WAS diagnosed)
good + corrupt         -> exit 1
all good               -> exit 0
```

That corrupt-file case is worth stating out loud since it is the subtle one: a present-but-corrupt file is diagnosed and comes out 1, while an absent file cannot be diagnosed and contributes to 2. That is the right line and it falls out of "findings, not exceptions" rather than being special-cased.

Gates clean, 655 passed, 3 skipped, lychee 0 errors.

**The fixup:** the contract you wrote is cross-command, naming `up`, `ingest`, and `stale` as well as `doctor`, but it lives in FORMAT.md. Someone scripting `hflow up` around exit codes has no reason to open the canonical-format reference. Added one pointer from the global-flags section of RUNTIME.md, where that reader already is. FORMAT.md stays the single definition; nothing is duplicated.

ayam04:
Thanks for the review and the docs fixup on merge, @kstonekuan - glad the exit-code precedence resolution landed right. On to the next one.

Issue #67, opened 2026-08-20
Tests: `hflow ingest` rejection of non-relative URIs has no coverage


kstonekuan:
`hflow ingest` rejects absolute paths and parent-directory escapes before it
makes an HTTP request. That guard has no test.

## Current behavior

The guard is at `src/hflow/cli.py:593-603`. A URI beginning with `/`, or whose
normalized path begins with `..`, prints `is not relative to the data root` and
returns exit code 2. The early failure is necessary because runtime URIs resolve
inside the container against the configured data root.

`tests/test_runtime_cli.py` covers ingest success paths and parser-level profile
validation, but it does not exercise this guard.

## Pattern to copy

Use `test_ingest_rejects_unknown_profile` at
`tests/test_runtime_cli.py:545-551`, including its `_rendered_bundle` setup.
Unlike the argparse rejection, this guard returns 2 from `main`, so assert on
the return value.

## Definition of done

- A test confirms that `/abs/x.mcap` returns 2.
- A test confirms that `../x.mcap` returns 2.
- The test asserts that `is not relative to the data root` appears on stderr.
- Do not mock HTTP. The guard runs before any request.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

WilliamK112:
I'd like to take this. I confirmed the issue is unassigned and there is no overlapping open PR.

I'll add focused CLI regression coverage using the existing rendered-bundle setup and exercise both rejected URI classes:

- `/abs/x.mcap` for an absolute path; and
- `../x.mcap` for a parent-directory escape.

The test will invoke the real `main` path, assert return code 2 and the `is not relative to the data root` stderr message, and will not mock HTTP because the guard must fail before any request. I'll run the focused test plus the repository's Ruff, formatting, type, and full pytest gates.

AI assistance disclosure: I am using OpenAI Codex to help inspect, implement, and validate the test. I will review the final diff and evidence before submitting.

WilliamK112:
Update: #123 was automatically closed by the repository's one-open-PR-per-outside-contributor policy because #122 is still awaiting review.

To respect the policy and keep starter issues available, I am not reserving #67 while #122 remains open; another contributor is welcome to take it. The tested branch remains available, and if #122 finishes first while this issue is still unclaimed, I can reopen #123 without redoing the work.

PR #122, opened 2026-08-23
fix(transform): reject pass-through video starting mid-GOP

### LINKED ISSUE #96: Transform: a pass-through video channel starting mid-GOP is accepted and stamped conforming, but doctor rejects it

WilliamK112:
Closes #96.

## Outcome

- Tracks the first successfully validated pass-through video message independently for each MCAP channel.
- Rejects a channel whose first H.264 access unit is not a keyframe, so the transform can no longer stamp a mid-GOP stream as canonical.
- Keeps accepted pass-through payloads byte-for-byte unchanged.

## Regression coverage

- Interleaves two camera channels and proves the second channel is rejected when its own first message starts mid-GOP, even after the first channel has already passed validation.
- Interleaves two valid channels that each start on a keyframe, then verifies their serialized payloads are unchanged and `diagnose()` reports the output conforming.

`TRANSFORM_BEHAVIOR_VERSION` is intentionally unchanged: this closes acceptance of an input that already violates the documented canonical format. Inputs that were valid before still produce identical canonical bytes.

## Validation

- `/Users/William/.local/bin/uv run ruff check --fix`
- `/Users/William/.local/bin/uv run ruff format`
- `/Users/William/.local/bin/uv run ty check`
- `/Users/William/.local/bin/uv run pytest -q` — 654 passed, 6 skipped
- `git diff --check`

AI assistance disclosure: I used OpenAI Codex to help inspect the existing transform/doctor invariants, implement the change, and run validation. I reviewed the final diff and test evidence.

### DIFF
--- a/src/hflow/transform.py
+++ b/src/hflow/transform.py
@@ -9,8 +9,9 @@ v1 behavior:
 - Channels already carrying ``foxglove.CompressedVideo`` pass through
   byte-for-byte, but only after every message is validated against the
   canonical video constraints (``format="h264"``, one AUD-delimited access
-  unit per message, SPS+PPS on keyframes) -- the provenance stamp asserts
-  conformance, so nonconforming video must fail loudly, never pass silently.
+  unit per message, SPS+PPS on keyframes, each channel starts on a keyframe) --
+  the provenance stamp asserts conformance, so nonconforming video must fail
+  loudly, never pass silently.
   Raw ``sensor_msgs/msg/Image``/``foxglove.RawImage`` is not supported in v1
   (explicit error).
 - Every other channel passes through byte-for-byte with its original schema.
@@ -226,7 +227,9 @@ def _resolve_decoder(topic: str, info: TopicInfo) -> Callable[[bytes], Any]:
     )
 
 
-def _validate_passthrough_video_payload(topic: str, decoded_message: Any) -> None:
+def _validate_passthrough_video_payload(
+    topic: str, decoded_message: Any, *, is_first_message: bool
+) -> None:
     """Enforce the canonical video constraints on a pass-through message.
 
     The provenance stamp asserts the whole file conforms (FORMAT.md), so a
@@ -252,10 +255,16 @@ def _validate_passthrough_video_payload(topic: str, decoded_message: Any) -> Non
             f"{len(access_units)} access units; the canonical convention requires "
             "exactly one decodable frame per message"
         )
-    if access_units[0].is_keyframe and not access_units[0].has_parameter_sets:
+    access_unit = access_units[0]
+    if access_unit.is_keyframe and not access_unit.has_parameter_sets:
         raise ValueError(
             f"pass-through video topic {topic!r} has a keyframe without SPS/PPS; "
             "the canonical convention requires parameter sets on every keyframe"
+        )
+    if is_first_message and not access_unit.is_keyframe:
+        raise ValueError(
+            f"pass-through video topic {topic!r} starts mid-GOP: its first message "
+            "is not a keyframe, so the stream is not decodable from the start"
         )
 
 
@@ -362,6 +371,7 @@ def write_canonical_episode(
             if info.schema_name in _PASSTHROUGH_VIDEO_SCHEMAS
         }
         passthrough_video_decoders: dict[int, Callable[[bytes], Any]] = {}
+        passthrough_video_channels_with_messages: set[int] = set()
         for batch in reader.iter_batches():
             if batch.channel_id in camera_payloads:
                 camera_log_times[batch.channel_id].extend(int(t) for t in batch.log_times)
@@ -375,7 +385,15 @@ def write_canonical_episode(
                         )
                     decode = passthrough_video_decoders[batch.channel_id]
                     for payload in batch.data:
-                        _validate_passthrough_video_payload(batch.topic, decode(payload))
+                        is_first_message = (
+                            batch.channel_id not in passthrough_video_channels_with_messages
+                        )
+                        _validate_passthrough_video_payload(
+                            batch.topic,
+                            decode(payload),
+                            is_first_message=is_first_message,
+                        )
+                        passthrough_video_channels_with_messages.add(batch.channel_id)
                 outgoing.extend(
                     _OutgoingMessage(
                         log_time=int(batch.log_times[index]),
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM. Merging, no changes.

Credit where it belongs: @promiseeuler did the design work on this in the issue thread, establishing that `iter_batches()` uses `log_time_order=True` so a per-channel first-message check is exactly correct rather than approximately so, and specifying the set-beside-the-decoder-cache shape including the detail that a channel must be marked seen only *after* its payload validates. This implements that faithfully.

What I verified beyond your tests:

**The original repro is dead.** My reproduction from the issue (canonical episode, drop the leading keyframe, rewrite as pass-through) now raises `pass-through video topic '/cam_0/compressed' starts mid-GOP` where it previously wrote an output and stamped it conforming.

**Your interleaving test genuinely binds to the per-channel design.** I degraded `is_first_message` from the per-channel set to a single global flag and re-ran: `test_transform_rejects_each_passthrough_video_channel_starting_mid_gop` fails with `DID NOT RAISE`. That is the difference between a test that looks thorough and one that is, since a single boolean passes a naive single-channel test.

**The invariant, not just the raise.** `test_transform_accepts_independent_interleaved_passthrough_video_channels` asserts byte-for-byte payload equality *and* `diagnose(output).conforming`, which is the property the issue was actually about: the transform must never stamp something `doctor` rejects.

Gates clean, 657 passed, 3 skipped.

Leaving `TRANSFORM_BEHAVIOR_VERSION` alone is right, and your reasoning is the reason: this only withdraws acceptance of input that already violated the documented format, and your byte-for-byte assertion proves previously-valid inputs still produce identical output. A bump would have marked every existing episode stale for no behavioural change.

On #67 and #123: declining to reserve an issue you could not immediately work was the right call and I would rather you had done that than hold it. #122 is merged now, so reopen #123 whenever you like; #67 is still unassigned.

PR #127, opened 2026-08-23
fix(catalog): coerce NumPy scalar measurements at the boundary; refuse the rest

### LINKED ISSUE #126: Catalog: NumPy scalar measurements are stored as all-NULL, silently excluding episodes from curated manifests
### Version or commit

main @ 08e9eac (built from source; also reproduces on any recent commit -- the branch in _insert_dependent_rows predates it)

### Environment

Ubuntu 24.04 (WSL2), CPython 3.12, x86_64. No Docker needed -- pure SDK surface (catalog + curation)

### Minimal reproduction

Two identical checks appended through the normal path (`CheckRunRow` -> `Catalog.append_episode`, i.e. what `@app.check()` results become), differing only in scalar type:

```python
import tempfile
from pathlib import Path

import numpy as np

import hflow
from hflow.catalog import Catalog, CheckRunRow
from hflow.curation import open_catalog_connection
from hflow.transform import EpisodeStamps

STAMPS = EpisodeStamps(
    schema_version="episode/v1", pipeline_version="repro",
    ffmpeg_version="n/a", robot_software_version=None,
)

work = Path(tempfile.mkdtemp())

def append(name, measurements):
    canonical = work / f"{name}.canonical.mcap"
    canonical.write_bytes(name.encode())
    row = CheckRunRow(
        check_name="camera_blackout", check_version="v1", critical=True,
        status=hflow.CheckStatus.MEASURED, duration_s=0.1,
        measurements=measurements,
    )
    print(name, Catalog(work / "catalog").append_episode(
        canonical_path=canonical, stamps=STAMPS,
        episode_metadata={}, check_rows=[row]).written)

# same logical outcome, only the types differ
append("numpy_episode", {"black_pct": np.float32(0.4),
                         "frame_count": np.int64(120), "camera_ok": np.bool_(True)})
append("python_episode", {"black_pct": 0.4,
                          "frame_count": 120, "camera_ok": True})

conn = open_catalog_connection(work / "catalog")
print(conn.execute("SELECT key, value_double, value_text, value_bool "
                   "FROM measurements ORDER BY episode_id, key").fetchall())
print(conn.execute("SELECT episode_id FROM episodes "
                   "WHERE black_pct < 1.0 AND status != 'quarantined'").fetchall())
```

**Expected behavior**
Both episodes pass the manifest filter. Any scalar type a check plausibly returns (Python or NumPy, and QC code lives in NumPy) should round-trip into `value_double`/`value_text`/`value_bool`, so the wide `episodes` view exposes it and SQL predicates like the README's `WHERE black_pct < 1.0` see it.

**Actual behavior**
The NumPy episode vanishes from the manifest query. Its measurement rows are written, keys present, but all three value columns are NULL, and in SQL `NULL < 1.0` is falsy. No error or warning anywhere:

```
== raw storage: the measurements table ==
  key=black_pct    value_double=None   value_text=None   value_bool=None   <- numpy episode
  key=camera_ok    value_double=None   value_text=None   value_bool=None
  key=frame_count  value_double=None   value_text=None   value_bool=None
  key=black_pct    value_double=0.4    value_text=None   value_bool=None   <- python episode
  key=camera_ok    value_double=None   value_text=None   value_bool=True
  key=frame_count  value_double=120.0  value_text=None   value_bool=None

SELECT ... WHERE black_pct < 1.0 AND status != 'quarantined'
  -> only the python-float episode comes back
```

Root cause: the storage branches at `src/hflow/catalog.py:307-314` accept exactly Python `int`/`float`/`str`/`bool`. `isinstance(np.float32(0.4), float)` -> False, `isinstance(np.int64(120), int)` -> False, `isinstance(np.bool_(True), bool)` -> False (only `np.float64` subclasses `float`, by accident). Each failed branch falls through to `None`.

**Additional context**
This sits on the README's six-line example: `result` in `hflow.CheckResult(measurements=result)` comes from user QC code, which returns NumPy scalars almost always (`np.mean()`, array indexing, `np.bool_` verdicts). Tenet #2 ("Evidence, not verdicts") fails quietly here. Coverage denominators lie and curated datasets shrink with nothing flagging it.

Checked before filing:
- `catalog.py:256-258` already knows these values arrive ("numpy integers/bools reach here untouched") but only routes them through `default=repr` for the fingerprint. Side effect: `np.float32(0.4)` and `float(0.4)` fingerprint differently, so the same logical outcome can split into two runs instead of replaying idempotently.
- `tests/test_catalog_curation.py:522` enshrines the loss: it asserts append succeeds and fingerprint is stable, but never reads values back. Its own comment (:535-536) says smuggling NumPy scalars past the declared type is "exactly what real check code does."
- `tests/test_runtime_bundle.py:877-879` claims catalog.py "accommodates numpy scalars in measurements because those are user data." Accommodated for fingerprints, discarded at storage.

Fix shape (not implemented): normalize at one choke point in `_insert_dependent_rows`. `np.generic` -> `.item()` is lossless for every NumPy scalar and also fixes the replay-split, since coerced values fingerprint identically. Refuse loudly, naming check/key/received type, for anything else. Extend `test_append_accepts_non_json_measurement_scalars` into a round-trip assertion, plus a regression that a NumPy-measured episode survives `WHERE black_pct < 1.0`. Existing catalogs keep their NULLs (append-only), worth a sentence in the PR.

Happy to take this if the coerce-vs-refuse split sounds right.

### Expected behavior

Both episodes survive the manifest filter. A measurement value of any scalar type a check plausibly returns (Python or NumPy scalars -- robotics QC code lives in NumPy) round-trips into value_double/value_text/value_bool, so the wide episodes view exposes it and SQL predicates like the README's own `WHERE black_pct < 1.0` see it.

### Actual behavior

Only the python-float episode comes back from the filter. The numpy one disappears silently.

Its measurement row IS written, just with every value column NULL:

    key=black_pct    value_double=None  value_text=None  value_bool=None   <- numpy episode
    key=camera_ok    value_double=None  value_text=None  value_bool=None
    key=frame_count  value_double=None  value_text=None  value_bool=None
    key=black_pct    value_double=0.4   value_text=None  value_bool=None   <- python episode
    key=camera_ok    value_double=None  value_text=None  value_bool=True
    key=frame_count  value_double=120.0 value_text=None  value_bool=None

And in SQL `NULL < 1.0` is NULL (falsy), so that row can never pass any threshold predicate.

Cause: the storage branches at src/hflow/catalog.py:307-314 accept exactly Python types:

    isinstance(np.float32(0.4), float) -> False
    isinstance(np.int64(120), int)     -> False
    isinstance(np.bool_(True), bool)   -> False
    isinstance(np.float64(0.4), float) -> True   # the one accidental survivor

Each failed branch falls through to None. No error or warning at any point.

### Additional context

This hits the README's six-line example directly: result in hflow.CheckResult(measurements=result) is whatever the user's QC function returned, and QC code returns numpy scalars constantly (np.mean(), array indexing, np.bool_ flags). Tenet #2 is "evidence, not verdicts" -- right now evidence gets recorded as a coverage lie instead of an error, and curated datasets quietly shrink.

What I checked before filing:

- catalog.py:256-258 already knows these values arrive ("numpy integers/bools reach here untouched") but only routes them through default=repr for fingerprinting. Side effect of that split: np.float32(0.4) and float(0.4) produce different fingerprints, so the same logical outcome appends twice instead of replaying idempotent.
- tests/test_catalog_curation.py:522 locks the current behavior in. It asserts written=True and stable fingerprints but never reads the values back, and its comment at :535 says smuggling numpy scalars past the declared type is "exactly what real check code does".
- tests/test_runtime_bundle.py:877 states catalog.py "accommodates numpy scalars in measurements because those are user data". True for fingerprints, false for storage.

Fix shape I'd suggest (not implemented): normalize at one choke point in _insert_dependent_rows. np.generic -> .item() is lossless for every numpy scalar and also fixes the fingerprint split, since coerced values hash identically. Anything outside int/float/str/bool/np.generic should raise naming the check, key, and received type rather than write NULLs, matching the registration-time refusals from #93/#86. Then extend test_append_accepts_non_json_measurement_scalars into a real round-trip test and add one regression through open_catalog_connection asserting a numpy-measured episode survives WHERE black_pct < 1.0. Existing catalogs keep their old NULL rows (append-only), worth a sentence in the PR.

Can take this one if the approach sounds right.

Sagar-024:
Closes #126

## Summary

Fixes silent data loss in the catalog: NumPy scalar measurements (`np.float32`, `np.int64`, `np.bool_`) were stored as NULL while their keys still landed, causing curation predicates like `WHERE black_pct < 1.0` to silently drop those episodes and splitting replay idempotence by scalar flavor. Now every scalar round-trips, replays converge on value equality, and exotic non-JSON types refuse loudly at the boundary.

## Why

Checks return NumPy scalars constantly (`np.mean()`, indexing, `np.bool_` verdicts), but measurement storage branched on exactly Python int/float/str/bool -- which only np.float64 subclasses. Every other flavor fell through to NULL columns while its key still landed, so the README's own `WHERE black_pct < 1.0` example silently dropped those episodes. The same logical outcome could also split into two runs depending on which scalar flavor a check happened to return that day (np.float32(0.4) vs float(0.4) fingerprinted differently; np.float64 survived by accident).

The fix is one choke point: `append_episode` normalizes measurements before anything consumes them -- both `_run_fingerprint` and `_insert_dependent_rows` now receive one normalized `check_rows`. `np.generic` coerces via `.item()`; anything else raises naming the check, key, and received type (#93/#86 tone), replacing silent NULLs. With values guaranteed JSON-native, `default=repr` is gone from the run fingerprint along with the comment describing the leak it papered over. Nothing but measurements ever fed the non-JSON path, and those were exactly the values being destroyed; exotic types now refuse loudly at the boundary instead of hashing successfully and storing NULLs.

One deliberate scope line: equal values across scalar flavors now replay as one run regardless of flavor, but a genuinely different value stays a distinct outcome -- float32 rounding of 0.4 differs from the float64 literal in the 8th decimal, and coercing cannot (and should not) fuse them. #126's replay example therefore converges on value equality, not flavor equality; happy to take a different ruling if you want float32-quantized identity.

Existing catalogs keep their old NULL rows: append-only, nothing migrates.

Enrichment labels reach the same path (`app.py` packs them into `CheckRunRow.measurements`), so they are covered for free -- with a test.

## Validation

```bash
uv run ruff check --fix src tests       # clean
uv run ruff format src tests            # clean
uv run ty check src tests               # clean
uv run pytest -q                        # 662 passed, 3 skipped
```

## Checklist

### DIFF
--- a/src/hflow/catalog.py
+++ b/src/hflow/catalog.py
@@ -35,11 +35,12 @@ import hashlib
 import json
 import tempfile
 from collections.abc import Sequence
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from datetime import UTC, datetime
 from pathlib import Path
 
 import duckdb
+import numpy as np
 
 from hflow.format import CATALOG_FORMAT_VERSION
 from hflow.steps import CheckResult, CheckStatus, Interval, MeasurementValue
@@ -255,6 +256,31 @@ def content_episode_id(canonical_path: Path) -> str:
     return digest.hexdigest()[:16]
 
 
+def _normalized_measurements(
+    check_name: str, measurements: dict[str, MeasurementValue]
+) -> dict[str, MeasurementValue]:
+    """Coerce NumPy scalars to Python scalars; refuse anything else loudly.
+
+    Real check code returns NumPy scalars (np.mean(), indexing, np.bool_
+    verdicts), and only np.float64 subclasses a Python type -- the rest used
+    to fall through every storage branch into all-NULL value columns while
+    the key still landed. Coercion happens here, before anything consumes
+    the values, so one np.float32(0.4) and float(0.4) outcome fingerprints
+    and stores identically instead of splitting into two runs.
+    """
+    normalized: dict[str, MeasurementValue] = {}
+    for key, value in measurements.items():
+        if isinstance(value, np.generic):
+            value = value.item()
+        if not isinstance(value, MeasurementValue):
+            raise ValueError(
+                f"check {check_name!r} measured {key!r} as {type(value).__name__}: "
+                "measurements hold one int/float/str/bool per key"
+            )
+        normalized[key] = value
+    return normalized
+
+
 def _run_fingerprint(
     episode_id: str,
     pipeline_version: str,
@@ -283,11 +309,8 @@ def _run_fingerprint(
         }
         for row in check_rows
     ]
-    # default=repr: measurement values are user data and may be non-JSON
-    # scalars (numpy integers/bools reach here untouched); repr keeps the
-    # fingerprint deterministic instead of crashing the whole append.
     check_outcomes.sort(
-        key=lambda outcome: json.dumps(outcome, sort_keys=True, separators=(",", ":"), default=repr)
+        key=lambda outcome: json.dumps(outcome, sort_keys=True, separators=(",", ":"))
     )
     payload = json.dumps(
         {
@@ -298,7 +321,6 @@ def _run_fingerprint(
         },
         sort_keys=True,
         separators=(",", ":"),
-        default=repr,
     )
     return hashlib.sha256(payload.encode()).hexdigest()[:12]
 
@@ -451,6 +473,14 @@ class Catalog:
         address for local roots).
         """
         episode_id = content_episode_id(canonical_path)
+        # One normalized shape feeds every consumer below -- the run
+        # fingerprint, the replay repair pass, and the dependent-table
+        # inserts. Normalizing any later would let np.float32(0.4) and
+        # float(0.4) hash as two outcomes while storing NULL columns.
+        check_rows = [
+            replace(row, measurements=_normalized_measurements(row.check_name, row.measurements))
+            for row in check_rows
+        ]
         run_fingerprint = _run_fingerprint(
             episode_id,
             stamps.pipeline_version,
--- a/tests/test_catalog_curation.py
+++ b/tests/test_catalog_curation.py
@@ -519,8 +519,8 @@ def test_constrained_curate_writes_the_manifest_but_refuses_outside_reads(
         )
 
 
-def test_append_accepts_non_json_measurement_scalars(tmp_path: Path) -> None:
-    """numpy scalars are user data: they must fingerprint, not crash the append."""
+def test_numpy_scalar_measurements_round_trip(tmp_path: Path) -> None:
+    """NumPy scalars from real check code store readable values, not NULLs."""
     import numpy as np
 
     canonical = tmp_path / "e.canonical.mcap"
@@ -532,9 +532,12 @@ def test_append_accepts_non_json_measurement_scalars(tmp_path: Path) -> None:
         critical=False,
         status=hflow.CheckStatus.MEASURED,
         duration_s=0.1,
-        # cast: deliberately smuggling user-supplied non-JSON scalars past the
+        # cast: deliberately passing user-supplied NumPy scalars past the
         # declared MeasurementValue type -- exactly what real check code does.
-        measurements=cast(dict, {"count": np.int64(3), "flag": np.bool_(True)}),
+        measurements=cast(
+            dict,
+            {"ratio": np.float32(0.4), "frames": np.int64(3), "flag": np.bool_(True)},
+        ),
     )
     result = catalog.append_episode(
         canonical_path=canonical,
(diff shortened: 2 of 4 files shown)


kstonekuan (approved):
LGTM, merging. You took the correction and put the coercion where it actually fixes both halves, and the replay test is careful in the way that matters — using `np.float64(0.5)`/`0.5` for the idempotence claim and saying in the docstring why float32 rounding of 0.4 is a genuinely different value rather than papering over it. Verified on top of current `main` (post-#131): 717 passed, ruff/format/ty clean.

Two things I checked that are not yours to fix, recorded so the next person does not rediscover them:

**A `None` measurement now raises where it used to write NULL.** Undeclared — `MeasurementValue` has no `None` member — but it worked, and a check returning `None` for "could not measure this episode" will now fail that episode's append. I think refusing is correct and I am keeping it: NULL is now the signal that nothing stored a value, and admitting `None` would recreate the exact ambiguity this PR removes. A check with nothing to say should omit the key. I will document it rather than ask you to change anything.

**Interval bounds have the same hole and are already broken.** `Interval.start_ns` declares `int`, and `np.int64` bounds fail on `main` today too — DuckDB throws `NotImplementedException` at insert. Your change moves the failure earlier with a clearer message (`not JSON serializable`), so removing `default=repr` regressed nothing. Separate issue if it ever bites someone.

Sagar-024:
Thanks for the review and the merge. Glad the None refusal is staying 
a check with nothing to say should omit the key, and NULL now means
"nothing stored a value."

The interval-bounds hole is the same bug class one screen over, so I can
file it as its own issue with a repro (np.int64 start_ns -> DuckDB
NotImplementedException on main) if that's useful. Otherwise I'll leave
it until it bites someone.

PR #133, opened 2026-08-23
fix(catalog): refuse measurement keys that shadow episodes columns

### LINKED ISSUE #130: A measurement key colliding with an episodes column is silently renamed, so SELECT returns the wrong value
## What happens

The wide `episodes` view pivots each numeric measurement key into a column beside the promoted episode columns (`task`, `uri`, `pipeline_version`, ...). When a measurement key equals one of those column names, DuckDB does not refuse it — it **silently renames** the pivoted column.

The episode column keeps the name. The measurement becomes `<key>_1`, a name nothing documents and no query would think to ask for. So `SELECT task FROM episodes` returns the episode metadata and never the measurement, with no error anywhere.

## Reproduction

```python
import hflow

app = hflow.App("reserved-key", data_root="./data")

@app.check()
def claims_task(ep: hflow.Episode) -> hflow.CheckResult:
    return hflow.CheckResult(measurements={"task": 99.0})

episode = synthesize_episode(
    "episode.mcap",
    SyntheticEpisodeSpec(duration_s=2.0, cameras=(), task="fold_napkin"),
)
app.process(episode, record=True)

connection = hflow.open_catalog_connection("./data/catalog")
columns = [d[0] for d in connection.execute("SELECT * FROM episodes LIMIT 0").description]
print([c for c in columns if "task" in c])
print(connection.execute("SELECT task FROM episodes").fetchall())
```

Observed:

```
['task', 'task_1']
[('fold_napkin',)]      # the metadata; the measurement said 99.0
```

The measurement is stored correctly in the long `measurements` table — only the wide view misroutes it.

## Why it matters

A wrong answer rather than a loud one. A curation query filtering on `task` silently reads metadata instead of the measurement, and nothing in the report says a column was renamed. The measurement is only reachable under a name the user never chose.

Related: two *checks* claiming one key is now refused at run time, but a key claiming an *episode column* still is not.

## Note on the docs

`docs/CATALOG.md` and `src/hflow/curation.py`'s module docstring both stated this "is an error". That was inaccurate and is corrected on the branch that files this issue — the accurate behaviour is the silent rename above.

## Suggested fix

Refuse the collision where the runner already refuses cross-step key collisions, naming the offending key and the column it shadows. The reserved set is the `episodes` view's own column list, so it can be derived rather than hardcoded.

Sagar-024:
Closes #130

A measurement key named like an episodes column (`task`, `uri`,
`pipeline_version`, ...) pivoted beside the real thing, and DuckDB resolved
the collision by silently renaming the measurement to `<key>_1`. Nothing
errored; `SELECT task` just returned the metadata forever while the
measurement sat behind a name no query would ask for. A wrong answer rather
than a loud one.

The append now refuses the run before anything is written, naming the
check, the key, and the column it shadows -- the same boundary treatment
measurement values got in #127. Keys compare case-insensitively, since
DuckDB identifiers do.

The reserved set is derived, not listed: it comes from the episodes table
DDL plus the wide view's one derived column (`status`), whose name now
lives in one place (`EPISODES_VIEW_STATUS_COLUMN`) and is interpolated into
both view definitions in curation.py. A future episode column is guarded
the moment it enters the DDL.

Docs updated to match the new behavior (curation.py module docstring,
CATALOG.md current-limits): appends written before this check keep their
renamed ghost columns -- append-only -- and such values remain queryable in
the long `measurements` table.

Validation:

    uv run ruff check src tests
    uv run ruff format --check src tests
    uv run ty check src tests
    uv run pytest -q          # 720 passed, 3 skipped

The issue's repro now raises at append time instead of printing
`['task', 'task_1']`.

### DIFF
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -344,12 +344,13 @@ itself. A catalog root contains the `format_version` marker.
   0/1). Text-valued measurements stay in the long `measurements` table;
   query them there.
 - A measurement key that collides with an episode column name (`task`,
-  `uri`, `pipeline_version`, ...) is **silently renamed** in the wide view, not
-  refused: the episode column keeps the name and the measurement is reachable
-  only as `task_1`, which no query would think to ask for. `SELECT task` then
-  returns the metadata and never the measurement. Follow the
-  [naming rules](#naming-measurement-keys) and it cannot arise; query the long
-  `measurements` table if you hit it in an existing corpus.
+  `uri`, `pipeline_version`, ...) or with the derived `status` column is
+  **refused at append time**. Pivoted beside the real column, DuckDB would
+  rename it to `task_1`, so `SELECT task` would silently return the metadata
+  instead of the measurement. Follow the
+  [naming rules](#naming-measurement-keys) and the collision cannot arise;
+  appends written before this check keep their renamed ghost columns -- query
+  such values in the long `measurements` table.
 - The wide view binds its measurement columns to the keys present when the
   connection was opened. `curate()` reopens per call, so this only matters
   if you hold a long-lived connection from `open_catalog_connection()` while
--- a/src/hflow/catalog.py
+++ b/src/hflow/catalog.py
@@ -103,6 +103,23 @@ _reconciled_append_stems: set[tuple[str, str]] = set()
 
 _FORMAT_MARKER_NAME = "format_version"
 
+# The quarantine state rendered as a queryable column by the curation
+# views (curation.py builds both episodes views with it). Declared here
+# because measurement keys are validated against the episodes view's full
+# column list -- the table's columns plus this derived one.
+EPISODES_VIEW_STATUS_COLUMN = "status"
+
+# Lowercased name -> canonical spelling of every column a measurement key
+# must not claim. Derived from the episodes DDL plus the derived column
+# above, so a new episode column is guarded without touching the check.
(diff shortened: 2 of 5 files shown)


kstonekuan (approved):
LGTM. Validated locally: ruff, format, ty clean, 720 passed, 3 skipped. The refusal lands before the fingerprint and before any write, and the end-to-end test asserts the episodes directory stays empty, so a refused append leaves nothing behind.

I confirmed the case-insensitive claim rather than taking it on trust, since it decides whether the lowercasing is over-broad. Pivoting a `Task` key beside `e.task` gives columns `[episode_id, 'task', 'Task_1']` and `SELECT task` returns the metadata. It is the same silent wrong answer, so refusing case-insensitively is right.

Two facts for whoever reads this later, neither a change request:

- `quarantined` is in the reserved set but is `EXCLUDE`d from the wide view, so a key by that name would not actually collide. Reserving it anyway is the better default: it is what `status` is derived from, and it keeps the rule "an episodes column name is not yours" true without exceptions.
- Deriving the set from `TABLE_COLUMN_DDL["'episodes"'].split(",")` assumes no column type carries a comma. True today, and a `DECIMAL(10,2)` would still guard the real column while adding a junk entry, so it degrades harmlessly. Worth remembering if the DDL ever grows a parameterized type.

Nice touch putting `EPISODES_VIEW_STATUS_COLUMN` in one place so both view definitions and the reserved set read from it.

PR #134, opened 2026-08-23
fix: reserve the built-in media step name

### LINKED ISSUE #129: A user step can take the built-in media step's name, writing two check_runs rows under one name
## What happens

`App._registered_step_names()` (`src/hflow/app.py`) collects the names of registered checks and enrichments so a duplicate is refused at registration. It does not reserve `MEDIA_CONTACT_SHEET_STEP_NAME` (`"media/contact_sheet"`), which `App.process()` constructs on the fly for the media stage.

So a user check named `media/contact_sheet` registers happily, and one run then writes **two `check_runs` rows under one `check_name`** — the user's check and the built-in media step.

## Reproduction

```python
import hflow
from hflow.app import MEDIA_CONTACT_SHEET_STEP_NAME

app = hflow.App("collide-with-media", data_root="./data")

@app.check(name=MEDIA_CONTACT_SHEET_STEP_NAME)
def shadows_the_media_step(ep: hflow.Episode) -> hflow.CheckResult:
    return hflow.CheckResult(measurements={"mine": 1})

episode = synthesize_episode(
    "episode.mcap", SyntheticEpisodeSpec(duration_s=2.0, cameras=("wrist_cam",))
)
app.process(episode, record=True)

connection = hflow.open_catalog_connection("./data/catalog")
print(connection.execute(
    "SELECT check_name, count(*) FROM check_runs GROUP BY check_name"
).fetchall())
```

Observed:

```
registration accepted: ['media/contact_sheet']
check_runs rows: [('media/contact_sheet', 2)]
```

## Why it matters

`check_name` is the unit of coverage reporting: `CurationReport` groups by it to answer "which checks ran on what fraction of episodes". Two producers under one name make that count wrong, and there is no signal that it happened. It is the same class of problem as two steps claiming one measurement key, which the runner now refuses.

## Suggested fix

Reserve the built-in name in `_registered_step_names()` so registration refuses it, in the same shape as the existing duplicate-step-name refusal. One line plus a test.

k4its1t:
## Summary

- reserve the built-in `media/contact_sheet` step name during user step registration
- reject both checks and enrichments that try to use the reserved name
- cover the public registration behavior with a regression test

## Why

The media stage creates `media/contact_sheet` at processing time. Allowing a user step to register with the same name can therefore write two `check_runs` rows under one `check_name`, which makes coverage reporting ambiguous. Reserving the name at registration prevents that invalid pipeline configuration before any data is processed.

Closes #129

## Validation

- `uv run pytest tests/test_enrichments.py -q` — 7 passed
- `uv run pytest -q` — 715 passed, 6 skipped
- `uv run ruff format` — 145 files left unchanged
- `uv run ty check` — passed

## Checklist

- [x] I added outcome-focused regression coverage for the changed registration behavior.
- [x] Documentation is unchanged because this restores the existing unique-step-name contract without adding a flag, format, or operational requirement.
- [x] I ran the relevant pytest suite and the full default suite.
- [x] Stored-data compatibility is preserved; invalid duplicate names are rejected before processing.

## AI assistance

OpenAI Codex assisted with implementation and PR preparation. The final diff was reviewed and all validation commands above were run locally.

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -751,11 +751,13 @@ class App:
         self.transform_override: TransformFunction | None = None
 
     def _registered_step_names(self) -> set[str]:
-        # Checks and enrichments share the catalog's check_name column, so
-        # names are unique across both.
-        return {registered.name for registered in self.checks} | {
-            registered.name for registered in self.enrichments
-        }
+        # Checks, enrichments, and the built-in media step share the catalog's
+        # check_name column, so names are unique across all three.
+        return (
+            {MEDIA_CONTACT_SHEET_STEP_NAME}
+            | {registered.name for registered in self.checks}
+            | {registered.name for registered in self.enrichments}
+        )
 
     @property
     def pipeline_version(self) -> str:
--- a/tests/test_enrichments.py
+++ b/tests/test_enrichments.py
@@ -7,6 +7,7 @@ from typing import cast
 import pytest
 
 import hflow
+from hflow.app import MEDIA_CONTACT_SHEET_STEP_NAME
 from hflow.curation import open_catalog_connection
 from hflow.testing import SyntheticEpisodeSpec, synthesize_episode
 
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM. Validated locally on top of main: ruff, format, ty clean, 718 passed, 3 skipped.

Checked the one thing that could have gone wrong: `_registered_step_names()` has exactly two call sites (`app.py:840`, `app.py:897`), both the duplicate-name refusal, so seeding it with the built-in name reserves the name without leaking into `pipeline_version` or the manifest.

Thanks for the clean first PR. If you want another, the good first issue list is here: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

PR #143, opened 2026-08-24
fix(cli): answer bad serve launch input with exit 2

### LINKED ISSUE #124: CLI: hflow serve answers bad launch input with a traceback and exit 1; ServerSettings accepts a non-int port
### What happens

`hflow serve` is the only command that answers bad launch input with a traceback and exit 1. Every other command prints one line and exits 2.

Measured on `main` (`3e1ff99`), same harness across all of them:

```
  serve --port 99999           exit=1  TRACEBACK
  up --api-port 99999          exit=2  up: api_port 99999 is not in 1-65535
  status bad bundle            exit=2  status: no rendered bundle at /tmp/not-a-bundle (missing ...)
  stale bad catalog            exit=2  stale: importing /tmp/nope.py failed: [Errno 2] ...
  manifest missing pipeline    exit=2  manifest: importing /tmp/nope.py failed: [Errno 2] ...
  doctor missing file          exit=2  doctor: /tmp/nope.mcap
  deploy missing pipeline      exit=2  deploy: [Errno 2] No such file or directory: '/tmp/nope.py'
```

Three inputs reproduce it, and they raise two different exceptions:

```
hflow serve --port 99999   -> ValueError: port 99999 is not in 1-65535
hflow serve --port 0       -> ValueError: port 0 is not in 1-65535
hflow serve --host not-a-host --port 4401
                           -> RuntimeError: no free port between 4401 and 4410 on not-a-host
```

The first two come from `ServerSettings.__post_init__`
(`packages/hflow-server/src/hflow_server/_settings.py:62`), the third from the port
probe (`packages/hflow-server/src/hflow_server/server.py:471`). `_command_serve`
(`src/hflow/cli.py:765`) wraps neither the `ServerSettings(...)` construction at
`cli.py:776` nor `serve(settings)` at `cli.py:784`, so both escape to the interpreter.

Exit code 1 is wrong here for the same reason it was wrong for `up` in #90: 1 means the
thing started and then failed, 2 means bad input and nothing happened. Nothing is
serving in any of these cases.

### Second half: the port type check never made it across

`ServerSettings.port` only has the range check. Its own comment
(`_settings.py:22-28`) says it mirrors "the same range ... that `hflow.runtime`'s
RuntimeConfig enforces", and it does, but it mirrors `RuntimeConfig` as of before #89,
which added the type check. So the two have drifted:

```
ServerSettings(port=True)     -> ACCEPTED, url=http://127.0.0.1:True
ServerSettings(port=8080.0)   -> ACCEPTED, url=http://127.0.0.1:8080.0
ServerSettings(port="4356")   -> TypeError: '<=' not supported between 'int' and 'str'

RuntimeConfig(api_port=True)    -> ValueError: api_port must be an int, not bool: True
RuntimeConfig(api_port=8080.0)  -> ValueError: api_port must be an int, not float: 8080.0
RuntimeConfig(api_port="4356")  -> ValueError: api_port must be an int, not str: '4356'
```

`bool` subclasses `int`, so `True` satisfies `1 <= port <= 65535` and is then
interpolated into the URL the server prints and hands to the browser. That is the
exact defect #85 described and #89 fixed, in the other package.

There is currently no test of `ServerSettings` port validation at all, in either
direction.

### Definition of done

1. `hflow serve` with an out-of-range port, port 0, or an unusable host prints one line
   to stderr prefixed `serve: ` and exits 2, with no traceback.
2. `ServerSettings` refuses a non-int port at construction with `ValueError`, matching
   `RuntimeConfig`'s message shape, and keeps accepting an `IntEnum` member (see #89 for
   why `isinstance` rather than `type(...) is int`).
3. Tests cover both: the CLI exit code and message, and the constructor for `bool`,
   `float`, `str`, and the accepted `int`/`IntEnum` cases.

Two things to get right:

- **`ValueError` for the type failure, not `TypeError`.** Whatever handler you add will
  be catching `ValueError`, and #89 settled this for the same field in the other package.
- **The host failure is a `RuntimeError` from the probe, not a validation error.** Decide
  whether to catch it in `_command_serve` or to validate the host earlier; catching is
  simpler and matches how `up` handles `ComposeError`. Do not widen the handler so far
  that a genuine mid-run server crash also becomes exit 2.

Once this lands, add `serve` to the exit-code section in `docs/FORMAT.md` (added in #121),
which currently names `doctor`, `stale`, `up`, and `ingest`.

### Pattern to copy

`_command_up` (`src/hflow/cli.py`) is the precedent for both halves: it wraps only the
`RuntimeConfig(...)` construction in `try/except ValueError -> return 2` (from #84), and
#95 added a separate `except FileNotFoundError` block rather than widening an existing
tuple. `RuntimeConfig.__post_init__` in `src/hflow/runtime/_bundle.py` is the type-check
to mirror, comment included. Server tests live in `packages/hflow-server/tests/`; CLI
exit-code tests live in `tests/test_runtime_cli.py`.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Not a starter issue: it crosses the `hflow` and `hflow-server` package boundary and the
right answer depends on a precedent set in a different package. Comment here and I will
assign it.

chiruu12:
Closes #124.

## What changed

`serve` was the only command answering bad launch input with a traceback and exit 1. All three
reported inputs now print one line to stderr and exit 2, and so does a fourth I hit while
reproducing.

```
serve --port 99999                      exit=2  serve: port 99999 is not in 1-65535
serve --port 0                          exit=2  serve: port 0 is not in 1-65535
serve --host not-a-host --port 4512     exit=2  serve: no free port between 4512 and 4521 on not-a-host
serve --data-root <a file>              exit=2  serve: Not a directory: /tmp/ws124file
serve --data-root <missing>             exit=2  serve: No such file or directory: /tmp/nope
```

The last two are the data-root half you mentioned. Before this, a data root that was a regular file
or was not there at all started the server, printed a URL, and served an empty workspace with
nothing said about why it was empty.

## The three pieces

**`ServerSettings.port` type check.** Mirrors `RuntimeConfig.api_port` from #89, message shape
included, and ordered before the range test for two reasons. `bool` subclasses `int`, so `True`
passes the range and reaches the URL as `http://127.0.0.1:True`. A `str` is worse: it raises
`TypeError` out of the comparison, and the handler I added catches `ValueError`, so a bad `--port`
would still escape as a traceback. `isinstance` rather than `type(...) is int` keeps `IntEnum`
working, same as #89.

**`_command_serve` guards.** Two separate blocks, following `_command_up`. `ValueError` around the
`ServerSettings(...)` construction, and the data-root check after it.

**`ServerStartupError`.** The issue warns against widening the handler until a mid-run server crash
also reads as bad input, and a bare `except RuntimeError` around `serve(settings)` does exactly
that. So the free-port probe now raises `ServerStartupError(RuntimeError)` instead. The probe runs
before `uvicorn.run` binds, so that type means "never got off the ground" and nothing else does. It
subclasses `RuntimeError` so anything already catching that is unaffected. There is a test asserting
a `RuntimeError` from a running server still propagates rather than becoming exit 2.

## One thing to check me on

I treated a **missing** data root as bad input alongside a file, since neither is a directory. That
made me change one line of an existing test: `test_ui_flags_land_in_the_launch_settings` passed
`tmp_path / "workspace"` without creating it. That test is about flags reaching the launch and its
sibling on the next case already passes a real `tmp_path`, so I created the directory rather than
weakening the check. If you would rather a missing root still serve, say so and I will narrow it to
`ENOTDIR` only and revert that line.

`docs/FORMAT.md` now names `serve` in the exit-code section.

## Validation

```
uv sync --locked --all-extras
uv run ruff check          All checks passed!
uv run ruff format --check 152 files already formatted
uv run ty check            All checks passed!
uv run pytest -q           808 passed, 9 skipped
```

Also ran each of the five commands above by hand against a real workspace to confirm the exit codes
and that no traceback reaches stderr.

### DIFF
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -157,8 +157,9 @@ Every `hflow` command follows the same three-value convention:
   `stale --exit-code`, stale episodes found; `up`, started then failed, so
   containers may still be running; `ingest`, episodes failed.
 - `2` - bad input, nothing useful happened: an invalid flag combination, an
-  unreachable endpoint, a missing catalog, or a `doctor` run in which no file
-  could be diagnosed at all.
+  unreachable endpoint, a missing catalog, a `serve` launch that never started
+  (a port outside 1-65535, a data root that is not a directory, a host with no
+  free port), or a `doctor` run in which no file could be diagnosed at all.
 
 `doctor` treats an unreadable file as at least as bad as a non-conforming one
 (exit 1), and reserves 2 for runs where nothing was diagnosed, so a batch run
--- a/packages/hflow-server/src/hflow_server/__init__.py
+++ b/packages/hflow-server/src/hflow_server/__init__.py
@@ -6,8 +6,8 @@ subcommand is a thin caller of exactly these two names.
 """
 
 from hflow_server._settings import ServerSettings
-from hflow_server.server import create_app, serve
+from hflow_server.server import ServerStartupError, create_app, serve
 
 __version__ = "0.1.0"
 
-__all__ = ["ServerSettings", "__version__", "create_app", "serve"]
+__all__ = ["ServerSettings", "ServerStartupError", "__version__", "create_app", "serve"]
--- a/packages/hflow-server/src/hflow_server/_settings.py
+++ b/packages/hflow-server/src/hflow_server/_settings.py
@@ -59,6 +59,17 @@ class ServerSettings:
         # the command line. Left to bind(2) instead, an out-of-range port
         # surfaces as an OverflowError from inside the port probe, and port 0
         # binds fine while printing a URL nobody can open.
+        # Type first, because the range test cannot do it. bool subclasses int,
+        # so True is 1 to a comparison and would pass the range on its way into
+        # the URL this launch prints and hands to the browser. Ordered first for
+        # a second reason: a str port reaching the range test raises TypeError
+        # from the comparison, and every caller here is catching ValueError.
+        #
+        # `isinstance` and not `type(...) is int`, so an IntEnum member is
+        # accepted; this mirrors RuntimeConfig.api_port in `hflow.runtime`,
+        # message shape included, because the two fields are the same field.
+        if not isinstance(self.port, int) or isinstance(self.port, bool):
+            raise ValueError(f"port must be an int, not {type(self.port).__name__}: {self.port!r}")
         if not MIN_PORT <= self.port <= MAX_PORT:
             raise ValueError(f"port {self.port!r} is not in {MIN_PORT}-{MAX_PORT}")
 
--- a/packages/hflow-server/src/hflow_server/server.py
+++ b/packages/hflow-server/src/hflow_server/server.py
@@ -428,6 +428,17 @@ def _contained_asset_response(assets_directory: Path, requested_path: str) -> Fi
     return FileResponse(resolved_candidate)
 
 
+class ServerStartupError(RuntimeError):
+    """The launch could not be started, and nothing is serving.
+
+    Its own type so ``hflow serve`` can answer a launch that never got off the
+    ground with exit 2 (bad input, nothing happened) without also catching a
+    RuntimeError raised out of ``uvicorn.run`` once the server is up, which is
+    exit 1 (started, then failed). Subclasses RuntimeError so callers that
+    already catch that keep working.
+    """
+
+
 def serve(settings: ServerSettings) -> None:
     """Run the workspace server: free port, printed URL, browser, uvicorn."""
     application = create_app(settings)
@@ -470,4 +481,6 @@ def _first_free_port(host: str, preferred_port: int) -> int:
             except OSError:
                 continue
         return candidate_port
-    raise RuntimeError(f"no free port between {preferred_port} and {last_candidate_port} on {host}")
+    raise ServerStartupError(
+        f"no free port between {preferred_port} and {last_candidate_port} on {host}"
+    )
--- a/packages/hflow-server/tests/test_server_settings.py
+++ b/packages/hflow-server/tests/test_server_settings.py
@@ -1,5 +1,7 @@
 """ServerSettings: the launch values parsed where they are set."""
 
+from enum import IntEnum
+
 import pytest
 from hflow_server import ServerSettings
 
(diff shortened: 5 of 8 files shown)


kstonekuan (approved):
LGTM. Validated on current main: ruff, format, ty clean, 811 passed, 6 skipped. Merging.

`ServerStartupError` is the right answer to the widening problem, and raising it from `_first_free_port` rather than around the `serve` call is what makes the type mean something: the probe runs before uvicorn binds, so "never got off the ground" is a property of where it is raised, not of what the handler chose to catch. `test_serve_does_not_turn_a_running_server_crash_into_bad_input` is the test that keeps that true.

Ordering the type check before the range check is correct for both reasons you give. The `str` one is the one that would actually have bitten: `"4356" >= MIN_PORT` raises `TypeError`, and your handler catches `ValueError`, so without the reorder a string port still escapes as a traceback and the fix would have missed its own case.

## On the missing data root

Merging as written, and flagging the `ENOENT` half for @kstonekuan rather than deciding it myself, because it is a first-run question and the UI work is live right now.

The fact that decides it: nothing creates the data root eagerly. `catalog.py:463` makes it on first append, so on a fresh clone `./data` does not exist until something ingests, and `hflow serve` today serves an empty workspace over it. The server is built for that state deliberately, but only the *empty* version of it: `empty_workspace_api` in the server tests is a data root that exists and has no catalog. Nothing covers absent.

So the two halves are not the same call. A data root that is a regular file is bad input under any reading. A missing one is bad input only if `serve` is not meant to be a valid first command. If it is, this turns "install, serve, look around" into exit 2.

Your instinct to create the directory in `test_ui_flags_land_in_the_launch_settings` rather than weaken the assertion was right either way. If the narrowing to `ENOTDIR` is wanted, it is one line plus that test line, and I will do it rather than send it back to you.

Eight merged PRs now, and this one picked up a fourth case the issue did not name. If you want something with more room in it than a bad-input fix: run HFlow against a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face) and report what breaks or drags. Discussion #48 is Sagar-024 doing that, and it was worth more than any issue on the list.

Issue #145, opened 2026-08-24
CLI: curate --sql-file and up --data-root answer bad path input with a traceback and exit 1


chiruu12:
### What happens

Two commands still answer bad path input with a traceback and exit 1. Measured on `ccaa8ad`,
same harness as #124:

```
curate --sql-file <a directory>    exit=1  IsADirectoryError: [Errno 21] Is a directory: '/tmp/probe/adir'
up --data-root <a file>            exit=1  NotADirectoryError: [Errno 20] Not a directory: '/tmp/probe/afile/runtime/dags'
```

Every other path flag on those commands prints one line and exits 2.

### curate --sql-file

`_command_curate` (`cli.py:788`) reads the file inside a try that catches
`(ValueError, FileNotFoundError)`:

```python
sql = arguments.sql if arguments.sql is not None else arguments.sql_file.read_text()
```

`read_text()` on a directory raises `IsADirectoryError`, which subclasses `OSError` and not
`FileNotFoundError`, so it walks straight past the handler at `cli.py:799`.

Note this is the same class boundary #102 called out, seen from the other side. There the
fix was to keep raising `FileNotFoundError` so existing handlers keep working. Here the
exception comes from `pathlib` rather than from our code, so the handler is what has to
move.

### up --data-root

A `--data-root` that is a regular file reaches `render_bundle` and fails inside `mkdir`
(`_bundle.py:876`):

```python
for directory in (dags_dir, logs_dir, user_dir):
    directory.mkdir(parents=True, exist_ok=True)
```

The default bundle dir is `<data-root>/runtime`, so a file data root makes every one of
those three `mkdir` calls raise `NotADirectoryError`. `_command_up` catches
`FileNotFoundError` and `(ComposeError, TimeoutError)`, neither of which covers it.

Nothing has been rendered or started when this happens, so it is bad input, exit 2, by the
same reasoning as #84, #90 and #95.

### Definition of done

1. `curate --sql-file` pointed at a directory prints one line prefixed `curate: ` and exits
   2, with no traceback.
2. `up --data-root` pointed at a regular file prints one line prefixed `up: ` and exits 2,
   with no traceback, before anything is rendered.
3. Tests cover both, in `tests/test_runtime_cli.py` beside the existing exit-code cases.

### Two things to get right

- **Do not widen either handler to bare `OSError`.** That would fold a genuine mid-run
  filesystem failure into "bad input" and exit 2. `up` in particular has a deliberate exit 1
[...]

PR #146, opened 2026-08-24
fix(storage): say 'Is a directory' at the two remaining sites

### LINKED ISSUE #144: storage: fetch_uri and LocalStorageRoot.fetch report a directory as 'No such file or directory'
### What happens

`storage.py` reports a directory as absent. Measured on `ccaa8ad`:

```
$ ls -d /tmp/probe/adir
/tmp/probe/adir
$ hflow doctor /tmp/probe/adir
doctor: /tmp/probe/adir
  [error] unreadable: [Errno 2] No such file or directory: '/tmp/probe/adir'
  verdict: NOT CONFORMING
$ echo $?
2
```

The path exists. It is a directory.

### Where

Two sites, both testing `is_file()` and raising `ENOENT` on the false branch, so a path
that exists but is the wrong kind gets the message for a path that is absent:

- `storage.py:344`, `LocalStorageRoot.fetch`
- `storage.py:656`, `fetch_uri`

```python
if not local_file.is_file():
    raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), str(local_file))
```

`fetch_uri` is the one `hflow doctor` reaches, which is how this surfaces on the command
line. `LocalStorageRoot.fetch` has the same shape and is reached by anything reading a
named file out of a data root.

### Why this is the residual of #102, not a new problem

#102 described exactly this defect and enumerated the sites it covered: `_bundle.py:861`
and `:896`, `_deploy.py:153` and `:169`. Those four are fixed and carry the reasoning in
comments. The two `storage.py` sites were never in that list, so they still raise `ENOENT`
for a directory.

The comment left at `_bundle.py:868` says "Three-argument form, as `storage.py` does",
so `storage.py` is already the file the fixed sites point at as precedent. It is the one
that did not get the fix.

### Definition of done

`fetch_uri` and `LocalStorageRoot.fetch` raise `EISDIR` when the path is a directory, and
keep `ENOENT` for a path that is absent. Both keep the `FileNotFoundError` class.

### The class stays FileNotFoundError

For the reason #102 settled and #120 carried into the four fixed sites: `IsADirectoryError`
subclasses `OSError`, not `FileNotFoundError`, so every existing `except FileNotFoundError`
stops catching it. `_command_doctor` at `cli.py:835` is one such handler, and the accurate
class would turn this into a traceback instead of a finding. The errno carries the accurate
text without moving the class:

```python
if local_file.is_dir():
    raise FileNotFoundError(errno.EISDIR, os.strerror(errno.EISDIR), str(local_file))
```

That prints `[Errno 21] Is a directory: '/tmp/probe/adir'`.

### Notes

Small and mechanical, and the decision it depends on is already made and documented in the
four sites #120 fixed. Tests for the doctor path live in `tests/`, storage tests in
`tests/test_storage.py`.

I am happy to do this one and will wait to be assigned first.

chiruu12:
Closes #144.

`storage.py` had two sites testing `is_file()` and raising `ENOENT` on the false
branch, so a path that exists but is a directory got the message for a path that
is absent. `hflow doctor <a directory>` is the one that surfaces on the command
line:

    $ hflow doctor /tmp/probe/adir
      [error] unreadable: [Errno 2] No such file or directory: '/tmp/probe/adir'

Now:

    $ hflow doctor /tmp/probe/adir
      [error] unreadable: [Errno 21] Is a directory: '/tmp/probe/adir'

Exit code is unchanged at 2.

Both sites keep `FileNotFoundError`. `IsADirectoryError` subclasses `OSError`
and not `FileNotFoundError`, so the accurate class would slip past every
existing handler. `cli.py:835` is one of them, and it is what turns this into a
doctor finding rather than a traceback. Sixteen more name the class under
`src/hflow/`. That is the decision #102 settled and #120 carried into the four
sites it covered; these two were never in that list.

`ENOENT` is unchanged for a path that is absent, and there is no caller in the
repo that branches on `errno`, so nothing beyond the message moves.

Two tests, one per site, placed beside the `ENOENT` tests they mirror. Both fail
on the unfixed code with `21 != 2` and pass on the fix. Full suite 593 passed,
9 skipped. `ruff check`, `ruff format --check` and `ty check` clean.

### DIFF
--- a/src/hflow/storage.py
+++ b/src/hflow/storage.py
@@ -343,9 +343,9 @@ class LocalStorageRoot:
         if local_file.is_dir():
             # Deliberately FileNotFoundError and not IsADirectoryError: the accurate
             # class subclasses OSError, not FileNotFoundError, so an `except
-            # FileNotFoundError` stops catching it. cli.py:835 is one such handler,
-            # reached from `hflow doctor` through fetch_uri below, and it turns this
-            # into a finding rather than a traceback; sixteen more name the class
+            # FileNotFoundError` stops catching it. `_command_doctor` is one such
+            # handler, reached through fetch_uri below, and it turns this into a
+            # finding rather than a traceback; sixteen more name the class
             # elsewhere under src/hflow/, and callers outside the repo are the real
             # unknown. The errno carries the accurate text without moving the class.
             # Same decision as the four sites #102 covered. #144.
--- a/src/hflow/storage.py
+++ b/src/hflow/storage.py
@@ -340,6 +340,16 @@ class LocalStorageRoot:
     def fetch(self, relative: str) -> Path:
         """The local file at ``relative`` (it already lives here)."""
         local_file = self.path / _validated_relative_key(relative)
+        if local_file.is_dir():
+            # Deliberately FileNotFoundError and not IsADirectoryError: the accurate
+            # class subclasses OSError, not FileNotFoundError, so an `except
+            # FileNotFoundError` stops catching it. cli.py:835 is one such handler,
+            # reached from `hflow doctor` through fetch_uri below, and it turns this
+            # into a finding rather than a traceback; sixteen more name the class
+            # elsewhere under src/hflow/, and callers outside the repo are the real
+            # unknown. The errno carries the accurate text without moving the class.
+            # Same decision as the four sites #102 covered. #144.
+            raise FileNotFoundError(errno.EISDIR, os.strerror(errno.EISDIR), str(local_file))
         if not local_file.is_file():
             raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), str(local_file))
         return local_file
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM. Validated on current main: ruff, format, ty clean, 813 passed, 6 skipped.

Both sites match the spec, the `is_dir()` test lands before the `is_file()` test so `ENOENT` is untouched for an absent path, and the two tests assert `errno` rather than only the message, which is what makes them mean something.

Pushed one fixup (`ee7ab89`): the comment named `cli.py:835`, and a line number in a checked-in comment goes stale the moment anything above it moves. It names `_command_doctor` now.

One thing on your numbers: 593 is `pytest tests`, which misses `packages/hflow-server/tests` entirely. Plain `uv run pytest -q` from the repo root collects both and gives 813. It did not matter here because this PR touches neither, but #143 did touch server code, so worth having in the muscle memory.

#145 is still yours and unblocked for the regular-file half.

PR #147, opened 2026-08-24
fix(cli): answer bad curate and up path input with exit 2

### LINKED ISSUE #145: CLI: curate --sql-file and up --data-root answer bad path input with a traceback and exit 1

chiruu12:
Closes #145.

Two commands answered bad path input with a traceback and exit 1. Both now print
one line and exit 2, like every other path flag on those commands.

    curate --sql-file <a directory>    was IsADirectoryError, exit 1
    up --data-root <a regular file>    was NotADirectoryError, exit 1

### curate

The read moves into its own block around the read alone, rather than widening
the handler below it. `read_text` on a directory raises `IsADirectoryError`,
which subclasses `OSError` and not `FileNotFoundError`, so it walked past the
existing handler.

That block catches `OSError`, which is only safe because of how narrow it is: it
spans one read of one path the caller named, and nothing `curate` itself does can
reach it. A mid-run filesystem failure is still an unhandled crash.

### up

The data root is checked before anything renders, not by catching the `mkdir`.
Catching the `mkdir` cannot tell a bad `--data-root` apart from a bundle dir that
is genuinely unwritable, and only the first of those is exit 2.

Only a root that exists and is not a directory is refused. A missing root is
deliberately left alone, per your note on the issue: `serve` refuses that case,
but whether it should is open on #143, and until that lands it is better to have
one command answering the case that is bad input under any reading than two
commands agreeing on an answer that may be withdrawn.

### Tests

Three, in `tests/test_runtime_cli.py` beside the existing exit-code cases.

The two defect tests fail on the unfixed code with the original traceback and
pass on the fix. The third, `test_up_still_accepts_a_data_root_that_is_not_there_yet`,
passes either way by design: it is a regression guard on the carve-out above, so
that a later change extending the check to a missing root has to fail a test
rather than pass quietly.

Full suite 596 passed, 9 skipped. `ruff check`, `ruff format --check` and
`ty check` clean.

### DIFF
--- a/src/hflow/cli.py
+++ b/src/hflow/cli.py
@@ -603,6 +603,23 @@ def _command_up(arguments: argparse.Namespace) -> int:
         # be false.
         print(f"up: {error}", file=sys.stderr)
         return 2
+    # A bucket data root has no local directory to check, the same distinction
+    # drawn for the bundle dir below. Only a root that exists and is not a
+    # directory is refused here: every one of the three `mkdir` calls in
+    # render_bundle raises NotADirectoryError against it, and nothing has been
+    # rendered or started when that happens, so it is bad input (2).
+    # A *missing* local root is deliberately left alone. `serve` refuses that
+    # case too, but whether it should is open on #143, and until that lands the
+    # two commands agreeing on the wrong answer is worse than only this one
+    # answering the case that is bad input under any reading.
+    if not is_bucket_url(arguments.data_root):
+        local_data_root = Path(arguments.data_root)
+        if local_data_root.exists() and not local_data_root.is_dir():
+            print(
+                f"up: {os.strerror(errno.ENOTDIR)}: {local_data_root}",
+                file=sys.stderr,
+            )
+            return 2
     # A bucket data root has no local directory to host the bundle: ./runtime.
     default_bundle_dir = (
         Path(RUNTIME_BUNDLE_DIRECTORY_NAME)
@@ -789,8 +806,22 @@ def _command_curate(arguments: argparse.Namespace) -> int:
     if (arguments.sql is None) == (arguments.sql_file is None):
         print("curate: pass exactly one of a SQL string or --sql-file", file=sys.stderr)
         return 2
+    if arguments.sql_file is not None:
+        try:
+            sql = arguments.sql_file.read_text()
+        except OSError as error:
+            # Its own block around the read alone, rather than widening the
+            # handler below. `read_text` on a directory raises
+            # IsADirectoryError, which subclasses OSError and not
+            # FileNotFoundError, so it walked past that handler. Catching
+            # OSError here is safe precisely because this block spans one
+            # read of one caller-named path: nothing curate does can reach it,
+            # so a mid-run filesystem failure is still an unhandled crash.
+            print(f"curate: {error}", file=sys.stderr)
+            return 2
+    else:
+        sql = arguments.sql
     try:
-        sql = arguments.sql if arguments.sql is not None else arguments.sql_file.read_text()
         report = curate(
             arguments.catalog,
             sql,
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM. Validated on current main: ruff, format, ty clean, 816 passed, 6 skipped.

The `up` carve-out is better than what I suggested. I said do the regular-file case and leave the missing one; you did that and added `test_up_still_accepts_a_data_root_that_is_not_there_yet`, which passes either way today and exists so that extending the check later has to fail a test rather than pass quietly. That turns a temporary carve-out into something with a tripwire on it, which is what makes it safe to leave open.

Catching `OSError` in the `curate` block is the right call for the reason you name: it spans one read of one caller-named path, so widening it there costs nothing a narrower catch would have protected. That is a different situation from the `RuntimeError` trap on #124, and you drew the line in the right place both times.

Issue #64, opened 2026-08-20
catalog.py: remove the dead `latest_quarantine_state()` helper (zero callers)


kstonekuan:
`latest_quarantine_state()` has no callers. Removing it is a small, provable
cleanup.

## Current behavior

`src/hflow/catalog.py:207-216` defines `latest_quarantine_state`, a thin
`bool | None` wrapper over `latest_quarantine`.

A repository-wide search finds only the definition:

```bash
rg -n "latest_quarantine_state" src tests docs examples README.md
```

The helper is not exported from `hflow.__init__`, has no test, and appears in no
documentation. The real consumer at `src/hflow/app.py:1119-1123` uses the richer
`latest_quarantine` result directly.

## Pattern to copy

No replacement is needed. Delete the unused function and use the search above
to confirm that nothing references it.

## Definition of done

- `latest_quarantine_state` no longer exists in the repository.
- No other behavior changes.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

WilliamK112:
I’d like to take this. I’ll keep the change scoped to removing the unused `latest_quarantine_state()` helper, confirm there are no remaining references, and run the repository’s lint, type-check, and test commands.

WilliamK112:
Update after checking the latest `main`: this cleanup has already landed as part of #116 (commit `87154c1`). That change removed `latest_quarantine_state()`, and a repository-wide search now finds no remaining definition or reference. There is no meaningful diff left for a PR, so this issue can be closed as already resolved.

kstonekuan:
Closing: you are right, and thank you for checking main before writing the patch. `87154c1` (#116) removed `latest_quarantine_state`, and it appears nowhere in `src/`, `packages/`, `tests/`, or `docs/` today. The issue outlived its defect and nobody noticed.

Checking first is the more useful contribution here, so it is not wasted effort. #66 is yours.

PR #148, opened 2026-08-24
test: cover plan_batches lower bounds

### LINKED ISSUE #66: Tests: `plan_batches` lower-bound validation has no coverage
`plan_batches` validates both lower bounds, but no test pins either branch.

## Current behavior

`src/hflow/batching.py` raises when:

- `batch_count < 1` at `src/hflow/batching.py:64-65`
- `target_batch_bytes < 1` at `src/hflow/batching.py:79-80`

`test_argument_validation` at `tests/test_batching.py:51-59` covers the
exactly-one rule and negative item sizes, but not these lower bounds.

## Pattern to copy

Add two `pytest.raises` blocks to `test_argument_validation`, following the
existing assertions in the same test. Use nonempty `item_sizes`,
`batch_count=0`, and `target_batch_bytes=0` so execution reaches both branches.

## Definition of done

- Tests exercise both lower-bound branches.
- Deleting either `raise` in `src/hflow/batching.py` makes a test fail.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

WilliamK112:
## Summary

- cover the `batch_count < 1` validation branch with a nonempty input
- cover the `target_batch_bytes < 1` validation branch with a nonempty input
- keep production behavior unchanged

Closes #66.

## Validation

- `uv run ruff check --fix`
- `uv run ruff format`
- `uv run ty check`
- `uv run pytest tests/test_batching.py -q` (7 passed)
- `uv run pytest -q` (810 passed, 9 skipped)

### DIFF
--- a/tests/test_batching.py
+++ b/tests/test_batching.py
@@ -55,6 +55,10 @@ def test_argument_validation() -> None:
         plan_batches({"a": 1}, batch_count=1, target_batch_bytes=1)
     with pytest.raises(ValueError, match="negative"):
         plan_batches({"a": -1}, batch_count=1)
+    with pytest.raises(ValueError, match="batch_count must be >= 1, got 0"):
+        plan_batches({"a": 1}, batch_count=0)
+    with pytest.raises(ValueError, match="target_batch_bytes must be >= 1, got 0"):
+        plan_batches({"a": 1}, target_batch_bytes=0)
     assert plan_batches({}, batch_count=3) == []
 
 


kstonekuan (approved):
LGTM. Validated on current main: ruff, format, ty clean, 813 passed, 6 skipped.

Both assertions match the messages at `batching.py:67` and `:82` exactly, and using a nonempty `item_sizes` is the right choice: it proves the guard fires on a real plan rather than on the empty short-circuit.

Thank you also for checking #64 against main before writing anything. It had already landed in `87154c1` and I have closed it. Finding that is worth more than the patch would have been.

One invitation, not a rule: the `good first issue` pool is best left for newcomers, and you are past that now. If you want something with more in it, run HFlow against a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face) and report what breaks, what is slow, or what is awkward. Discussion #48 is what that looks like.

WilliamK112:
Thank you for the review and for the steer. That makes sense—I’ll leave the good-first-issue pool for newcomers.

I’ve read Discussion #48 and its follow-ups. The LeRobot converter and reproducible stress scripts are already owned and in flight, so I won’t duplicate that work. I’ll first check the current discussions and open PRs, then choose a distinct real-corpus or workload investigation and report reproducible observations about what breaks, what is slow, or what is awkward before proposing a code change.

I appreciate the direction.

PR #150, opened 2026-08-24
feat(resample): add quaternion slerp policy

### LINKED ISSUE #97: Resample: add policy="slerp" so rotation channels can be interpolated correctly

WilliamK112:
## Summary

- add public `policy="slerp"` support for exactly four-component rotation fields
- normalize source samples, reject non-finite/zero-norm inputs, and apply sequential hemisphere alignment before shortest-arc interpolation
- use normalized linear interpolation when `dot > 1 - 1e-7` to avoid the near-parallel slerp precision loss
- preserve component-order agnosticism (`xyzw` and `wxyz` both work) and normalize every output sample
- update quaternion-shaped `interpolate` errors to recommend `slerp`, and reject non-4-column fields explicitly
- bump `RESAMPLE_POLICY_VERSION` to `"2"`

## Regression coverage

- 180-degree midpoint and unit norm
- antipodal sign flip with exact-source rotational equality (`abs(dot) ≈ 1`)
- near-parallel normalized-lerp fallback
- explicit non-4-column rejection
- updated quaternion interpolation guidance

## Validation

- `ruff check .`
- `ruff format --check .`
- `ty check`
- full test suite: 817 passed, 9 skipped
- randomized 200-path smoke check: finite/unit outputs and exact-source rotational equivalence

Closes #97

Signed-off-by: WilliamK112 <164879897+WilliamK112@users.noreply.github.com>

### DIFF
--- a/src/hflow/format.py
+++ b/src/hflow/format.py
@@ -51,7 +51,7 @@ NANOSECONDS_PER_SECOND = 1_000_000_000
 # versioned: bump this when the grid or selection semantics change.
 # Annotated as ``str`` rather than inferred as a literal, for the same reason
 # TRANSFORM_BEHAVIOR_VERSION is: the whole point is that it changes.
-RESAMPLE_POLICY_VERSION: str = "1"
+RESAMPLE_POLICY_VERSION: str = "2"
 
 # ``provenance/v1`` keys written when derived channels are present:
 # ``derived/<topic>`` holds each derived channel's content-hash version, and
--- a/src/hflow/resample.py
+++ b/src/hflow/resample.py
@@ -6,7 +6,7 @@ versioned: ``hflow.format.RESAMPLE_POLICY_VERSION`` names the semantics
 below and is stamped into ``provenance/v1`` whenever derived channels are
 written.
 
-Resample policy version "1":
+Resample policy version "2":
 
 - The grid steps by ``1e9 / rate_hz`` nanoseconds and is anchored to integer
   multiples of the step (epoch-anchored): it runs from the first multiple at
@@ -17,6 +17,8 @@ Resample policy version "1":
 - ``interpolate``: linear interpolation per column (``numpy.interp``).
 - ``nearest``: the value of the source sample nearest in time to each grid
   point; the earlier sample wins an exact tie.
+- ``slerp``: shortest-arc spherical interpolation of four-component rotation
+  fields after source normalization and sequential hemisphere alignment.
 """
 
 import math
@@ -31,7 +33,9 @@ from hflow.format import NANOSECONDS_PER_SECOND
 if TYPE_CHECKING:
     from hflow.episode import ChannelData
 
-ResamplePolicy = Literal["interpolate", "nearest"]
+ResamplePolicy = Literal["interpolate", "nearest", "slerp"]
+
+_SLERP_NEAR_PARALLEL_DOT = 1.0 - 1e-7
 
 
 @dataclass(frozen=True)
@@ -120,10 +124,9 @@ def _source_columns(channel: "ChannelData", field: str | None) -> dict[str, np.n
 
 
 # Componentwise interpolation of quaternions produces unnormalized rotations
-# and is wrong across antipodal sign flips; refuse the obvious cases until a
-# proper slerp alignment policy lands (tracked follow-up). Name-based
-# detection is best-effort: it cannot catch every quaternion field, but it
-# catches the conventional ones (orientation/quaternion/rotation, 4 columns).
+# and is wrong across antipodal sign flips. Name-based detection is best-effort:
+# it cannot catch every quaternion field, but it catches the conventional ones
+# (orientation/quaternion/rotation, 4 columns).
 _QUATERNION_NAME_HINTS = ("quat", "orientation", "rotation")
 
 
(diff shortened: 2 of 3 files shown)


kstonekuan (approved):
LGTM. Validated on current main: ruff, format, ty clean, 820 passed, 6 skipped.

I checked the math against an independent textbook slerp rather than by reading yours. Over 2000 random quaternion pairs, a third of them sign-flipped, worst-case `1 - |dot|` was 2.2e-16, which is float64 noise. Angular velocity is constant to 9e-15 across a rotation, exact-source grid points come back as the same rotation, and every output is unit norm. The near-parallel threshold is also placed well: I walked half-angles across the crossover at 4.47e-4 and both branches are exact on either side, so there is no discontinuity where they meet.

The bracketing is right for a reason worth stating, since it is not obvious from the diff: `_grid_timestamps_ns` uses `ceil` on the first index and `floor` on the last, so the grid is always inside the source span and `later_indices` clipped to `len - 1` can never extrapolate past the final sample. A grid point landing exactly on a source timestamp gets `fraction == 1.0` and returns that sample. A single-sample channel degenerates to `earlier == later` and returns the normalized sample instead of dividing by zero.

## A correction to the issue, not to your PR

The issue told you the version bump was free:

> `stale_episodes` compares `pipeline_version` and `schema_version` only (`curation.py:382`), not `resample_policy_version`, so bumping it does **not** mark any existing corpus stale.

That is wrong, and you should know it before this lands. `stale_episodes` does compare `pipeline_version`, and `compute_pipeline_version` folds `RESAMPLE_POLICY_VERSION` into `pipeline_version` for any pipeline that has derived channels (`transform.py:175`). Measured on your branch:

```
 WITH derived channels: v1=937cb7b22d71  v2=901dcd4d84a7  CHANGED
   NO derived channels: v1=5fb7151b3d5c  v2=5fb7151b3d5c  unchanged
```

So corpora with derived channels do go stale, and because `pipeline_version` is stamped inside the bytes `episode_id` hashes, re-ingesting them re-mints episode identities. Corpora without derived channels are untouched.

Your bump is still correct: the comment at `transform.py:169` says a bump **must** make those episodes stale, because the policy decides their samples and no step hash can see it. The instruction was right and the reason given for it was not. I am recording it here rather than asking you to change anything, and flagging the re-ingest to @kstonekuan as a release-note item.

Nice work on the hardest issue open. If you want more of this kind: running HFlow against a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face) and reporting what breaks is the thing we most want and cannot get from issues.

Issue #68, opened 2026-08-20
Tests: `hflow stale --pipeline` with no App variable has no coverage


kstonekuan:
`hflow stale --pipeline` has several invalid-pipeline paths. The import-failure
path is tested; the App-resolution paths are not covered at the CLI boundary.

## Current behavior

Since the opinionated-defaults merge (`039943b`) the two pipeline loaders are
unified into one, and resolving the App is a three-outcome operation rather than
one lookup. The raise lives in `hflow.app.load_pipeline_application`
(`src/hflow/app.py:2186`), reached from `_import_pipeline_app`
(`src/hflow/cli.py:662`); `_command_stale` (`src/hflow/cli.py:705-716`) converts
the error to exit code 2.

Measured on `039943b`, `hflow stale --pipeline <spec>`:

```
no_app.py                 exit 2  no_app.py defines no hflow.App -- a pipeline file
                                  assigns one at module level, e.g. `app = hflow.App("my-pipeline")`
no_app.py:custom_name     exit 2  no_app.py has no hflow.App named 'custom_name'
two_apps.py               exit 2  two_apps.py defines 2 hflow.App objects ('kitchen',
                                  'garage'); address the one you mean as two_apps.py:kitchen
```

`tests/test_catalog_curation.py:641-651` covers a pipeline file that raises
during import. Nothing covers these three.

## Why the original recipe no longer works

This issue used to say: write `app = 42`, invoke `stale --pipeline`, assert
`has no hflow.App named` on stderr. Sole-App discovery changed that. A file with
no `App` and no `:name` in the address now falls through to the discovery path
and reports **"defines no hflow.App"**, so a test written to the old spec fails.
Only an explicit `:custom_name` still produces the old message.

## Pattern to copy

Extend the sibling test at `tests/test_catalog_curation.py:641-651`.

## Definition of done

Three tests at the `hflow stale` CLI boundary, each asserting exit code 2 and
the message on stderr:

- `--pipeline file.py:custom_name` against a file binding no such App ->
  `has no hflow.App named 'custom_name'`
- `--pipeline file.py` where the file binds no App at all (e.g. `value = 42`) ->
  `defines no hflow.App`
- `--pipeline file.py` where the file binds two Apps -> a message naming both
  candidates

The raises themselves already have unit coverage in `tests/test_runtime_cli.py`,
so the work here is specifically the exit-code and stderr boundary that
`hflow stale` owns.

## Validation

```bash
[...]

Issue #61, opened 2026-08-20
CLI: `hflow down`, `ingest`, and `status` print a bare `--help` with no description


kstonekuan:
Three subcommands show only a usage block and options under `--help`, while the
other six explain what they do.

## Current behavior

In `src/hflow/cli.py`, these parsers pass `help=` but no `description=`:

- `down` at `src/hflow/cli.py:277-280`
- `ingest` at `src/hflow/cli.py:296-299`
- `status` at `src/hflow/cli.py:330-333`

Running `uv run hflow down --help`, `uv run hflow ingest --help`, or
`uv run hflow status --help` therefore prints no explanatory paragraph.

## Pattern to copy

Copy the parser shape and tone from `up` at `src/hflow/cli.py:180-187` or
`doctor` at `src/hflow/cli.py:150-157`. Use two or three concise sentences that
say what the command does, what it needs, and what it prints.

## Definition of done

- `hflow down --help` includes a useful description.
- `hflow ingest --help` includes a useful description.
- `hflow status --help` includes a useful description.
- No command behavior changes beyond help text.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

kstonekuan:
Still reproduces on `039943b`: all three parsers still pass only `help=` (`down` at `src/hflow/cli.py:462`, `ingest` at `:481`, `status` at `:536`), and `--help` on each still prints usage plus options with no explanatory paragraph. `down` and `status` are byte-identical to before.

`ingest` isn't. The merge gave it a third executor and two new flags, so the description this issue asks for now has to cover more than it did, and its existing one-liner is wrong:

> trigger the master ingest DAG over episode URIs

That's only true when a runtime is addressed. With no bundle and no `--airflow-url`, `hflow ingest` imports the pipeline and runs the stages in this process. The new description should cover all three executors (rendered bundle, remote `--airflow-url`, in-process when neither is addressed), plus `--all-stages` and the per-episode planning it opts out of, and `--pipeline` defaulting to `hflow.toml`'s `pipeline`. The inaccurate `help=` line wants correcting in the same PR.

Line references drifted: `up` is `cli.py:356-363` now, `doctor` `:319-330`. And "the other six commands" is seven now, `dataset` landed with a description.

PR #154, opened 2026-08-24
fix: Coerce NumPy scalars and protect artifact label prefix

### LINKED ISSUE #151: A user label can claim the framework's artifact/ prefix, and ships as fake media in a snapshot
### What happens

`artifact/` is a namespace the framework owns, and nothing stops a user from writing into it.

When an enrichment publishes artifacts, `app.py:476-481` packs their URIs into the same
measurements dict the enrichment's own labels go into:

```python
labels: dict[str, float | int | str | bool] = (
    dict(enrichment_result.labels) if enrichment_result is not None else {}
)
if enrichment_result is not None:
    labels.update(
        {
            f"{ARTIFACT_MEASUREMENT_KEY_PREFIX}{artifact_name}": artifact_uri
            for artifact_name, artifact_uri in enrichment_run.artifact_uris.items()
        }
    )
```

An enrichment that returns `EnrichmentResult(labels={"artifact/notes": "s3://..."})` therefore
lands a row indistinguishable from a real published artifact, and `labels.update` means a real
artifact of the same name silently overwrites the label.

### Why it matters downstream

`snapshot.py:357` collects **every** text measurement whose key starts with the prefix, from any
step, and exports each one as a media artifact with a guessed MIME type:

```python
WHERE starts_with(measurements.key, ?)
  AND measurements.value_text IS NOT NULL
```

So a user label under that prefix ships inside a portable dataset snapshot as though it were
media the pipeline produced. The server's per-episode query is not exposed the same way:
`_catalog.py:680` scopes its lookup to `check_name = MEDIA_CONTACT_SHEET_STEP_NAME`, so only the
snapshot path is affected.

### Why this is the same defect as #129 and #130, a third time

Both of those were a name the framework reserves without enforcing:

- #129: the built-in media step's *step name* was takeable by a user step.
- #130: an `episodes` *column name* was takeable by a measurement key.
- This one: the `artifact/` *key prefix* is takeable by a measurement key or enrichment label.

The fix shape is settled by #133, which refuses at append time in `catalog.py` rather than at
registration, because that is the only layer that sees measurement keys at all.

### Definition of done

`Catalog.append_episode` refuses a run whose measurement key starts with
`ARTIFACT_MEASUREMENT_KEY_PREFIX`, naming the check and the key, **except** for the keys the
framework itself writes.

That exception is the whole difficulty, so be explicit about it: `app.py:476` builds those keys
and hands them to the same `CheckRunRow.measurements` a user's labels arrive in, so by the time
`catalog.py` sees the row the two are indistinguishable. Pick one and say why in a comment:

1. Refuse in `app.py` where user labels are still separate from framework artifact keys, before
   the `labels.update` merges them. Simplest, and the only place the distinction still exists.
2. Carry the artifact URIs beside the measurements rather than inside them, so the prefix stops
   being a naming convention at all. Bigger change; do not attempt it under this issue.

Option 1 is the expected answer. If you find a third, say so in the PR.

### Pattern to copy

`_raise_if_measurement_keys_shadow_episode_columns` (`catalog.py:301`) for the message shape:
name the check, name the key, say what it collides with, say what to do instead. Its test
`test_measurement_key_claiming_an_episode_column_is_refused`
(`tests/test_catalog_curation.py:891`) for the test shape, including the assertion that nothing
was written.

1. An enrichment label or check measurement key starting with `artifact/` is refused with a
   message naming the check and the key.
2. A real published artifact still lands, unchanged. There is a test proving both in the same
   file, because the whole point is that the two are told apart.
3. Nothing is written when a run is refused.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### LINKED ISSUE #152: A NumPy scalar in an Interval bound fails with an error naming neither the check nor the field
### What happens

A check that computes interval bounds from a NumPy array dies with an error that names nothing:

```python
Interval(label="segment", start_ns=np.int64(0), end_ns=np.int64(5))
```

```
TypeError: Object of type int64 is not JSON serializable
```

No check name, no field name, no mention of intervals. It surfaces from inside the run
fingerprint, which is nowhere near where the value was written.

Two other fields on `CheckRunRow` have the same hole, both reachable only from a custom runner
rather than from user check code:

```
duration_s=np.float32(0.1)   NotImplementedException: Unable to transform python value of
                             type '<class 'numpy.float32'>' to DuckDB LogicalType
critical=np.bool_(True)      TypeError: Object of type bool is not JSON serializable
```

`Interval` bounds are the one a user reaches, because a check computing segment boundaries from
`channel.to_numpy()` gets `np.int64` from indexing without doing anything unusual.

### Why it is only intervals that are exposed

Every built-in wraps its bounds by hand, for example `checks.py:118-121`:

```python
start_ns = int(stamps_ns[run_start_index])
end_ns = int(stamps_ns[run_end_index])
```

That is why no test caught this: the built-ins are all disciplined, and only user checks hit it.
`Interval` at `steps.py:111-116` annotates `start_ns: int` / `end_ns: int`, and a NumPy scalar
satisfies neither the annotation nor the runtime, but nothing checks.

### Why #127 stopped short

#127 added exactly this coercion for `measurements`, at `catalog.py:548-551`:

```python
check_rows = [
    replace(row, measurements=_normalized_measurements(row.check_name, row.measurements))
    for row in check_rows
]
```

`_normalized_measurements` (`catalog.py:276`) coerces `np.generic` to its Python scalar with
`.item()` and refuses anything else, naming the check and the key. Intervals were never in
scope. The choke point already exists; it just does not cover this field.

### Definition of done

A NumPy scalar in an `Interval` bound is coerced to its Python equivalent at the same boundary
`measurements` are, before the run fingerprint. Anything that is not a NumPy scalar and not an
`int` is refused, naming the check and which bound.

Two things to get right:

- **Placement matters, not just the coercion.** The comment at `catalog.py:544-547` explains why
  normalization happens before `_run_fingerprint`: normalizing later would let `np.int64(5)` and
  `5` hash as two different outcomes while writing the same row, which is the replay-identity
  split #126 was about. Put this in the same place for the same reason.
- **Refuse a float bound.** `start_ns` is nanoseconds and an `int`. `np.float64(5.0).item()` is a
  `float`, which is not a timestamp; coercing it silently would be the wrong kind of helpful.

`duration_s` and `critical` are in scope if the same pass covers them cheaply, and out of scope
if it does not. Say which you did. Neither is reachable from user check code, so do not
contort the fix to reach them.

### Pattern to copy

`_normalized_measurements` (`catalog.py:276`) for both the coercion and the refusal message.
Its tests are in `tests/test_catalog_curation.py`; put yours beside them.

1. `Interval(start_ns=np.int64(0), end_ns=np.int64(5))` round-trips and is queryable.
2. A `float` bound is refused with a message naming the check and the bound.
3. A test proves `np.int64(5)` and `5` produce the **same** run fingerprint, which is the half
   that is easy to miss.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

VedantMadane:
## Summary

Coerce NumPy scalars and protect artifact label prefix

## Changes

- app.py: strip artifact/ user labels

Fixes #152
Fixes #151

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -470,7 +470,7 @@ def _check_run_rows(report: "TestReport") -> list[CheckRunRow]:
     for enrichment_run in report.enrichments:
         enrichment_result = enrichment_run.result
         labels: dict[str, float | int | str | bool] = (
-            dict(enrichment_result.labels) if enrichment_result is not None else {}
+            {k: v for k, v in dict(enrichment_result.labels).items() if not str(k).startswith(ARTIFACT_MEASUREMENT_KEY_PREFIX)} if enrichment_result is not None else {}
         )
         if enrichment_result is not None:
             labels.update(


kstonekuan (changes requested):
Thanks for jumping on these, and welcome. I need to ask for changes on this one, so let me be specific about what I checked rather than leave you guessing.

## The PR closes two issues but only touches one

The diff is one line in `app.py`. #152 is about NumPy scalars in `Interval` bounds, which live in `catalog.py`, and nothing here reaches them. I ran the case on your branch:

```
Interval(label="s", start_ns=np.int64(0), end_ns=np.int64(5))
  -> TypeError: Object of type int64 is not JSON serializable
```

Still failing, same as `main`. Since the body says `Fixes #152`, merging this would close an open bug without fixing it, which is the part I most want to avoid.

**Please do one issue per PR.** They touch different files and different reviewers' worth of reasoning, and #151 in particular has a decision in it worth arguing about on its own thread.

## The #151 approach drops data instead of refusing it

```python
{k: v for k, v in dict(enrichment_result.labels).items() if not str(k).startswith(ARTIFACT_MEASUREMENT_KEY_PREFIX)}
```

This silently discards the label. A user who writes `EnrichmentResult(labels={"artifact/notes": "..."})` gets no error, no row, and no way to find out why their value vanished.

That is the exact failure mode the catalog spent #126, #127 and #133 removing. The rule those settled: a value that cannot be stored is refused loudly, naming the check and the key, because a silently missing row and a deliberately absent one are indistinguishable afterwards. `_raise_if_measurement_keys_shadow_episode_columns` (`catalog.py:301`) is the shape to copy, message included.

Two more things the issue asks for that this does not cover yet:

- **Check measurements too, not only enrichment labels.** A check can return `measurements={"artifact/x": "..."}` and reach the same place.
- **A real artifact must still land.** The whole difficulty is telling a framework-written `artifact/` key apart from a user-written one, which is why the issue suggests refusing in `app.py` *before* the `labels.update` at `:476` merges them. Your instinct to work in `app.py` was right; it is the timing relative to that merge that matters.

## Two mechanical things

- `uv run ruff format --check src` fails on this line, so CI would go red on `main`. Run `uv run ruff format` before pushing.
- No tests. Both issues ask for them, and for #151 the important one is that a *real* published artifact still lands unchanged, in the same test file, since that is the behaviour the refusal must not break.

## Where to go from here

Pick one issue and take it properly. #152 is the more self-contained of the two if you want a cleaner first run at the codebase; #151 has the design question in it.

Note also that non-collaborators are capped at one open PR at a time, so if you repoint this branch at a single issue you keep your slot. Everything in `CONTRIBUTING.md` applies, and `uv run pytest -q` from the repo root is the full suite.

Happy to review again as soon as you push.

kstonekuan:
@VedantMadane update, and one correction that is mine to own: I pointed you at #152 as the cleaner of the two to take. That was bad advice on timing. #152 has since been fixed by #155, which was a complete implementation with tests and landed first. Nothing you had written for it is lost, because this branch never touched it, but I sent you toward work that was about to be taken and I should not have.

**#151 is still open, still unclaimed, and it is the one this branch actually changes.** If you want it, that is the one to take, and the review above still describes what it needs: refuse rather than silently drop, cover check measurements as well as enrichment labels, and keep a real published artifact landing unchanged. Refusing in `app.py` before the `labels.update` at `:476` is the placement to aim for, since that is the last point where a framework-written `artifact/` key and a user-written one are still distinguishable.

Say the word and I will assign #151 to you so nobody else picks it up while you work. Repoint this branch at that issue alone, drop `Fixes #152` from the body, run `uv run ruff format`, and I will review it properly.

kstonekuan:
@VedantMadane closing this, and you deserve a straight account of why rather than a silent close.

Both issues this PR claimed are now fixed by other people: #152 landed as #155, and #151 landed as #158 about an hour ago. Neither of those was a judgment about your work. Our rule is that only an *assignee* reserves an issue, so an unassigned issue with an open PR on it can be picked up by someone else, and the author risks being preempted. That is what happened here, twice.

Where I did badly by you: I offered you #151 and said I would assign it if you wanted it, and I did not simply assign it there and then. Had I done that, it would have been held for you. That is on me, not on you or on the person who took it.

So let me do the thing I should have done the first time. **Tell me which of these you want and I will assign it to you immediately**, which means nobody can take it while you work:

- #61 - three CLI subcommands print a bare `--help` with no description. Small, self-contained, touches the argparse setup you would need for almost anything else in the CLI.
- #68 - `hflow stale --pipeline` with no `App` variable has no test coverage. If you would rather write a test than change behaviour.
- #9 - one reference page for the `HFLOW_*` environment variables. Bigger, and the most useful of the three to an actual user.

Reply here or on the issue with which one, and it is yours before you write a line.

For what it is worth, the instinct in this PR was sound: you went to `app.py`, which is the right file, and #158 refuses in the same function you were editing. What it needed was to refuse rather than drop, and to cover checks as well as enrichments. That is a review comment, not a talent problem.

PR #157, opened 2026-08-25
Test episodes view reserved-column drift

### LINKED ISSUE #153: Tests: the reserved-column set is not pinned to the episodes view it protects
### What we want

A test that pins the reserved-column set to the `episodes` view it protects, so the two cannot
drift apart silently.

### Current behavior

#133 refuses a measurement key that would collide with an `episodes` column. The set it checks
against is derived by string-splitting the table DDL (`catalog.py:115-121`):

```python
_EPISODES_VIEW_RESERVED_COLUMNS: dict[str, str] = {
    **{
        column.split()[0].lower(): column.split()[0]
        for column in TABLE_COLUMN_DDL["episodes"].split(",")
    },
    EPISODES_VIEW_STATUS_COLUMN.lower(): EPISODES_VIEW_STATUS_COLUMN,
}
```

That is correct today and nothing tests that it stays correct. Two ways it drifts:

1. **A parameterized column type.** `split(",")` assumes no type contains a comma. A future
   `DECIMAL(10,2)` column splits into `"amount DECIMAL(10"` and `"2)"`, adding a junk `2)` entry.
   The real column is still guarded, so this one degrades harmlessly, but it is a trap left lying
   around.
2. **A second derived column.** The set is the table's columns plus one hand-added derived name,
   `status` (`curation.py:318` and `:344`). A future view that derives another column gets no
   entry, and a measurement key can then shadow it, which is exactly the defect #130 described.

The second is the one that actually bites, because it fails the way #130 failed: silently, with
`SELECT <column>` returning the wrong value forever.

### Definition of done

A test that opens a catalog, reads the `episodes` view's real column list, and asserts it matches
the reserved set. When they disagree, the failure message should say which name is in one and not
the other, because the whole value of this test is being told *what* drifted.

One wrinkle to handle deliberately rather than by weakening the assertion: `quarantined` is in the
reserved set but is `EXCLUDE`d from the wide view (`curation.py:318`), so it is legitimately in
the set and not in the view. Assert the relationship that is actually true, and say in a comment
why that one name is exempt. Do not delete it from the reserved set to make the test pass;
reserving it is deliberate, because `status` is derived from it.

The wide view also carries one pivoted column per measurement key, so build the test's catalog
with no measurements, or subtract the known measurement keys. Either is fine; say which.

### Pattern to copy

`open_catalog_connection` (`curation.py:196`) to get a connection with the views registered.
`connection.execute("SELECT * FROM episodes").description` gives the column names.
`tests/test_catalog_curation.py` for placement, beside
`test_measurement_key_claiming_an_episode_column_is_refused` (`:891`), which is the check this
test defends.

1. The test passes on current `main`.
2. Adding a fake derived column to the view without adding it to the reserved set fails the test.
   Prove this to yourself before opening the PR; a test that cannot fail is worse than no test.
3. No production code changes. This is test-only.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

HarperZ9:
## Summary
- add a characterization test that compares the queryable `episodes` view columns with `_EPISODES_VIEW_RESERVED_COLUMNS`
- report both drift directions so a future derived column or stale reserved entry fails clearly
- leave production code unchanged

Closes #153.

## Why
The reserved-column guard rejects measurement keys that would collide with curated episode columns. This test pins that guard to the actual `episodes` view shape, with `quarantined` as the explicit stored-column exemption because `status` is derived from it.

## Validation
- `uv sync --locked`
- `uv run pytest -q tests/test_catalog_curation.py::test_episodes_view_reserved_columns_match_queryable_episode_columns tests/test_catalog_curation.py::test_measurement_key_claiming_an_episode_column_is_refused tests/test_catalog_curation.py::test_measurement_key_shadowing_is_case_insensitive` → 3 passed
- no-source mutation probe with a fake `future_derived_column` → expected `view_not_reserved=['future_derived_column']` failure observed
- `uv run ruff check --fix`
- `uv run ruff format`
- `uv run ty check`
- `PATH="$(dirname "$HFLOW_FFMPEG"):$PATH" HFLOW_FFMPEG=/root/.cache/hflow/ffmpeg/autobuild-2026-08-16-13-00-x86_64/ffmpeg HFLOW_FFPROBE=/root/.cache/hflow/ffmpeg/autobuild-2026-08-16-13-00-x86_64/ffprobe uv run pytest -q` → 826 passed, 6 skipped
- `git diff --check`
- diff secret-pattern check over `tests/test_catalog_curation.py` → no hits

## Checklist
- [x] Focused on one issue
- [x] Tests verify observable catalog behavior
- [x] No generated data, credentials, or runtime artifacts committed
- [x] AI assistance disclosed: I used OpenAI Codex to inspect the issue/repository state, implement this test-only change, and run validation.

### DIFF
--- a/tests/test_catalog_curation.py
+++ b/tests/test_catalog_curation.py
@@ -8,7 +8,13 @@ import duckdb
 import pytest
 
 import hflow
-from hflow.catalog import TABLE_COLUMN_DDL, Catalog, CheckRunRow, content_episode_id
+from hflow.catalog import (
+    _EPISODES_VIEW_RESERVED_COLUMNS,
+    TABLE_COLUMN_DDL,
+    Catalog,
+    CheckRunRow,
+    content_episode_id,
+)
 from hflow.cli import main as cli_main
 from hflow.curation import CurationReport, curate, open_catalog_connection
 from hflow.testing import SyntheticEpisodeSpec, synthesize_episode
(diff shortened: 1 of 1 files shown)


kstonekuan (approved):
LGTM. Validated on current main: ruff, format, ty clean, 826 passed, 6 skipped.

I checked the thing the issue asked you to check, because a characterization test that cannot fail is worse than none: I added `1 AS injected_drift_column` to the wide view and the test failed with

```
assert (not [] and not ['injected_drift_column'])
```

which names the drifted column, so the failure tells the next person what moved rather than only that something did. Reverted, and it passes again.

Reporting both directions rather than one is the right call, and handling `quarantined` as a named exemption with the reason in a comment, instead of quietly relaxing the assertion, is exactly what the issue was asking for.

Thanks for the clean first PR. If you want another: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

PR #141, opened 2026-08-24
feat(cli): add `hflow curate --dry-run` and render the no-output report clearly

ayam04:
## What

Adds `hflow curate --dry-run` and fixes the report's no-output rendering.

- `--dry-run` is mutually exclusive with an explicit `--output` (argparse group, per the issue's pattern-to-copy). When set, the CLI passes `output=None` to the library so the query still runs and reports row count + check coverage without writing a manifest. The library (`curate(..., output=None)`, documented at `src/hflow/curation.py:438-454`) remains the single owner of dry-run execution.
- Omitting both flags preserves the existing environment-aware default output (`$HFLOW_DATA_ROOT/manifest.parquet`).
- `CurationReport.summary()` now renders the no-output case as `manifest: (not written; dry run)` instead of the literal `None`.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix      # clean
uv run ruff format           # clean (1 file reformatted by the gate)
uv run ty check              # 1 pre-existing diagnostic only: unresolved import cv2 in
                             # src/hflow/motion.py:71 (optional opencv dep absent in this env;
                             # file untouched by this PR)
uv run pytest -q             # 781 passed, 7 skipped
```

The 4 failures + 13 errors are all pre-existing environment issues on this machine, proven against a clean-main worktree: `test_ffmpeg.py` needs ffmpeg/ffprobe on PATH, and the two `test_checks.py` failures come from the new `motion.py` module needing `cv2` (both reproduce identically with all my changes stashed). Targeted tests for this change: `pytest tests/test_catalog_curation.py -k "curate or report or manifest"` → 10 passed (3 new: dry-run CLI behavior, dry-run + `--output` mutual exclusion, no-output report rendering).

## Tests added

- `test_cli_curate_dry_run_reports_without_writing_a_manifest` — exit 0, report shows `(not written; dry run)`, no manifest created under `HFLOW_DATA_ROOT`.
- `test_cli_curate_rejects_dry_run_with_an_explicit_output` — argparse exits 2.
- `test_curation_report_summary_renders_the_no_output_case` — no `manifest: None`.

Closes #8

PR #158, opened 2026-08-25
refuse measurement keys that claim the artifact/ namespace

### LINKED ISSUE #151: A user label can claim the framework's artifact/ prefix, and ships as fake media in a snapshot

ayam04:
Fixes #151

## What changed

`_check_run_rows` now refuses, before anything is written, any enrichment label or check measurement key that claims the framework's `artifact/` prefix:

- **Enrichments**: guarded on the user labels **before** `labels.update` merges published artifact URIs into the same dict — the one layer that can still tell the two apart (the merge erases the distinction, which is why the catalog layer cannot do this).
- **Checks**: guarded on the row's measurements (those never carry framework keys, but the guard keeps the message shape uniform).
- Message shape mirrors `_raise_if_measurement_keys_shadow_episode_columns`: names the check and the key, states the collision, says what to do instead.

## Why

A user label like `artifact/notes` lands a row indistinguishable from a real published artifact, `snapshot.py:357` exports every measurement key under the prefix as media, and `labels.update` means a real artifact of the same name silently overwrites the label. Same defect family as #129/#130, third instance, settled fix shape per #133.

## Tests

Three new tests in `tests/test_enrichments.py`, in the same file as the existing artifact-landing test so the two behaviors are proven side by side:

- `test_enrichment_label_claiming_the_artifact_namespace_is_refused` — refused with a message naming check + key; nothing written (no episode parquet).
- `test_check_measurement_claiming_the_artifact_namespace_is_refused` — same gate on the check path.
- `test_labels_near_the_artifact_namespace_still_land_with_real_artifacts` — `artifact_notes` (close, not claiming) and a real `artifact/segments` URI both land unchanged.

## Validation

```bash
uv run pytest tests/test_enrichments.py -q      # 10 passed
uv run pytest -q                                # 813 passed, 6 skipped
uv run ruff check --fix                         # clean
uv run ruff format                              # clean
uv run ty check                                 # clean
```

The 2 failed + 13 errors in the full run are all in `tests/test_ffmpeg.py` and are the known environment gap (no ffmpeg binary in this WSL; `HFLOW_FFMPEG` unset), same as noted on #141 — zero overlap with changed files.

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -467,12 +467,18 @@ def _check_run_rows(report: "TestReport") -> list[CheckRunRow]:
         )
         for run in report.checks
     ]
+    _raise_if_measurement_keys_claim_artifact_namespace(
+        (row.check_name, key) for row in check_rows for key in row.measurements
+    )
     for enrichment_run in report.enrichments:
         enrichment_result = enrichment_run.result
         labels: dict[str, float | int | str | bool] = (
             dict(enrichment_result.labels) if enrichment_result is not None else {}
         )
         if enrichment_result is not None:
+            _raise_if_measurement_keys_claim_artifact_namespace(
+                (enrichment_run.enrichment.name, key) for key in labels
+            )
             labels.update(
                 {
                     f"{ARTIFACT_MEASUREMENT_KEY_PREFIX}{artifact_name}": artifact_uri
@@ -492,6 +498,37 @@ def _check_run_rows(report: "TestReport") -> list[CheckRunRow]:
             )
         )
     return check_rows
+
+
+def _raise_if_measurement_keys_claim_artifact_namespace(
+    claimed: Iterable[tuple[str, str]],
+) -> None:
+    """Refuse a key that claims the framework's ``artifact/`` namespace.
+
+    ``ARTIFACT_MEASUREMENT_KEY_PREFIX`` is reserved for URIs the framework
+    itself publishes: this module packs them into the same measurements dict
+    under that prefix, and snapshot.py exports every key under it as media.
+    A user label reaching that dict under the prefix is indistinguishable
+    from a real published artifact, and the ``labels.update`` merge below
+    would let a real artifact of the same name silently overwrite the label.
+    The enrichment call site runs before that merge, while the two are still
+    told apart; the check call site runs on rows whose measurements never
+    carry framework keys at all.
+    """
+    claimed_names = [
+        f"{key!r} from {check_name!r}"
+        for check_name, key in claimed
+        if key.startswith(ARTIFACT_MEASUREMENT_KEY_PREFIX)
+    ]
+    if not claimed_names:
+        return
+    raise ValueError(
+        f"measurement keys claim the framework's artifact/ namespace: "
+        f"{'; '.join(claimed_names)}. Keys under artifact/ are reserved for the "
+        "URIs the framework publishes (snapshot.py exports every such key as "
+        "media), so a real artifact of the same name would silently overwrite "
+        "the label. Rename the key."
+    )
 
 
 def _raise_if_measurement_keys_collide(check_rows: Sequence[CheckRunRow]) -> None:
--- a/tests/test_enrichments.py
+++ b/tests/test_enrichments.py
@@ -166,3 +166,75 @@ def test_enrichment_uses_alias_is_preflighted(source_episode: Path, tmp_path: Pa
 
     with pytest.raises(ValueError, match="captioner"):
         app.test(source_episode, verbose=False)
+
+
+def test_enrichment_label_claiming_the_artifact_namespace_is_refused(
+    source_episode: Path, tmp_path: Path
+) -> None:
+    """A user label under `artifact/` is indistinguishable from a published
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM on the code. Validated on current main: ruff, format, ty clean, 828 passed, 6 skipped.

The placement is right and the docstring says why: the enrichment guard runs before `labels.update`, which is the last moment a framework-written `artifact/` key and a user-written one are still distinguishable. `test_labels_near_the_artifact_namespace_still_land_with_real_artifacts` is the test that matters, because the refusal is only correct if it does not disturb a real published artifact.

## On taking this one

@VedantMadane had an open PR against #151 four hours before you claimed it, and #151 is labelled `good first issue`. By our own rule you were within your rights: only an assignee reserves an issue, nobody was assigned, and a competing PR is fair game. That is why I am merging this rather than holding it.

But I have asked you before, and I am asking again more plainly: **leave the `good first issue` pool alone.** With nine merged PRs you are one of the most productive people here, and every starter issue you take is one a newcomer cannot cut their teeth on. That pool is the only on-ramp this project has. It is not a rule and I will not enforce it, which is exactly why it depends on you honouring it.

What we actually want from you cannot come from the issue list at all: run HFlow against a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face) and report what breaks, what is slow, or what is awkward. Discussion #48 is Sagar-024 doing that, and it produced more useful work than any starter issue has. If you want something concrete instead, ask me and I will find you one that is not in the newcomer pool.

ayam04:
Thanks for the review and merge, Kingston — and for validating on current main (828 passed). The placement point (enrichment guard before `labels.update`, the last moment framework-written and user-written artifact keys are distinguishable) is exactly the invariant I structured the fix around. On to the next one.

Issue #162, opened 2026-08-25
Docs: CATALOG.md's key-naming rules describe the pre-#133 behavior and omit the artifact/ reservation


kstonekuan:
### What we want

`docs/CATALOG.md` "Naming measurement keys" (`docs/CATALOG.md:148`) is the page a check author
reads before naming a key. It now describes the wrong consequence for one rule, and is missing a
third reserved name added since it was written.

### The stale claim

The section says, of a key clashing with an episode column:

> The topic prefix is what keeps two cameras (or two checks) from colliding, and it also keeps a
> key from clashing with an episode column like `task` or `uri`, **which makes the wide view
> error**.

The wide view does not error. Before #133 the collision was silent: DuckDB renamed the measurement
to `task_1` and `SELECT task` returned the episode metadata forever. Since #133 the **append is
refused**, naming the check, the key, and the column it shadows
(`catalog.py:301`, `_raise_if_measurement_keys_shadow_episode_columns`).

The "current limits" list further down the same file (`docs/CATALOG.md:363-368`) was updated by
#133 and is correct. Only this earlier section still describes the old behavior, and it is the one
someone reads while naming keys.

### The missing rule

#151 landed a third reserved name: a measurement key or enrichment label may not start with
`artifact/`. That prefix is reserved for URIs the framework publishes, and `snapshot.py` exports
every key under it as media, so a user key there ships inside a portable snapshot as fake media.
It is refused at `app.py` (`_raise_if_measurement_keys_claim_artifact_namespace`).

The section is written as "Three rules, in decreasing order of how much it hurts to get them
wrong." The reserved-name rule is now two reserved namespaces, not one: episode column names, and
the `artifact/` prefix.

### Definition of done

1. The "makes the wide view error" clause describes what actually happens: the append is refused,
   naming the check, the key, and the column.
2. The `artifact/` reservation appears. Folding it into the existing name rule and keeping "three
   rules" is fine; so is a fourth rule. Choose whichever reads better and keep the count in the
   intro sentence consistent with what follows.
3. No production code changes.

Keep it short. This section is deliberately three tight rules with a worked example, not a
reference manual, and the value is that someone reads all of it before naming a key.

### Verify before you write

Every claim you add should be checked against the code, not against this issue. The two refusals
are at `catalog.py:301` and in `app.py`; the messages they raise are the ground truth for what the
docs should promise.

### Validation

Markdown only, so the link check is the relevant gate. The command is in
[...]

PR #163, opened 2026-08-25
Update CATALOG.md

Rezzhwxc:
Clarify naming conventions for measurement keys and their restrictions.

## Summary

- Corrects the outdated claim that a key clashing with an episode column "makes the wide view error" to the actual behavior: the append is refused, naming the check, the key, and the column.
- Adds the missing reserved namespace `artifact/` (measurement keys and enrichment labels may not start with this prefix) as a third restriction, consistent with the code in `app.py`.
- The section now accurately reflects the two refusal checks in `catalog.py:301` and `app.py`, and keeps the "Three rules" structure intact.

## Why

The docs were out of sync with #133 (refusal instead of error) and #151 (new artifact/ reserved name). This fixes the mismatch so check authors have accurate, up‑to‑date rules.

## Validation

- Since this is a documentation-only change and I don't have the local toolchain set up, I relied on manual review of the changes.

## Checklist

- [ ] I ran `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check`.
- [ ] I ran the relevant pytest suite.

### DIFF
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -161,13 +161,12 @@ episode-scoped, prefix it with the check instead.
 
 **Name keys `<topic>/<metric>_<unit>`.** The topic prefix is what keeps two
 cameras (or two checks) from colliding, and it also keeps a key from clashing
-with an episode column like `task` or `uri`. If a measurement key shadows an
-episode column, the append is refused, naming the check, the key, and the
-column (`catalog.py:301`). Additionally, a measurement key or enrichment label
-must not start with `artifact/`; that prefix is reserved for URIs published by
-the framework, and user keys in that namespace would be treated as fake media
-– such keys are refused at `app.py`
-(`_raise_if_measurement_keys_claim_artifact_namespace`).
+with an episode column like `task` or `uri`. Two namespaces are not yours to
+take, and both are refused at append time rather than resolved silently: an
+episode column name, refused naming the check, the key, and the column it
+shadows; and anything starting with `artifact/`, which is reserved for the URIs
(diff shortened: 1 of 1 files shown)


Rezzhwxc:
close #162

kstonekuan (approved):
LGTM, merging. Both corrections are right and you got to them within twenty minutes of the issue going up.

The content was correct as written, so I fixed the presentation myself rather than sending it back (`a4f7213`). Two things, both style rather than substance:

- The passage introduced the only `catalog.py:301` style citation in the whole docs tree. Line numbers in prose go stale the moment anything above them moves, and this section is stable enough that a reader will trust a wrong number. Docs name the function; issues and reviews carry the line numbers, because those are dated by nature.
- It also introduced the only en dash in `docs/`. Everything here uses `--`.

I also folded the two reserved namespaces into one sentence, so the unit rule that follows is not pushed behind a long aside. The rules are meant to be read in one go, which is why there are only three of them.

Validated: lychee reports 293 OK, 0 errors, and the `#naming-measurement-keys` anchor still resolves, which matters because `ARCHITECTURE.md`, `PORTING.md`, and this file all link to it.

One note for next time, no action needed now: the checklist in your PR body ticked the tests and toolchain boxes. For a docs-only change nobody expects you to have run pytest, and saying so plainly is better than a tick. Being able to trust those boxes is the only thing that makes them worth having.

Thanks for the quick first PR. The list, if you want another: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

PR #174, opened 2026-08-25
fix: Reject inverted or negative interval bounds on append

### LINKED ISSUE #161: An interval whose end precedes its start is stored without complaint
### What happens

An interval whose end precedes its start is accepted and stored. Measured on `e6103d0`:

```python
Interval(start_ns=10, end_ns=5, label="segment")   # appended, no complaint
```

So is a negative bound. Every consumer that subtracts the two gets a negative duration, and
nothing downstream can tell an inverted interval from a real one after the fact, because both are
just two integers in a row.

### Why now

#152 just landed `_normalized_intervals` (`catalog.py:301`), which walks both bounds of every
interval at append time to coerce NumPy scalars and refuse a float. It is already the one place
that looks at both bounds together, one commit after it was written, and it does not yet look at
their order.

### Definition of done

`Catalog.append_episode` refuses an interval whose `end_ns` is less than its `start_ns`, naming the
check, the label, and both bounds.

Two decisions to make deliberately rather than by accident:

- **Is `end_ns == start_ns` allowed?** A zero-length interval is a real thing to record: a single
  instant, an event with no duration. The recommendation is to allow it and refuse only
  `end_ns < start_ns`, but say which you chose in a comment, because the next reader will wonder.
- **Are negative bounds refused?** Log time is nanoseconds since the epoch, so a negative bound is
  a bug in the check. But it is a separate rule from ordering and adding it is optional here.
  Scoping this issue to ordering alone is fine; say so if you do.

Include the label in the message. A check that emits many intervals is otherwise a needle-in-a-
haystack, and `Interval.label` exists precisely to tell them apart.

### Pattern to copy

`_normalized_intervals` (`catalog.py:301`) for placement and message shape; it already raises a
`ValueError` naming the check and which bound. Its tests are in `tests/test_catalog_curation.py`,
beside `test_a_float_interval_bound_is_refused_naming_the_check_and_bound`.

1. `Interval(start_ns=10, end_ns=5)` is refused, with the check name, the label, and both bounds
   in the message.
2. `Interval(start_ns=5, end_ns=5)` behaves the way you decided, with a test either way.
3. Ordinary intervals are unaffected, and the built-in checks that emit intervals
   (`timestamp_regularity`, `camera_frame_stats`, `trajectory_segments`, `action_integrity`) still
   pass. They all build bounds in ascending order already, so a failure there means the rule is
   too strict.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

VedantMadane:
﻿## Summary

Reject inverted or negative interval bounds on append

## Changes

- In _normalized_intervals, refuse end_ns < start_ns
- Refuse negative start_ns/end_ns bounds with a clear ValueError
- Prevents negative durations and unreadable inverted spans in the catalog

Fixes #161

### DIFF
--- a/src/hflow/catalog.py
+++ b/src/hflow/catalog.py
@@ -328,14 +328,18 @@ def _normalized_intervals(check_name: str, intervals: list[Interval]) -> list[In
             bounds[bound_name] = value
         start_ns = bounds["start_ns"]
         end_ns = bounds["end_ns"]
+        # The label, because a check emitting many intervals is otherwise a
+        # needle in a haystack: the bounds alone do not say which one is wrong.
+        # end_ns == start_ns is allowed on purpose -- an instant is a real thing
+        # to record, and only end_ns < start_ns is nonsense.
         if start_ns < 0 or end_ns < 0:
             raise ValueError(
-                f"check {check_name!r} set a negative interval bound "
+                f"check {check_name!r} set a negative bound on interval {interval.label!r} "
                 f"(start_ns={start_ns}, end_ns={end_ns}): bounds are non-negative nanoseconds"
             )
         if end_ns < start_ns:
             raise ValueError(
-                f"check {check_name!r} set inverted interval "
+                f"check {check_name!r} set an inverted interval {interval.label!r} "
                 f"(start_ns={start_ns}, end_ns={end_ns}): end must be >= start"
             )
         normalized.append(replace(interval, **bounds))
--- a/src/hflow/catalog.py
+++ b/src/hflow/catalog.py
@@ -326,6 +326,18 @@ def _normalized_intervals(check_name: str, intervals: list[Interval]) -> list[In
                     f"{type(value).__name__}: interval bounds are int nanoseconds"
                 )
             bounds[bound_name] = value
+        start_ns = bounds["start_ns"]
+        end_ns = bounds["end_ns"]
+        if start_ns < 0 or end_ns < 0:
+            raise ValueError(
+                f"check {check_name!r} set a negative interval bound "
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging. Validated on current main (post-0.2.1): ruff, format, ty clean, 949 passed, 6 skipped.

The logic is right, including the part the issue left to you: you allowed `end_ns == start_ns` and refused only `end_ns < start_ns`. An instant is a real thing to record, so that was the correct call.

I pushed a fixup (`d10a249`) rather than sending this back, because the code was correct and only two specified pieces were missing:

- **The label in the message.** The issue asked for it, and the reason is worth keeping in mind: a check that emits twenty intervals gives the reader no way to tell which one is wrong from the bounds alone. `Interval.label` exists for exactly that.
- **The tests.** Three: the inverted case, the negative case, and a zero-length case that pins your decision. That last one is the one I would most want kept, because the ordering rule is one `<` away from rejecting every event with no duration, and nothing else in the suite would notice.

I checked they are real tripwires before pushing: with your `catalog.py` change stashed, the inverted and negative tests both fail, and they pass with it.

## For next time

Both your PRs have had the same gap, so it is worth saying plainly rather than fixing silently a second time: **a behaviour change needs a test in the same PR.** Not for ceremony. The test is what stops the next person deleting your rule by accident, and on this project it is also the thing a reviewer checks first. The pattern to copy is always the tests already sitting next to the function you changed.

I wrote them this time because your logic was right and you had already had one PR closed out from under you. Next one is yours to write.

Glad this one landed.

PR #175, opened 2026-08-25
Cache frame_stats per video to fix doubled decode cost

### LINKED ISSUE #173: Wrapping camera_frame_stats costs two ffmpeg decode passes per camera, forever
### What happens

Six checks auto-register on every App. The documented way to configure one is to wrap it
under a name of your own (`docs/how-to/enable-built-in-checks.md`), and when you do, the
automatic copy stands down and records `superseded`.

But supersession is decided from what a check actually MEASURED, which is only knowable
after it has run. `App._yield_defaults_superseded_by_the_pipeline` runs after every check
has executed and sets `run.result = None`. So the automatic copy runs in full and its
output is thrown away.

For five of the six defaults that is free. For `camera_frame_stats` it is **two ffmpeg
decode passes per camera per episode**, every episode, forever. `Episode.video`'s MP4 remux
is cached in the workdir; `hflow.ffmpeg.frame_stats` is not.

### Why it is sharp

`DEFAULT_CHECKS`' own membership argument admits `camera_frame_stats` on the grounds that
it "earns" its single decode, and excludes `camera_signal_quality` precisely because it
"would cost a SECOND decode for a deeper reading of the same footage". The documented
configuration path spends that second decode anyway -- on a strictly worse trade, since the
first pass's result is discarded rather than used.

The cost is now documented in the how-to, with the zero-cost spelling beside it:

```python
app = hflow.App(
    "my-pipeline",
    default_checks=[c for c in hflow.checks.DEFAULT_CHECKS if c is not camera_frame_stats],
)
```

That is a workaround, not a fix. It asks a user to know an implementation detail in order
to avoid paying for it.

### Why it was not fixed with the merge

Because the obvious fix is wrong. Deciding supersession at REGISTRATION time would need to
know which measurement keys a check emits without running it, and a check is an arbitrary
Python function -- `required_topics` emits keys derived from its `topics=` argument,
`camera_frame_stats` from the episode's camera list. Nothing static can answer it.

### Possible shapes

- **Declared keys.** Let a check optionally declare the keys it emits, and decide
  supersession from declarations when both sides have them. Opt-in, so nothing existing
  breaks, and the built-ins can all declare.
- **Order the defaults last and short-circuit.** Run the pipeline's own steps first, collect
  their keys, then skip any default whose declared or previously-observed keys collide.
  Needs the same declaration to be useful on episode one.
- **Memoize the instrument.** Make `hflow.ffmpeg.frame_stats` cache per (episode, camera)
  the way the MP4 remux already does. Narrower -- it fixes this check rather than the
  general shape -- but it removes the actual cost, which is the thing users feel.

The third is the only one that helps a pipeline calling the built-in twice for its own
reasons, and it is probably the cheapest to land.

### Definition of done

A pipeline that wraps `camera_frame_stats` under its own name performs one ffmpeg decode
pass per camera per episode, not two, with no `default_checks=` incantation. The paragraph
in `docs/how-to/enable-built-in-checks.md` that currently documents the doubled cost comes
out when it stops being true.

Sagar-024:
Closes #173

`hflow.ffmpeg.frame_stats` runs one ffmpeg decode pass per video. A pipeline that calls it twice — the `camera_frame_stats` and `camera_signal_quality` built-in pair, or a wrapper that re-reaches for the instrument — paid two decodes per camera per episode. `Episode.video()`'s MP4 remux is already cached in the workdir; the instrument was not.

Memoize the raw `metadata=print` instrument stdout in a `.instrument.txt` sibling of the workdir MP4, the same storage pattern the remux cache uses: hit detection by `.exists()`, atomic `.tmp` + `os.replace` write, no public API change. Aggregation thresholds re-run from the cached text on every call, so a threshold change is a free no-decode re-aggregation. Non-`.mp4` paths skip the cache so a raw `.h264` source is never silently shadowed. A corrupt or truncated cache self-heals by deleting the file and re-decoding.

**Why option 3, not 1 or 2:** option 1 (declared measurement keys) is a public API change; option 2 (run user steps first, short-circuit defaults) needs the same declaration to be useful on the first episode. Option 3 fixes the actual cost — the thing users feel — and is the same fix for a pipeline that calls the built-in twice for its own reasons.

**Definition of done from the issue:**
- A pipeline calling the instrument twice (the issue's "wrapping under your own name", or the `camera_frame_stats` + `camera_signal_quality` pair it implies) → 1 ffmpeg decode per camera per episode. Covered by `test_registering_both_camera_checks_caches_the_instrument`.
- Default configuration → 1 decode. Same test, no `default_checks=` arg used.
- No `default_checks=` incantation needed to get the fix.

**Tests added (12 in total):** cache-path naming, atomic write, temp cleanup on failure, missing/corrupt-cache self-heal, hit/miss/separate-key behavior, threshold-change re-aggregation, per-process persistence, and the integration test that runs both built-ins on one `Episode` and asserts the cache file plus exactly one ffmpeg decode.

**Docs updated:** `docs/how-to/enable-built-in-checks.md` and `camera_signal_quality`'s docstring now correctly say the two checks share one ffmpeg decode per camera per episode (the old text claimed it without the cache backing the claim).

### DIFF
--- a/docs/how-to/enable-built-in-checks.md
+++ b/docs/how-to/enable-built-in-checks.md
@@ -130,7 +130,7 @@ query later rather than pass/fail decisions baked into your corpus.
 | `camera_frame_stats` | Blackout, freeze, exposure, and stored frame count versus the rate the stream claims -- all from one decode pass. |
 | `joint_discontinuity` | Does any joint move faster than a limit you set? |
 | `idle_fraction` | How much of the episode had nothing moving? |
-| `camera_signal_quality` | Coding range, range-gated exposure, impulse noise, and stillness -- from the same decode pass. |
+| `camera_signal_quality` | Coding range, range-gated exposure, impulse noise, and stillness. Shares the same one ffmpeg decode pass per camera per episode as `camera_frame_stats` -- the instrument's raw output is cached in the workdir, so registering both checks against one episode pays a single decode, not two. |
 | `camera_stability` | How much footage is shaky rather than deliberately moving. Needs the `motion` extra. |
 | `action_integrity` | Are the recorded values sound -- no NaNs, no publisher stalls repeating a sample, no dimension that never moves? |
 | `trajectory_metrics` | How far and fast did it move, how much stood still, was it still settling when recording stopped? |
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -831,8 +831,10 @@ def camera_signal_quality(
     a check is the unit of three things at once -- one gate decision, one
     coverage denominator, and one content-hash version. Twenty-odd measurements
     under one name would mean a threshold on any of them gating all of them, and
-    one changed constant re-versioning the lot. Both checks read the same cached
-    remux, so the second costs a decode and no more.
+    one changed constant re-versioning the lot. The instrument is memoized in
+    the workdir (``hflow.ffmpeg.frame_stats`` writes a ``.instrument.txt``
+    sibling of the workdir MP4 it reads), so registering both checks against
+    the same episode pays one ffmpeg decode per camera per episode, not two.
 
     The coding range is measured from the pixels, never read from the
     container's declared range. That tag lies in practice -- a corpus can
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -832,9 +832,13 @@ def camera_signal_quality(
     coverage denominator, and one content-hash version. Twenty-odd measurements
     under one name would mean a threshold on any of them gating all of them, and
     one changed constant re-versioning the lot. The instrument is memoized in
-    the workdir (``hflow.ffmpeg.frame_stats`` writes a ``.instrument.txt``
-    sibling of the workdir MP4 it reads), so registering both checks against
-    the same episode pays one ffmpeg decode per camera per episode, not two.
+    the workdir (``hflow.ffmpeg.frame_stats`` caches its raw output beside the
+    workdir MP4 it reads, keyed on the filter graph), so registering both
+    checks against the same episode pays one ffmpeg decode per camera per
+    episode, not two. Both checks default the three graph parameters
+    identically, which is what makes them share the decode; override one on
+    only one of the checks and you are back to two decodes, correctly, because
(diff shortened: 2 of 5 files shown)


kstonekuan (approved):
Option 3 was the right pick, and the storage shape is right: atomic `.tmp` write, self-healing on a corrupt cache, caching off for a non-MP4 source. Validated on a merge with current main: 967 passed, ruff and ty clean.

**One defect, fixed on your branch in `dc0d92c`.** The cache was keyed on the video path alone, but three of `frame_stats`' parameters are compiled into the filter graph rather than applied after it: `black_pixel_threshold`, `freeze_noise_db`, and `freeze_min_duration_s` (`_instrument.py:359-360`). They change what ffmpeg measures. So this returned the first call's numbers:

```python
frame_stats(video, freeze_min_duration_s=0.5)
frame_stats(video, freeze_min_duration_s=5.0)   # cache hit, wrong freezes
```

Four distinct graphs, one decode, three answers for thresholds nobody asked for. That is the one thing this instrument must never do, and it lands on the documented path: the way to configure a built-in is to wrap it under your own name with your own thresholds, so two wrappers over one camera with different values is ordinary.

The fix hashes the graph into the cache filename, so a parameter added to the graph later is covered with no edit at the key site. `bright_luma_threshold` and `black_frame_amount_pct` stay out of the key, which keeps your free-re-aggregation property and your test for it intact. Both new tests fail with the digest stubbed out. Your tests that spelled the filename by hand now compute it, and the integration test asserts exactly one cache file, which is a stronger statement than the old name equality.

**Correcting your note:** `DEFAULT_CHECKS` auto-registration does exist, at `App._register_default_checks` (`app.py:1119`), added in #167. `camera_signal_quality` is not in it. So the doubled cost is the one #173 describes: wrap `camera_frame_stats`, the default copy runs anyway, two decodes. Your cache fixes that for a wrapper that keeps the defaults, which is the common case and the paragraph the issue wanted gone.

A wrapper that changes a graph threshold still pays two decodes and throws one away, because the superseded default still runs. That is the declared-keys shape (#173's options 1 and 2) and not yours to carry here. I will file it separately; say the word if you want it.

PR #176, opened 2026-08-25
fix: refuse non-finite catalog measurements

### LINKED ISSUE #159: A NaN measurement is stored, poisoning corpus statistics and dropping the episode from both sides of a threshold
### What happens

A check may measure `float("nan")` or `float("inf")` and the append accepts it. Measured on
`e6103d0`, over a three-episode catalog where one episode measured NaN and the other two
measured 0.0 and 2.0:

```
avg / min / max over the wide view   ->  (nan, 0.0, nan)
SELECT count(*) ... WHERE "video/blur_pct" < 10   ->  2
SELECT count(*) FROM episodes                     ->  3
```

Two separate harms, and the second is the worse one:

1. **One episode destroys a cohort statistic.** `avg` and `max` are NaN for the whole corpus.
   Coverage reporting still says the check ran on 3 of 3, so nothing looks wrong.
2. **The episode falls out of both sides of a threshold.** NaN compares false against everything,
   so it is in neither `< 10` nor `>= 10`. A curation cut and its complement do not add up to the
   corpus, and nothing says so. That is the same class of silent wrong answer as #130.

`inf` is accepted too, and makes any threshold on that key meaningless in the other direction.

### Why this is the float spelling of a settled question

`steps.py:31-37` already records the rule, in the comment above `MeasurementValue`:

> A check with nothing to say omits the key rather than measuring None: a NULL value column
> means nothing stored a value, so admitting None would make "not measured" and "silently
> dropped" the same row (see #126).

NaN is that same "nothing to say", spelled as a float instead of as `None`, and it arrives the
same way: `np.mean([])`, a divide by a zero denominator, a fit that did not converge. #126 refused
the `None` form. The float form still gets in.

### Definition of done

`Catalog.append_episode` refuses a measurement whose value is NaN or infinite, naming the check and
the key, and saying what to do instead (omit the key).

The message matters more than usual here, because the author of the check has to understand that
omitting is the supported way to say "no value for this episode". State it.

### Pattern to copy

`_normalized_measurements` (`catalog.py:276`) is the choke point and already refuses a value it
cannot store, naming the check and key. Add the check there, after the NumPy coercion, so
`np.float64("nan")` is caught too. Its tests are in `tests/test_catalog_curation.py`.

Note it is `float` values only: `math.isnan` on an `int`, `str`, or `bool` is either wrong or a
`TypeError`. `math.isfinite` covers both NaN and infinity in one test.

1. `measurements={"k": float("nan")}` is refused, naming the check and the key.
2. `float("inf")` and `float("-inf")` are refused the same way.
3. `np.float64("nan")` is refused too, which is the spelling real check code produces.
4. Ordinary floats, ints, strings, and bools are unaffected. `0.0` in particular must still be
   accepted, since it is a real measurement and falsy.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

sharonyao1127:
Fixes #159

## Problem
Catalog measurement normalization accepted non-finite float values such as NaN and infinity. Those values can poison aggregate statistics and make threshold predicates drop an episode from both sides of a cut.

## Fix
After NumPy scalar coercion, `_normalized_measurements` now refuses non-finite Python floats with an error that names the check/key and tells the check author to omit the key when there is no finite value. Ordinary finite floats, ints, strings, and bools are unchanged.

## Tests
- `.venv/bin/pytest tests/test_catalog_curation.py -q`
- `.venv/bin/ty check`
- `.venv/bin/ruff check src/hflow/catalog.py tests/test_catalog_curation.py`
- `.venv/bin/ruff format --check src/hflow/catalog.py tests/test_catalog_curation.py`

## Agent Disclosure
This PR was prepared with AI agent assistance. I reviewed the final diff and test results before submission.

### DIFF
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -186,6 +186,13 @@ SELECT episode_id, "/wrist_cam/compressed/black_frame_pct" AS black_pct
 FROM episodes WHERE black_pct < 5.0
 ```
 
+**Omit the key rather than measuring NaN or infinity.** A non-finite float is
+refused at append time, naming the check and the key. NaN compares false against
+everything, so an episode carrying one falls out of both a threshold and its
+complement: it is in neither `black_pct < 1.0` nor `black_pct >= 1.0`, and it
+poisons any `avg()` over the corpus. A check with nothing to say about an episode
+says nothing, and the coverage denominator already records that it ran.
+
 **End the key with a unit the tooling knows.** The workspace UI labels a value
 by the token after the key's last `_`, so `_s`, `_ms`, `_ns`, `_hz`, `_pct`,
 `_ratio`, `_count`, `_bytes`, `_deg`, `_rad`, and `_m` render with their unit
--- a/packages/hflow-server/tests/ui_test_fixtures.py
+++ b/packages/hflow-server/tests/ui_test_fixtures.py
@@ -6,8 +6,11 @@ repository's root test conftest.
 """
 
 import time
+from collections.abc import Iterator
+from contextlib import contextmanager
 from dataclasses import dataclass
 from pathlib import Path
+from unittest import mock
 
 import pytest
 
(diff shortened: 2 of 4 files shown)


kstonekuan (approved):
Correct, and it hits all four items in #159's checklist, including `0.0` still being accepted. Placement after the NumPy coercion is right, which is what catches `np.float64("nan")`.

**One thing to carry forward: run `uv run pytest -q` from the repo root, not a single file.** Plain `pytest` collects `packages/hflow-server/tests` too, and your change took that suite down: 126 errors, the whole thing, because its shared fixture appends `nan_metric` and `inf_metric` on purpose. Three tests pin that the API nulls them instead of returning a body no client can parse.

Fixed on your branch in `d07b8b2`, keeping those tests. Worth knowing why they stay rather than getting deleted to fit the new rule: the server reads catalogs rather than writing them, so a corpus appended before this rule still reaches it, and one of the guards is about a finite subtraction overflowing to infinity rather than a stored NaN at all. So the fixture now writes what an older hflow wrote, suspending the normalization guard for the two appends that carry the illegal doubles.

Also added the rule to `docs/CATALOG.md` beside the other append-time refusals, and to the function's docstring.

Validated on the full gate: 958 passed, ruff, ty, and lychee clean.

Thanks for the agent disclosure, and for reviewing the diff before sending it. If you want another, the open list is here: https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

Issue #177, opened 2026-08-25
A superseded default check still runs, and with non-default parameters still costs a discarded decode


kstonekuan:
### What happens

`App._yield_defaults_superseded_by_the_pipeline` (`src/hflow/app.py:987`) runs at
`app.py:1945`, after every check has already executed. Supersession is decided from what a
check actually measured, which is only knowable after it ran, so the automatic copy runs in
full and its result is discarded.

#175 took the cost out of the common case: `hflow.ffmpeg.frame_stats` now caches its raw
output beside the workdir MP4, so a wrapper that keeps the default thresholds shares the
default copy's decode. One decode per camera per episode, as #173 asked.

The cache is keyed on the video path plus the filter graph, because three of `frame_stats`'
parameters are compiled into the graph and change what ffmpeg measures. That is correct, and
it leaves a case behind:

```python
app = hflow.App("my-pipeline")

@app.check()
def camera_health(episode):
    return hflow.checks.camera_frame_stats(episode, freeze_min_duration_s=5.0)
```

Two graphs, two decodes. One of them is the automatic `camera_frame_stats` measuring at the
default 2.0 seconds, and its result is thrown away at `app.py:987`. The decode is real work
that nobody reads.

Binding a parameter to a built-in is the documented reason to wrap one
(`docs/how-to/enable-built-in-checks.md`), so this is not an exotic spelling. It is the
whole point of wrapping.

### Why the obvious fix is wrong

Deciding supersession at registration time needs to know which measurement keys a check
emits without running it, and a check is an arbitrary Python function. `required_topics`
derives its keys from its `topics=` argument, `camera_frame_stats` from the episode's camera
list. Nothing static can answer it. #173 recorded this and it has not changed.

### Possible shapes

Both are #173's options 1 and 2, which #175 deliberately did not carry:

- **Declared keys.** Let a check optionally declare the keys it emits, and decide
  supersession from declarations when both sides have them. Opt-in, so nothing existing
  breaks, and the built-ins can all declare. `camera_frame_stats` would declare a pattern
  over the episode's cameras rather than a literal list.
- **Order the defaults last and short-circuit.** Run the pipeline's own steps first, collect
  the keys they emitted, then skip any default whose declared keys collide. Needs the same
  declaration to be useful on the first episode, and changes step ordering, which step
  identity may care about.

The first is the smaller change and does not touch execution order. Neither is obviously
right, so treat the shape as open rather than specified.

### Definition of done

[...]

PR #178, opened 2026-08-25
fix(checks): skip defaults whose keys the pipeline already emitted (#177)

Sagar-024:
## What

A pipeline step that wraps one of the default checks under a name of its own now short-circuits the default before it runs. The pre-execution check uses an in-source key-pattern registry (`_DEFAULT_KEY_PATTERNS` in `src/hflow/checks.py`) that mirrors the `measurements[...] = ...` writes in each default, line for line. When any default's predicted key set overlaps keys a pipeline step has already emitted in this run, the default is recorded as `superseded` with the same reason and the same key list as the post-execution path, and the ffmpeg decode is never paid.

## Why

#175 cached `frame_stats` per video path so a same-parameter wrapper no longer costs two decode passes, but the default was still running in full and only being thrown away at `_yield_defaults_superseded_by_the_pipeline`. For a wrapper calling `camera_frame_stats` with non-default parameters, the decode is paid and the result discarded. The wrapper alone should decode; the default should stand down before that work happens.

## What stays the same

- No public API change. `@app.check()`, `App(...)`, `CheckResult`, the catalog, and the report shape are all unchanged.
- No new `keys=` parameter on `@app.check()`; the registry is private (`_DEFAULT_KEY_PATTERNS`) and a user step without a registered pattern still uses the post-execution backstop.
- `SupersededByPipeline.reason` and `superseded_keys` are identical regardless of which path fired; the only behavioral change is that `duration_s` on the superseded default is now ~0 instead of the ~100ms the decode took.
- Defaults that no pipeline step has covered in a given run still execute exactly as before.

## Tests

`tests/test_default_checks.py` gains:

- `test_a_wrapper_with_non_default_parameters_supersedes_the_default_without_running_it` -- the issue-177 case; counts ffmpeg decode passes and asserts the default is superseded with `duration_s == 0`.
- `test_a_wrapper_with_default_parameters_uses_the_post_execution_backstop` -- same-key wrapper; the pre-execution path fires.
- `test_a_non_overlapping_user_step_does_not_supersede` -- a wrapper with a different key set leaves the default running.
- `test_superseded_default_reports_the_overlapping_keys` -- the `superseded_keys` list is non-empty and contains `/decoded_frame_count`.
- `test_partial_camera_coverage_drops_the_uncovered_camera_too` -- a wrapper that covers only one of two cameras still supersedes the default for both; this matches the pre-#177 post-execution behavior, pinned so a future change to the pattern set can't silently flip it.
- `test_every_default_in_the_pattern_registry_is_drift_free` -- the patterns are exact, enforced by running each default and comparing emitted vs. predicted. A drift here fails the build rather than misfiring on a production corpus.
- `test_quarantined_episode_still_carries_default_measurements` -- defaults are exempt from the blanket `report.quarantined` skip: a critical user check that trips on its own measurement does not cascade-skip the engine's own baseline, so the rows the quarantine tag was derived from (`content_digest`, `media_digest`) survive on the episode you most need to diagnose.

978 of 984 existing tests still pass (6 unrelated skips, no new failures). `ruff check`, `ruff format`, and `ty check` are clean on the whole `src/` tree.

### DIFF
--- a/docs/how-to/enable-built-in-checks.md
+++ b/docs/how-to/enable-built-in-checks.md
@@ -99,13 +99,19 @@ copies would record the same measurement keys and the catalog would keep only
 one of them. A default is the exception: registering or wrapping one of those
 supersedes it, as above.
 
-Supersession is decided from what a check actually measured, which is only
-knowable after it runs. The automatic copy therefore runs before HFlow discards
-its result. Camera frame measurements cache their raw instrument output in the
-workdir, so a wrapper with the same graph still pays one FFmpeg decode per
-camera. A wrapper that changes a graph parameter correctly causes another
-decode because it requested a different measurement. You can still remove the
-automatic copy when you do not want it to run:
+One cost worth knowing before you wrap `camera_frame_stats` specifically.
+Supersession of a default happens before the default runs when the pipeline's
+wrapper is registered in the same run: HFlow knows the keys each default would
+emit, and any default whose predicted key set overlaps the wrapper's emitted
+keys is recorded as `superseded` without ever calling ffmpeg. The wrapper
+alone decodes the video, the default does no work, and the catalog records
+both rows -- the wrapper's measurements and the `superseded` notice on the
+default, naming the keys the wrapper covered. Every other built-in is cheap
+enough that this never mattered; the doc only called `camera_frame_stats` out
+because it is the one whose cost is ffmpeg.
+
+If you prefer to drop the default and own the measurement yourself, that works
+the same as before:
 
 ```python
 app = hflow.App(
@@ -115,7 +121,8 @@ app = hflow.App(
 ```
 
 `functools.partial` costs the same, and so does registering it bare under its
-own name -- that is a replacement, not a wrapper, so nothing runs twice.
+own name -- that is a replacement, not a wrapper, so the default does not
+exist to be superseded in the first place.
 
 [`examples/stress/synthetic.py`](../../examples/stress/synthetic.py) registers
 several this way over a generated corpus, if you want a running reference.
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -1844,7 +1844,21 @@ class App:
             for registered in checks_to_run:
                 run = CheckRunReport(check=registered)
                 report.checks.append(run)
-                if report.quarantined:
+                # Quarantine stops the user's own steps from piling more
+                # work onto an already-rejected episode, but the defaults
+                # are cheap diagnostic evidence and the episode you most
+                # need to diagnose is the one that just tripped a gate:
+                # ``content_digest`` and ``media_digest`` are exactly the
+                # rows that help tell whether the rejected recording is
+                # the same as a previous one. They are also what the
+                # quarantine tag itself was derived from, so skipping
+                # them on the same run that produced the tag blanks the
+                # catalog row you'd grep for. Defaults still respect the
+                # gate -- their result carries the ``failed:<name>`` or
+                # is recorded normally, and a failing critical default
+                # *adds* a quarantine tag rather than gating later
+                # ones. They just do not get blanket-skipped.
+                if report.quarantined and registered.name not in self._default_check_names:
                     run.not_run = SkippedByQuarantine(tuple(report.quarantine_tags))
                     continue
                 # A default with a registered key pattern: if any pipeline
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -1486,10 +1486,19 @@ class App:
 
     def _ordered_checks(self) -> list[RegisteredCheck]:
         # Cheap-first: steps that need no special resources run before steps
-        # declaring requires/uses. Stable within each class.
+        # declaring requires/uses. Within each class, the pipeline's own
+        # steps run before the automatic defaults -- so a wrapper registered
+        # under its own name with non-default parameters has a chance to
+        # emit its measurement keys before the default would run, and the
+        # pre-execution short-circuit at the top of the check loop can
+        # supersede the default without paying its ffmpeg decode.
+        # Stable within each class.
         return sorted(
             self.checks,
-            key=lambda registered: bool(registered.requires) or registered.uses is not None,
+            key=lambda registered: (
+                bool(registered.requires) or registered.uses is not None,
+                registered.name in self._default_check_names,
+            ),
         )
 
     def _ordered_enrichments(self) -> list[RegisteredEnrichment]:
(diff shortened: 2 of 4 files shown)


kstonekuan (changes requested):
The pre-execution short-circuit works and the drift guard is the right instinct. One blocking regression, and one claim to retract.

## The reorder skips the defaults on every quarantined episode

`_ordered_checks` now sorts defaults after the pipeline's own steps. Defaults were registered in `__init__` (`app.py:985`), so they used to run first. The check loop skips everything once `report.quarantined` is set (`app.py:1841`), and that flag is now set before any default is reached.

So a user's critical check that quarantines takes all six defaults with it:

```python
app = hflow.App("quarantining", data_root=...)   # defaults on

@app.check(critical=True, gate=RECOMMENDED_CAMERA_INTEGRITY)
def blackout(ep):
    return hflow.CheckResult(measurements={"black_frame_pct": 99.0})
```

On `125b82e`: `episode_duration: measured`, `content_digest: measured`. On this branch both are `SkippedByQuarantine`. Every default measurement is gone from the episodes you most need to diagnose, `content_digest` and `media_digest` included.

`SkippedByQuarantine` is documented as conditional and does lift on retune, so nothing is permanently lost. It is still the wrong answer: the defaults are the cheap evidence, they cost nothing to keep, and quarantining an episode should not blank its digests.

Shape is yours to choose. The one that keeps this PR's win is to decide the quarantine skip from a step's pre-reorder position, so the reorder only moves the supersession decision and not the quarantine boundary. Whatever you pick, a test that a quarantined episode still carries its default measurements should come with it.

## The good news on reordering

`_run_fingerprint` sorts `check_outcomes` before hashing (`catalog.py:416`), so the reorder does not re-mint episode identities and existing corpora do not go stale. That is the expensive thing a reorder usually costs, and it does not apply here. Worth knowing before you pick a fix.

## Retract the "out of scope" paragraph

There is no cross-wrapper cache bug to track. You describe the cache as path-keyed, which was true of `b3b076c`; the fixup on your own #175 branch keys it on the video path plus a digest of the filter graph, and this branch contains it. Two wrappers with different `freeze_min_duration_s` get two cache files:

```
wrist_cam_compressed.instrument.536b5ebb9b107de7.txt
wrist_cam_compressed.instrument.ddd5d5cefa1a5ab4.txt
```

Separately, two steps emitting the same measurement keys is refused outright by `_raise_if_measurement_keys_collide`, so the scenario needs prefixed keys before it can even be built.

## One note, not blocking

`_DEFAULT_KEY_PATTERNS` is a second implementation of six functions, which is the cost #177 flagged. The drift guard is what makes it acceptable, so keep it exact. But `_keyframe_interval_keys` walks `channel.raw` and parses every payload to predict a key set. That is real work added to a path whose purpose is to avoid work, and it now runs on every episode where any default gets superseded. Worth predicting from something cheaper, or noting why the scan is acceptable.

Gate is otherwise clean: 978 passed, ruff, format, and ty green.

Sagar-024:
@kstonekuan Fixed all three:

1. **Quarantine regression:** Defaults are now exempt from the quarantine skip (they're cheap diagnostic evidence). The reorder only moves the supersession decision; the quarantine boundary stays where it was. New test `test_quarantined_episode_still_carries_default_measurements` pins this behavior.

2. **Retracted the "Out of scope" paragraph.** You're right — the cache is path+graph-keyed and `_raise_if_measurement_keys_collide` refuses the scenario anyway. I was reading stale #175 code, my mistake.

3. **`_keyframe_interval_keys`:** added a comment in the docstring rather than refactoring — the keyframe count is what decides `median_keyframe_interval_s` (needs >=2) and there is no cheaper structural signal that distinguishes the empty / single / multi cases. The scan is acceptable because the work it adds (one payload walk per camera on the supersede path) is dwarfed by the ffmpeg decode it prevents (~2-5s per camera on a fresh episode). Trade is documented in source so a future change to `keyframe_interval` can revisit the prediction.

Full suite still green, ruff/format/ty clean.

kstonekuan:
The quarantine fix is right, and I confirmed it: `episode_duration` and `content_digest` both come back `measured` on a quarantined episode again. Retraction and the keyframe note are both fine.

**This now needs a rebase, and not a trivial one.** `a478b95` ("refactor: incubate video measurements") landed after you branched and moved the ground under this PR:

- `hflow/ffmpeg/_instrument.py` is gone. It is now `hflow/_video_measurements/_frame_statistics.py`, and `tests/test_default_checks.py:15` imports the dead path, so `ty` fails on the merge.
- `docs/how-to/enable-built-in-checks.md` conflicts.
- `src/hflow/checks.py` changed by 163 lines, which is where your mirrors live.

**The mirror is already out of date because of it.** `camera_frame_stats` now emits `camera_measurement_definition` alongside `camera_instrument` (`checks.py:320` on main), and `_camera_frame_stats_keys` does not predict it. Your drift guard should catch that the moment it runs, which is the guard doing its job. Re-run it against the new `checks.py` rather than reasoning about the diff.

Worth knowing while you rebase: the refactor kept the graph-keyed cache and strengthened it. `_frame_statistics_cache_path` now folds `FRAME_STATISTICS_DEFINITION_VERSION` and the ffmpeg version into the digest alongside the filter graph, so a toolchain bump invalidates the cache too.

One line to drop from the new comment: it says a failing critical default adds a quarantine tag. Defaults are registered through `self.check()` with `critical=False` (`app.py:1131`), so a default is never critical and that case cannot arise. The rest of the reasoning holds.

Sagar-024:
Hey Kingston,

Rebased onto dfbbc82. You were right that it would not be trivial. The refactor
moved the instrument module and the mirrors went stale with it.

The drift guard earned its keep. The moment I ran it against the new checks.py
it flagged camera_measurement_definition as the one key my pattern did not
predict. I fixed that, then re-checked all six mirrors by hand against the new
code, and the guard is green again.

Everything else held up. The quarantine test still pins the
defaults-on-quarantined behavior, full suite is 976 passed and 6 skipped, and
ruff, format, and ty are all clean. The few ty diagnostics you would see on a
Windows box (fcntl, os.getuid) are pre-existing on main and disappear on Linux.

Dropped the "critical default" line from the comment too. Good catch. Defaults
register with critical=False so that case cannot happen.

Ready for another look.

Sagar

kstonekuan (approved):
Rebase is clean and the quarantine fix holds. Validated on a merge with current main (`8c17dd4`, which added `unverified`): 990 passed, ruff, format, and ty green.

**One correction: the drift guard did not catch `camera_measurement_definition`.** It could not have. I named that key in my last comment, which is where it came from.

The guard is the entire reason a shadow reimplementation of six defaults is acceptable, so I went looking for whether it can actually fail. Two defects, the second hidden by the first:

1. **It ran only on a camera-less episode** (`source_episode` is `cameras=()`). Three of the six mirrors emit nothing at all without a camera, so predicted and emitted were both the empty set. I deleted `camera_measurement_definition` from `_camera_frame_stats_keys` and all fourteen tests stayed green.
2. **It replayed each pattern against the raw source episode** while comparing against what the checks emitted from the canonical one. Keyframe detection reads payload bytes, so on a raw episode the mirror finds no keyframes and drops `max_keyframe_gap_s`. That was invisible until a camera was present, at which point it reported drift production does not have, because `App` passes `canonical_episode`.

Fixed on your branch in `e7c5c2d`: parametrized over zero, one, and two cameras, and replaying against `report.canonical_path`. It now fails on that deleted key, in both camera cases.

**Your mirrors were right.** Once the guard compares like with like, all six are exact on one and two cameras. Two cameras as well as one on purpose: a per-topic mirror that accidentally hardcodes the first camera reads correct at one.

Merging. The keyframe scan note and dropping the critical-default line are both fine.

Worth saying plainly, because it is the useful part of this round: a guard that cannot fail is worse than no guard, since it buys the confidence without the coverage. The instinct to write one was right, and it is the reason this design is mergeable at all. When you write the next one, delete something it should catch and watch it go red before you trust it.

Next one is a good place to use it: #177 stays open behind this only in the sense that `_DEFAULT_KEY_PATTERNS` now has to be maintained by hand. If you want a follow-up with real teeth, making the defaults declare their own keys (option 1 from the issue) would delete the registry and the guard with it.

Sagar-024:
Thanks Kingston, and thanks for fixing the guard on my branch. That is above
and beyond.

You are right on both counts. The guard as written could not have caught the new key, and the raw-vs-canonical replay meant it was comparing unlike things.
"A guard that cannot fail is worse than no guard" is going to stick with me.
The mutation step (delete a key it should catch, watch it go red) is now part
of how I write any invariant test.

I would like to take the follow-up. Making the defaults declare their own keys
(option 1) deletes the hand-maintained registry and the guard together, which
is the cleaner end state. Given what this round taught me, I will write a short design note first so we agree on the shape before I build it out, then open the PR against #177.

Appreciate the time you have put into these reviews. They have taught me more than the merges have.

Sagar and gn

kstonekuan:
Yes, take it. Opened #182 for it, since #177 is closed and a design note wants a home.

Design note first is the right instinct and I have written the issue that way: it lists what the note should settle and nothing more. The shape is genuinely open, including whether it is worth the API surface. A note concluding "the registry plus a working guard is the better trade" would close #182 legitimately, and I would rather have that than a design chosen because the issue implied it.

Comment on #182 and I will assign you. GitHub silently drops an assignee who has not posted on the issue yet, which is why it is not already yours.

PR #180, opened 2026-08-25
fix(catalog): report a crashed critical check as unverified

### LINKED ISSUE #164: An episode whose critical check crashed reports status 'ok', identical to one that passed
### What happens

An episode whose **critical check crashed** is indistinguishable in the catalog from one whose
critical check ran and passed. Both come back as `status = 'ok'`.

Measured on `e6103d0`, three episodes appended with one critical `blur` check each, one passing,
one erroring, one failing:

```
2e7d2c03  quarantined     <- critical check returned verdict False
3e23e816  ok              <- critical check CRASHED
ca978112  ok              <- critical check passed
```

So `status = 'ok'` currently means "not quarantined", which covers both *verified clean* and
*never verified*. Every curation query that means "episodes I have checked for blur" silently
includes episodes where the blur check never produced an answer.

### Why it cannot quarantine today, precisely

`app.py:1445` decides quarantine with:

```python
if run.result is not None and run.result.verdict is False:
    if registered.critical:
        report.quarantine_tags.append(f"quarantined:{registered.name}")
```

A crashed check sets `run.error` and leaves `run.result` as `None` (`app.py:1440-1442`), so it
fails the first condition **even when `critical=True`**. A critical check that crashed cannot
quarantine, by construction rather than by oversight: the comment there says "Infrastructure, not
data: never recorded as a quality outcome", and that is the right instinct. An infrastructure
crash is not a quality verdict, and treating it as one would let a flaky endpoint mark a corpus
bad.

The mistake is not that it fails to quarantine. It is that the *absence* of the verdict is spelled
the same way as a *good* verdict.

### The decision, already made

Add a third `status` value rather than conflating it into either existing one. A crash is neither
"this episode is bad" nor "this episode is fine": it is "nobody knows", and the catalog should be
able to say so.

The value is **`'unverified'`**, naming what is true about the *episode* rather than what happened
to the check. `check_runs.status = 'error'` already records the check's side of it.

### No new stored column is needed

The fact is already in the catalog. `check_runs` stores `critical BOOLEAN` and `status VARCHAR`
per step (`catalog.py:70-74`), so the episode-level state derives from a join. This works
retroactively on episodes already appended, and needs no format-version bump. Verified on
`e6103d0`:

```sql
CASE WHEN e.status = 'quarantined' THEN 'quarantined'
     WHEN EXISTS (SELECT 1 FROM check_runs c
                  WHERE c.episode_id = e.episode_id
                    AND c.run_fingerprint = e.run_fingerprint
                    AND c.critical AND c.status = 'error')
     THEN 'unverified' ELSE 'ok' END
```

```
2e7d2c03  today=quarantined  derived=quarantined
3e23e816  today=ok           derived=unverified
ca978112  today=ok           derived=ok
```

Scope it to the episode's own `run_fingerprint`, as above. Without that, an error from any
historical run of the episode would mark the current one unverified forever, which is wrong: a
later successful run is exactly how a crash gets resolved.

Do not add a stored column unless you find a reason the derivation cannot work, and say what that
reason is if you do.

### Rules to pin down

- **Precedence: `quarantined` wins.** An episode that is both quarantined and had a critical error
  is quarantined. It is the stronger claim and the more conservative one.
- **Critical steps only.** A non-critical check crashing does not undermine a verification claim,
  and widening this to any step would mark most of a corpus unverified during one flaky enrichment
  outage. Enrichments are never critical in the sense that matters here; scope it to checks with
  `critical = true`.
- **`SKIPPED` is not `ERROR`.** A step skipped because the episode was already quarantined
  upstream is not an unverified episode, it is a quarantined one, and precedence already handles
  it. Do not fold `skipped` in.

### The query idiom changes, and that is in scope

`status != 'quarantined'` is the idiom throughout the repo and it would still include unverified
episodes, so the new value buys nothing until the recommended idiom becomes `status = 'ok'`.
**Changing what these queries return is accepted and intended.**

All 15 occurrences are in scope for this PR, because landing the column without landing the idiom
leaves the repo teaching the unsafe query:

- `README.md:183`
- `docs/CATALOG.md:81`, `:187`
- `docs/RUNTIME.md:147`
- `docs/how-to/enable-built-in-checks.md:179`
- `examples/egocentric/curate.sql:12`
- `tests/test_catalog_curation.py:389`, `:429`, `:458`, `:814`

Re-verify that list against `main` before you start; line numbers move.

`docs/CATALOG.md` also needs the three-value vocabulary written down, including why an
infrastructure crash is not a quality verdict. That prose is the actual deliverable here. The SQL
is the easy part.

### Also check

- `EPISODES_VIEW_STATUS_COLUMN` and `_EPISODES_VIEW_RESERVED_COLUMNS` (`catalog.py:110-121`) still
  hold: the column name does not change, only its domain.
  `test_episodes_view_reserved_columns_match_queryable_episode_columns` should still pass.
- The server renders `status` (`packages/hflow-server/`). A third value must not fall through a
  two-branch conditional into a blank cell or a crash. Grep for the existing values and check
  every site.
- Both `episodes` view definitions in `curation.py:318` and `:344` (the constrained and
  unconstrained builds) need the same derivation. They are separate SQL strings and it is easy to
  update one.

### Definition of done

1. An episode whose critical check errored reports `status = 'unverified'`.
2. A quarantined episode still reports `'quarantined'`, including when it also had a critical
   error.
3. An episode with a non-critical error still reports `'ok'`.
4. A later successful run of a previously errored episode reports `'ok'` again.
5. Both view definitions agree, proven by a test that opens a constrained connection too.
6. Every occurrence of the old idiom is updated, and `docs/CATALOG.md` documents the vocabulary.
7. The server renders the new value.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

`packages/hflow-server/tests`, and this change reaches them. Markdown changes also want the
lychee command from CONTRIBUTING.md.

chiruu12:
Closes #164.

An episode whose critical check crashed reported `status = 'ok'`, the same as one
whose critical check ran and passed. `status` now takes a third value,
`unverified`, for exactly that case.

The rule lives in one place, `episode_status_case_sql` in `hflow/catalog.py`, and
every site that renders the status column builds it from there: both `episodes`
view shapes and both snapshot sample queries. That is what makes item 5 hold by
construction rather than by four edits staying in step. A `check_runs_latest`
view joins the fact in; it ranks the same way `measurements_latest` does, by the
owning episode's `recorded_at`, so it cannot pick a check run from a different
generation than the episode row it decorates.

Precedence is quarantined, then a critical check whose status is specifically
`error`, then ok. Not "not passed": `skipped` and `superseded` are not crashes
and still read `ok`.

The server derives the same status without the SQL, so
`episode_status_for_quarantine_flag` became `episode_status_for_flags` and takes
the second fact. It reads the same view with the same predicate.

Definition of done, item by item:

1. errored critical check reports `unverified` - `test_errored_critical_check_reports_unverified`
2. quarantined still wins when both apply - `test_quarantine_outranks_a_critical_error`
3. non-critical error still reports `ok` - `test_non_critical_error_still_reports_ok`
4. a later good run clears it - `test_successful_rerun_clears_unverified`
5. both view definitions agree, constrained too - `test_both_view_definitions_agree_on_unverified`
6. old idiom updated, vocabulary documented - four SQL sites, the server contract, `docs/CATALOG.md`
7. the server renders it - `test_dossier_reports_unverified_when_a_critical_check_crashed`

I checked which of the new tests actually fail against unfixed source by
reverting `src` and `packages/*/src` and re-running, rather than assuming.
`test_errored_critical_check_reports_unverified`,
`test_successful_rerun_clears_unverified`,
`test_both_view_definitions_agree_on_unverified`, the dossier test and the
snapshot test all fail there. The rest assert a status that was already correct
before this change, so they are regression guards and I am not claiming they
prove anything.

Three things worth your judgement rather than mine:

- `docs/CATALOG.md` recommended `status != 'quarantined'` in two examples. Once
  `unverified` exists that filter keeps episodes nobody checked, so I changed both
  to `status = 'ok'`. The same pattern is still taught in `README.md:192`,
  `docs/RUNTIME.md:210`, `docs/how-to/enable-built-in-checks.md:229` and
  `examples/egocentric/curate.sql:12`. I left those alone, since changing a shipped
  example reads like your call, but item 6 arguably covers them. Say the word and
  they go in.
- `hflow dataset create` is unchanged and still filters `!= 'quarantined'`. Its
  default policy already excludes an episode whose critical check only ever
  crashed, through the settled-steps rule rather than the status one. The gap is
  narrower: a settled run in history plus a later crash satisfies the step
  requirement while the status column reads `unverified`. I documented that
  distinction instead of tightening the filter, which would silently shrink
  existing datasets.
- Every `SELECT` from `episodes` now pays a correlated `EXISTS` over
  `check_runs_latest`. I did not measure it. If that cost matters on a large
  catalog, a join would probably be cheaper than the subquery and I am happy to
  redo it that way.

`EpisodeStatus` in the server contract gained a third member, so a client
generated from the previous OpenAPI schema can now receive a value its enum does
not have. Additive, but it belongs in a release note.


kstonekuan (approved):
Centralising the rule in `episode_status_case_sql` makes item 5 hold by construction, and `_episode_id_scope` refusing an unqualified column closes a failure mode that's silent and total.

Verified rather than taken on trust:

- The rule renders correctly. `CheckStatus.ERROR` interpolates to `'error'`, not `'CheckStatus.ERROR'`, which is the one way a status enum in an f-string goes quietly wrong.
- Your tripwires are real. I kept the symbol and neutered the CASE back to two-way instead of reverting `src`, since reverting breaks collection and proves nothing. Six fail that way, including `test_one_errored_episode_does_not_mark_its_neighbours_unverified`, which catches the correlation bug your guard prevents.
- Behaviour on a real catalog: 60 episodes, every 7th with a crashed critical check, gives 51 `ok` and 9 `unverified`.
- `?status=unverified` works on the list endpoint for free, because the filter compares against the derived column.

**Your three questions.**

**The other four idiom sites: yes, and done.** Pushed as `a8e02eb`, covering README, `docs/RUNTIME.md`, the built-in checks how-to, and `examples/egocentric/curate.sql`. Item 6 covers them, and leaving them was the worse option: CATALOG.md now calls that filter a trap while four other pages still taught it as the everyday cut.

**The `EXISTS` cost isn't a problem.** I checked the plan instead of guessing: DuckDB flattens it into a `LEFT_DELIM_JOIN` with `Join Type: MARK`, so it's already a hash semi-join and not a per-row subquery. Writing the join by hand would buy nothing. Right to flag it unmeasured instead of asserting either way.

**`dataset.py` stays as you left it.** Documenting the distinction beats silently shrinking existing datasets. The two notions do need to agree eventually, but they aren't the same rule (dataset membership drops any errored registered step, this is critical-only), so converging them is a decision and not a cleanup. Filing it separately.

One consequence to flag rather than fix: `EpisodeStatus` gaining a third member means a client generated from the old OpenAPI schema can receive a value its enum lacks. You called that out yourself and it belongs in a release note.

Validated on a merge with current main: 981 passed, ruff, ty, and lychee all clean.

PR #183, opened 2026-08-25
fix(dataset): select only ok episodes by default

### LINKED ISSUE #181: hflow dataset create still filters on 'quarantined', not the new 'unverified' status
**Decided: add `status = 'ok'` to the default dataset policy.** The shape below is settled, so
implement it rather than reopening the choice. The alternatives are recorded at the bottom so
nobody has to rediscover why they lost.

### What happens

`status = 'unverified'` landed in #180, so the catalog now says out loud when an episode's
critical check crashed and nobody actually checked it. `hflow dataset create` does not use it.
`default_dataset_sql` still opens with `status != 'quarantined'` (`src/hflow/dataset.py:203`),
which is the filter `docs/CATALOG.md:277` now names as a trap.

The default policy is not broken. It reaches almost the same outcome by a second rule: it
requires every registered step to have a `check_runs` row with a settled status, and `error` is
not in `SETTLED_STATUSES` (`src/hflow/steps.py:158`). An episode whose critical check only ever
crashed has no settled row for it and is dropped by that rule, not by the status one.

### The gap that closes

One case survives both rules, and an ordinary re-ingest reaches it:

> A settled run of the critical check exists earlier in the episode's history, and a later run
> crashed. The settled-step requirement is satisfied by the older row, while the status column
> reads `unverified`. The episode lands in the dataset.

`docs/CATALOG.md:289-293` describes exactly this today and tells the reader to add
`status = 'ok'` themselves. Asking every user to remember that is what the default is for.

Why this matters more than an ordinary filter change: `create_dataset` writes its SQL verbatim
into a `manifests/<slug>-<ts>.json` provenance sidecar that is never overwritten. Every dataset
created before this changes preserves the loose filter permanently.

### What to do

1. In `default_dataset_sql` (`src/hflow/dataset.py:203`), replace the
   `status != 'quarantined'` predicate with `status = 'ok'`. Keep the settled-steps requirement
   exactly as it is: it is strictly stronger for the case it covers, and this is an addition to
   the policy's strictness, not a replacement for it.
2. Update the docstring above it. It currently explains the two-rule arrangement; it should now
   explain that the status predicate covers the history case the step requirement cannot see.
3. Rewrite `docs/CATALOG.md:283-293`. That paragraph argues today that the divergence is the end
   state and hands the user a workaround. Both halves stop being true. Do not simply delete it:
   the two rules still differ in scope (dataset membership drops *any* errored registered step,
   `unverified` is critical-only), and that difference is worth keeping on the page.
4. `tests/test_dataset.py:100` pins the SQL literal (`assert "status != 'quarantined'" in sql`).
   It moves with the change.

### Definition of done

- `default_dataset_sql` filters `status = 'ok'`, and the settled-steps requirement is unchanged.
- A test builds the case above (a settled run of a check version, then a later run of the same
  version that errored) and pins that the episode is **excluded**. This is the one test that
  matters: without it the change is indistinguishable from a no-op, because every other
  crash-only episode was already excluded by the settled-steps rule.
- A test pins that an episode with a non-critical error is still **included**, so the change
  does not quietly borrow the status rule's critical-only scope for something wider.
- `docs/CATALOG.md` matches what lands, and still records that the two rules differ in scope.
- `tests/test_dataset.py`'s pinned literal is updated.

New datasets get smaller by exactly the case above. Existing manifests are immutable sidecars
and are not rewritten, so nothing already created changes.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

`packages/hflow-server/tests`. Markdown changes also want the lychee command from
CONTRIBUTING.md.

### Alternatives considered

- **Widen the status rule to any registered step, matching dataset membership.** Rejected:
  `unverified` would then fire when something non-critical crashed, which is a weaker claim than
  "nobody checked this". `status` is the user-facing vocabulary and is the wrong place to absorb
  a looser rule.
- **Leave both rules and keep documenting the divergence.** Rejected: `default_dataset_sql` is
  library code that writes its own SQL into a permanent provenance sidecar, so every dataset
  created from now on would preserve the trap filter.

### Notes

- Not a good first issue: it changes a shipped default's behaviour.
- Raised while reviewing #180. @chiruu12 examined this deliberately there and chose to document
  rather than tighten, which was the right call for that PR. Comment and it is yours.

chiruu12:
Closes #181.

Implements the shape settled in the issue body: `default_dataset_sql` filters on `status = 'ok'`, and the settled-steps requirement stays exactly as it was.

## The test

This is the part that decides whether the change is real, so it goes first.

`TestSettledThenCrashed` builds the one case the step requirement cannot see. A pipeline with a critical check ingests the episode once and settles, and the dataset picks it up. The check then crashes on a re-run and the dataset drops it.

The re-run goes through `ingest --all-stages`, which is what re-runs a step the catalog already records at its current version. That detail is load-bearing. Bumping the check version would also produce a crashed run, but rule 3 would then catch the episode on its own and the test would pass without the predicate change. Keeping the version fixed is what leaves only the status column doing the work.

Verified against unfixed source. With the predicate reverted to `status != 'quarantined'` and everything else in place, the test fails. It also asserts the two `check_runs` rows really are `measured` and `error`, so a future change that stops recording the crash cannot make it pass vacuously.

`ingest --all-stages` returns 1 here, not 0. The crash is a processing error and the CLI reports it as one. The test asserts that rather than working around it.

## The change

`src/hflow/dataset.py` uses `EPISODE_STATUS_OK` from the catalog module rather than a second copy of the literal, so the vocabulary keeps one owner.

The docstring's rule 2 changes from "not quarantined" to what the status column now means, and says which half of it rule 3 cannot see. I added a short paragraph on why both rules stay: an episode that only ever crashed is dropped by rule 3 alone, an episode that settled then crashed is dropped by rule 2 alone, so neither subsumes the other.

## Docs

`docs/CATALOG.md` had a paragraph telling the reader that the divergence was the end state and handing them `status = 'ok'` as a workaround. Both halves stop being true. I kept the scope difference, since the dataset policy and a hand-written selection query still are not the same thing, and dropped the workaround.

One existing test pinned the old string and now pins the new one.

## Gates

```
uv run ruff check --fix     All checks passed!
uv run ruff format          168 files left unchanged
uv run ty check             All checks passed!
uv run pytest -q            988 passed, 9 skipped
```

I did not run lychee. It is not installed here, and the docs edit changed prose and inline code only, so no link in the file is new or different.

## One thing I did not do

The four surfaces outside this issue still teach `status != 'quarantined'`: `README.md:192`, `docs/RUNTIME.md:210`, `docs/how-to/enable-built-in-checks.md:229` and `examples/egocentric/curate.sql:12`. I raised these on #180 and left them alone again here. Say the word and I will take them in a separate PR.

### DIFF
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -280,17 +280,20 @@ SELECT episode_id FROM episodes WHERE status != 'quarantined'
 SELECT episode_id FROM episodes WHERE status = 'ok'
 ```
 
-`hflow dataset create` is mostly already strict here, but by a different rule.
-Its default policy keeps `status != 'quarantined'` *and* requires every
-registered step to have a settled `check_runs` row, and `error` is not settled
-(see `SETTLED_STATUSES`). An episode whose critical check only ever crashed has
-no settled row for it and is excluded by that second rule, not the first.
+`hflow dataset create` applies `status = 'ok'` for you, so its default policy
+does not need this filter written by hand. It also requires every registered
+step to have a settled `check_runs` row, and `error` is not settled (see
+`SETTLED_STATUSES`).
 
-The gap is narrower than it looks, and it is worth knowing. If a settled run of
-that check version exists anywhere in the episode's history and a *later* run
-crashed, the step requirement is satisfied by the old row while the status
-column reads `unverified`, so the episode lands in the dataset. Filter on
-`status = 'ok'` in your selection SQL when you want the stricter reading.
+Those two rules overlap, and neither replaces the other. An episode whose
+critical check *only ever* crashed has no settled row at all, so the step
+requirement excludes it on its own. An episode that settled once and crashed on
+a later run is different: the step requirement is satisfied by the older row,
+and only the status column still reads `unverified`. That case is why the
+default policy filters on status rather than leaning on the step rule alone.
+
+Your own selection SQL gets neither rule for free. `status = 'ok'` is the one
+worth carrying over.
 
 Pinning a check version by hand (the mixed-version-corpus reality).
 `check_version` is a content hash, not ordered, so pin exact values (or filter
--- a/src/hflow/dataset.py
+++ b/src/hflow/dataset.py
@@ -28,7 +28,7 @@ from datetime import UTC, datetime
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from hflow.catalog import EPISODES_VIEW_STATUS_COLUMN
+from hflow.catalog import EPISODE_STATUS_OK, EPISODES_VIEW_STATUS_COLUMN
 from hflow.curation import CheckCoverage, CurationReport, curate
 from hflow.format import EPISODE_FORMAT_VERSION
 from hflow.steps import SETTLED_STATUSES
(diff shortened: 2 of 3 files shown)


kstonekuan (approved):
Merging. The test is the right one and it works: with the predicate reverted to `status != 'quarantined'` and everything else in place, it fails at `assert 1 == 0`. Holding the check version fixed by driving the crash from an environment variable rather than editing the source is the detail that makes it prove anything, and asserting the two `check_runs` rows are `measured` and `error` closes the vacuous-pass route. Ran lychee for you: 299 OK, 0 errors.

`EPISODE_STATUS_OK` instead of a second copy of the literal is better than what the issue asked for.

**Two corrections, one of them mine.**

**The four surfaces are already done.** I pushed them as `a8e02eb` while merging #180, onto your own branch, which is an easy thing to miss when it lands in someone else's commit. `README.md`, `docs/RUNTIME.md`, the built-in checks how-to, and `examples/egocentric/curate.sql` all say `status = 'ok'` on main today. Nothing left to take.

**The definition of done asked you for a test that cannot exist, and you were right to skip it.** I wrote that `an episode with a non-critical error is still included` should be pinned. It cannot be: `ERROR` is not in `SETTLED_STATUSES` (`steps.py:158`), so rule 3 drops that episode for want of a settled row no matter what rule 2 says. The property I actually wanted was that this change does not widen exclusion beyond critical-only, and that holds trivially, because such an episode was already excluded before the change. Skipping it was correct. Saying so would have been better than leaving me to find it.

The docs paragraph is exactly the shape I asked for, and the closing line about your own selection SQL getting neither rule for free is a better ending than the one I suggested.

Issue #171, opened 2026-08-25
A camera-less episode is planned for the media stage on every re-ingest


kstonekuan:
**Decided: ask the catalog whether the episode has cameras, and plan `media` only when it
does or when the catalog cannot say.** The shape below is settled; implement it rather than
reopening the choice. The two rejected alternatives are recorded at the bottom.

### What happens

The re-ingest planner asks, per episode and per step, whether the catalog already records that
step at its current version. The media stage's step is `media/contact_sheet`, and `App.process`
only runs it when the episode has cameras (`src/hflow/app.py:2057`):

```python
if Stage.MEDIA in enabled_stages and canonical_episode.cameras:
```

So a camera-less episode records no `media/contact_sheet` row, ever. The planner cannot tell "no
row because nothing ran" from "no row because there was nothing to render", and it schedules
`media` for that episode on every pass. `plan_outstanding_stages` states this as a known
limitation at `src/hflow/stage_planning.py:205-211`.

### What it costs, and where that changed

#195 gave the Airflow lane the same planner, so this now costs in both lanes. On a LOCAL root the
media sub-DAG's camera probe drops the episode before any work happens. On a BUCKET root that
probe is deliberately skipped, because reading a channel list downloads the whole episode
(`src/hflow/runtime/_bundle.py`, the `is_bucket_url` carve-out). So the episode is handed to
`process_batch`, which fetches the entire canonical file to discover there are no cameras, once
per episode per ingest.

Visible as `media: N processed` on a corpus where every other stage reports `0 processed, N
already current`.

### What to build

The catalog already knows. `camera_frame_stats` is a default and records a `check_runs` row on
every episode, including camera-less ones, while emitting no per-camera measurement keys:

```
camera-less episode -> check_runs: [('camera_frame_stats', 'measured'), ('media_digest', 'measured')]
                       per-camera measurement keys: none
```

So the question "does this episode have cameras" is answerable from stored rows, with no new
column and no `CATALOG_FORMAT_VERSION` bump:

1. In `plan_outstanding_stages`, when deciding whether `media` is outstanding for an episode,
   ask whether a `camera_frame_stats` row exists for that `episode_id` at the version
   `_DEFAULT_CHECK_VERSION_BY_FUNCTION[camera_frame_stats]` records (`checks.py:1912`), and
   whether any measurement key for that episode matches the per-camera shape the check emits.
2. If a row exists and no per-camera keys do, the episode has no cameras. Do not plan `media`.
[...]

PR #185, opened 2026-08-26
refuse empty and whitespace-only measurement keys before the wide-view pivot

### LINKED ISSUE #160: An empty measurement key becomes a wide-view column named after the SQL that made it
### What happens

An empty measurement key is accepted, and the wide view then carries a column whose *name is the
SQL expression that produced it*. Measured on `e6103d0`, appending
`measurements={"": 1.0}` and listing the `episodes` view columns:

```
['episode_id', 'run_fingerprint', 'orchestrator_run_id', 'uri', ..., 'status',
 '"first"("value") FILTER (WHERE (CAST("key" AS VARCHAR) IS NOT DISTINCT FROM \'\'))']
```

The pivot has no name to give the column, so DuckDB falls back to rendering the expression. The
value is stored and is not reachable by any name a person would write, and the column name leaks
the view's internals into a surface documented as a stable queryable contract.

### Why it matters beyond being ugly

`docs/CATALOG.md` ("Naming measurement keys") states the rule this violates:

> Because every numeric key becomes a column of the wide view, key names are a queryable surface
> with no rename path: old rows keep the old name forever.

An empty key is that surface with nothing on it. It cannot be selected, cannot be renamed later,
and unlike the collisions #130 and #151 already refuse, nothing warns at any point.

### Definition of done

`Catalog.append_episode` refuses an empty or whitespace-only measurement key, naming the check.

Whitespace-only is included deliberately: a key of `" "` has the same "no name a person would
write" problem, and stripping it silently would be a rename, which the doc above says never
happens.

Consider whether the same rule belongs on tag strings and interval labels, which are also
accepted empty today. If you include them, say so and test them; if you leave them, say why. An
empty tag or label is less harmful than an empty key, because neither becomes a column, so
scoping this to measurement keys alone is a defensible answer.

### Pattern to copy

`_normalized_measurements` (`catalog.py:276`) already refuses a value it cannot store, naming the
check and key. This is the same boundary and the same message shape. Its tests live in
`tests/test_catalog_curation.py`.

1. `measurements={"": 1.0}` is refused, naming the check.
2. `measurements={"   ": 1.0}` is refused too.
3. A test asserts the refusal happens before anything is written, the way
   `test_measurement_key_claiming_an_episode_column_is_refused` does.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

ayam04:
Fixes #160

## What changed

`_normalized_measurements` (catalog.py) now refuses an empty or whitespace-only measurement key, naming the check and the key in the same message shape as its existing value-coercion refusals. The check runs inside `append_episode` before anything is written (the episodes file is written manifest-last), so a refused append leaves no partial state.

Tags and interval labels are intentionally left out, per the issue's own reasoning: neither becomes a wide-view column, so the "no name a person would write" surface problem does not apply to them. Stripping whitespace silently was rejected as a rename, which docs/CATALOG.md ("Naming measurement keys") says never happens.

## Tests

Two new tests in tests/test_catalog_curation.py, mirroring `test_measurement_key_claiming_an_episode_column_is_refused`:

- `test_empty_measurement_key_is_refused` — `measurements={"": 1.0}` raises ValueError naming the check; no parquet written.
- `test_whitespace_only_measurement_key_is_refused` — `measurements={"   ": 1.0}` same.

## Validation

```
.venv/bin/python -m pytest tests/test_catalog_curation.py -q   # 65 passed
.venv/bin/python -m pytest -q                                  # 940 passed; 2 failed + 16 errors ALL confined to tests/test_ffmpeg.py (pre-existing env: no ffmpeg binary in WSL; reproduced in isolation)
.venv/bin/ruff check src/hflow/catalog.py tests/test_catalog_curation.py   # clean
.venv/bin/ruff format --check src/hflow/catalog.py tests/test_catalog_curation.py  # clean
.venv/bin/ty check                                              # clean
```

### DIFF
--- a/src/hflow/catalog.py
+++ b/src/hflow/catalog.py
@@ -339,6 +339,12 @@ def _normalized_measurements(
     threshold and its complement at once, so storing one drops the episode
     from both sides of a cut while the key still claims to be measured.
 
+    An empty or whitespace-only key is refused as well: the wide view pivots
+    every key into a column, so an empty key becomes a column whose name is
+    the SQL expression that produced it -- a queryable surface with no name
+    a person would write and no rename path (docs/CATALOG.md, "Naming
+    measurement keys"). Refusing it here means nothing has been written yet.
+
     Real check code returns NumPy scalars (np.mean(), indexing, np.bool_
     verdicts), and only np.float64 subclasses a Python type -- the rest used
     to fall through every storage branch into all-NULL value columns while
@@ -359,6 +365,12 @@ def _normalized_measurements(
             raise ValueError(
                 f"check {check_name!r} measured {key!r} as {value!r}: "
                 "omit the key when a check has no finite value for this episode"
+            )
+        if not key.strip():
+            raise ValueError(
+                f"check {check_name!r} measured {key!r}: measurement keys must "
+                "not be empty or whitespace-only (an empty key becomes a "
+                "wide-view column named after the SQL that made it)"
             )
         normalized[key] = value
     return normalized
--- a/src/hflow/catalog.py
+++ b/src/hflow/catalog.py
@@ -354,6 +354,15 @@ def _normalized_measurements(
     """
     normalized: dict[str, MeasurementValue] = {}
     for key, value in measurements.items():
+        # Before the value checks: the key is wrong independently of what it
+        # holds, and reporting the value first would send a check author to
+        # fix a NaN only to hit the real problem on the next run.
+        if not key.strip():
+            raise ValueError(
+                f"check {check_name!r} measured {key!r}: measurement keys must "
+                "not be empty or whitespace-only (an empty key becomes a "
+                "wide-view column named after the SQL that made it)"
+            )
         if isinstance(value, np.generic):
             value = value.item()
         if not isinstance(value, MeasurementValue):
@@ -365,12 +374,6 @@ def _normalized_measurements(
             raise ValueError(
                 f"check {check_name!r} measured {key!r} as {value!r}: "
                 "omit the key when a check has no finite value for this episode"
-            )
-        if not key.strip():
-            raise ValueError(
-                f"check {check_name!r} measured {key!r}: measurement keys must "
-                "not be empty or whitespace-only (an empty key becomes a "
-                "wide-view column named after the SQL that made it)"
             )
         normalized[key] = value
     return normalized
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
Correct, and the reasoning for leaving tags and interval labels alone is right: neither becomes a column, so the problem does not exist there. Not stripping whitespace silently is also right, for the reason you give.

Both tests are real tripwires. I stubbed the condition out and they go red. Your ffmpeg failures were environmental as you suspected: 958 passed here with a working binary.

Pushed one change, `c622cc8`. The key check now runs before the value checks, since an empty key is wrong whatever it holds. Previously `{"": float("nan")}` reported the NaN, so a check author would fix that and hit the real cause on the next run.

On the pool: that is nine merged now, so you are well past needing starter issues, and they are most useful to people arriving for the first time. No rule, and I am not going to police it.

The more valuable thing, if you want it: run HFlow against a real corpus and report what breaks. Someone did exactly that this week and it produced a genuine transform limitation nobody had found from the inside, which is now a PR. Egocentric-10K or Egocentric-100K on Hugging Face are reasonable starting points. That kind of report is worth more than anything currently in the tracker, and only someone actually using the tool can write it.

ayam04:
Thank you for the merge @kstonekuan — 12th one in, and the camera-less corner case is closed. Next lane ready whenever you want it (the LeRobot v3 converter issues look like a good deeper cut; happy to take #171's design question to a concrete answer too).

kstonekuan:
Take the LeRobot ones. They are the deeper cut you are asking for, and they are labelled `advanced` rather than starter, which is where you should be at this point.

They read as a chain, so order matters:

- **#189** generalizes the PushT converter to real multi-camera v3 datasets. Everything else assumes it.
- **#190** exports an HFlow-curated selection back out as a loadable v3 dataset, which is the return leg.
- **#191** ties both into one reproducible real-camera example.

#189 first unless you see a reason otherwise. Comment on whichever you take and I will assign it, since GitHub silently drops an assignee who has not posted on the issue.

On #171: it carries a `good first issue` label but it is really a design question with three answers and no obvious winner, so it is not the deeper cut it looks like from the title. Taking it would mean proposing a direction rather than implementing one. I have just added a note there about how #195 changed where it costs, in case you want the context either way.

One correction: #185 closed the empty-measurement-key hole. The camera-less corner is #171 and is still open.

PR #186, opened 2026-08-26
test(storage): cover bucket create-if-absent races

### LINKED ISSUE #165: Concurrent appends to a bucket-backed catalog have no test coverage
### What we want

Prove the catalog's append idiom holds on a **bucket** root. It is claimed there, implemented
there, and tested only on local roots.

This is a coverage issue, not a bug report. I looked for a race and did not find one; the
implementations on both backends are correct as far as reading goes. What is missing is any test
that says so for the backend where the guarantee is hardest.

### Current state

`Catalog.append_episode` rests on one invariant, stated at `catalog.py:596-601`:

> the episodes file is written last (manifest-last, in miniature), so its existence proves the
> whole append completed

Exclusive create is what makes that true, and each backend implements it differently:

- `LocalStorageRoot.write_bytes_if_absent` (`storage.py:281`) writes a temporary file and then
  `os.link`s it into place, treating `FileExistsError` as "lost the race".
- `BucketStorageRoot.write_bytes_if_absent` (`storage.py:482`) uses `PutMode::Create` and treats
  `AlreadyExistsError` the same way. Its docstring makes the strongest claim in the file:

  > the store itself refuses the second writer, so two workers appending the same content-named
  > file can never corrupt or duplicate it

`test_concurrent_append_of_the_identical_outcome_keeps_one_recorded_at`
(`tests/test_catalog_curation.py:1268`) proves the invariant properly, with a barrier-gated
storage double forcing a real interleaving rather than hoping for one. It runs against a local
root.

No test in `tests/test_catalog_curation.py` uses a bucket root at all. `BucketStorageRoot` appears
only in `tests/test_storage.py` and `tests/test_processing_regressions.py`, neither of which races
appends.

### Why the bucket case is the one worth proving

The local guarantee is a filesystem primitive that has behaved the same way for decades. The
bucket guarantee is a much younger contract: `PutMode::Create` is honoured differently across
object stores, and obstore's `file://` backend (what a test would use) is a third implementation
again. If any of them treats "create" as "put", the failure is not a crash. Two writers both
believe they won `episodes`, both proceed to publish dependents with `publish` (an unconditional
overwrite, deliberately, because the winner is supposed to be unique), and the catalog ends up
with one episode row and dependents from two different appends. That is the exact corruption
`recorded_at` alignment exists to prevent, arriving through a door nothing watches.

### Definition of done

A test that races two `append_episode` calls for the identical outcome against a
`BucketStorageRoot`, and asserts the same invariants the local test does:

1. Exactly one caller reports `written=True`.
2. Every table file for the append carries one `recorded_at`.
3. Every dependent table file exists afterwards.

Plus one assertion the local test cannot make: that `write_bytes_if_absent` on the bucket backend
actually refused the second writer, rather than the second writer having been serialized by luck.
Count the refusals; a test that passes because the threads never overlapped is not evidence.

### Pattern to copy

`test_concurrent_append_of_the_identical_outcome_keeps_one_recorded_at`
(`tests/test_catalog_curation.py:1268`) for the whole shape, especially its `GatedLocation`
barrier double, which is what turns an intermittent race into a deterministic one. Point the same
technique at a bucket root.

`bucket_over_tmp` (`tests/test_storage.py:33`) builds a `BucketStorageRoot` over a local directory
with `file://`, which is how you get a bucket backend without a network or credentials. Lift it
into a shared fixture rather than importing across test modules, or duplicate it with a comment
saying why.

### Notes

- If obstore's `file://` backend turns out not to implement `PutMode::Create` faithfully, that is
  a finding worth its own issue, not a reason to weaken the test. Say so in the PR and we will
  decide what to do about it.
- `_reconciled_append_stems` (`catalog.py:102`) is process-local, so two threads share it and two
  processes do not. If you find that distinction matters for what you are asserting, note it;
  making the test multi-process is out of scope here.
- Do not add locking or change `catalog.py` under this issue. If the test finds a real defect,
  report it and we will scope the fix separately.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

Run plain `uv run pytest -q` from the repo root. Run the new test repeatedly (`pytest ... --count`
or a loop) before opening the PR: a concurrency test that passes once has told you very little.

WilliamK112:
## Summary

Add deterministic bucket-backed concurrency coverage for the three create-if-absent publishers named in #165:

- catalog episode appends
- dataset manifest writes
- ingest-failure ledger records

Each race holds two callers at the real `BucketStorageRoot.store_file_if_absent` operation, counts the losing conditional put, and verifies the published business outcome in the backing store. The existing `file://` bucket helper is now a shared fixture so these tests and the storage suite exercise the same obstore-backed seam.

## Why

The local-storage races did not prove that the bucket conditional-put path arbitrates concurrent writers. These tests make both callers pass their early existence checks, release them together at the actual store operation, and require exactly one refusal. They then verify the resulting artifacts: one complete catalog outcome with a single `recorded_at`, one readable one-row manifest, and one queryable ledger row.

This is intentionally test-only. No catalog, dataset, ledger, or storage production behavior changes.

Closes #165.

## Validation

- `python -m pytest -q tests/test_bucket_create_if_absent.py tests/test_storage.py` — 44 passed, 1 skipped
- 50 consecutive runs of `python -m pytest -q tests/test_bucket_create_if_absent.py` — all passed
- `python -m pytest -q` — 967 passed, 9 skipped
- `ruff check .` — passed
- `ruff format --check .` — passed
- `ty check` — passed

## Checklist

- [x] Documentation is not applicable: this is a test-only change with no changed behavior, flags, formats, or requirements.
- [x] I ran `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check` (using the existing project virtual environment).
- [x] I preserved stored-data compatibility; production and stored formats are unchanged.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,9 +5,30 @@ never trigger the pinned-build download (a per-machine, network-bound step).
 Set at import time, before any test imports resolve the (cached) binary path.
 """
 
+from __future__ import annotations
+
 import os
 import shutil
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+import pytest
+
+if TYPE_CHECKING:
+    from hflow.storage import BucketStorageRoot
 
 _system_ffmpeg = shutil.which("ffmpeg")
 if _system_ffmpeg is not None:
     os.environ.setdefault("HFLOW_FFMPEG", _system_ffmpeg)
+
+
+@pytest.fixture
+def bucket_over_tmp(tmp_path: Path) -> tuple[BucketStorageRoot, Path]:
+    """A real bucket root over obstore's local backend and its remote dir."""
+    pytest.importorskip("obstore", reason="bucket tests need the hflow[bucket] extra")
+    from hflow.storage import BucketStorageRoot
+
+    remote_dir = tmp_path / "bucket"
+    remote_dir.mkdir(parents=True, exist_ok=True)
+    root = BucketStorageRoot(f"file://{remote_dir}", mirror=tmp_path / "bucket-mirror")
+    return root, remote_dir
--- a/tests/test_bucket_create_if_absent.py
+++ b/tests/test_bucket_create_if_absent.py
@@ -1,0 +1,203 @@
+"""Deterministic races for bucket-backed create-if-absent publishers."""
+
+from collections.abc import Callable
+from concurrent.futures import ThreadPoolExecutor
+from pathlib import Path
+from threading import Barrier, Lock
+
+import duckdb
+import pytest
+
+import hflow
+from hflow.catalog import AppendResult, Catalog, CheckRunRow
+from hflow.curation import open_catalog_connection
+from hflow.dataset import ManifestAlreadyExistsError, write_dataset_manifest
+from hflow.ingest_ledger import IngestFailure, record_ingest_failure
+from hflow.storage import BucketStorageRoot
+from hflow.transform import EpisodeStamps
+from hflow.workspace import Workspace
+
+pytest.importorskip("obstore", reason="bucket tests need the hflow[bucket] extra")
+
+
+class _CreateIfAbsentRace:
+    """Hold two matching puts at the real storage operation and count losses."""
+
+    def __init__(
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+        target: Callable[[BucketStorageRoot, str], bool],
+    ) -> None:
+        self.attempts = 0
+        self.refusals = 0
+        self._barrier = Barrier(2)
+        self._lock = Lock()
+        store_file_if_absent = BucketStorageRoot.store_file_if_absent
+
+        def gated_store_file_if_absent(
+            root: BucketStorageRoot, local_file: Path, relative: str
+        ) -> bool:
+            if not target(root, relative):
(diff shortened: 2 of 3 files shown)


kstonekuan (approved):
This closes the gap the issue actually described, which is the part I was unsure anyone would get right. The `file://` root still goes through `obstore.put(..., mode="create")` rather than the local `os.link` path, so these tests exercise the conditional-put seam and not a lookalike. That was the whole point of #165 and it would have been easy to miss.

The barrier is placed correctly: both callers clear their early existence check, meet at the real store operation, and only then race. Counting refusals is what stops accidental serialization from passing as a result.

I checked they would actually fail against a backend without exclusivity, by dropping `mode="create"` so the put becomes last-writer-wins. All three go red, on the refusal count and on the published outcome. Then 40 consecutive runs of the file, no flakes, and 970 passed on the full suite.

Filtering the patch by relative key so unrelated puts pass through untouched is a nice detail. A blanket patch on the class method would have made every other bucket write in the test process part of the race.

Merging. #165 was mine, filed when I went looking for a race in this code and found the implementation correct but untested. Better outcome than the bug I was expecting.

Issue #191, opened 2026-08-26
LeRobot: add a reproducible real-camera import, quality, curate, and export example


kstonekuan:
## Goal

Provide one reproducible real-camera workflow that proves the LeRobot adapter works beyond a simulated PushT fixture: import source episodes, run HFlow quality checks, curate from recorded evidence, and export a standard LeRobot Dataset v3 selection.

## Current behavior

Rewritten 2026-08-28, after #189 landed as #230. The original text described a PushT-only converter, which is no longer what the repository has.

`examples/lerobot/prepare.py` now converts any LeRobot Dataset v3 repository, reading the feature schema, frame rate, episode boundaries, and video paths from `meta/info.json` rather than constants, and converting every selected camera into its own channel. `examples/README.md:98-113` documents the supported subset and both commands. So import is solved, and this issue no longer needs to argue for it.

What is still missing is the rest of the workflow. Converting one episode is a command; there is no bounded run that a contributor can execute from a clean checkout to go from a pinned source revision through quality checks and a curation query to an exported selection, and reproduce the same result twice. That composition is what this issue is for, and it is the only place the pieces are shown working together.

## Dependencies

- #189 provides general LeRobot v3 import and multi-camera conversion. **Done**, merged as #230.
- #190 provides export of an HFlow curation result back to LeRobot Dataset v3. Still open and unassigned, and it is the one thing blocking this.

Do not duplicate either implementation in this example. This issue composes their public callables and commands.

## Acceptance corpus

Use this pinned public dataset:

- Repository: lerobot/svla_so101_pickplace
- Revision: f641879e22172be7e8161d5e6c1503c2d2feb657
- License: Apache-2.0
- Shape: 50 real SO-100 episodes, 30 frames per second, two 640 by 480 cameras, and six-dimensional state and action
- Total download: about 86 MB

Process a bounded, explicitly listed episode subset. The workflow must make the subset easy to change without embedding this repository or its feature names into shared adapter code.

## What to build

Add a focused runnable example under examples/lerobot that:

1. Downloads the pinned source revision and verifies the resolved commit.
2. Converts an explicitly listed episode subset to canonical MCAP with every declared camera.
3. Runs hflow doctor on the converted inputs.
4. Processes the episodes through an HFlow App using the applicable default checks and contact-sheet enrichment.
5. Cuts a training selection with an explicit SQL query over recorded quality evidence.
[...]

Issue #190, opened 2026-08-26
LeRobot: export an HFlow-curated selection as a loadable v3 dataset


kstonekuan:
## Goal

Turn an HFlow curation result over LeRobot-originated episodes into a standard LeRobot Dataset v3 repository that existing LeRobot training and visualization tools can load.

## Current behavior

The converter in examples/lerobot/prepare.py records source_dataset, source_revision, and source_episode_index in episode/v1 at examples/lerobot/prepare.py:742-748. HFlow preserves unpromoted episode metadata in the catalog metadata_json column, as documented in docs/CATALOG.md:42.

This provides a traceable path from a curated HFlow episode back to its original LeRobot episode. No callable or command currently uses that path to materialize a selected LeRobot dataset.

## Dependency

#189 is merged (by #230), so the metadata contract is settled and no longer needs agreeing first. The converter stamps `source_dataset`, `source_revision`, `source_episode_index`, `task`, and `embodiment` on every output episode, with the revision resolved to an immutable commit rather than a branch name. Read those five keys. Adding to them is a change to #189's contract, so say why here before you do.

## What to build

Add a standalone export callable and command under examples/lerobot. The exporter must:

1. Accept an HFlow catalog plus either a curation SQL query or a manifest produced by hflow curate.
2. Resolve each selected episode to source_dataset, source_revision, and source_episode_index from the catalog.
3. Require all selected source references to be immutable and fail clearly when a source revision or episode index is missing.
4. Read the selected episodes through the supported LeRobot Dataset v3 interface.
5. Create a new local LeRobot Dataset v3 repository containing only those episodes.
6. Preserve the selected cameras, state, action, task, feature schema, frame timing, and source ordering.
7. Finalize the dataset before validation.
8. Record the source repository, source commit, HFlow version, curation query or manifest digest, and selected source episode indexes in machine-readable provenance and in a generated dataset-card draft.
9. Produce local output only. Uploading to Hugging Face is not part of this issue.

Scope the first version to episodes that entered HFlow through the LeRobot adapter and still reference one source repository and one immutable source revision. Converting arbitrary MCAP schemas to LeRobot, merging incompatible feature schemas, and publishing to the Hub are out of scope.

## Failure behavior

Fail before writing a publishable output when:

- a selected episode lacks a LeRobot source reference;
- source repositories or revisions are mixed;
- the selected source feature schemas are incompatible;
- a selected source episode does not exist;
[...]

Issue #189, opened 2026-08-26
LeRobot: generalize the PushT converter to real multi-camera v3 datasets


kstonekuan:
## Goal

Convert selected episodes from general LeRobot Dataset v3 repositories into canonical MCAP without encoding PushT assumptions into the adapter.

## Current behavior

examples/lerobot/prepare.py proves the Parquet, video, timestamp, and MCAP mapping end to end, but the public callable is still dataset-specific:

- DEFAULT_REPO and DEFAULT_CAMERA_KEY at examples/lerobot/prepare.py:35-38 select lerobot/pusht and one camera.
- Episode count and duration are fixed at 206 and 10 seconds at examples/lerobot/prepare.py:447-448.
- State and action are fixed to two float32 values at examples/lerobot/prepare.py:651-652.
- The source file roles and dataset license are assembled from PushT constants rather than repository metadata.

A real LeRobot v3 repository can use different episode boundaries, frame rates, vector widths, tasks, and multiple video features. The current callable accepts a repository argument but cannot correctly convert those layouts.

## What to build

Keep the adapter standalone and preserve the existing lerobot_to_mcap callable and command behavior for PushT. Generalize its inputs as follows:

1. Read the v3 feature schema, frames per second, episode boundaries, tasks, data paths, and video paths from repository metadata instead of constants.
2. Resolve and stamp the immutable source commit, not only the requested branch name.
3. Discover video features from the schema. Convert every selected camera into a separate foxglove.CompressedVideo channel. Let the command select one or more camera keys, and provide a documented default.
4. Derive fixed-width numeric state and action schemas from the declared dtype, shape, and names. Fail before conversion with the feature name, dtype, and shape when a feature is unsupported.
5. Derive timestamps and episode lengths from the source metadata while preserving the existing fail-loud comparison between tabular timestamps and video presentation timestamps.
6. Preserve source_dataset, source_revision, source_episode_index, task, and embodiment metadata on every output episode.
7. Keep LeRobot-specific dependencies out of the hflow core package.

Scope this issue to LeRobot Dataset v3, video camera features, and fixed-width numeric state and action vectors. Image-only features, language tensors, depth maps, and arbitrary nested features are out of scope unless the contributor proposes and agrees on a small extension first.

## Acceptance corpus

Validate against this pinned public real-camera dataset:

- Repository: lerobot/svla_so101_pickplace
- Revision: f641879e22172be7e8161d5e6c1503c2d2feb657
[...]

PR #192, opened 2026-08-26
test(cli): cover stale pipeline app errors

### LINKED ISSUE #68: Tests: `hflow stale --pipeline` with no App variable has no coverage

WilliamK112:
Closes #68

## Summary

- cover an explicitly named App that is absent from a pipeline module
- cover a pipeline module that defines no `hflow.App`
- cover an ambiguous pipeline module and require stderr to name both App candidates

All three tests exercise the `hflow stale --pipeline` CLI boundary, including exit code 2 and the user-facing stderr contract. Production code is unchanged.

## Validation

- focused three-test replay: 3 passed
- mutation check: changing `_command_stale`'s pipeline-resolution error return from 2 to 0 made all three new tests fail; restoring it returned them to green
- `uv run ruff check --fix`
- `uv run ruff format`
- `uv run ty check`
- `uv run pytest -q` — 976 passed, 9 skipped

Disclosure: I used Codex to assist with implementation and verification, and reviewed the final diff and test evidence.

### DIFF
--- a/tests/test_catalog_curation.py
+++ b/tests/test_catalog_curation.py
@@ -656,6 +656,62 @@ def test_cli_stale_reports_a_broken_pipeline_file_instead_of_crashing(
     assert "boom at import time" in capsys.readouterr().err
 
 
+def test_cli_stale_reports_an_explicitly_named_missing_app(
+    tmp_path: Path, capsys: pytest.CaptureFixture[str]
+) -> None:
+    Catalog(tmp_path / "catalog")
+    pipeline = tmp_path / "pipeline.py"
+    pipeline.write_text("value = 42\n")
+
+    exit_code = cli_main(
+        [
+            "stale",
+            "--catalog",
+            str(tmp_path / "catalog"),
+            "--pipeline",
+            f"{pipeline}:custom_name",
+        ]
+    )
+
+    assert exit_code == 2
+    assert "has no hflow.App named 'custom_name'" in capsys.readouterr().err
+
+
+def test_cli_stale_reports_a_pipeline_with_no_apps(
+    tmp_path: Path, capsys: pytest.CaptureFixture[str]
+) -> None:
+    Catalog(tmp_path / "catalog")
+    pipeline = tmp_path / "pipeline.py"
+    pipeline.write_text("value = 42\n")
+
+    exit_code = cli_main(
+        ["stale", "--catalog", str(tmp_path / "catalog"), "--pipeline", str(pipeline)]
+    )
+
+    assert exit_code == 2
+    assert "defines no hflow.App" in capsys.readouterr().err
+
+
+def test_cli_stale_reports_every_app_when_a_pipeline_is_ambiguous(
+    tmp_path: Path, capsys: pytest.CaptureFixture[str]
+) -> None:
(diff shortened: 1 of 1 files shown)


kstonekuan (approved):
Matches #68 exactly, and the ambiguous case asserting both App names rather than just the count is the right call. 979 passed.

I re-ran your mutation independently. Changing the `return 2` at `cli.py:731` to `0` fails all three, plus the pre-existing broken-pipeline test, so the exit-code contract is genuinely pinned rather than incidentally satisfied.

On the pool: that is five merged, so starter issues are better left to people arriving for the first time. Not a rule.

The higher-leverage work is what you have already been doing this week. #149 started as you running HFlow on a real corpus and ended as a transform capability nobody would have found from the inside. There are three new LeRobot issues (#189, #190, #191) that want the same instinct, and they are not starter issues.

PR #195, opened 2026-08-26
feat(runtime): plan outstanding stages in the Airflow lane

### LINKED ISSUE #172: Stage planning is direct-executor only; the Airflow lane re-runs everything
### What happens

`hflow.stage_planning` makes a re-ingest cost only what the catalog does not already record
as current, and `run_stages_directly` applies it per episode. The Airflow lane does not: the
master DAG resolves a run profile to a stage set and triggers those sub-DAGs over every
episode it is handed, whatever their state.

So the two executors now have materially different costs for the same command. Re-ingesting
an unchanged corpus in-process skips the decode passes and the contact sheets; the same
corpus through `hflow ingest --bundle-dir ...` pays for all of them.

### Why it was left out

Two reasons, both real:

- **The conf cannot express it.** The trigger conf carries a profile NAME, validated against
  a `RUN_PROFILES` literal baked into the DAG at render time. A per-episode stage set is not
  a profile name, and adding one means `hflow up` again on every deployed bundle.
- **The plan is built between stages.** In-process, the plan is computed after `sync` has
  appended its rows and before `meta` runs -- that ordering is what makes it correct, because
  nothing in the catalog hashes the source, so a plan built before `sync` would skip `meta`
  for a recording whose bytes changed underneath it. Across sub-DAGs there is no equivalent
  moment: each stage is a separate DAG run.

### Possible shape

The second constraint is the interesting one, and it may already be solved: each stage
sub-DAG maps `process_batch` over its own batch, so the filtering could happen **inside the
sub-DAG** at map time rather than in the master's conf. A `meta` sub-DAG that asks the
catalog which of its `uris` still have outstanding checks needs no new conf vocabulary and no
re-render of the master -- and it runs after `sync` completed, which is exactly the ordering
that makes the answer sound.

Worth checking whether that holds for `labels` and `media` too, and what it does to the
mass-failure budgets, which currently apply to what a stage was handed.

### Definition of done

- A scheduled re-ingest of an unchanged corpus does not re-run stages whose steps the catalog
  already records at their current versions.
- The escape hatch exists on this path too (`--all-stages` is in-process only today).
- The ordering argument in `hflow/stage_planning.py`'s module docstring is either satisfied
  by the chosen design or explicitly amended for the scheduled case.
- `docs/RUNTIME.md` currently says outright that this applies to the in-process executor
  only; that paragraph is the one to update.

chiruu12:
Closes #172.

A scheduled re-ingest now costs what an in-process one costs. Each stage sub-DAG filters its own `uris` at plan time and skips outright when nothing is left, so an unchanged corpus stops paying for decode passes and contact sheets the catalog already records.

The predicate is `outstanding_stage_uris` in `hflow.stage_planning`, sharing `plan_outstanding_stages` with `run_stages_directly` so the two lanes cannot answer differently. What the new function adds is the translation: conf `uris` are data-root-relative strings, the catalog files rows under `source_uri`, and computing that identity a second way is how a planner queries for rows filed under another name and reads everything as outstanding.

It reaches the DAGs through the slot `MEDIA_PLAN_FILTER_TEMPLATE` already uses, on every stage but sync. Sync records no steps and is its own cache, so `Stage.SYNC` raises rather than returning an answer. On media both filters apply and the catalog query goes first, since it does not open episodes and the camera probe does.

Bucket roots keep the filter. The media probe skips them because reading a channel list downloads the whole episode; a catalog query syncs catalog parquet through the mirror and never opens an episode, so that reason does not carry over.

`all_stages` is new trigger conf on the master and on every sub-DAG, the scheduled counterpart of `--all-stages`. The master passes it down, and a directly triggered sub-DAG takes it in its own conf. A string value is read the way every other flag here reads one, since a hand-typed conf arrives as text.

Two behaviour changes worth stating rather than leaving to be discovered.

`plan` now loads the user pipeline. It already ran in the user venv and these are the same three helpers `process_batch` calls, so no new machinery, but a pipeline that will not import now fails at `plan` instead of at `process_batch`.

A recording with no episode row is handed on, not dropped. This is the one place the scheduled filter differs from `plan_outstanding_stages`, and the difference is deliberate. That function drops such a recording because the direct executor ran sync itself and so knows the row is missing because sync failed. A stage sub-DAG cannot tell that from a corpus nobody synced, and dropping it would turn the loud `FileNotFoundError` a meta-only run raises today into a silent skip. Filtering is here to stop paying for finished work, not to swallow failures.

Budgets are computed after the filter, which was already the case: `_tally_batch_counts` counts what `process_batch` reported, not what the stage was handed. Below 800 episodes `run_failure_budget` is 8 either way. The gate that does change is `errors and errors == total`, which fires outside the budget: filter a run down to two episodes and have both error and it fails loudly, where 200 handed over with 2 errors would not. Everything that ran failed, so that is the right answer, but it is a change.

On evidence. I cannot run Airflow end to end, so the tests split three ways rather than claiming a full path was executed. `tests/test_stage_planning.py` runs the predicate against a real catalog. `tests/test_runtime_bundle.py` pins the wiring: which sub-DAGs carry the filter, which stage each asks about, the ordering against the camera probe, and the conf reaching it. Between them, `TestTheRenderedPlanTask` extracts the rendered `plan` body, strips its Airflow decorator and calls it, which is what catches a bad name or a missed substitution in an injected string; that snippet's contract is that it runs under plain CPython in the user venv, so calling it directly is a fair test of it. What none of them cover is Airflow scheduling the task, mapping the batches, or `skip_on_exit_code` turning exit 99 into a SKIPPED task.

Validation:

```
uv run pytest -q                                  992 passed, 9 skipped
uv run --python 3.11 --locked pytest -q           992 passed, 9 skipped
uv run --python 3.14 --locked pytest -q           992 passed, 9 skipped
uv run ruff check                                 All checks passed
uv run ruff format --check                        169 files already formatted
uv run ty check                                   All checks passed
```

The new tests fail against unpatched `src/` (15 failures), including the three that execute the rendered plan.

No `TRANSFORM_BEHAVIOR_VERSION` bump: nothing here changes the bytes the transform writes, only which episodes a stage is handed.

`docs/RUNTIME.md` said outright that planning applied to the in-process executor only. That paragraph is rewritten.

### DIFF
--- a/docs/RUNTIME.md
+++ b/docs/RUNTIME.md
@@ -63,10 +63,15 @@ step whose rows still say it ran -- a cleaned `media/` directory keeps its
 hflow ingest --all-stages episodes-in/*.mcap
 ```
 
-This applies to the in-process executor only. The Airflow lane runs the whole
-profile on every episode it is handed, because a stage set that varies per
-episode is not something a trigger conf can express without re-rendering every
-bundle.
+The Airflow lane plans too, and asks the same question in the same place. Each
+stage sub-DAG filters its own `uris` at plan time, after the sync sub-DAG has
+finished, so it reads the catalog state the in-process plan would read. A stage
+left with nothing outstanding skips rather than reporting a hollow success.
+
+`--all-stages` has a counterpart there: pass `all_stages: true` in the trigger
+conf and every stage runs over everything it was handed, whatever the catalog
+says. The master passes it down to every sub-DAG it triggers, and a directly
+triggered sub-DAG takes it in its own conf.
 
 ## Prerequisites
 
--- a/src/hflow/runtime/_bundle.py
+++ b/src/hflow/runtime/_bundle.py
@@ -81,6 +81,7 @@ from hflow.runtime._templates import (
     EXTERNAL_PYTHON_BOOTSTRAP_REQUIREMENT,
     MASTER_DAG_TEMPLATE,
     MEDIA_PLAN_FILTER_TEMPLATE,
+    OUTSTANDING_PLAN_FILTER_TEMPLATE,
     SUB_DAG_ERROR_GATE_TEMPLATE,
     SUB_DAG_QUARANTINE_GATE_TEMPLATE,
 )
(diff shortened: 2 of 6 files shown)


kstonekuan (approved):
Merging. This is the most careful PR the project has had, and the thing that makes it reviewable is that the PR body states the behaviour changes instead of leaving them to be found.

`TestTheRenderedPlanTask` is the part worth copying elsewhere. Parsing the rendered sub-DAG, stripping the Airflow decorator off `plan` and calling the real body closes the exact gap that made this feature hard to trust: an injected string compiles fine and only fails when a task runs. I mutated `outstanding_stage_uris` to filter nothing and four tests went red including that one, so it is pinning behaviour rather than shape.

Keeping a recording the catalog cannot vouch for is the right call, and the reason is the good one: a stage sub-DAG genuinely cannot tell a failed sync from a corpus nobody synced, and dropping it would convert a loud `FileNotFoundError` into a silent skip. Filtering to stop paying for finished work, not to swallow failures.

**One gap, fixed on your branch in `0895b56`.** The `all_stages` coercion was covered for `"true"` only. That is the safe direction: reading a true spelling as false costs one wasted run and you see it immediately. The other way round is silent. A stage that reads `"false"` as true just processes everything, which is what it did before this filter existed, so the corpus stays correct and the run keeps costing exactly what #172 was about, with nothing saying so. `bool(all_stages)` on the raw string is the obvious wrong implementation and it survived your suite; it now fails on five of the six spellings I pinned.

That is the same shape as the drift guard we fixed on #178 earlier this week, and it is worth carrying: when a flag's two directions have asymmetric failure modes, test the quiet one first.

Verified on a merge with current main: 1001 passed, ruff, format, ty, and lychee clean. Ordering is right, the bucket reasoning holds, and `app_variable` reaches the injected snippet the same way `process_batch` already takes it.

Issue #182, opened 2026-08-25
Replace the hand-maintained default key registry with checks that declare their own keys


kstonekuan:
### Where this came from

#178 landed the pre-execution short-circuit that #177 asked for: a default whose measurement
keys a pipeline step has already emitted is superseded before it runs, so a wrapper with
non-default parameters no longer pays for an ffmpeg decode that gets discarded.

It pays for that with `_DEFAULT_KEY_PATTERNS` (`src/hflow/checks.py:1529`), a private registry of
six functions that predict what each default would emit without running it. Each one mirrors the
`measurements[...] = ...` writes in the default's body, line for line. `App` reads it at
`src/hflow/app.py:1875`.

That registry is a second implementation of six functions, kept honest only by a drift guard
(`tests/test_default_checks.py:359`). It has already drifted once, when the video-measurement
refactor added `camera_measurement_definition` to `camera_frame_stats`. The guard did not catch
it, because it ran on a camera-less episode where three of the six mirrors emit nothing at all.
That specific hole is closed, but the shape that produced it is not: every future change to a
default's key set is one forgotten registry edit away from a silently wrong supersession
decision.

### What to build

Option 1 from #177: let a check declare the keys it emits, and decide supersession from
declarations rather than from a hand-maintained mirror. The registry and the drift guard both
go away, because there is only one statement of the fact left.

The hard part is that a check is an arbitrary Python function and its keys are usually not
static. `camera_frame_stats` derives them from the episode's camera list, `required_topics` from
its `topics=` argument. Any workable declaration has to express "one of these per camera" rather
than a literal list, and it has to stay optional so that a user check that declares nothing still
falls back to the post-execution comparison in
`App._yield_defaults_superseded_by_the_pipeline` (`src/hflow/app.py:987`).

### Design note first

Open the design note as a comment here before writing the implementation. It should settle:

- **How a declaration is spelled**, and whether it is a `@app.check()` parameter, an attribute on
  the function, or something else. This is public API surface if users can use it, so say
  explicitly whether they can.
- **How a per-topic key set is expressed** without running the check, given that the topic list
  comes from the episode.
- **What happens when a declaration is wrong.** The current design fails a build via the drift
  guard. If declarations replace the mirror, is there still anything that catches a declaration
  that disagrees with the body, or does the fallback path absorb it?
- **What the fallback is** for checks that declare nothing, and whether the post-execution path
  stays as a backstop or is deleted.
[...]

PR #197, opened 2026-08-27
refactor(checks): camera_frame_stats owns its key set (#182)

Sagar-024:
Part 1 of #182. This removes the hand-maintained `_DEFAULT_KEY_PATTERNS`
mirror for `camera_frame_stats` and replaces it with one function that owns
the key set.

Instead of a mirror that restates the body, `camera_frame_stats_keys` is now
the single place that states which keys the check emits. The body iterates
over the fact's output and routes each key through a small dispatcher,
`_camera_value`. A key the fact does not name cannot be emitted, and a key
with no branch raises where it happens instead of writing a null into a
`record=True` catalog row.

## What changed

- `camera_frame_stats_keys(episode, *, cameras=None)`: the single statement
  of the key set. Per topic it emits `message_count` always,
  `expected_frame_count` and `frame_deficit_pct` only when there are 2 or
  more stamps, and the seven decoded-evidence keys always. `camera_instrument`
  and `camera_measurement_definition` are part of the per-topic selection, so
  a zero-camera episode emits nothing, intervals included.

- `_CameraIntermediates` and `_camera_intermediates`: computes the
  expected-period precedence (declared `expected_hz` per topic, else median
  inter-message delta) once per topic, copied verbatim from the old body.

- `_camera_value`: the dispatcher. It uses `rpartition("/")` rather than
  `partition`, because camera topics are themselves paths like
  `/wrist_cam/compressed`, so only the last separator splits topic from
  measurement name. Unknown keys raise `ValueError`.

- The body now builds intermediates first, then runs one dict comprehension
  over `sorted(camera_frame_stats_keys(...))`. `_DEFAULT_KEY_PATTERNS`
  routes to the fact.

## What stays the same

Every measurement key, value, and the freeze-interval mapping are unchanged
for existing callers. The fact is called with default parameters on automatic
registration, exactly as before.

## Behavior notes

Edge cases the single-key-set shape makes explicit:

- `camera_instrument` provenance is read from the first selected camera topic
  via `next(iter(intermediates_by_topic.values()))`. The value is identical on
  every topic in a run because one pinned ffmpeg build measures the whole
  episode.

- Duplicate entries in a `cameras=` list behave exactly as before this
  refactor. Each listed occurrence is decoded and contributes its freeze
  intervals. The fact's measurement keys dedupe (they are a set), so
  measurement values are written once, but interval emission follows the
  pre-refactor per-occurrence behavior.

- All instrument decodes run before any value is computed. If a decode fails,
  the run raises before partial measurements exist, rather than leaving a
  half-populated result.

## Tests

Value-asserting fixtures over zero, one, and two cameras plus a two-stamp
topic. These assert the full measurement dict, values included, against
hand-written expectations.

The one-camera fixture puts a black run inside the 1-second window (the spec
default of `(2.0, 3.0)` misses it), so `black_frame_pct` is pinned to
`[30, 40]` and `overexposed_frame_pct` to exactly `0.0`. A black/overexposed
branch swap goes red (proved by mutation).

A unit-level guard feeds the dispatcher a stub with `decoded_frame_count=15`
against 13 stamps, so a message/decoded branch swap goes red too (proved by
mutation). The integration fixtures hold the two equal at 15 by construction
(one access unit per canonical message), which is exactly why the unit guard
exists.

Also covered: the poison test (withdrawing a key from the fact must withdraw
it from the body's output), the `expected_hz` precedence test, the existing
one-stamp error contract, and the unknown-key raise end to end.

## Out of scope

The other five defaults (`episode_duration`, `timestamp_regularity`,
`keyframe_interval`, `content_digest`, `media_digest`) still use their mirrors
and follow in their own PRs. Two user wrappers sharing one path-keyed cache is
#177 follow-up territory, unchanged here.

### DIFF
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -34,6 +34,7 @@ from hflow._video_measurements import (
     FrameStatisticsSettings,
     InsufficientVideoFrames,
     LumaRangeEvidence,
+    VideoFrameStatistics,
     measure_camera_motion,
 )
 from hflow.episode import Episode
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
Merging. Taking the hardest default first was the right pick, and every condition from the issue is met and verified rather than asserted.

**The refactor is output-identical.** I ran `camera_frame_stats` on main and on this branch over the same canonical episodes at zero, one, and two cameras, plus the duplicate-`cameras=` case, and diffed measurements, interval tuples, and tags. Byte-identical, including the per-occurrence interval behaviour you documented.

**My objection is answered.** The mismapped-field class was the whole reason I asked for value fixtures, so I mutated it three ways:

```
black_frame_percent  -> overexposed_frame_percent   2 failed
average_luma_minimum -> average_luma_maximum        3 failed
message_count        -> decoded_frame_count         1 failed
```

The third is the one worth calling out. Your integration fixtures hold those two equal by construction, so only the unit-level guard with a stub catches it. Noticing that a fixture cannot distinguish two fields and writing a separate test for exactly that is the part most people skip.

You also caught a bug in your own design note. The sketch used `partition("/")`, which would have parsed `/wrist_cam/compressed/luma_avg_max` as topic `""`. `rpartition` is right and the comment explains why.

One nit, not blocking: the two `assert inter.expected_frame_count is not None` statements are coupling assertions between the fact and the dispatcher, and `python -O` strips them. The failure then degrades to a `None` refused at append rather than a crash at the branch, which is survivable but is the quieter of the two. A raise would hold under `-O`.

1011 passed on a merge with current main, gates clean.

#182 stays open for the remaining five. The shape is settled now, so those should read as mechanical; if any of them fights the pattern, that is worth saying in the PR rather than forcing it.

PR #198, opened 2026-08-27
refactor(checks): the remaining five defaults own their key set (#182)

Sagar-024:
# refactor(checks): the remaining five defaults own their key set (#182)

Follows up on #197. The registry used to keep a mirror for each default's body, and we had to keep them in sync. Now every default owns its key set in one public fact function. The body just iterates that fact's output through a dispatcher. If a key isn't in the fact, it can't be emitted. If the dispatcher doesn't know a key, it raises immediately.

## What changed

- `episode_duration`, `timestamp_regularity`, `keyframe_interval`, `content_digest`, and `media_digest` now each have a public `<name>_keys(episode)` function for the key set and a `_value` dispatcher. The body iterates the sorted keys and routes them through the dispatcher.
- `keyframe_interval` got a shared payload-scan helper `_keyframe_indices(channel)` that both the fact and body use. The scan is the irreducible cost here, and the pre-decode supersession path only runs it when the default is going to run anyway.
- `timestamp_regularity_keys` now takes a `topics=` argument so the fact and body share the same selection rule. Previously, passing `topics=` to the check could give a broader key set than expected.
- Renamed `_DEFAULT_KEY_PATTERNS` to `_DEFAULT_KEY_FACTS` since it's just routing now. Every default points to the same statement the body iterates.
- Replaced the drift-guard test `test_every_default_in_the_pattern_registry_is_drift_free` with `test_every_default_iterates_its_fact_for_measurements`. This one just asserts the body emits a subset of the fact.

## What stays the same

- Measurement values are byte-for-byte identical. New tests pin exact digests and float counts, and the existing `tests/test_checks.py` suite passes without changes.
- Pre-decode supersession in `app.py:1988-1997` works exactly as before. The post-execution backstop `_yield_defaults_superseded_by_the_pipeline` at `app.py:1035` is unchanged. Same-parameter wrappers and user-check supersession paths are also unchanged.

## Tests

Each default has a full-dict value fixture, a poison test (withdrawing a key from the fact withdraws it from the body), and an unknown-key raise test. I also verified each default catches branch swaps via mutation (e.g., swapping `period_violation_pct` to return `np.median` fails the pin; reversing the `content_digest` hex fails the pin).

## Out of scope

- Caching the keyframe scan on the channel to avoid a second scan during pre-decode supersession. It runs once per invocation for now.
- A pre-decode supersession test for the user-check-overlapping-default case. It still falls back to the post-execution path.

## Validation

- `uv run pytest -q` passed 1033 tests, 6 skipped.
- `uv run ruff check src tests` is clean.
- `uv run ruff format --check src tests` is clean.
- `uv run ty check src tests` is clean.

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -1953,7 +1953,7 @@ class App:
             # (first episode, or no overlapping user step) falls through to
             # the regular path and runs as before.
             pipeline_emitted_keys: set[str] = set()
-            from hflow.checks import _DEFAULT_KEY_PATTERNS
+            from hflow.checks import _DEFAULT_KEY_FACTS
 
             for registered in checks_to_run:
                 run = CheckRunReport(check=registered)
@@ -1986,7 +1986,7 @@ class App:
                 # only fires when both the wrapper has run and the key sets
                 # actually overlap.
                 if registered.name in self._default_check_names and pipeline_emitted_keys:
-                    pattern = _DEFAULT_KEY_PATTERNS.get(registered.function)
+                    pattern = _DEFAULT_KEY_FACTS.get(registered.function)
                     if pattern is not None:
                         predicted = pattern(canonical_episode)
                         superseded_keys = sorted(predicted & pipeline_emitted_keys)
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -1932,28 +1932,6 @@ def _default_check_version_for_automatic_registration(check_function: CheckFunct
         ) from None
 
 
-# Key-set predictor for every default: what ``measurements`` keys would this
-# function emit for a given episode, without actually running it.
-#
-# Used by ``hflow.app.App`` to decide whether a pipeline step has already
-# covered a default's keys, so the default can be short-circuited before the
-# ffmpeg decode it would otherwise pay for. The patterns mirror the
-# ``measurements[...] = ...`` writes in the function bodies above, line for
-# line -- a separate, internal contract that the drift-guard test in
-# ``tests/test_default_checks.py`` enforces: a default whose actual key set
-# diverges from its pattern is a regression in this engine, not in the
-# default itself.
-#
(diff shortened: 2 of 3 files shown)


kstonekuan (approved):
Merging, and #182 closes with it. `_DEFAULT_KEY_PATTERNS` and its guard are gone from `src/` and `tests/` entirely.

**Output-identical.** I ran all five defaults on main and on this branch over the same canonical episodes at zero, one, and two cameras and diffed measurements, interval tuples, and tags. Identical. Taking `topics=` into `timestamp_regularity_keys` so fact and body share one selection rule is a real fix, not just plumbing: passing `topics=` used to give a broader predicted key set than the body would emit.

**One change, pushed as `09c0859`.** The new guard asserted the body emits a *subset* of its fact. That leaves the over-naming direction uncovered, and it is the direction with teeth. A fact naming a key the body never emits usually crashes, because the body iterates the fact and the dispatcher has no branch for it. The exception is the pre-decode gate: it reads the fact and skips the body, so an over-named key that a pipeline step happens to write supersedes a default that never collided, and its measurements are silently gone.

I checked before tightening it rather than assuming equality was achievable. It holds for all six defaults on every shape I could build: zero, one, two and three cameras, a sub-second episode, a 1 Hz camera, state-only, and a camera dropped for carrying no messages. So the loosening bought nothing. An over-naming fact now fails at every camera count.

1033 passed on a merge with current main, gates clean.

Two notes on your out-of-scope list, both agreed. The keyframe scan running twice when the gate fires is worth leaving until someone measures it; it is bounded and the decode it avoids is far larger. And the pre-decode path for a user check that overlaps a default genuinely has no mechanism to test, since user checks have no fact, so the post-execution backstop is the only answer and it already has coverage.

Good sequence of work. The design note, the hardest default first, then the mechanical five, with the shape settled before any of it was written.

Issue #200, opened 2026-08-27
Tests: the incubated video-measurement package has ten unexercised refusals


kstonekuan:
### What happens

`hflow/_video_measurements/` was incubated as its own package in `a478b95`, and its input
validation has no tests. Ten refusals, none of which any test exercises:

| Site | Refuses |
|---|---|
| `_toolchain.py:24` | an `ffmpeg` path that is not a file |
| `_toolchain.py:26` | an `ffprobe` path that is not a file |
| `_toolchain.py:28` | an empty `ffmpeg_version` |
| `_toolchain.py:30` | an empty `ffprobe_version` |
| `_frame_statistics.py:57` | a `start_seconds` that is negative or not finite |
| `_frame_statistics.py:59` | an `end_seconds` before `start_seconds`, or not finite |
| `_frame_statistics.py:77-92` | five out-of-range or non-finite `FrameStatisticsSettings` fields |
| `_camera_motion.py:54` | a `frames_per_second` that is not finite and positive |
| `_camera_motion.py:59` | a `horizontal_field_of_view_degrees` outside `(0, 360]` |
| `_raw_frames.py:62` | a `long_edge_pixels` below 2 |
| `_raw_frames.py:152` | a `frames_per_second` that is not positive |

`tests/test_ffmpeg.py:115` and `:225` do cover a missing binary, but for the *resolver*
(`FfmpegNotFoundError`), not for `VideoMeasurementToolchain.__post_init__`. Different code, same
words, so a grep for the message is misleading.

This is a package the transform and two built-in checks now depend on, and every one of these
refusals is a contract someone could break without a test noticing.

### Pattern to copy

`tests/test_catalog_curation.py:309` and `:935` are the shape: construct the thing invalidly and
assert `pytest.raises(ValueError, match=...)` against a fragment of the real message.

```python
with pytest.raises(ValueError, match="ffmpeg_version must not be empty"):
    VideoMeasurementToolchain(...)
```

Match on a distinctive fragment rather than the whole string, so rewording the message later does
not break the test while still pinning which refusal fired.

### Definition of done

1. One test per row above. Where a row covers several fields, one test per field.
2. Each test asserts on a message fragment that identifies which refusal fired, not just
   `pytest.raises(ValueError)`. A bare type assertion would pass if the wrong branch raised.
3. The valid case is covered too: a toolchain built from real paths and non-empty versions
   constructs without error, and `FrameStatisticsSettings()` with no arguments is valid.
[...]

PR #201, opened 2026-08-27
fix(video-measurements): refuse bool and wrong types in settings dataclasses

### LINKED ISSUE #199: Video-measurement settings validate ranges but not types, so bool passes every one
### What happens

The settings dataclasses in `hflow/_video_measurements/` validate their ranges but not their
types, and `bool` is a subclass of `int`, so a boolean satisfies every range test. Verified on
`02c973b`:

```python
FrameStatisticsSettings(black_pixel_luma_threshold=True)                  # accepted
FrameStatisticsSettings(black_frame_minimum_pixel_share_percent=True)     # accepted
FrameStatisticsSettings(freeze_minimum_duration_seconds=True)             # accepted
FrameStatisticsSettings(overexposed_average_luma_threshold=True)          # accepted
VideoTimeInterval(start_seconds=True, end_seconds=True)                   # accepted
CameraMotionSettings(frames_per_second=True)                              # accepted
```

An `int`-declared field also accepts a float, for the same reason nothing checks the type:

```python
FrameStatisticsSettings(black_frame_minimum_pixel_share_percent=98.5)     # accepted, field is int
```

### Why it is worth fixing rather than shrugging at

`black_pixel_luma_threshold` goes into the measuring filter graph as a literal
(`_frame_statistics.py:541`), so `True` reaches ffmpeg as text:

```
blackframe=amount=0:threshold=True
```

ffmpeg rejects that, so the failure is not silent, but it arrives as an ffmpeg parse error with
no mention of which parameter was wrong. It also mints a distinct cache digest
(`_video_measurement_toolchain.py:_frame_statistics_cache_path`), so the bad value gets its own
cache entry that can never hold a real measurement.

This is the same hole the project has now closed four separate times: the server's port check,
NumPy/`bool` measurements (#127, #133), interval bounds (#174), and non-finite floats (#176).
Each time it was found by probing a boundary with a value the annotations do not forbid.

### Pattern to copy

`hflow/catalog.py:398-410` is the canonical form, comment included:

```python
# bool subclasses int, so True satisfies the test below and would
# store as 1 nanosecond. Excluded explicitly, the same way the port
# and measurement boundaries do it.
if not isinstance(value, int) or isinstance(value, bool):
    raise ValueError(...)
```

For a float-declared field the equivalent is `isinstance(value, int | float) and not
isinstance(value, bool)`. Note only `np.float64` subclasses `float`, so do not assume a NumPy
scalar arrives as a Python one; `catalog.py:_normalized_measurements` shows how that is handled
there. Deciding whether these dataclasses should also coerce NumPy scalars is part of this
issue: either is defensible, but say which you chose and why.

### Sites

- `FrameStatisticsSettings.__post_init__` (`_video_measurements/_frame_statistics.py:76`), five
  fields: two declared `int`, three declared `float`.
- `VideoTimeInterval.__post_init__` (`_video_measurements/_frame_statistics.py:56`), two `float`
  fields.
- `CameraMotionSettings.__post_init__` (`_video_measurements/_camera_motion.py:53`), two `float`
  fields.

### Definition of done

1. Every field above refuses a `bool`, naming the field and the type it got.
2. Every `int`-declared field refuses a non-integer; every `float`-declared field accepts an
   `int` (an integer is a perfectly good float here) and refuses a `str` or `None`.
3. The existing range and finiteness refusals keep working unchanged.
4. A test per refusal, including the `bool` case for each field. `0` and `0.0` must still be
   accepted where they are in range, since they are real values and falsy.
5. No behaviour change for any valid value: `FrameStatisticsSettings()` with no arguments, and
   the values `hflow.checks.camera_frame_stats` passes, are unaffected.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Touches the same two files as the sibling coverage issue. If both are open at once, whoever
takes the second should rebase rather than resolve conflicts by hand.

afonsojanu:
## Summary

`FrameStatisticsSettings`, `VideoTimeInterval`, and `CameraMotionSettings` now refuse a `bool` (or any wrong type) for every numeric field, instead of silently accepting it because `bool` subclasses `int` and satisfies the range checks.

## Why

Fixes #199

Range checks alone (`0 <= x <= 100`) don't reject the wrong *type*. `FrameStatisticsSettings(black_pixel_luma_threshold=True)` was accepted, then reached ffmpeg's filter graph as the literal text `True` — ffmpeg rejects it, but with no mention of which parameter was wrong. A `str`/`None` raised a bare `TypeError` from the comparison itself instead of a message naming the field.

Added `require_int`/`require_float` in a small shared `_field_guards.py`, called at the top of every affected `__post_init__` before the existing range checks. int-declared fields refuse `bool` and any non-`int`; float-declared fields refuse `bool` and accept `int` (a perfectly good float value, and `0`/`0.0` must still pass since they're real values, just falsy) but refuse `str`/`None`.

On the NumPy-coercion question the issue raises: I deliberately did **not** add NumPy-scalar coercion here, unlike `catalog.py`'s measurement/interval normalization. These are caller-constructed configuration, not measurement-pipeline output — there's no established path for a NumPy scalar to reach these constructors, so a caller building a threshold from one can call `.item()` itself.

## Validation

```
uv run ruff check --fix   # all checks passed
uv run ruff format        # 2 files reformatted (the two touched)
uv run ty check           # all checks passed (two intentional wrong-type test calls marked `# ty: ignore`, matching the existing convention in tests/test_ffmpeg.py)
uv run pytest -q          # 1045 passed, 9 skipped, 1 failed
```

The one failure (`test_content_digest_emits_exactly_the_documented_measurements`, a pinned content-digest hash) is pre-existing and unrelated — confirmed by stashing this change and re-running just that test, same failure either way.

Added `tests/test_video_measurement_settings_types.py`: a bool-refusal test for all 9 affected fields (5 on `FrameStatisticsSettings`, 2 on `VideoTimeInterval`, 2 on `CameraMotionSettings`), int-field refuses float/str, float-field refuses str/None, float-field accepts a plain int, `0`/`0.0` still accepted, and defaults are unaffected. Confirmed all 16 new tests fail against `main` (13 didn't raise `ValueError` at all, the other 3 raised an uncaught `TypeError` instead) and pass with this fix.

Kept this scoped to the type guards only (issue #199) — the sibling coverage issue for the pre-existing range refusals (#200) is separate, per the issue's own note that whoever takes the second of the two should rebase onto the first.

## Checklist

- [ ] I updated documentation for changed behavior, flags, formats, or requirements. — not applicable, no documented behavior changed (invalid inputs already failed; now they fail with a clearer error).

### DIFF
--- a/src/hflow/_video_measurements/_camera_motion.py
+++ b/src/hflow/_video_measurements/_camera_motion.py
@@ -7,6 +7,7 @@ from types import ModuleType
 
 import numpy as np
 
+from ._field_guards import require_float
 from ._raw_frames import LUMA_FRAME_FILTER_GRAPH, luma_frames
 from ._toolchain import VideoMeasurementToolchain
 
@@ -51,6 +52,8 @@ class CameraMotionSettings:
     horizontal_field_of_view_degrees: float = DEFAULT_HORIZONTAL_FIELD_OF_VIEW_DEGREES
 
     def __post_init__(self) -> None:
+        require_float(self.frames_per_second, "frames_per_second")
+        require_float(self.horizontal_field_of_view_degrees, "horizontal_field_of_view_degrees")
         if not math.isfinite(self.frames_per_second) or self.frames_per_second <= 0:
             raise ValueError(
                 f"frames_per_second must be finite and positive, got {self.frames_per_second}"
--- a/src/hflow/_video_measurements/_field_guards.py
+++ b/src/hflow/_video_measurements/_field_guards.py
@@ -1,0 +1,30 @@
+"""Shared numeric-field validation for the video-measurement settings dataclasses.
+
+Range checks alone (``0 <= x <= 100``) do not reject the wrong *type*: ``bool``
+subclasses ``int``, so ``True``/``False`` satisfy every numeric comparison and
+range test the settings below already run, and a ``str``/``None`` would raise
+a bare ``TypeError`` from the comparison itself instead of a clear message
+naming the field. This closes that hole the same way ``catalog.py`` already
+does for interval bounds and measurement values.
+
+These settings are caller-constructed configuration, not measurement-pipeline
+output, so unlike ``catalog.py``'s NumPy-scalar coercion, no NumPy handling is
+added here: a caller building a ``np.float64`` threshold can call ``.item()``
+itself, the same way any other non-native-Python value would need to.
+"""
+
+
+def require_int(value: object, name: str) -> None:
+    """Refuse anything but a plain ``int`` - ``bool`` included."""
+    if isinstance(value, bool) or not isinstance(value, int):
+        raise ValueError(f"{name} must be an int, got {type(value).__name__}")
+
+
+def require_float(value: object, name: str) -> None:
+    """Refuse anything but a plain ``int`` or ``float`` - ``bool`` included.
+
+    An ``int`` is accepted for a float-declared field: it is a perfectly good
+    float value (``0`` is a real, falsy value that must still pass).
+    """
+    if isinstance(value, bool) or not isinstance(value, int | float):
+        raise ValueError(f"{name} must be an int or float, got {type(value).__name__}")
--- a/src/hflow/_video_measurements/_field_guards.py
+++ b/src/hflow/_video_measurements/_field_guards.py
@@ -15,13 +15,13 @@ itself, the same way any other non-native-Python value would need to.
 
 
 def require_int(value: object, name: str) -> None:
-    """Refuse anything but a plain ``int`` - ``bool`` included."""
+    """Refuse anything but a plain ``int``, ``bool`` included."""
     if isinstance(value, bool) or not isinstance(value, int):
         raise ValueError(f"{name} must be an int, got {type(value).__name__}")
 
 
 def require_float(value: object, name: str) -> None:
-    """Refuse anything but a plain ``int`` or ``float`` - ``bool`` included.
+    """Refuse anything but a plain ``int`` or ``float``, ``bool`` included.
 
     An ``int`` is accepted for a float-declared field: it is a perfectly good
     float value (``0`` is a real, falsy value that must still pass).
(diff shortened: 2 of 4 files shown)


kstonekuan (approved):
Welcome, and merging. Validated locally since first-time CI is held behind an approval gate: 1049 passed, ruff, format and ty clean on a merge with current main.

I probed all three dataclasses field by field against `bool`, `str` and `None` rather than counting your tests, and every one of the nine refuses with the field named. Valid values still pass, including the ones easy to break: `0` and `0.0` where they are in range, a plain `int` for a float-declared field, and `VideoTimeInterval(0.0, 0.0)`. All six pre-existing range and finiteness refusals still fire.

Your answer on the NumPy question is the right one, and the reason you give is the reason: these are caller-constructed configuration, not measurement-pipeline output, so there is no path for a NumPy scalar to arrive the way there is in `catalog.py`. Putting that reasoning in the module docstring is better than putting it in the PR body, where it would be lost.

Pushed one two-character change as `bef2067`: two docstrings used `` ``int`` - ``bool`` included``, and the repo avoids dashes in prose, so those are commas now.

**On the failing test.** I cannot reproduce it. `test_content_digest_emits_exactly_the_documented_measurements` passes on current main here, and the full suite is 1033 passed, 0 failed. Your run also shows 9 skipped against 6 here, so something differs in the environment.

That test pins a literal digest for a joints-only episode, no video and no ffmpeg involved, so if it genuinely fails for you then the synthetic writer is not as deterministic across platforms as its comment claims. That would be worth its own issue. What platform and Python version are you on? If it turns out to be platform-dependent, a contributor seeing a red test they did not cause is a bad first experience and I would rather fix it than have the next person work around it.

Nothing blocking either way, since it is unrelated to your change.

#200 is the sibling: the same package's ten pre-existing refusals have no tests at all. It is a separate issue and yours to take if you want it.

afonsojanu:
Thanks for the thorough review and the merge!

Platform/Python for the record: macOS-26.0.1-arm64-arm-64bit, Python 3.11.4 (Clang 14.0.3), Apple Silicon.

Happy to dig into the digest mismatch a bit if it'd help narrow it down — didn't want to go poking at something unrelated to #199 without asking first. If you'd rather I open a separate issue with the two digest values and repro steps so it's tracked properly, just say the word.

Issue #10, opened 2026-08-19
Docs: document the doctor finding codes; FORMAT.md claims a B-frame check that does not exist


kstonekuan:
`docs/FORMAT.md` overstates one doctor check, and the doctor's stable finding
codes have no reference.

### The documentation claims a B-frame check that does not exist

The video conformance item at `docs/FORMAT.md:142` includes `no B-frames`, and
`docs/FORMAT.md:145-147` says `hflow doctor` executes the listed checks.

`src/hflow/doctor.py:94-136` checks the H.264 format, access-unit delimiter
(AUD) framing, one access unit per message, Sequence Parameter Set (SPS) and
Picture Parameter Set (PPS) presence on keyframes, and keyframe-first ordering.
It does not detect B-frames.

Keep no B-frames as a canonical-format rule, but state that the doctor does not
yet verify it.

### The finding codes are undocumented

`Finding.code` is a stable kebab-case identifier at `src/hflow/doctor.py:47`.
The doctor currently emits 16 codes between `src/hflow/doctor.py:101-274`:

- `video-format`
- `video-not-aud-delimited`
- `video-multiple-access-units`
- `video-keyframe-missing-parameter-sets`
- `video-stream-starts-mid-gop`
- `unreadable`
- `no-summary`
- `no-statistics`
- `no-chunk-indexes`
- `chunk-missing-message-indexes`
- `chunk-mixes-video-and-state`
- `missing-provenance`
- `provenance-missing-key`
- `missing-episode-record`
- `topic-time-order`
- `read-failed`

Add a reference table to `docs/FORMAT.md`, or to a linked reference page, with
each code's level and one-line meaning.

## Pattern to copy

Use the `collector.add(...)` calls in `src/hflow/doctor.py:101-274` as the
authoritative source for code, level, and meaning. `chunk-mixes-video-and-state`
and `missing-episode-record` are warnings; the other current codes are errors.

## Definition of done

- Every code emitted by `doctor.py` appears in the documentation with its
  level and meaning.
- `docs/FORMAT.md` no longer implies that the doctor detects B-frames.

## Validation

```bash
lychee --no-progress --include-fragments \
  --exclude '^https://github\.com/Hebbian-Robotics/hflow/(issues|security/advisories/new)$' \
  --exclude-path references/mcap-spec.md \
  --exclude-path references/foxglove-CompressedVideo.proto .
```

kstonekuan:
[...]

PR #202, opened 2026-08-27
docs: document input MCAP converter contract

### LINKED ISSUE #166: Docs: no page tells a converter author what an input MCAP must satisfy
### What we want

A short page telling someone writing a converter what their **input** MCAP must satisfy, as
distinct from what HFlow's **canonical output** satisfies.

### Why this gap exists

Converters are the input story. `ARCHITECTURE.md` lists "MCAP input only" and the status table
calls more readers deferred, but the design is that a third-party tool converts a vendor format to
MCAP and hands it to `hflow ingest`. That makes the input contract a public interface, and it is
the one interface with no page.

What we have instead is `docs/FORMAT.md`, whose "Conformance" section (`docs/FORMAT.md:135`) lists
five rules a file "claiming this convention" must satisfy, including:

> 4. **Video constraints**: every `foxglove.CompressedVideo` message is one Annex B access unit
>    beginning with an AUD; every IDR access unit contains SPS and PPS; no B-frames; `format="h264"`.
> 5. **Stamps**: `provenance/v1` present with `schema_version` and `pipeline_version`.

Those describe what HFlow **produces**. A converter author reading that page reasonably concludes
they must produce it too, which is wrong and expensive: `provenance/v1` is stamped by the
transform stage, and the video and chunking constraints are what the transform stage exists to
establish. Someone can spend a day implementing guarantees the pipeline was going to give them.

The reverse mistake is worse. Nothing states what the transform stage genuinely **requires** of an
input, so a converter can produce something that fails deep inside a pipeline run with a
`ValueError` from `transform.py` rather than at the door.

### Definition of done

A short page (`docs/how-to/write-a-converter.md` or similar, your call) that answers:

1. **What must be true of an input MCAP** for `hflow ingest` to accept it. Derive this from the
   code, not from `FORMAT.md`: `src/hflow/transform.py` raises a `ValueError` for each thing it
   cannot handle (unsupported compressed-image format at `:193`, mixed image formats on one topic
   at `:287`, and others nearby). Those raises are the real contract. Read them all and write down
   what they collectively require.
2. **What HFlow adds, so a converter should not bother.** Provenance stamps, topic-group chunking,
   GOP structure, and the canonical identifiers all come from the transform stage.
3. **What is merely advisable.** Topic naming, units, anything that makes the built-in checks
   useful rather than merely make ingest succeed.
4. A pointer to `examples/lerobot/prepare.py`, which is a real converter, and a sentence on which
   parts of it are the pattern versus which are specific to that dataset.

`FORMAT.md` should gain one line near the Conformance list saying it describes the canonical
output and pointing at the new page for inputs. That cross-reference is half the value of this
issue: without it the same wrong turn stays one click away.

### Get this right

Every requirement on the page must come from a line of code you can point at, and the PR should
cite them. A converter contract that is aspirational rather than actual is worse than no page,
because someone will build against it.

If you find that the input contract is genuinely "anything valid MCAP that the transform stage's
raises do not reject", say exactly that. It is a legitimate and useful answer, and it is much more
useful stated than left for each author to discover.

### Validation

Markdown only. Run the lychee link command from [CONTRIBUTING.md](../CONTRIBUTING.md) from the
repo root, and `uv run pytest -q` if you touch anything outside prose. If you add a new file under
`docs/how-to/`, check whether an index or the README links that directory and add it there too.

tunglambk:
## Summary

Add a converter guide that separates the source MCAP accepted by ingest from the canonical MCAP that HFlow writes. The guide covers the required input, transform-owned output work, useful source conventions, ingest-failure diagnostics, and the reusable boundary in the LeRobot example.

Link the guide from the documentation index and from the canonical format's Conformance section so converter authors do not implement output guarantees before ingest.

## Why

Fixes #166.

The documented requirements follow the current implementation:

- `src/hflow/reader.py:119-150` requires a readable MCAP with a summary and permits schema-less channels.
- `src/hflow/transform.py:210-224` and `:336-458` define the supported compressed-image formats, camera decoder/encoder paths, and pass-through video constraints.
- `src/hflow/transform.py:495-514` rejects raw-image schemas and derived-topic collisions, while `:595-635` defines empty, single-frame, and multi-frame compressed-image behavior.
- `src/hflow/transform.py:670-835` shows global ordering, data-derived grouping and chunk sizing, source record handling, and canonical provenance.
- `src/hflow/ingest_ledger.py:80-100` explains why converter-facing transform errors currently appear under the fallback `infrastructure` failure kind.
- `src/hflow/checks.py:1885-1922` is the source for the six automatic checks described in the conventions section.

## Validation

- `lychee --no-progress --include-fragments --exclude '^https://github\.com/Hebbian-Robotics/hflow/(issues|security/advisories/new)$' --exclude-path references/mcap-spec.md --exclude-path references/foxglove-CompressedVideo.proto .` — 327 total, 324 OK, 3 excluded, 0 errors.
- `git diff --check`

## Checklist

- [x] Outcome-focused tests are not applicable because this is a Markdown-only change.
- [x] I added documentation for the converter input contract.
- [x] Python formatting and type checks are not applicable because no Python changed.
- [x] I ran the repository's required Markdown link check.
- [x] Stored-data behavior and compatibility are unchanged.

### DIFF
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -143,6 +143,8 @@ Stored-data identifiers (metadata record names, their keys, group names) are **n
 
 ## Conformance
 
+These rules describe the canonical output written by HFlow. A converter should follow the separate [input MCAP contract](./how-to/write-a-converter.md) instead of implementing these guarantees before ingest.
+
 A file claiming this convention must satisfy, in increasing strictness:
 
 1. **It is valid MCAP**: readable by the stock [`mcap` package](https://mcap.dev/docs/python/) with CRC validation on, with a complete summary section.
--- a/docs/README.md
+++ b/docs/README.md
@@ -46,6 +46,7 @@ the whole workflow before adapting it.
 Use these when you already know the outcome you need.
 
 - [Enable the built-in quality checks](./how-to/enable-built-in-checks.md)
+- [Write an input MCAP converter](./how-to/write-a-converter.md)
 - [Port existing processing and quality-check code](./PORTING.md)
 - [Serve a workspace over HTTP](./SERVE.md)
 - [Call an OpenAI vision endpoint from a step](./how-to/call-openai-vision.md)
--- a/docs/how-to/write-a-converter.md
+++ b/docs/how-to/write-a-converter.md
@@ -1,0 +1,42 @@
+# Write an input MCAP converter
+
+A converter only needs to produce a source MCAP that the built-in transform can read. It does not need to produce a [canonical episode](../FORMAT.md): HFlow establishes that output contract during ingest.
+
+## Required input contract
+
+The source must be a valid MCAP file with a summary section. HFlow opens it with the stock Python `mcap` reader and reads channels from the summary; an unindexed or truncated file is rejected. Finishing the file with a conforming MCAP writer normally supplies the required summary. See [`PythonMcapEpisodeReader`](../../src/hflow/reader.py).
+
+Non-camera channels have no HFlow-specific payload contract. Their topic, schema when present, message encoding, payload bytes, log time, and publish time pass through unchanged. A schema-less channel is allowed, as are multiple channels sharing one topic. The transform operates on channel IDs rather than assuming that a topic identifies one channel. See [`write_canonical_episode`](../../src/hflow/transform.py).
+
+Camera handling is selected by exact schema name. A custom schema name that happens to contain images is treated as an ordinary pass-through channel, so it will not appear in `Episode.cameras` and camera checks will not inspect it. The accepted camera inputs are:
+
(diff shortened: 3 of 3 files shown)


kstonekuan (approved):
Welcome, and merging. Lychee clean at 327 links, and the FORMAT.md pointer is placed well: a converter author who lands on the canonical output rules first is exactly the reader this page exists for.

I checked the claims most likely to be stale or subtly wrong, and all of them hold on current main:

The AUD paragraph is current, including its condition. #184 landed yesterday and changed a refusal into a lossless repair, and you documented both that and the fact that the repair needs the message to be re-serializable by its protobuf or ROS 2 CDR encoder. A page written from a week-old mental model would have said it was refused.

The failure-kind warning is exact. `classify_ingest_failure` returns `source-unreadable` only for `McapError` and `source-missing` only for `SourceNotFound`, so a transform `ValueError` about non-conforming video really does land in the `infrastructure` fallback. Telling a converter author to read `error_type` and `message` instead of `failure_kind` is the correct advice and I would not have thought to include it.

The B-frame note is honest about a real gap: nothing in the validator inspects for B-frames, so a converter must disable them without HFlow enforcing it. That is tracked separately in #10.

The camera-specific split is right too: `camera_frame_stats`, `keyframe_interval` and `media_digest` are the three that need the recognised schema names, and the other three read any populated topic.

One thing your PR surfaced that is worth its own issue: a transform `ValueError` about the source file's content is a fact about the recording, not about the machine, so classifying it as `infrastructure` is arguably wrong at the source rather than something the docs should route around. I am filing that. Your page should stay as it is either way, since it describes what happens today.

Thanks for reading the code instead of the existing prose. Every line reference in your PR body pointed where you said it did.

PR #207, opened 2026-08-27
fix: validate batch planning inputs

### LINKED ISSUE #204: plan_batches accepts a bool count, a bool file size, and a negative or NaN stagger
### What happens

`plan_batches` (`src/hflow/batching.py:40`) checks that `batch_count` and `target_batch_bytes`
are at least 1, and that no item size is negative. It does not check types, and it does not check
`stagger_interval_s` at all. Verified on current main:

```python
plan_batches(sizes, batch_count=True)                      # accepted, one batch
plan_batches(sizes, target_batch_bytes=True)               # accepted, one batch PER ITEM
plan_batches({"a.mcap": True}, batch_count=1)              # accepted, a bool as a file size
plan_batches(sizes, batch_count=1, stagger_interval_s=-1.0)   # accepted, start_delay_s = -0.0
plan_batches(sizes, batch_count=1, stagger_interval_s=float("nan"))  # accepted, start_delay_s = nan
```

`bool` is a subclass of `int`, so `True` satisfies every `>= 1` test.

### Why each one is worth refusing

**`target_batch_bytes=True`** means a 1-byte target, so every item gets its own batch. On a
corpus of ten thousand episodes that is ten thousand batches, each a mapped Airflow task. The
call succeeds and the plan looks deliberate.

**`stagger_interval_s` negative or NaN** fails later, somewhere worse. `start_delay_s` is carried
into the rendered sub-DAG and consumed by `time.sleep` (`src/hflow/runtime/_templates.py:565`):

```
time.sleep(-1.0)  -> ValueError: sleep length must be non-negative
time.sleep(nan)   -> ValueError: Invalid value NaN (not a number)
```

So a bad stagger passes planning, ships into the bundle, and raises inside a mapped task on a
worker. The traceback names `time.sleep`, not the parameter that was wrong.

**A `bool` item size** is a file size that is neither a file nor a size, and it packs as 1 byte.

### Pattern to copy

`src/hflow/_video_measurements/_field_guards.py` was added in #201 for exactly this hole, and
`src/hflow/catalog.py:398-410` is the original, comment included:

```python
# bool subclasses int, so True satisfies the test below and would
# store as 1 nanosecond. Excluded explicitly, the same way the port
# and measurement boundaries do it.
if not isinstance(value, int) or isinstance(value, bool):
    raise ValueError(...)
```

Do not import `_field_guards` from `batching.py`. It lives under `_video_measurements/` and is
that package's own helper; reaching across would couple two unrelated modules. Copy the shape.

### Definition of done

1. `batch_count` and `target_batch_bytes` refuse a `bool` and refuse a non-`int`, naming the
   parameter and the type received. The existing `>= 1` refusals are unchanged.
2. Item sizes refuse a `bool` and refuse a non-`int`, naming the offending uri the way the
   existing negative-size refusal does.
3. `stagger_interval_s` refuses a `bool`, a non-number, a negative value, and a non-finite value,
   naming the parameter. `0.0` stays valid and is the default.
4. A test per refusal. `0` stays accepted as an item size, since a zero-byte file is a real
   thing, and `stagger_interval_s=0` stays accepted.
5. `plan_batches_from_files` (`batching.py:109`) forwards to the same checks, so its callers get
   the same messages. Add a test that it does.
6. Nothing valid changes: existing batching tests pass untouched.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Before you finish, break one of the new refusals and confirm its test goes red. A validation test
that cannot fail is worse than none, because it buys the confidence without the coverage.

anisayakmitra-in:
## Summary

Validate all `plan_batches` inputs at the planning boundary so malformed values fail with parameter-specific errors.

## Why

Issue #204 showed that booleans pass integer checks, item sizes accept non-integers, and negative or non-finite stagger values reach runtime. This adds focused validation while preserving valid zero-byte items and zero stagger values.

## Validation

- `uv run ruff check` — passed via native Ruff 0.16.2 on the changed files.
- `uv run ruff format --check` — passed after formatting the changed files.
- Direct module checks under WSL Python 3.14 — passed for bool/non-int parameters, bool/non-int item sizes, negative/non-finite/non-number stagger values, and valid zero-byte/zero-stagger input.
- Full pytest suite was not runnable in this Windows checkout because HFlow requires POSIX `fcntl`; the upstream Ubuntu CI will run `uv sync --locked` and `uv run pytest -q`.

## Checklist

- [x] I ran `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check` (Ruff check/format completed; native full typecheck is blocked by the same platform dependency).
- [ ] I ran the relevant pytest suite (blocked by native Windows `fcntl`; covered by CI).

Closes #204

### DIFF
--- a/src/hflow/batching.py
+++ b/src/hflow/batching.py
@@ -23,8 +23,10 @@ least as well as random jitter and reproduces exactly.
 """
 
 import heapq
+import math
 from collections.abc import Iterable, Mapping
 from dataclasses import dataclass
+from numbers import Real
 from pathlib import Path
 
 
@@ -53,8 +55,36 @@ def plan_batches(
     if (batch_count is None) == (target_batch_bytes is None):
         raise ValueError("pass exactly one of batch_count or target_batch_bytes")
     for uri, size_bytes in item_sizes.items():
+        if not isinstance(size_bytes, int) or isinstance(size_bytes, bool):
+            raise ValueError(
+                f"item {uri!r} has type {type(size_bytes).__name__}, expected int bytes"
+            )
         if size_bytes < 0:
             raise ValueError(f"item {uri!r} has negative size {size_bytes}")
+
+    if not isinstance(stagger_interval_s, Real) or isinstance(stagger_interval_s, bool):
+        raise ValueError(
+            f"stagger_interval_s must be a number, got {type(stagger_interval_s).__name__}"
+        )
+    if not math.isfinite(stagger_interval_s):
+        raise ValueError(f"stagger_interval_s must be finite, got {stagger_interval_s}")
+    if stagger_interval_s < 0:
+        raise ValueError(f"stagger_interval_s must be nonnegative, got {stagger_interval_s}")
+
+    if batch_count is not None:
+        if not isinstance(batch_count, int) or isinstance(batch_count, bool):
+            raise ValueError(f"batch_count must be an int, got {type(batch_count).__name__}")
+        if batch_count < 1:
+            raise ValueError(f"batch_count must be >= 1, got {batch_count}")
+    else:
+        assert target_batch_bytes is not None
+        if not isinstance(target_batch_bytes, int) or isinstance(target_batch_bytes, bool):
+            raise ValueError(
+                f"target_batch_bytes must be an int, got {type(target_batch_bytes).__name__}"
+            )
+        if target_batch_bytes < 1:
+            raise ValueError(f"target_batch_bytes must be >= 1, got {target_batch_bytes}")
+
     if not item_sizes:
         return []
 
@@ -63,8 +93,6 @@ def plan_batches(
 
     batches: list[tuple[list[str], int]]
     if batch_count is not None:
-        if batch_count < 1:
-            raise ValueError(f"batch_count must be >= 1, got {batch_count}")
         # Least-loaded assignment: push each item onto the currently-lightest
         # batch. Heap entries are (total_bytes, batch_index) so byte ties
         # break on index, deterministically.
@@ -78,8 +106,6 @@ def plan_batches(
             heapq.heappush(heap, (total + size_bytes, index))
     else:
         assert target_batch_bytes is not None
-        if target_batch_bytes < 1:
-            raise ValueError(f"target_batch_bytes must be >= 1, got {target_batch_bytes}")
         # First-fit-decreasing: first open batch with room, else a new one.
         # An item exceeding the capacity still gets (its own) batch.
         batches = []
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
Welcome, and merging. Validated here since you could not: 1062 passed, ruff, format and ty clean.

All five items in #204 are covered, including the one easiest to skip: `plan_batches_from_files` forwarding, with a test for it. `numbers.Real` for the stagger is a better choice than `int | float`, since it accepts a NumPy float without a special case.

I broke the negative-stagger refusal and its test went red, so the case with the real consequence is genuinely pinned. That was the one worth checking: a negative stagger used to pass planning and fail inside `time.sleep` on a worker.

**One behaviour change you did not mention.** Moving the range checks above the `if not item_sizes: return []` early return means an empty mapping with an invalid count now raises where it returned `[]`:

```
main:   plan_batches({}, batch_count=0) -> []
branch: plan_batches({}, batch_count=0) -> ValueError
```

I think your order is the right one, since invalid input should be refused whether or not there is work to do, and the runtime never reaches it (the plan task returns early on empty uris). Worth stating in the PR body next time. A reviewer who spots it themselves has to work out whether you meant it.

Pushed `35a5aed`, removing one line: the `assert target_batch_bytes is not None` you added duplicates one already at the top of the packing branch, and `isinstance` narrows the type for the range check anyway. `ty` passes without it.

**On the Windows problem.** You are the third contributor this week to hit a platform wall, so it is ours and not yours. HFlow imports `fcntl` unconditionally somewhere in the path a test collection touches, which makes the suite unrunnable on native Windows. I am filing that separately. Until it is fixed, WSL is the workaround, and skipping the suite while saying so plainly, as you did, is the right way to handle it. Do not let it stop you sending a PR.

kstonekuan:
Correcting myself on the Windows point: there is nothing to file, and I should have checked before saying there was.

`CONTRIBUTING.md:30` already documents it, and accurately:

> Native Windows does not work: HFlow imports `fcntl` for file locking (`src/hflow/storage.py`), and that module does not exist on Windows, so `import hflow` fails before any test can run. Work inside WSL2 with an Ubuntu distribution.

So it is deliberate rather than an oversight. The lock it needs is an advisory `flock` around the etag-and-bytes pairing in the object-store mirror, and POSIX is the documented requirement. `README.md:223` says the same at a higher level.

Your instinct to say plainly which checks you could not run was still exactly right, and it is what let me verify the rest quickly. The one thing worth doing differently is reading the Platform support section first: it would have saved you the attempt.

PR #208, opened 2026-08-27
test(video-measurements): cover the range and value refusals

afonsojanu:
## Summary

Adds one test per unexercised range/value refusal in the incubated `hflow._video_measurements` package: `VideoMeasurementToolchain` (missing/invalid executable paths, empty version strings), `VideoTimeInterval` (negative/non-finite `start_seconds`, non-finite or before-start `end_seconds`), `FrameStatisticsSettings` (all five range/finiteness checks), and `CameraMotionSettings` (frame rate, field of view).

## Why

Addresses #200: these `__post_init__` refusals had no tests, so a regression would only surface as a confusing FFmpeg or filter-graph error far from the constructor that let the bad value through.

Two of the issue's ten rows - `_raw_frames.py`'s `long_edge_pixels < 2` and `rgb_frames`'s non-positive `frames_per_second` - already have coverage in `test_ffmpeg.py` (`test_scaled_frame_shape_refuses_a_degenerate_long_edge`, `test_rgb_frames_reject_a_non_positive_rate`), so they aren't duplicated here.

No production code changes - as requested in the issue, I left the checks themselves alone even where a stricter or looser bound might arguably be better, since that's a separate discussion.

## Validation

```
uv run ruff check --fix   # All checks passed!
uv run ruff format        # 174 files left unchanged
uv run ty check           # All checks passed!
uv run pytest -q          # 1098 passed, 9 skipped, 1 failed
```

The one failure (`test_content_digest_emits_exactly_the_documented_measurements`) is pre-existing and unrelated - the same platform-dependent digest mismatch @kstonekuan already flagged as not reproducible on their machine in #201's review thread. Confirmed it fails identically with my new test file removed.

Also did the issue's own self-check: temporarily disabled the `horizontal_field_of_view_degrees` range check and confirmed its 5 parametrized tests went red, then reverted.

## Checklist

- [ ] I updated documentation for changed behavior, flags, formats, or requirements. (test-only change, no behavior changed)
- [x] I preserved stored-data compatibility or documented an explicit version change. (n/a, test-only)

### DIFF
--- a/tests/test_video_measurement_settings_ranges.py
+++ b/tests/test_video_measurement_settings_ranges.py
@@ -1,0 +1,247 @@
+"""Range and value refusals on the video-measurement dataclasses (#200).
+
+Every constructor in ``hflow._video_measurements`` refuses an out-of-range or
+non-finite value, but the refusals themselves were never exercised - a
+regression here would only surface as a confusing FFmpeg or filter-graph
+error far from the constructor that let a bad value through. These tests are
+dependency-free (no ffmpeg, no video fixtures, no ``cv2`` extra) except for
+the ``VideoMeasurementToolchain`` executable-path checks, which only need a
+``tmp_path`` file to stand in for a binary.
+
+``_raw_frames.py``'s two refusals (``long_edge_pixels`` below 2,
+``frames_per_second`` not positive in ``rgb_frames``) already have coverage
+in ``test_ffmpeg.py`` (``test_scaled_frame_shape_refuses_a_degenerate_long_edge``,
+``test_rgb_frames_reject_a_non_positive_rate``) and are not duplicated here.
+"""
+
+from pathlib import Path
+
+import pytest
+
+from hflow._video_measurements._camera_motion import CameraMotionSettings
+from hflow._video_measurements._frame_statistics import (
+    FrameStatisticsSettings,
+    VideoTimeInterval,
+)
+from hflow._video_measurements._toolchain import VideoMeasurementToolchain
+
+
+@pytest.fixture
+def real_binaries(tmp_path: Path) -> tuple[Path, Path]:
+    ffmpeg_executable = tmp_path / "ffmpeg"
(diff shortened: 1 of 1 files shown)


kstonekuan (approved):
Merging. 1102 passed on a merge with current main, gates clean. I neutered the empty-`ffmpeg_version` refusal and its two tests went red, so they hold.

**You corrected my issue and you are right.** #200 claimed ten unexercised refusals; two of them, `long_edge_pixels < 2` and the non-positive `frames_per_second` in `rgb_frames`, are already covered by `test_scaled_frame_shape_refuses_a_degenerate_long_edge` and `test_rgb_frames_reject_a_non_positive_rate`. My scan matched on message text and those two tests match on other fragments, so it counted them as uncovered. Checking rather than copying the list was the right instinct.

Doing the issue's self-check unprompted, disabling the field-of-view range check and confirming its five tests went red, is the part I most wanted and least expected.

**Your digest failure is real, and it is our bug.** I found the cause. `synthesize_episode` generates joint positions with `math.sin` and `math.cos` (`src/hflow/testing.py:584-586`), and libm is not bit-identical across platforms. Different last-bit results give different payload bytes, which give a different `content_digest`. So `test_content_digest_emits_exactly_the_documented_measurements` (`tests/test_default_checks.py:1239`) pins a literal that is only valid on the platform that produced it.

It passes on Linux and on CI, which is why I could not reproduce it, and it fails anywhere else. `CONTRIBUTING.md` says macOS is supported for native development, so this is a supported platform failing a test nobody broke. Filing it now.

Two apologies owed on that: I said in #201's review that if it turned out to be platform-dependent I would rather fix it than have the next person work around it, and then you were the next person and worked around it twice. And I asked you for your platform without doing the ten minutes of digging that would have answered it myself.

kstonekuan:
Filed as #209, with the cause and your two reports as the evidence. If you can post the digest you actually get, that confirms the diagnosis directly instead of by inference from the source.

PR #210, opened 2026-08-27
docs(benchmarks): measure chunk floor on manipulation data

### LINKED ISSUE #170: The derived chunk-target floor of 800 KB has no measurement behind it
### What happens

`hflow.format.derived_chunk_size_bytes` clamps each group's target to
`[MINIMUM_DERIVED_CHUNK_SIZE_BYTES, MAXIMUM_DERIVED_CHUNK_SIZE_BYTES]` = `[800 KB, 8 MB]`.
The ceiling has a measurement behind it. The floor does not: 800 KB is where the benchmark
happens to start, inherited from the old flat default, not a measured optimum.

`docs/BENCHMARKS.md` now says this outright, and the table it says it under is the first
evidence the floor is too low:

| chunk size | fetches x MB |
|---|---|
| 800 KB flat | 61.0 |
| 8 MB flat | 41.3 |
| **derived per group** (the default) | **42.5** |
| 5.41 MB flat | 38.0 |

Every difference between the derived row and the flat 5.41 MB row is the state group. The
formula puts the camera group at ~5.4 MB and everything else on the floor; a flat target
lifts the state group off it. The training sample reads a state topic, so the derived layout
pays 3.79 fetches against 3.08. The full `/imu` scan shows it more sharply: **16 fetches
derived against 3** at flat 5.41 MB, for the same 4 MB.

### Why this is not just "tune the constant"

The reference recording's state group is 20 topics of nuScenes vehicle telemetry, which is
far busier than a robot's proprio channel -- the workload HFlow is actually for. So this
table cannot settle the floor on its own; raising it to fit nuScenes would be the same
mistake as the flat 800 KB default, one recording's number shipped as a principle.

What is needed is a second measurement, on a recording whose state group is proprio-rate.
If the floor turns out to be right there and wrong here, the answer may be that the floor
should not be a constant at all.

### Definition of done

- The read benchmark run against a manipulation-style recording (camera + proprio, state
  group in the kilobytes-per-second range, not megabytes), with `--chunk-size-bytes derived`
  and at least two flat targets for comparison.
- A stated conclusion about `MINIMUM_DERIVED_CHUNK_SIZE_BYTES`: keep it, move it, or replace
  it with something that is not a constant.
- `docs/BENCHMARKS.md`'s "the floor is the part with the least evidence behind it" paragraph
  updated to say what the second measurement showed.
- If the floor moves, `TRANSFORM_BEHAVIOR_VERSION` bumps with it -- the epoch note in
  `src/hflow/behavior.py` explains why.

`benchmarks/_real_input.py` documents how to fetch the pinned nuScenes input; a second
reference recording would need the same treatment (URL plus sha256, not vendored).

sylvesterkaczmarek:
## Summary

- Add a pinned real manipulation reference episode from `RobotisAI/evButtonPush-260615-0-MCAP`.
- Measure the derived chunk-size floor on a three-camera + compact state/proprio workload across 200 KB, 400 KB, 800 KB, 1.6 MB, and 3.2 MB flat targets.
- Cross-check a 400 KB candidate floor against both the manipulation episode and the existing nuScenes input on the same runner.
- Keep the shipped 800 KB floor because the smaller candidate improves one balanced metric on manipulation data but increases round trips there and regresses the nuScenes cross-check.

Closes #170.

## Why

The existing floor had only been exercised on the sensor-heavy nuScenes corpus. The added manipulation recording measures the camera + compact state/proprio workload the default is intended to serve.

The pinned episode is 75,946,038 bytes and 11.6167 s long. Its schema-default state group contains `/arm/tactile_broadcaster/gpio_states`, `/joint_states`, `/leader/joint_states`, and `/tf`, totalling 4,769,296 payload bytes, or 410.6 kB/s.

On the manipulation episode, a 400 KB floor lowers `fetches x MB` from 3.25 to 2.75, but increases training fetches from 2.88 to 3.77 and state-only scan fetches from 7 to 13.

On the same runner, lowering the floor also worsens nuScenes from 42.11 to 44.52 `fetches x MB` and raises the `/imu` scan from 16 to 21 fetches.

No fixed lower floor dominates across the two measured corpus shapes and access patterns, so this PR documents the evidence and keeps `MINIMUM_DERIVED_CHUNK_SIZE_BYTES` unchanged. There is no transform behavior change and no `TRANSFORM_BEHAVIOR_VERSION` bump.

## Validation

Reference manipulation input:

- Repository: `RobotisAI/evButtonPush-260615-0-MCAP`
- Episode: `107/107_0.mcap`
- Revision: `93a0ba73c6b82689696d9c4909b3b48af0294cf8`
- Download: 75,946,038 bytes
- SHA-256: `51bec431162844b2bb239f65ea158ea1030bb3fce5312ad81768978dc68f82c3`

Held constant:

- `GopPreset.VLA` with a 1 s window
- schema-default camera/state grouping for the manipulation corpus
- 200 seeded 1 s training windows
- the same state-only `/joint_states` scan

Varied:

- flat chunk targets `200000`, `400000`, `800000`, `1600000`, and `3200000`
- shipped `derived`
- then only `MINIMUM_DERIVED_CHUNK_SIZE_BYTES` between 800 KB and 400 KB for the same-run cross-check

Repository validation:

- `uv run ruff check` passed
- `uv run ruff format --check` passed
- `uv run ty check` passed
- `uv run pytest -q` passed on Python 3.11 and 3.14
- Markdown link check passed

## Checklist

- [x] No business logic changed, so no outcome-focused test update is required.
- [x] Benchmark documentation and reproducibility metadata are updated.
- [x] Lint, formatting, and typecheck pass.
- [x] The full pytest suite passes on Python 3.11 and 3.14.
- [x] No recordings, generated media, credentials, private URLs, or runtime artifacts were added.
- [x] Stored-data compatibility is unchanged; `TRANSFORM_BEHAVIOR_VERSION` remains 7.

### DIFF
--- a/benchmarks/_real_input.py
+++ b/benchmarks/_real_input.py
@@ -6,14 +6,18 @@ content". Both benchmark scripts accept ``--input <mcap>`` and use the
 helpers here to discover the camera and state topics of an arbitrary
 recording instead of assuming the synthetic fixture's layout.
 
-The reference real input (not redistributed -- fetch it yourself):
+The reference real inputs are not redistributed. Fetch the nuScenes sample:
 
-    curl -fLo nuscenes-mini-sample.mcap \\
+    curl -fLo nuscenes-mini-sample.mcap \
         https://mcap-proxy.lichtblick.workers.dev/NuScenes-v1.0-mini-scene-sample.mcap
 
 nuScenes v1.0-mini scene sample as converted by nuscenes2mcap and hosted by
 the Lichtblick project. License: CC BY-NC-SA 4.0 (non-commercial research).
 Attribution: "Adapted from nuScenes dataset. Copyright 2020 nuScenes."
+
+Issue #170 also uses one manipulation episode from RobotisAI's public MCAP
+corpus, pinned to an immutable repository revision. The URL and checksum are
+constants below so the exact camera + proprio recording can be reproduced.
 """
 
 from dataclasses import dataclass
@@ -29,6 +33,15 @@ NUSCENES_SAMPLE_URL = (
 )
 NUSCENES_SAMPLE_SHA256 = "089b5be708d2536a8c7e341f4d2b62dc618a581525e4db8d15c93ec5ce446547"
 NUSCENES_ATTRIBUTION = "Adapted from nuScenes dataset. Copyright 2020 nuScenes."
+
+ROBOTIS_BUTTON_PUSH_SAMPLE_REVISION = "93a0ba73c6b82689696d9c4909b3b48af0294cf8"
+ROBOTIS_BUTTON_PUSH_SAMPLE_URL = (
+    "https://huggingface.co/datasets/RobotisAI/evButtonPush-260615-0-MCAP/"
+    f"resolve/{ROBOTIS_BUTTON_PUSH_SAMPLE_REVISION}/107/107_0.mcap?download=true"
+)
+ROBOTIS_BUTTON_PUSH_SAMPLE_SHA256 = (
+    "51bec431162844b2bb239f65ea158ea1030bb3fce5312ad81768978dc68f82c3"
+)
 
 # The training-sample "state" analog, in preference order: proprioception-like
 # telemetry first, densest non-camera topic as the fallback.
--- a/docs/BENCHMARKS.md
+++ b/docs/BENCHMARKS.md
@@ -16,14 +16,15 @@ alongside the million-hour results Dyna published in
 | Four-camera synthetic training windows | **2.42x fewer chunk fetches** than per-topic chunking | Shows how grouping topics by read pattern reduces sample assembly work |
 | Six-camera real footage with 8 MB chunks | **2.81x fewer fetches and 3.21x fewer bytes fetched** than per-topic chunking | Demonstrates that chunk size and grouping policy must be tuned together |
 | Selective state scans | Naive schema grouping fetched **230 MB** for a 0.2 MB `/imu` stream | Shows why HFlow exposes per-topic group overrides instead of treating grouping as a fixed schema rule |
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging. This is what #170 asked for and could not assume it would get: a measurement on the workload the default is meant to serve, with the conditions stated, and a conclusion that follows from the numbers rather than from wanting a result.

**The pin resolves and is exact.** I checked the revision without downloading it, and both facts you recorded match the Hub:

```
x-linked-size: 75946038
x-linked-etag: "51bec431162844b2bb239f65ea158ea1030bb3fce5312ad81768978dc68f82c3"
```

Keeping the 800 KB floor is the right conclusion, and stating it as "no fixed floor dominates across the two measured shapes" is more useful than a new number would have been. The state-scan column is what makes the case: 7 to 13 fetches on manipulation and 16 to 21 on nuScenes is the cost that a fetches-times-bytes improvement hides.

Gates clean: 1102 passed, lychee 324 OK.

**One thing to double-check rather than a change request.** In the cross-check table, manipulation at the 400 KB candidate and nuScenes at the shipped 800 KB both read `3.77` fetches/sample. Two corpora agreeing to three significant figures is possible, and it is also what a copy-paste looks like. You have the run output and I do not, so a glance is cheaper for you than a re-run is for anyone else.

I also confirmed the two numbers that look inconsistent between your PR body and the doc are not: the body quotes the cross-check row (2.75, 3.77) and the sweep table has the flat 400 KB row (2.77, 3.80). Different configurations, both present. Worth naming which table a number comes from when you quote it, since a reader without both open sees a contradiction.

For the record, neither checksum constant is read by anything; `NUSCENES_SAMPLE_SHA256` is documentary too. Yours matches the existing precedent, so nothing to fix here, but the constants do not enforce what they look like they enforce.

PR #211, opened 2026-08-27
docs: add contract docstrings for exported API names (#60)

### LINKED ISSUE #60: Docs: docstrings for the last undocumented exported names (outside doctor.py)
Several exported names are the remaining public API members without useful
docstrings. Documenting them improves `help()` and editor hovers and makes them
consistent with adjacent public names.

## Current behavior

- `FfmpegNotFoundError` is a bare class at
  `src/hflow/ffmpeg/_binary.py:90-91`; its subclasses at lines 94-99 are
  documented.
- `InstrumentParseError` is a bare class at
  `src/hflow/ffmpeg/_instrument.py:30-31`.
- `AirflowClientError` has no class docstring at
  `src/hflow/runtime/_client.py:26-30`.
- `CheckCoverage` and its `fraction` property have no docstrings at
  `src/hflow/curation.py:62-69`.
- `CurationReport` has no class docstring at `src/hflow/curation.py:73-79`.
  The `manifest_path` tri-state exists only in comments at lines 74-75.
- `EpisodeReader.close` is the only undocumented protocol method at
  `src/hflow/reader.py:113`.
- `TestReport.quarantined` is undocumented at `src/hflow/app.py:358-359`, while
  `has_errors` at lines 362-366 is documented.

These names are exported through `hflow`, `hflow.ffmpeg`, or `hflow.runtime`.
`doctor.py` and `CanonicalMcapWriter` remain out of scope because #11 covers
them.

## Pattern to copy

Use the adjacent documented siblings, such as `FfprobeNotFoundError` at
`src/hflow/ffmpeg/_binary.py:94-95` and the one-line contracts on other
`EpisodeReader` methods at `src/hflow/reader.py:79-110`.

## Definition of done

- Every name listed above has a concise contract docstring.
- The redundant `pass` statements are removed.
- The `manifest_path` tri-state moves from comments into the
  `CurationReport` class docstring.
- No behavior changes.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

NiranjanKumar001:
Fixes #60

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -869,6 +869,7 @@ class TestReport:
 
     @property
     def quarantined(self) -> bool:
+        """Whether a critical check verdict marked this run as quarantined."""
         return bool(self.quarantine_tags)
 
     @property
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -869,7 +869,12 @@ class TestReport:
 
     @property
     def quarantined(self) -> bool:
-        """Whether a critical check verdict marked this run as quarantined."""
+        """Whether this run carries any quarantine tag.
+
+        A failed critical check adds one. So does a quarantine already
+        recorded in the catalog, when this run skipped the meta stage and
+        read the episode's cataloged state instead.
+        """
         return bool(self.quarantine_tags)
 
     @property
--- a/src/hflow/curation.py
+++ b/src/hflow/curation.py
@@ -131,8 +131,9 @@ class CheckCoverage:
 class CurationReport:
     """Outcome of a curation SQL query and manifest generation.
 
-    manifest_path is a Path for local manifests, a str object URL for bucket
-    manifests, or None when curate() ran with dry_run=True.
+    ``manifest_path`` is a Path for local manifests, a str object URL for
+    bucket manifests, and None when ``curate()`` ran with ``output=None``,
+    which reports row count and coverage without writing anything.
     """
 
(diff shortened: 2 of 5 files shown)


kstonekuan (approved):
Merging. Every name #60 listed is covered, and the contracts say what the code does.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1102 passed, 6 skipped.

Two fixups pushed to your branch (2c41d6c), both accuracy rather than style:

- `CurationReport` said the manifest is None "when curate() ran with `dry_run=True`". There is no `dry_run` parameter. The condition is `output=None` (`curation.py:537`), which still runs the query and reports row count and coverage.
- `TestReport.quarantined` said a critical check verdict marks the run. That is one of two sources. When a run skips the meta stage it reads the episode's cataloged quarantine and extends the same list (`app.py:2038-2043`), so a media-only run can report `quarantined` with no check verdict in it at all. Worth knowing because that is the case the property exists to catch.

One thing from the issue you correctly did not do: `InstrumentParseError` is listed there but no longer exists, since the ffmpeg instrument module was refactored after the issue was filed. Nothing missing.

If you want another, the open list is here: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

Issue #209, opened 2026-08-27
A pinned content digest makes the test suite fail on any platform whose libm differs


kstonekuan:
### What happens

`test_content_digest_emits_exactly_the_documented_measurements`
(`tests/test_default_checks.py:1225`) pins a literal digest:

```python
# Pin the exact digest for the joints-only source episode. The synthetic
# writer is deterministic, so a dispatcher that returns a transformed
# value (e.g. reversed hex) will not match this literal.
assert digest == "1ee3e9bcad4d5a12112b196f7a3d4e272329cc41a5b5efe7e8751339ab1a4722"
```

The comment's premise is wrong. `synthesize_episode` generates every joint position and velocity
with `math.sin` and `math.cos` (`src/hflow/testing.py:584-586`):

```python
angle_rad = 2.0 * math.pi * frequency_hz * sampled_time_s + phase_rad
positions.append(JOINT_SINE_AMPLITUDE_RAD * math.sin(angle_rad) + jump_offset_rad)
velocities.append(JOINT_SINE_AMPLITUDE_RAD * 2.0 * math.pi * frequency_hz * math.cos(angle_rad))
```

`math.sin` and `math.cos` call the platform's libm, and libm is not bit-identical across
platforms. glibc, musl, macOS, and the Windows CRT disagree in the last bit for some inputs.
Those floats are encoded into the joint payloads, `content_digest` hashes the payload bytes, so
the digest is a property of the platform that produced it.

### What it has cost

It passes on Linux and on CI. It fails elsewhere, and a contributor who sees it cannot tell
whether they broke something:

- @afonsojanu hit it twice, in #201 and #208, and worked around it both times by removing their
  own test file to prove the failure was pre-existing.
- I could not reproduce it and asked them for their platform instead of finding the cause, which
  cost another round trip.

`CONTRIBUTING.md:27` says macOS works for native development, so this is a supported platform
failing a test nobody broke.

### What the test is actually for

It was added in #198 to prove the `content_digest` dispatcher returns the digest rather than a
transformed value, "e.g. reversed hex". The literal is one way to prove that and not the only
one. Anything that pins the relationship between the emitted value and the digest of the same
bytes works, without depending on which bytes libm produced.

### Possible shapes

Not obvious which is right, which is why this is an issue rather than a patch:

1. **Compute the expected digest in the test** from the same canonical file, using
   `hashlib.sha256` over the same inputs `content_digest` reads. Proves the dispatcher returns
[...]

PR #212, opened 2026-08-27
refactor(runtime): parse the conf flag where mode is parsed

chiruu12:
Follow-up to #195, on the `all_stages` conf flag it added.

The string coercion lived inside the sub-DAG template. `plan`'s own docstring says the lane semantics live in `hflow.stage_execution`, one owner for every execution backend, and this was the one piece of conf parsing that did not. It moves next to where `mode` is parsed, and the template calls it.

That also makes it directly testable. The coercion was only reachable by extracting the rendered `plan` body and calling it, which works but is a heavy way to cover four spellings.

The direction that matters is false, not true, which is the point #195 made when it pinned that side. Reading a true spelling as false costs one wasted run and is obvious. Reading a false spelling as true turns the filter off silently: the corpus stays correct and the re-ingest just keeps costing what it cost before the filter existed. `bool()` on the raw value gets that wrong for every non-empty string, `"false"` included.

`tests/test_stage_execution.py` covers the yes spellings, the no spellings including the empty and whitespace forms a cleared trigger-form field sends, and native values, since `render_template_as_native_obj` hands a real bool straight through and an unset param arrives as `None`.

No behaviour change. The rendered DAGs do what they did after #195.

Validation:

```
uv run pytest -q          1111 passed, 9 skipped, 1 failed
uv run ruff check         All checks passed
uv run ruff format --check  174 files already formatted
uv run ty check           All checks passed
```

The failure is `tests/test_default_checks.py::test_content_digest_emits_exactly_the_documented_measurements`. It fails on a clean checkout of `c81a48e` too, so it is not from this branch and I have left it alone.

### DIFF
--- a/src/hflow/runtime/_templates.py
+++ b/src/hflow/runtime/_templates.py
@@ -522,15 +522,13 @@ def ingest_stage():
         catalog cannot see -- an artifact deleted out from under a step whose
         rows still say it ran.
         """
-        from hflow.stage_execution import plan_stage_batches
+        from hflow.stage_execution import parse_conf_flag, plan_stage_batches
 
         if not uris:
             return []
-        # Conf values arrive rendered; a native bool passes straight through
-        # and a hand-typed string is read the way every other flag in the
-        # project reads one.
-        if isinstance(all_stages, str):
-            all_stages = all_stages.strip().lower() in {"1", "true", "yes", "on"}
+        # The conf vocabulary is parsed at the library boundary, like mode --
+        # this task only feeds it what Airflow rendered.
+        all_stages = parse_conf_flag(all_stages)
 $stage_plan_filter        return plan_stage_batches(
             [str(uri) for uri in uris],
             mode=mode,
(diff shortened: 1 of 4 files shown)


kstonekuan (approved):
Merging. The flag now has one owner and the false direction is pinned, which is the side that fails quietly.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1115 passed, 6 skipped (13 new).

I checked the no-behaviour-change claim rather than taking it: the old code left `None` alone and the new code returns `False` for it, but `_templates.py:697` reads the value as `if not all_stages`, so the two are the same at the only site that consumes it. `test_sub_dag_sources_compile_and_encode_contract` still pins the import line, so a rename cannot quietly detach the template from the parser.

One fixup pushed (557ca2c): the ` -- ` dashes became commas and a period, and "It matters more than it looks" came out. House style on both, and the two sentences after that clause already carry it.

You are right that the digest failure is not yours. It is #209, and your third data point there settled the part that was open. Assigning it to you.

PR #213, opened 2026-08-28
fix: avoid replanning camera-less media

### LINKED ISSUE #171: A camera-less episode is planned for the media stage on every re-ingest

tunglambk:
## Summary

- Use the current default `camera_frame_stats` catalog result to identify camera-less episodes before media planning.
- Keep media planning unchanged when camera presence is unknown or a camera-bearing episode still lacks a contact sheet.

## Why

Camera-less episodes never record `media/contact_sheet`, so every re-ingest currently schedules a media stage that cannot produce output. The default camera check already records enough evidence to skip that work without changing the catalog schema, while preserving the existing fallback when the evidence is unavailable.

Closes #171.

## Validation

- `uv sync --locked --all-extras` — completed successfully.
- `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check` — passed.
- `uv run pytest -q tests/test_stage_planning.py` — 37 passed.
- `uv run pytest -q` — 1,117 passed and 6 skipped; one unrelated local ffmpeg test failed because the installed ffmpeg rejects an apostrophe in a drawtext font path. The same `main` commit passes upstream CI.

## Checklist

### DIFF
--- a/src/hflow/stage_planning.py
+++ b/src/hflow/stage_planning.py
@@ -40,7 +40,7 @@ from dataclasses import dataclass
 from enum import StrEnum
 from typing import TYPE_CHECKING
 
-from hflow.steps import SETTLED_STATUSES, Stage
+from hflow.steps import RAN_STATUSES, SETTLED_STATUSES, Stage
 
 if TYPE_CHECKING:
     import duckdb
@@ -179,6 +179,50 @@ def _settled_step_names(
     return settled
 
 
+def _camera_presence_by_episode(
+    connection: "duckdb.DuckDBPyConnection",
+    application: "App",
+    episode_ids: Sequence[str],
+) -> dict[str, bool]:
+    """Whether the default camera check measured at least one camera."""
+    from hflow.checks import (
+        _default_check_version_for_automatic_registration,
+        camera_frame_stats,
+    )
+
+    check_name = camera_frame_stats.__name__
+    check_version = _default_check_version_for_automatic_registration(camera_frame_stats)
+    if not episode_ids or not any(
+        registered.function is camera_frame_stats
+        and registered.name == check_name
+        and registered.version == check_version
+        for registered in application.checks
+    ):
+        return {}
+
+    episode_placeholders = ", ".join("?" for _ in episode_ids)
+    statuses = ", ".join(_quote_sql_string(status.value) for status in RAN_STATUSES)
+    rows = connection.execute(
+        f"""
+        SELECT runs.episode_id, count(measurements.key) > 0
+        FROM check_runs AS runs
+        LEFT JOIN measurements
+          ON measurements.episode_id = runs.episode_id
+         AND measurements.run_fingerprint = runs.run_fingerprint
+         AND measurements.check_name = runs.check_name
+         AND measurements.check_version = runs.check_version
+         AND ends_with(measurements.key, '/message_count')
+        WHERE runs.status IN ({statuses})
+          AND runs.episode_id IN ({episode_placeholders})
+          AND runs.check_name = ?
+          AND runs.check_version = ?
+        GROUP BY runs.episode_id
+        """,
+        [*episode_ids, check_name, check_version],
+    ).fetchall()
+    return {str(episode_id): bool(has_camera) for episode_id, has_camera in rows}
+
+
 def plan_outstanding_stages(
     application: "App",
     source_identities: Sequence[str],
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
Merging. This implements the shape #171 specified, and it got the subtle part right on its own.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1119 passed, 6 skipped.

Reading `RAN_STATUSES` rather than `SETTLED_STATUSES` is the decision that matters here, and the issue did not call it out. A superseded `camera_frame_stats` emitted no per-camera keys because it stood down, so counting that row as an answer reads a camera-bearing episode as camera-less and never renders its contact sheet. Your version excludes it and falls through to planning, which is the right side of rule 3.

Nothing pinned it, though. I swapped `RAN_STATUSES` for `SETTLED_STATUSES` and all 37 tests still passed, so I pushed one test (9aafbf3) that supersedes `camera_frame_stats` on a camera-bearing episode and asserts `media` is still planned. It fails on that swap and on nothing else.

Two things I checked rather than assumed:

- `_default_check_version_for_automatic_registration` is the same accessor `app.py:1186` uses at registration, so a version bump cannot silently turn the inference off. That was the concern behind the issue's note about constants.
- `ends_with(key, '/message_count')` is a correct marker: `_camera_frame_stats_keys` emits `{topic}/message_count` for every camera topic unconditionally, and the two non-topic keys do not match the suffix.

On your local ffmpeg failure: HFlow pins its own ffmpeg build (`PINNED_BUILDS_BY_MACHINE` in `ffmpeg/_binary.py`), so a system ffmpeg on `PATH` can behave differently from what CI runs. `_quote_filter_value` does handle the apostrophe close-escape-reopen form. If it still fails with the pinned binary, that is worth its own issue and I would want the ffmpeg version in it.

One note since this is your third contribution: the `good first issue` pool is best left for people arriving, and you are past needing it. Higher-leverage work if you want it: run HFlow against a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face are good targets) and report what breaks, what is slow, or what is awkward. Reports like that have produced better issues than anything in the pool. #166 is still yours either way.

Issue #215, opened 2026-08-28
Pinned ffmpeg rejects apostrophes in contact-sheet font paths


tunglambk:
### Version or commit

`01b92a8e409f07acd2b05c73d7149150d4348a22`

### Environment

Ubuntu 22.04.4 LTS, Python 3.13.13, x86_64, pinned ffmpeg `n8.1.2-44-g7c533d0f86-20260816`. `fc-match` resolves a usable system font on this host.

### Minimal reproduction

Run the existing regression test with HFlow's pinned ffmpeg installed:

```console
uv run pytest -q tests/test_ffmpeg.py -k test_contact_sheet_accepts_apostrophes_in_external_paths
```

### Expected behavior

The test passes and produces a contact sheet when both the extracted-frame directory and copied font filename contain apostrophes.

### Actual behavior

The concat demuxer accepts the quoted frame paths, but drawtext rejects the copied font path:

```text
Error initializing filter 'drawtext' with args 'fontfile=.../wearer's assets/worker's-font.ttf:text=+0.0s:...'
Error reinitializing filters!
Failed to inject frame into filter network: Invalid argument
```

The test fails with `RuntimeError: ffmpeg contact sheet failed` at `_contact_sheet.py:187`.

### Additional context

This test is environment-dependent. It only copies a font into the apostrophe-containing directory when `_find_usable_font_file()` returns a path. On an environment with no discoverable font, drawtext is disabled and the test can pass without exercising the fontfile quoting branch. That likely explains why current `main` CI is green while the same test fails here with the pinned binary.

`_quote_filter_value()` currently uses close-escape-reopen quoting for all filter values. The failure is specific to the drawtext `fontfile` option's parsing layers; the frame paths using the concat demuxer's equivalent quoting are accepted. A regression should force a known usable font path so this branch is covered consistently.

PR #216, opened 2026-08-28
fix(video-measurements): keep empty freeze totals float

### LINKED ISSUE #214: Egocentric example errors on healthy episodes with no freeze intervals
### Version or commit

`01b92a8e409f07acd2b05c73d7149150d4348a22`

### Environment

Ubuntu 22.04.4 LTS, Python 3.13.13, x86_64, pinned ffmpeg `n8.1.2-44-g7c533d0f86-20260816`.

### Minimal reproduction

I found this while running HFlow against a real Egocentric-10K excerpt from the public Apache-2.0 [`Voxel51/Egocentric_10K_subset`](https://huggingface.co/datasets/Voxel51/Egocentric_10K_subset) at revision `8a136de2d005a7041d80d2f78ac0478ce5f2fb54`. I adapted three 20-second windows with `hflow.testing.write_video_episode`; all three reproduced the same result.

Run the repository's documented pipeline against any healthy adapted excerpt with no detected freeze intervals:

```console
uv run python examples/egocentric/pipeline.py path/to/healthy-episode.mcap
```

The smaller underlying condition is a `camera_frame_stats` result with no freeze intervals. `_FrameStatsAccumulator.finish()` computes `freeze_total_seconds` with an unseeded `sum(...)`, so the empty case produces `int 0` even though `VideoFrameStatistics.freeze_total_seconds` is declared as `float`.

### Expected behavior

The example's `camera_health` check should record a measured result for a healthy excerpt. `freeze_total_s` should be `0.0`, matching the `VideoFrameStatistics` field type and the non-empty case.

The example README says the 90 unmodified excerpts pass `camera_health` while the six injected-fault excerpts are quarantined.

### Actual behavior

The check records an error because the example correctly expects the typed field to be a float:

```text
! camera_health [error]
  File "examples/egocentric/pipeline.py", line 42, in camera_health
    assert isinstance(freeze_total_seconds, float)
AssertionError
```

The catalog contains `('camera_health', 'error', ...)`, but the command still exits 0 and continues with enrichments, which makes the problem easy to miss. The built-in `camera_frame_stats` output in the same run shows `freeze_total_s = 0` and otherwise healthy evidence: 200 decoded frames, 0% black frames, and 0% frame deficit.

### Additional context

Using `sum((...), 0.0)` for `freeze_total_seconds` would preserve the declared core type for the empty case. A regression test should check the type, not only equality with `0.0`, because `0 == 0.0` allowed the current test to pass. Changing only the example to accept `int` would leave the core dataclass's runtime value inconsistent with its annotation.

tunglambk:
## Summary

- Seed the empty freeze-duration sum with `0.0` so `freeze_total_seconds` always matches its declared `float` type.
- Strengthen the existing no-freeze regression to check the runtime type as well as numeric equality.

## Why

Running the documented egocentric pipeline against three healthy excerpts from a real Egocentric-10K video made `camera_health` error on each episode. With no detected freeze intervals, Python's unseeded `sum(...)` returned `int 0`; the example expects the core field's declared `float` type and rejected it. Seeding the aggregation fixes the core invariant rather than weakening the example.

Closes #214.

## Validation

- The new type assertion fails on `main` with `type(0) is float` and passes with this change.
- `uv run python examples/egocentric/pipeline.py <healthy-real-episode.mcap>` — `camera_health` changed from `error` to `passed` with 200 decoded frames and `freeze_total_s = 0`.
- The pinned 96-episode Egocentric-10K example corpus completed with `camera_health` passing 90 clean episodes and failing exactly the six injected faults; the catalog had zero check errors and zero ingest failures. Re-ingest completed in 5.02 seconds with all expensive work current.
- `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check` — passed with no changes.
- `uv run pytest -q -k 'not test_contact_sheet_accepts_apostrophes_in_external_paths'` — 1,118 passed, 6 skipped, 1 deselected. The excluded test also fails unchanged on `main` with the pinned ffmpeg and is tracked separately in #215.

## Checklist

- [x] I updated documentation for changed behavior, flags, formats, or requirements. (Not applicable: this restores the field's declared `float` type.)

### DIFF
--- a/src/hflow/_video_measurements/_frame_statistics.py
+++ b/src/hflow/_video_measurements/_frame_statistics.py
@@ -382,7 +382,11 @@ class _FrameStatisticsAccumulator:
             overexposed_frame_percent=100.0 * self.overexposed_frame_count / frame_count,
             freeze_intervals=tuple(self.freeze_intervals),
             freeze_total_seconds=sum(
-                interval.end_seconds - interval.start_seconds for interval in self.freeze_intervals
+                (
+                    interval.end_seconds - interval.start_seconds
+                    for interval in self.freeze_intervals
+                ),
+                0.0,
             ),
             average_luma_mean=self.average_luma.mean(),
             average_luma_minimum=self.average_luma.minimum(),
--- a/tests/test_ffmpeg.py
+++ b/tests/test_ffmpeg.py
@@ -371,6 +371,7 @@ def test_synthetic_output_aggregates_exactly() -> None:
     assert stats.average_luma_maximum == pytest.approx(100.0)
     assert stats.freeze_intervals == ()
     assert stats.freeze_total_seconds == 0.0
+    assert type(stats.freeze_total_seconds) is float
     # Default threshold 235.0: none of 100, 10, 40 are >= 235
     assert stats.overexposed_frame_count == 0
     assert stats.overexposed_frame_percent == 0.0


kstonekuan (approved):
Merging. Fixing the core invariant instead of loosening the example is the right call, and `type(...) is float` is the right assertion: `isinstance` would also accept `np.float64`, which is the other way this field could drift.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1119 passed, 6 skipped.

I checked whether this has siblings. It does not: every other aggregate in `src/` either sums ints into an int-declared field or wraps `np.sum` in an explicit `float()`, so `freeze_total_seconds` was the only float-declared field reachable by an unseeded `sum()`.

The apostrophe test passes here, which supports your read in #215 that the branch only runs where `_find_usable_font_file()` finds something. That is now labelled and open.

This is the kind of report the corpus steer was for, and you turned it around in a couple of hours. #214 and #215 are both better issues than anything currently in the starter pool.

PR #217, opened 2026-08-28
refactor(transform): type pass-through video messages

### LINKED ISSUE #206: The pass-through video path passes an untyped decoded message and checks h264 twice
### What happens

The pass-through video path in `transform.py` carries a decoded message as `Any` and each function
reaches into it for the fields it needs:

- `_insert_missing_passthrough_video_aud(topic: str, decoded_message: Any)` (`transform.py:383`)
  reads `.format`, reads `.data`, and assigns `.data`
- `_validate_passthrough_video_payload(topic: str, decoded_message: Any, ...)`
  (`transform.py:401`) reads `.format` again and reads `.data` again
- `_resolve_encoder`'s `encode_protobuf(message: Any)` (`transform.py:360`) re-serializes it
- the transcode path assigns `.data` and `.format` on a different object of the same shape
  (`transform.py:622-623`)

Two of those independently test `decoded_message.format != "h264"` (`:385` and `:409`), so the
knowledge that a conforming message is h264 lives in two places. `mcap_ros2` and the protobuf
decoder return different classes with the same duck-typed surface, which is why the annotation is
`Any` and why nothing catches a typo in a field name.

### Why now

The path grew a third caller last week. #184 added the AUD repair, which introduced the mutation
(`decoded_message.data = canonical_data`) and the re-serialization step, and its correctness
depends on the encoder round-tripping fields nobody has named. Each addition duplicates the same
assumptions about the message's shape, and the type system is not watching any of them.

`.data` is also mutated in place. The style guide's preference for parsing into a value and
passing it forward would make the repair produce a new payload rather than edit a decoded object
that the caller still holds.

### What to build

Parse the decoded message once at the boundary into a small frozen type carrying the two fields
this path actually uses, then let the three functions take that type:

```python
@dataclass(frozen=True)
class PassthroughVideoMessage:
    format: str
    data: bytes
```

The boundary is where `decode(payload)` is called (`transform.py:568` area). One parse site, and
the h264 check becomes one place rather than two. The repair returns a new instance instead of
mutating. `_resolve_encoder` keeps taking the decoder's own object, since re-serialization
genuinely needs the original class; hand it the repaired bytes rather than a mutated message.

This is the style guide's "parse external input into domain types at system boundaries and trust
refined types downstream", and its rule against shotgun parsing, where business functions
repeatedly validate fragments of raw input.

### Definition of done

1. `_insert_missing_passthrough_video_aud` and `_validate_passthrough_video_payload` take a typed
   value, not `Any`.
2. The `format != "h264"` test exists once.
3. No function mutates a decoded message in place. The repair returns bytes or a new value.
4. Both encodings still work: a protobuf `foxglove.CompressedVideo` source and a ROS 2 CDR
   `foxglove_msgs/msg/CompressedVideo` source. `tests/test_processing_regressions.py` already
   covers both AUD-repair paths; they must pass untouched.
5. Byte-for-byte output is unchanged for a conforming source and for a repaired source. A
   repaired message is still exactly six bytes longer with the original payload as its suffix.
6. No `TRANSFORM_BEHAVIOR_VERSION` bump, because nothing about the written bytes changes. If you
   find yourself wanting one, something in the refactor is not behaviour-preserving.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Not a good first issue: it sits on the canonical-video path, and item 5 wants a before-and-after
byte comparison.

The other `Any` annotations in the codebase are fine and out of scope. `_load_obstore() -> Any`,
the MediaPipe result, and the obstore handle are all loose representations of optional
dependencies at their own boundaries, which the guide explicitly allows.

WilliamK112:
## Summary

Parse pass-through video once at the decode boundary into a frozen `PassthroughVideoMessage` with an H.264 format type and immutable payload bytes. AUD repair returns a new value, and the validator accepts that value instead of re-reading an untyped decoded object.

The codec boundary receives the original decoded message and the repaired bytes separately. Protobuf serialization copies the message, preserving unknown fields. CDR serialization uses a read-only payload override that forwards the remaining fields: `copy.copy` is not sufficient for mcap_ros2's slotted `SimpleNamespace` messages, as the unchanged CDR regression demonstrated by losing `frame_id` and `format` during development.

There is now one `format != "h264"` check. Conforming serialized payloads still bypass re-encoding. No decoded input is mutated, no public behavior is changed, and **no `TRANSFORM_BEHAVIOR_VERSION` bump** is made.

Closes #206

## Verification

- Left the existing protobuf and CDR AUD-repair regressions unchanged, including the six-byte insertion / original-payload suffix assertions.
- Added one parameterized ownership regression for both encodings, with conforming and repaired messages. The two repaired cases fail against the original implementation because the decoder-owned input is mutated. It also pins that an additional CDR field named `original` survives serialization.
- `uv sync --locked` — complete development environment as documented in current CONTRIBUTING.md; no optional endpoint/model extras needed.
- `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check` — pass. Final check-only equivalents also pass.
- `uv run pytest -q tests/test_processing_regressions.py tests/test_identity_stability.py -o tmp_path_retention_count=0` — **47 passed** on Python 3.14.7. The same 47 tests pass on Python 3.11.4 using the existing Python 3.11 environment with this checkout first on `PYTHONPATH`.
- Root suite, including `packages/hflow-server/tests`: `PYTHONUNBUFFERED=1 uv run pytest -q -o tmp_path_retention_policy=failed -o tmp_path_retention_count=0` — **1,119 passed, 9 skipped, 1 failed**. The only failure is the already tracked macOS digest literal in #209. It produces exactly the same `164675bb...468117c2` value on clean baseline `01b92a8`, reproduced separately with the same interpreter/dependencies. No tests were deselected or assertions changed. The retention options only free temporary files after passing tests on a disk-constrained machine.
- `git diff --check` — pass.
- [Hosted Linux CI](https://github.com/Hebbian-Robotics/hflow/actions/runs/33137091313) — **1,123 passed, 6 skipped on both Python 3.11 and 3.14**. Lint, format, typechecking, links, and the one-open-PR check all pass.

## Before/after byte comparison

Before editing, wrote canonical outputs at `01b92a8` for four fixed two-frame inputs (keyframe + delta). Compared **complete MCAP files**, not just decoded payloads, after the refactor. The protobuf corpus includes an unknown wire field, and the CDR corpus includes an extra `source_label` field.

| Input | Canonical bytes | Before/after |
| --- | ---: | --- |
| protobuf, conforming | 1,971 | identical |
| protobuf, missing AUD | 1,971 | identical |
| CDR, conforming | 1,560 | identical |
| CDR, missing AUD | 1,560 | identical |

The full-file SHA-256 remains `d5cba2e6f160c56bc3930b77f809d30445931b5e2c8193b2d7281d46cb17343c` for both protobuf cases and `2e9b94adc6c40227c238b0db60c9d38bf797d51caab5e29a9f5b37aebff6cc8b` for both CDR cases. Pipeline version remains `9cd5d34f513d`. The same comparisons also pass under Python 3.11.4.

## Checklist

- [x] Added outcome-focused coverage and kept the existing repair tests intact.
- [x] Ran the Python quality gates and root test suite; documented the known baseline failure.
- [x] Preserved stored bytes and identity without a behavior-version bump.
- [x] No recordings, generated media, credentials, or runtime artifacts included.
- [x] No documentation behavior change: this is an internal representation refactor.

Implemented and validated with OpenAI Codex assistance.

### DIFF
--- a/src/hflow/transform.py
+++ b/src/hflow/transform.py
@@ -400,26 +400,26 @@ def _resolve_encoder(topic: str, info: TopicInfo) -> Callable[[Any, bytes], byte
 
 
 @dataclass(frozen=True)
-class PassthroughVideoMessage:
+class _PassthroughVideoMessage:
     """The H.264 payload parsed at the codec boundary, independent of its encoding."""
 
     format: Literal["h264"]
     data: bytes
 
 
-def _parse_passthrough_video_message(topic: str, decoded_message: Any) -> PassthroughVideoMessage:
+def _parse_passthrough_video_message(topic: str, decoded_message: Any) -> _PassthroughVideoMessage:
     if decoded_message.format != "h264":
         raise ValueError(
             f"pass-through video topic {topic!r} carries format "
             f"{decoded_message.format!r}; the canonical convention requires 'h264' "
             "(re-encode upstream -- v1 does not transcode CompressedVideo)"
         )
-    return PassthroughVideoMessage(format="h264", data=bytes(decoded_message.data))
+    return _PassthroughVideoMessage(format="h264", data=bytes(decoded_message.data))
 
 
 def _insert_missing_passthrough_video_aud(
-    topic: str, message: PassthroughVideoMessage
-) -> PassthroughVideoMessage:
+    topic: str, message: _PassthroughVideoMessage
+) -> _PassthroughVideoMessage:
     """Repair the one canonical constraint that is lossless to bridge."""
     try:
         canonical_data = video_module.ensure_access_unit_delimiter(message.data)
@@ -427,11 +427,11 @@ def _insert_missing_passthrough_video_aud(
         raise ValueError(
             f"pass-through video topic {topic!r} violates the canonical convention: {error}"
         ) from error
-    return PassthroughVideoMessage(format=message.format, data=canonical_data)
+    return _PassthroughVideoMessage(format=message.format, data=canonical_data)
 
 
 def _validate_passthrough_video_payload(
-    topic: str, message: PassthroughVideoMessage, *, is_first_message: bool
+    topic: str, message: _PassthroughVideoMessage, *, is_first_message: bool
 ) -> None:
     """Enforce the canonical video constraints on a pass-through message.
 
--- a/src/hflow/transform.py
+++ b/src/hflow/transform.py
@@ -57,6 +57,7 @@ import hashlib
 import json
 import logging
 from collections.abc import Callable, Mapping, Sequence
+from copy import copy
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Literal
@@ -353,12 +354,25 @@ def _resolve_decoder(topic: str, info: TopicInfo) -> Callable[[bytes], Any]:
     )
 
 
-def _resolve_encoder(topic: str, info: TopicInfo) -> Callable[[Any], bytes]:
+@dataclass(frozen=True)
+class _Ros2VideoMessageWithData:
+    """Override only the payload while forwarding every other field to its owner."""
+
+    _original: object
+    data: bytes
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self._original, name)
+
+
+def _resolve_encoder(topic: str, info: TopicInfo) -> Callable[[Any, bytes], bytes]:
     """An encoder matching a decoded pass-through video message."""
     if info.message_encoding == "protobuf":
 
-        def encode_protobuf(message: Any) -> bytes:
-            serialize = getattr(message, "SerializeToString", None)
+        def encode_protobuf(message: Any, data: bytes) -> bytes:
+            repaired_message = copy(message)
+            repaired_message.data = data
+            serialize = getattr(repaired_message, "SerializeToString", None)
             if not callable(serialize):
                 raise ValueError(
                     f"camera topic {topic!r} decoded a protobuf message that cannot "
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
Merging. Every item in #206's definition of done is met, and the verification in the PR body holds up under independent checking.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1123 passed, 6 skipped.

What I checked rather than took from the description:

**Byte identity.** I built conforming and repaired pass-through sources from the repo's own fixtures and hashed the complete canonical files on `01b92a8` and on this branch. Identical both ways: `8463e975...f133ed4f` conforming, `d15c88e2...442b7ac5` repaired. Different corpus from yours, same result.

**The ownership tests are real tripwires.** I reverted each encoder to mutate the decoded message in place. `encode_protobuf` mutating fails `[repaired-protobuf]` and nothing else; `encode_cdr` mutating fails `[repaired-cdr]` and nothing else. Each parametrization catches exactly its own path, which is what makes item 3 enforceable rather than reviewed once.

**`copy.copy` on the protobuf message.** This was my main worry, since a shallow copy that aliased would make the no-mutation property silently false. It does not: the copy is a distinct object, assigning `.data` on it leaves the original at its old bytes, and other fields survive. Your unknown-wire-field corpus covers the half I did not.

The `_Ros2VideoMessageWithData` wrapper is the interesting part of this PR. Reaching for `copy.copy` first and finding it drops `frame_id` and `format` on mcap_ros2's slotted messages is the kind of thing that is invisible in a diff, so thank you for putting it in the description. Pinning a CDR field named `original` is a good instinct, given the wrapper holds `_original`.

One fixup pushed (aa043f2): `PassthroughVideoMessage` is now `_PassthroughVideoMessage`. In this file an unprefixed name means exported from `hflow` (`TransformConfig` and `EpisodeStamps` both are), and this type is internal to the pass-through path. The unprefixed spelling came from my issue body, so that one is on me.

PR #218, opened 2026-08-28
fix(contact-sheet): escape apostrophes for drawtext

### LINKED ISSUE #215: Pinned ffmpeg rejects apostrophes in contact-sheet font paths

tunglambk:
## Summary

- Escape apostrophes through both the filtergraph and `drawtext` option parsers.
- Make the existing regression require an actual font path and verify that timestamps were burned.

## Why

The pinned ffmpeg rejects contact sheets when the selected font path contains an apostrophe because the existing escaping is consumed by the first parsing layer. Preserving the escape for the second parser lets `drawtext` open the intended font without changing the fallback behavior on systems without a usable font or filter support.

Closes #215.

## Validation

- `uv run pytest -q tests/test_ffmpeg.py -k test_contact_sheet_accepts_apostrophes_in_external_paths` — 1 passed, 56 deselected.
- `uv run pytest -q` — 1,123 passed, 6 skipped.
- `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check` — passed with no changes.

## Checklist

- [x] I updated documentation for changed behavior, flags, formats, or requirements. (Not applicable: this fixes escaping for an already supported input.)

### DIFF
--- a/src/hflow/ffmpeg/_contact_sheet.py
+++ b/src/hflow/ffmpeg/_contact_sheet.py
@@ -96,10 +96,10 @@ def _evenly_sampled_indices(total_count: int, max_count: int) -> list[int]:
 def _quote_filter_value(value: str) -> str:
     """Quote a value for use inside an ffmpeg filtergraph option.
 
-    Inside filtergraph single quotes a backslash is LITERAL (av_get_token),
-    so an embedded apostrophe must close-escape-reopen: ``'`` -> ``'\\''``.
+    ``drawtext`` parses option values after the filtergraph parser, so an
+    embedded apostrophe needs one escaping layer for each parser.
     """
-    return "'" + value.replace("'", "'\\''") + "'"
+    return "'" + value.replace("'", r"'\\\''") + "'"
 
 
 def _drawtext_filters(
--- a/tests/test_ffmpeg.py
+++ b/tests/test_ffmpeg.py
@@ -854,6 +854,12 @@ def test_contact_sheet_accepts_apostrophes_in_external_paths(
     tmp_path: Path,
     monkeypatch: pytest.MonkeyPatch,
 ) -> None:
+    system_font_path = _find_usable_font_file()
+    if system_font_path is None:
+        pytest.skip("requires a usable system font")
+    if not _contact_sheet._ffmpeg_supports_drawtext(ffmpeg_path()):
+        pytest.skip("requires ffmpeg drawtext support")
+
     external_path_directory = tmp_path / "wearer's assets"
     external_path_directory.mkdir()
     copied_frames: list[ExtractedFrame] = []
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging, with one thing you should know: I could not reproduce #215, and I went looking properly.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1124 passed, 6 skipped.

`fc-match` resolves DejaVuSans here and drawtext is available, so the fontfile branch does run on this machine, and my ffmpeg reports the same pinned build as yours (`n8.1.2-44-g7c533d0f86-20260816`). On `6d80961` the apostrophe path renders a sheet with `timestamps_burned=True`. Both the old escaping and yours produce a byte-identical 3339-byte sheet. So the difference between our hosts is not the ffmpeg build, and I cannot say which escaping is right in general. Yours does not regress anything here, and it fixes a real failure you can reproduce, which is enough to merge.

What I did find is that the test cannot tell either way, and that is worth more than the escaping question:

```
apostrophe path, real font    32f5a615b461ae08
clean path, real font         32f5a615b461ae08
MISSING font path             32f5a615b461ae08   <- same sheet
alternate face, clean path    a2a92d2640df9de1   <- font choice IS observable
```

When ffmpeg cannot open the `fontfile` it names, drawtext falls back to a default face and still reports `timestamps_burned=True`. So `assert sheet.timestamps_burned is True` holds with a font path that does not exist. On a host where the bad quoting silently substitutes instead of erroring, the strengthened test still passes.

So I pushed one test (551aa5d) that renders a face which is not the default (DejaVuSerif) from an apostrophe path and compares the output against the same font at a plain path. It skips unless the host can actually tell the two faces apart, which it checks by rendering first. Deleting the escaping entirely fails the new test and leaves the existing one green.

Your change to make the old test skip instead of silently passing is the right half of this. That was the part of #215 I would have missed.

Issue #205, opened 2026-08-27
CheckRunReport can hold two outcomes at once, or none, and an empty one reports 'measured'


kstonekuan:
### What happens

`CheckRunReport` (`src/hflow/app.py:503`) carries three optional fields where exactly one should
ever be set:

```python
check: RegisteredCheck
result: CheckResult | None = None
error: str | None = None
not_run: StepNotRun | None = None
```

`status` resolves them by precedence (`app.py:513-527`): `not_run` first, then `error`, then the
result's verdict, then `measured` as the fallback. Nothing prevents a contradictory combination,
and the fallback means the *absence* of information reports the most confident answer. Verified on
current main:

```python
CheckRunReport(check=rc, result=CheckResult(...), error="boom",
               not_run=SkippedByQuarantine(("quarantined:x",))).status
# -> CheckStatus.SKIPPED, with a result and an error silently outranked

CheckRunReport(check=rc).status
# -> CheckStatus.MEASURED, from a report holding nothing at all
```

### Why the second one matters more

An empty report is reachable. The runner constructs it and appends it to the
report before running anything (`app.py:2001-2002`):

```python
run = CheckRunReport(check=registered)
report.checks.append(run)
```

So between those two lines and the assignment that follows, the report says `measured`. Anything
that reads `report.checks` after an exception escapes that window sees a check claiming it
measured an episode it never opened. `measured` is also what lands in `check_runs.status`, so the
claim is durable.

### What to build

One field holding a variant, so the states are exclusive by construction and `status` becomes an
exhaustive match with no precedence and no fallback:

```python
CheckOutcome = Measured(CheckResult) | Errored(str) | NotRun(StepNotRun)
```

The runner then has to produce an outcome before the report is readable, which removes the empty
window. `StepNotRun` and its `SupersededByPipeline` /
`SkippedByQuarantine` members already exist and stay as they are; this wraps them rather than
replacing them.

This is the style guide's rule about putting recoverable cases in an explicit success-side
variant callers must handle, and its rule against contradictory field combinations standing in
for one state type.

### `EnrichmentRunReport` has the same two defects

`EnrichmentRunReport` (`src/hflow/app.py:531`) carries the same three optional fields and resolves
them the same way (`src/hflow/app.py:542-556`): `not_run` by match, then `error`, then `MEASURED`
as the fallback. Verified on current main:

```
CheckRunReport(check=...)                          -> measured
[...]

Issue #203, opened 2026-08-27
A transform refusal about the recording is filed as an infrastructure failure


kstonekuan:
### What happens

`classify_ingest_failure` (`src/hflow/ingest_ledger.py:80`) sorts an ingest failure into one of
three kinds. It returns `source-missing` for `SourceNotFound`, `source-unreadable` for any
`McapError`, and `infrastructure` for everything else.

Everything else includes every refusal the transform makes about the file's contents:

- non-conforming pass-through video (`format` is not h264, more than one picture per message, a
  keyframe without SPS and PPS, a channel starting mid-GOP)
- a raw-image schema (`sensor_msgs/msg/Image`, `foxglove.RawImage`)
- an unsupported compressed-image format
- a derived topic colliding with a source topic

Each of those is a `ValueError` naming exactly what is wrong with the recording, and each is
filed under a kind that says the problem is with the machine.

### Why it matters

The `infrastructure` kind is what a retry is for. A corrupt or unsupported recording will fail
identically on every retry, so an operator draining the ledger by kind retries a batch that
cannot succeed, and a converter author debugging their output is told the fault is on our side.

`docs/how-to/write-a-converter.md` (added in #202) currently routes around this by telling
converter authors to read `error_type` and `message` and ignore `failure_kind`. That is good
advice for today and a poor place to leave it: the column exists so nobody has to parse
messages.

### Why it is not just a missing branch

`classify_ingest_failure`'s own docstring says the classification is a heuristic, stored beside
the verbatim `error_type` and `message` so a wrong guess stays visible. That design is
deliberate and I think worth keeping. The question is whether the heuristic can be improved without
becoming a list of exception messages, which is the thing it was written to avoid.

Three shapes, and the choice is not obvious:

1. **A dedicated exception type for source-content refusals.** The transform raises something
   like `SourceNotConforming(ValueError)` for the refusals above, and the classifier tests for
   it the way it already tests for `SourceNotFound`. Most precise, and it means the transform
   states the classification at the point that knows it. Costs a new public exception name and
   touching every raise site.
2. **A fourth kind, `source-unsupported`,** distinguishing "this recording is not something v1
   can ingest" from "this recording is not readable at all". Useful to an operator, since the
   first is a converter bug and the second is a corrupt file. Needs the same reclassification
   work as (1) to route anything into it.
3. **Leave the kinds and document them.** Defensible, and it is where #202 left things, but it
   makes `failure_kind` a column nobody can act on.

[...]

PR #224, opened 2026-08-28
fix: validate transform config numeric settings

### LINKED ISSUE #220: TransformConfig accepts a bool crf and an out-of-range crf, and folds both into pipeline_version
### Current behavior

`TransformConfig` (`src/hflow/transform.py:114`) validates none of its numeric fields. Its
content hash becomes `pipeline_version`, and three of its fields are written into
`provenance/v1` (`src/hflow/transform.py:193-195`), so a nonsense value becomes part of an
episode's recorded identity.

Measured against current main (`6d80961`), transforming a one-second synthetic episode:

| Config | Result |
| --- | --- |
| `chunk_size_bytes=True` | accepted; a 1-byte chunk target, output grows from 15,578 to 39,925 bytes |
| `chunk_size_bytes=0` | refused: `chunk_size must be >= 1, got 0` |
| `chunk_size_bytes=-1` | refused: `chunk_size must be >= 1, got -1` |
| `crf=True` | `VideoEncodeError: ffmpeg encode failed (exit 234)` from libx264's expression parser |
| `crf=999` | accepted; transcodes, and `pipeline_version` becomes `5833eecf1054` instead of `9cd5d34f513d` |
| `gop_seconds=0.0` | accepted |
| `gop_seconds=-1.0` | accepted |
| `gop_seconds=float("nan")` | `ValueError: cannot convert float NaN to integer` |
| `compression="gzip"` | refused: `unknown compression 'gzip'; expected one of ['zstd', 'none']` |

Two things stand out.

`chunk_size_bytes=True` is the `bool` subclasses `int` hole this repo has now closed in four
other places (the API port, measurement values, interval bounds, and the video-measurement
settings in #199). Zero and negative are already refused, so the guard exists and `True` walks
past it as 1.

`crf=999` is outside H.264's 0 to 51 range. It transcodes, and it produces a different
`pipeline_version` from the default. I did not check what libx264 does with the value, so I do
not know whether the output bytes differ from a valid crf. Either way an episode records a
distinct pipeline identity for a setting that is not a setting.

The two `cannot convert float NaN to integer` and libx264 expression errors are the shape a
caller sees today: an error from three layers down that does not name the field they set.

### Pattern to copy

`src/hflow/_video_measurements/_field_guards.py` (`require_int`, `require_float`) was written for
this exact problem, and `src/hflow/checks.py`'s settings dataclasses show the call shape. Do not
import from that package: it is deliberately dependency-free so it can be extracted
(`src/hflow/_video_measurements/__init__.py:1-6`). Copy the two functions into core, or use the
inline form `catalog.py:398-410` uses.

`compression` shows the message style to match: name the field, the value, and the accepted set.

### Definition of done

1. `TransformConfig` refuses a `bool` for `crf`, `chunk_size_bytes`, and `gop_seconds`.
2. `crf` is bounded to 0 through 51 with a message naming the range.
3. `gop_seconds` must be positive and finite when it is not `None`.
4. `chunk_size_bytes` must be a positive int when it is not `None`. The existing
   `chunk_size must be >= 1` refusal stays where it is; this one fails earlier and names the
   config field.
5. Validation runs when the config is constructed, so the caller learns at the call site instead
   of partway through a transform.
6. **No output bytes change for any currently valid config.** `TRANSFORM_BEHAVIOR_VERSION`
   (`src/hflow/behavior.py`) is not bumped. Prove it: transform the same source before and after
   and compare the full canonical file's SHA-256, not just its size.
7. One test per refusal, each asserting the message names the field.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

`packages/hflow-server/tests`. `tests/test_identity_stability.py` is the file that fails first if
item 6 is broken, so watch it.

### Notes

`gop_preset` and `topic_groups` are out of scope. `gop_preset` is a `GopPreset` enum and already
constrained; `topic_groups` maps topic names to group names and has no numeric range to check.

anisayakmitra-in:
## Summary

- Reject booleans, invalid CRF values, non-positive or non-finite GOP durations, and non-positive chunk sizes when `TransformConfig` is constructed
- Keep all valid configuration values and transform behavior unchanged
- Add focused regression coverage for each invalid boundary

Closes #220

## Validation

- `uv run ruff check --fix`
- `uv run ruff format`
- `uv run ty check`
- `uv run pytest tests/test_transform.py -q` (18 passed)
- `uv run pytest -q` completed; 19 ffmpeg checks initially failed only because ffmpeg was absent from WSL. After installing the documented prerequisite, `uv run pytest --lf -q` passed all 19.

<sub>Body reformatted by a maintainer: the original was submitted with literal `` `n `` in place of newlines, which a squash-merge would have carried into main. Content unchanged.</sub>

### DIFF
--- a/src/hflow/transform.py
+++ b/src/hflow/transform.py
@@ -56,6 +56,7 @@ v1 behavior:
 import hashlib
 import json
 import logging
+import math
 from collections.abc import Callable, Mapping, Sequence
 from copy import copy
 from dataclasses import dataclass, field
@@ -132,6 +133,31 @@ class TransformConfig:
     # topic -> group name, beating the defaults; topics not listed go to
     # ``cameras`` by schema, ``bulk`` by mean message size, else ``state``.
     topic_groups: dict[str, str] = field(default_factory=dict)
+
+    def __post_init__(self) -> None:
+        """Reject invalid settings before they become a pipeline identity."""
+        if isinstance(self.crf, bool) or not isinstance(self.crf, int):
+            raise ValueError(f"crf must be an int, got {type(self.crf).__name__}")
+        if not 0 <= self.crf <= 51:
+            raise ValueError(f"crf must be between 0 and 51, got {self.crf}")
+
+        if self.gop_seconds is not None:
+            if isinstance(self.gop_seconds, bool) or not isinstance(self.gop_seconds, int | float):
+                raise ValueError(
+                    f"gop_seconds must be an int or float, got {type(self.gop_seconds).__name__}"
+                )
+            if not math.isfinite(self.gop_seconds) or self.gop_seconds <= 0:
+                raise ValueError(f"gop_seconds must be positive and finite, got {self.gop_seconds}")
+
+        if self.chunk_size_bytes is not None:
+            if isinstance(self.chunk_size_bytes, bool) or not isinstance(
+                self.chunk_size_bytes, int
+            ):
+                raise ValueError(
+                    f"chunk_size_bytes must be an int, got {type(self.chunk_size_bytes).__name__}"
+                )
+            if self.chunk_size_bytes <= 0:
+                raise ValueError(f"chunk_size_bytes must be positive, got {self.chunk_size_bytes}")
 
 
 @dataclass(frozen=True)
--- a/tests/test_transform.py
+++ b/tests/test_transform.py
@@ -1,5 +1,6 @@
 """Transform: synthetic input MCAP -> canonical episode."""
 
+from collections.abc import Callable
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging. Every refusal the issue asked for is here, at construction, with the field named.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1135 passed, 6 skipped.

**DoD item 6 was the one the description did not report, so I ran it.** Full canonical-file SHA-256 on `de139b2` and on this branch, over a two-second camera episode, six configs including both new `crf` bounds:

| Config | pipeline_version | SHA (first 24) |
| --- | --- | --- |
| default | `9cd5d34f513d` | `39ceb3b756297a04e8332387` |
| `crf=0` | `83bae6656eb1` | `9ed2db25c148d8cc2ff9861f` |
| `crf=51` | `528524017340` | `051324a735bc22ab1b2b0f09` |
| `gop_seconds=2.5` | `130a7b5f8b90` | `b6c7f1fedd3eb1b567bbaf6e` |
| `chunk_size_bytes=1` | `352341617209` | `6ec4ac775ff785391b87c8b4` |
| `chunk_size_bytes=800_000` | `b54bb2a347dd` | `f75c30b8331b6dd2f2f4409c` |

Identical on both trees. Worth doing rather than assuming, because a `__post_init__` on the dataclass whose content hash *is* `pipeline_version` is exactly the change that could move an identity without anyone noticing.

One thing you did that I want to name: `crf` refuses a `float` while `gop_seconds` accepts one. That reads as an inconsistency and it is not, because it follows the declared annotations (`crf: int`, `gop_seconds: float | None`). `TransformConfig(crf=23.0)` used to work and now raises, which is the right call for a field that ends up on an ffmpeg command line.

I reformatted your PR body. It arrived with literal `` `n `` instead of newlines, which is what PowerShell does to a `\n` in a double-quoted string, and a squash-merge would have carried it into main's history. If you are on `gh pr create` under PowerShell, `--body-file` avoids it. Content unchanged.

Since this is your second contribution: the `good first issue` pool is best left for people arriving, and you are past needing it. Higher-leverage work, if you want it, is running HFlow against a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face are good targets) and reporting what breaks or what is slow. @tunglambk did exactly that this week and it produced two better issues than anything I found by reading code. The non-starter backlog is also open: #203 and #205 are both unassigned.

PR #226, opened 2026-08-28
fix(testing): refuse invalid SyntheticEpisodeSpec in synthesize_episode

Soham-47:
## Summary

- `synthesize_episode` refuses bad `SyntheticEpisodeSpec` values before it writes an episode.
- It refuses non-positive `duration_s`, `image_hz`, `joint_hz`, image dimensions, and `joint_count`.
- It refuses bools and non-finite floats.
- The `ValueError` names the field.
- It also refuses `black_segment`, `joint_freeze_segment`, and `timestamp_offset_segment` past `duration_s`.
- Default fault segments past a short `duration_s` still write, so current call sites keep working.
- Valid specs and episode contents stay the same.

## Why

- A typo in duration or rate wrote a syntactically valid episode with no messages.
- Later steps then failed with a weak error about missing data.
- `VideoEpisodeSpec` already refuses bad values at write time.
- This change applies the same rule to `synthesize_episode`.
- Checks run in `synthesize_episode`, not in `SyntheticEpisodeSpec.__post_init__`.
- Odd image dimensions stay allowed. The synthetic path does not need the yuv420p even-size rule.

## Validation

- `uv run ruff check --fix`
- `uv run ruff format`
- `uv run ty check`
- `uv run pytest -q` (1118 passed, 6 skipped)

## Checklist

### DIFF
--- a/src/hflow/testing.py
+++ b/src/hflow/testing.py
@@ -409,6 +409,12 @@ def _validate_synthetic_episode_spec(spec: "SyntheticEpisodeSpec") -> None:
         segment = getattr(spec, segment_name)
         if segment is None:
             continue
+        # A segment still holding its class default is exempt when it runs past
+        # a shortened duration. The defaults are sized for duration_s=8.0, and
+        # most callers shorten the episode without clearing them, so enforcing
+        # this on defaults refuses specs that have always worked: removing this
+        # branch fails 78 tests and errors 112 more. An explicitly chosen
+        # segment is checked, which is the case the caller can act on.
         default = SyntheticEpisodeSpec.__dataclass_fields__[segment_name].default
         if segment == default and segment[1] > spec.duration_s:
             continue
--- a/src/hflow/testing.py
+++ b/src/hflow/testing.py
@@ -31,6 +31,9 @@ Implementation notes:
   ``MSG: pkg/msg/Type`` headers) so any ROS 2-aware reader decodes them.
 - Everything is deterministic: fixed ``start_time_ns``, seeded noise, no
   wall-clock reads.
+- ``_require_int`` / ``_require_float`` are copied here rather than imported
+  from ``hflow._video_measurements`` so that incubating package can stay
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
Merging. Clean implementation, and you were right to push back on the issue.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1153 passed, 6 skipped.

**My definition of done was wrong and you fixed it.** Item 4 said segments outside `duration_s` should be refused, full stop. It cannot be: `black_segment` and `timestamp_offset_segment` default to spans sized for `duration_s=8.0`, there are 87 construction sites, and a large share of them shorten the episode without clearing the defaults. Exempting a segment that still holds its default is the right call, and it is the part of this PR I would have got wrong.

I measured what that exemption is holding up. Deleting the three lines:

```
78 failed, 961 passed, 6 skipped, 112 errors
```

Nothing in the diff said so, and the refusal tests all still pass without it, so the next person to tidy that branch gets 190 failures and no explanation. One fixup pushed (d380921): a comment naming what it protects and the number, plus two tests. One asserts a shortened episode keeping its default segments still writes; the other asserts an explicitly chosen segment past the end is still refused, which is the line the exemption draws.

Two things I checked rather than assumed:

- The exemption compares the segment's value to the class default, so `SyntheticEpisodeSpec(duration_s=1.0, black_segment=(2.0, 3.0))` is also exempt even though it was chosen deliberately. I left that alone. It is behaviour that already worked, and closing it would mean tracking which fields were explicitly passed, which is more machinery than the case is worth.
- Odd image dimensions stay allowed, as the issue asked. Good that you read that note rather than copying `_validate_video_episode_spec` wholesale.

Copying `_require_int` and `_require_float` into `testing.py` with a module-docstring note explaining why is exactly right, and the note is better than the one I wrote in the issue.

Welcome, and thanks for a first contribution that argued with the spec. If you want another, the open list is here: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

PR #227, opened 2026-08-28
fix(ffmpeg): validate contact sheet sizing arguments

### LINKED ISSUE #221: contact_sheet validates columns and max_tiles but not tile_width, and a bool slips past all three
### Current behavior

`contact_sheet` (`src/hflow/ffmpeg/_contact_sheet.py:134`) checks two of its three sizing
arguments and skips the third. `columns` and `max_tiles` are guarded at
`src/hflow/ffmpeg/_contact_sheet.py:145-148`. `tile_width` goes straight into the filter string
at `src/hflow/ffmpeg/_contact_sheet.py:155`:

```python
filter_chain: list[str] = [f"scale={tile_width}:-1"]
```

Measured on current main (`95da1c2`), two 640x480 source frames at `columns=2`:

| Call | Result |
| --- | --- |
| `tile_width=320` (default) | 640x240 sheet, 3,339 bytes |
| `tile_width=0` | 1280x480 sheet, 5,172 bytes |
| `tile_width=-320` | 1280x480 sheet, 5,172 bytes |
| `tile_width=1` | 2x1 sheet, 204 bytes |

`scale=0:-1` and `scale=-320:-1` are both ffmpeg's "keep the source width", so asking for
zero-width or negative-width tiles returns tiles at full source size. The sheet is four times the
pixels the caller asked for, and nothing says so.

The `bool` case goes past the guards that do exist, because `bool` subclasses `int`:

| Call | Result |
| --- | --- |
| `columns=True` | passes `columns < 1`, then `RuntimeError: ffmpeg contact sheet failed` |
| `tile_width=True` | `RuntimeError: ffmpeg contact sheet failed` |
| `max_tiles=True` | accepted; keeps 1 of 2 frames |

`columns=True` fails because `src/hflow/ffmpeg/_contact_sheet.py:159` formats it into the filter
as `tile=Truex2`. The caller sees an ffmpeg error, not the argument they got wrong.

`max_tiles=True` is the least bad of the three: it reduces to 1, and `frames_sampled_from` on the
result still reports 2, so the drop is visible if the caller looks.

### Pattern to copy

The two guards already in the function at `src/hflow/ffmpeg/_contact_sheet.py:145-148` set the
message style: name the argument, the bound, and the value.

For the type half, `src/hflow/_video_measurements/_field_guards.py` has `require_int`, written for
this exact hole. Do not import it from there: that package is deliberately dependency-free so it
can be extracted (`src/hflow/_video_measurements/__init__.py:1-6`). Use the inline form
`src/hflow/catalog.py:398-410` uses, which is two lines.

### Definition of done

1. `tile_width` must be a positive int, refused with a message naming the argument and the value.
2. `columns`, `tile_width`, and `max_tiles` each refuse a `bool`.
3. The three refusals raise `ValueError` before ffmpeg runs, so no caller sees
   `ffmpeg contact sheet failed` for an argument problem.
4. Every currently valid call still produces the same sheet. `tile_width=320, columns=2` on two
   frames is still a 640x240, 3,339-byte JPEG.
5. One test per refusal.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

`packages/hflow-server/tests`. `tests/test_ffmpeg.py` holds the contact-sheet tests.

### Notes

Do not add a lower bound above 1 or an even-number rule for `tile_width`. I checked: odd widths
render fine here, and 1 produces a legal if useless sheet. Positive is the property that is
actually violated.

Iamnotphage:
## Summary

Make `contact_sheet` fail fast on invalid sizing arguments:

* require `tile_width` to be a positive integer
* reject booleans for `columns`, `tile_width`, and `max_tiles`
* add regression coverage proving that validation happens before FFmpeg is invoked

Valid inputs continue through the existing rendering path unchanged.

## Why

FFmpeg interprets zero and negative scale widths as “keep the source width.” As a result, an invalid `tile_width` can appear to succeed while quietly producing a much larger sheet than the caller requested.

Python’s `bool`-is-an-`int` relationship creates a second trap: `True` can leak into an FFmpeg filter or silently reduce `max_tiles` to one. Validating these arguments at the API boundary turns those surprises into immediate, actionable `ValueError`s without changing valid rendering behavior.

Closes #221

## Validation

* `uv sync --locked --all-extras` — resolved 63 packages; checked 60 packages
* `uv run ruff check --fix` — all checks passed
* `uv run ruff format` — 175 files left unchanged
* `uv run ty check` — all checks passed
* `uv run pytest tests/test_ffmpeg.py -q` — 62 passed, 1 skipped
* `uv run pytest -q` — 1,158 passed, 6 skipped

## Checklist

* [x] I added or updated outcome-focused tests for changed business logic.
* [x] Documentation was reviewed; no update is needed because this fix only changes invalid-argument handling.
* [x] I ran `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check`.
* [x] I ran the relevant pytest suite.
* [x] I did not add recordings, generated media, credentials, private URLs, or runtime artifacts.
* [x] I preserved stored-data compatibility.

### DIFF
--- a/src/hflow/ffmpeg/_contact_sheet.py
+++ b/src/hflow/ffmpeg/_contact_sheet.py
@@ -142,8 +142,16 @@ def contact_sheet(
     """Composite ``frames`` (from ``Episode.frames()``) into one JPEG grid."""
     if not frames:
         raise ValueError("contact_sheet needs at least one frame")
+    if not isinstance(columns, int) or isinstance(columns, bool):
+        raise ValueError(f"columns must be an int, got {columns!r}")
     if columns < 1:
         raise ValueError(f"columns must be >= 1, got {columns}")
+    if not isinstance(tile_width, int) or isinstance(tile_width, bool):
+        raise ValueError(f"tile_width must be an int, got {tile_width!r}")
+    if tile_width < 1:
+        raise ValueError(f"tile_width must be >= 1, got {tile_width}")
+    if not isinstance(max_tiles, int) or isinstance(max_tiles, bool):
+        raise ValueError(f"max_tiles must be an int, got {max_tiles!r}")
     if max_tiles < 1:
         raise ValueError(f"max_tiles must be >= 1, got {max_tiles}")
     selected_frames = [frames[index] for index in _evenly_sampled_indices(len(frames), max_tiles)]
--- a/tests/test_ffmpeg.py
+++ b/tests/test_ffmpeg.py
@@ -960,6 +960,54 @@ def test_contact_sheet_rejects_empty_input(tmp_path: Path) -> None:
         contact_sheet([], tmp_path / "never.jpg")
 
 
+def _refuse_ffmpeg_for_invalid_contact_sheet_arguments(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    monkeypatch.setattr(_contact_sheet, "_find_usable_font_file", lambda: None)
+
+    def fail_if_ffmpeg_runs(*_args: object, **_kwargs: object) -> None:
+        pytest.fail("invalid contact_sheet arguments must be refused before ffmpeg runs")
+
+    monkeypatch.setattr(_contact_sheet.subprocess, "run", fail_if_ffmpeg_runs)
+
+
+def _unreadable_frame(tmp_path: Path) -> ExtractedFrame:
+    return ExtractedFrame(path=tmp_path / "must-not-be-read.jpg", log_time_ns=0)
+
+
+@pytest.mark.parametrize("tile_width", [0, -320])
+def test_contact_sheet_rejects_non_positive_tile_width_before_ffmpeg(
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1158 passed, 6 skipped.

The tests are better than the issue asked for. Item 3 said the refusal must happen before ffmpeg runs, and rather than trusting that, you monkeypatched `subprocess.run` to fail the test if it is reached, and passed a frame path that does not exist. Both halves are load-bearing: the fake path means a refusal that slipped through could not accidentally succeed either.

Using `{columns!r}` in the type message so it reads `got True` is the right choice over the type name. `True` is what the caller wrote.

PR #228, opened 2026-08-28
test(checks): stop pinning a libm-dependent content digest

### LINKED ISSUE #209: A pinned content digest makes the test suite fail on any platform whose libm differs

chiruu12:
Closes #209.

`test_content_digest_emits_exactly_the_documented_measurements` pinned a digest that only that machine could produce. The joint payloads come from `math.sin` and `math.cos`, which call the platform's libm, so the literal was a property of the runner. That test now asserts shape alone, 64 lowercase hex characters, which is all it can honestly claim about an episode built from floats. The comment above it already said the value was runner dependent, so the two disagreed.

The pin moves rather than disappears, onto a fixture the test file writes itself. Every payload is an integer rendered as JSON, so the bytes are identical everywhere and the literal is a property of the content.

Not the shape I proposed. I said shape 2 in the issue and it does not hold up, because it cannot catch the mutation you asked to keep caught. Reversing the hex keeps every property shape 2 asserts: still 64 hex characters, still changes with the content, still stable across containers. Measured, with `_content_digest_value` returning `inter.digest_hex[::-1]`:

```
1 failed, 5 passed
FAILED TestContentDigestIsAPropertyOfTheContent::test_the_emitted_value_is_the_digest_and_not_a_transform_of_it
```

Only the literal catches it. So this is shape 2 for the claims plus shape 3's reasoning applied to one fixture instead of every synthetic one, which keeps the guarantee without re-minting the existing fixtures.

`TestContentDigestIsAPropertyOfTheContent` covers the two things the check's docstring actually promises:

- a changed message changes the digest
- chunk size, compression, and an extra metadata record do not, which is what the duplicate-hunt query rests on. The two files differ in size, asserted, so the container really varies.

Validation:

```
uv run pytest -q                                   1151 passed, 11 skipped
uv run --python 3.11 --locked pytest tests/test_default_checks.py -q   52 passed
uv run --python 3.14 --locked pytest tests/test_default_checks.py -q   52 passed
uv run ruff check                                  All checks passed
uv run ruff format --check                         175 files already formatted
uv run ty check                                    All checks passed
```

Run on macOS 26.5.1 arm64, the platform that could not pass this file before. CI covers Linux, and I have not run it there.

### DIFF
--- a/tests/test_default_checks.py
+++ b/tests/test_default_checks.py
@@ -1232,11 +1232,118 @@ def test_content_digest_emits_exactly_the_documented_measurements(
     assert run.result is not None
     measurements = dict(run.result.measurements)
     assert set(measurements) == {"content_digest"}
-    digest = measurements["content_digest"]
-    # Pin the exact digest for the joints-only source episode. The synthetic
-    # writer is deterministic, so a dispatcher that returns a transformed
-    # value (e.g. reversed hex) will not match this literal.
-    assert digest == "1ee3e9bcad4d5a12112b196f7a3d4e272329cc41a5b5efe7e8751339ab1a4722"
+    digest = str(measurements["content_digest"])
+    # Shape only. This episode's joint payloads come from math.sin and math.cos
+    # (hflow.testing), which call the platform's libm, and libm is not
+    # bit-identical across platforms, so its digest is a property of the
+    # machine that produced it. The value is pinned instead against a fixture
+    # this file writes itself, in TestContentDigestIsAPropertyOfTheContent.
+    assert len(digest) == 64
+    assert set(digest) <= set("0123456789abcdef")
+
+
+class TestContentDigestIsAPropertyOfTheContent:
+    """What the digest promises, tested without depending on libm.
+
+    The check's docstring makes two claims: the digest identifies the recorded
+    content, and it is independent of container layout, so two files differing
+    only in chunking, compression, or metadata digest the same. Both are tested
+    here against episodes this class writes itself.
+
+    Writing them here is the point. The synthetic fixtures generate joint
+    values with math.sin and math.cos, which call the platform's libm, and libm
+    is not bit-identical across platforms, so any digest of theirs belongs to
+    the machine that computed it. Every payload below is an integer rendered as
+    JSON, so the bytes are the same everywhere and a literal can be pinned.
+    """
+
+    TOPIC = "/joints"
+    FIRST_LOG_TIME_NS = 1_000_000
+    PAYLOADS = (b'{"v":0}', b'{"v":1}', b'{"v":2}')
+    # Minted from the fixture below. No floating point reaches these bytes, so
+    # this is a property of the payloads and not of the platform: the same
+    # value comes back on Linux and on macOS.
+    PINNED_DIGEST = "61bd180b08e0c3f00713e60c1d54e3d2f048f4c569ee86d073fa6745b467b05a"
+
+    @classmethod
+    def _write(
+        cls,
+        path: Path,
+        *,
+        payloads: Sequence[bytes] | None = None,
+        chunk_size: int = 1 << 20,
+        compress: bool = True,
+        metadata: bool = False,
+    ) -> Path:
+        """One episode, with the container knobs the docstring says do not count."""
+        from mcap.writer import CompressionType, Writer
+
+        with path.open("wb") as handle:
+            writer = Writer(
+                handle,
+                chunk_size=chunk_size,
+                compression=CompressionType.ZSTD if compress else CompressionType.NONE,
+            )
+            writer.start()
+            if metadata:
+                writer.add_metadata("recording", {"operator": "someone"})
+            schema_id = writer.register_schema(
+                name="std_msgs/msg/String", encoding="jsonschema", data=b'{"type": "object"}'
+            )
+            channel_id = writer.register_channel(
+                topic=cls.TOPIC, message_encoding="json", schema_id=schema_id
+            )
+            for index, payload in enumerate(cls.PAYLOADS if payloads is None else payloads):
+                log_time = cls.FIRST_LOG_TIME_NS + index
+                writer.add_message(
+                    channel_id=channel_id, log_time=log_time, data=payload, publish_time=log_time
+                )
+            writer.finish()
+        return path
+
+    @staticmethod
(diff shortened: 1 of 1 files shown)


kstonekuan (approved):
Merging. This is better than the shape I specified, and the reason is worth writing down.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1156 passed, 6 skipped.

Shape 2 in the issue traded away the literal, and I accepted that trade because I could not see how to keep it. Your fixture keeps it: every payload is an integer rendered as JSON, so no floating point reaches the bytes, and the digest of that episode is a property of the payloads rather than of the machine. So the one thing only a literal can catch, a dispatcher returning a transform of the digest instead of the digest, stays pinned. Your comment says exactly that, which is the part I would have lost.

**The mutation check, run rather than taken on trust.** I changed `_content_digest_value` (`checks.py:899`) to `return inter.digest_hex[::-1]` and ran the whole suite:

```
1 failed, 1155 passed, 6 skipped
FAILED TestContentDigestIsAPropertyOfTheContent::test_the_emitted_value_is_the_digest_and_not_a_transform_of_it
```

One test, the intended one, and nothing else. The reversed hex still passes the shape assertions, still changes with content, and still holds across containers, so that literal is doing work the other three tests cannot do.

The container-invariance test is the one I would have written last and it is the one that matters most, since a duplicate hunt rests on it. Varying chunk size, compression, and a metadata record together, then asserting the file sizes differ before asserting the digests match, means it cannot pass by accident on two identical files.

On the platform question: I can only confirm the Linux half here, and the literal reproduces. Your macOS run is the other half and I am taking that from you. The argument that it holds anywhere is the integers-only payload, which is checkable by reading the fixture, so nobody has to re-run it to trust it.

Thanks for the three data points on #209 as well. Separating the platform from the Python version was the measurement that closed it.

Issue #225, opened 2026-08-28
step_kind and transform_kind are finite string sets typed as str, one beside an enum that already exists


kstonekuan:
### Current behavior

Two finite sets of strings in `app.py` are typed as `str` and spelled as literals at every site.

**`step_kind`.** `_STEP_EXAMPLE_BY_KIND` (`src/hflow/app.py:595-600`) is keyed by three literals:

```python
_STEP_EXAMPLE_BY_KIND = {
    "check": ("hflow.CheckResult", "check"),
    "enrichment": ("hflow.EnrichmentResult", "enrichment"),
    "derived channel": ("hflow.DerivedSeries", "series"),
}
```

`_raise_if_step_cannot_take_only_an_episode` takes `step_kind: str`
(`src/hflow/app.py:606`) and indexes that dict at `src/hflow/app.py:618`. The three callers pass
literals: `src/hflow/app.py:1385`, `:1439`, `:1477`. A misspelled kind is a `KeyError` raised
while building an error message, so a registration mistake would surface as a crash in the
diagnostic rather than the diagnostic.

`StepKind` (`src/hflow/manifest.py:33`) already exists, is exported from `hflow`, and holds
`CHECK = "check"` and `ENRICHMENT = "enrichment"`. Two of the three keys are values of an enum
this repo already owns; the strings are just not connected to it.

**`transform_kind`.** `_SyncCompletion.transform_kind` is `str | None`
(`src/hflow/app.py:266`). It is written as `"override" if self.transform_override is not None
else "default"` (`src/hflow/app.py:1937-1939`) and read back by comparing against a literal
(`src/hflow/app.py:1124`):

```python
if completion.transform_kind != "default":
    return None
```

Nothing else in the tree reads it. A typo on either side compiles, passes the type checker, and
silently disables canonical-episode reuse: the comparison fails, the function returns `None`, and
every re-ingest re-runs the transform. There is no error, only work.

### What to build

Give each set a type, and keep `StepKind` doing the one job it already does.

For `transform_kind`, a small `StrEnum` next to `_SyncCompletion` with `DEFAULT = "default"` and
`OVERRIDE = "override"`.

For `step_kind`, add a private enum rather than widening `StepKind`. `StepKind` is the serialized
vocabulary of `PipelineManifest` (`PIPELINE_MANIFEST_VERSION = 3`, `src/hflow/manifest.py:30`),
and a derived channel is not a manifest step kind. Adding a member to a versioned, serialized enum
[...]

Issue #223, opened 2026-08-28
trajectory_metrics has a guard that cannot be false, hiding a conditional measurement key


kstonekuan:
### Current behavior

`trajectory_metrics` (`src/hflow/checks.py:1022`) emits `final_pose_speed` unconditionally and
`final_pose_unsettled_ratio` behind a guard (`src/hflow/checks.py:1105-1107`):

```python
measurements[f"{topic}/final_pose_speed"] = final_speed
if isinstance(mean_speed, float) and mean_speed > 0:
    measurements[f"{topic}/final_pose_unsettled_ratio"] = final_speed / mean_speed
```

Two separate things are wrong with those two lines.

**The `isinstance` half can never be False.** `mean_speed` is read from the measurements dict it
was written to 26 lines earlier, at `src/hflow/checks.py:1079`:

```python
measurements[f"{topic}/mean_velocity"] = float(
    np.sum(measured_speeds * measured_durations_s) / valid_velocity_s
)
```

That `float(...)` is unconditional, and nothing reassigns the key in between. Verified on current
main (`95da1c2`) over two synthetic episodes: `mean_velocity` came back `type=float` in both.

**The `> 0` half is real, and the omission it causes is undocumented.** A motionless episode
gives `mean_velocity == 0.0`, and the ratio key disappears while `final_pose_speed` stays:

```
normal motion       mean_velocity=2.114332923731631   final_pose_speed=True  unsettled_ratio=True
frozen throughout   mean_velocity=0.0                 final_pose_speed=True  unsettled_ratio=False
```

(The second is `SyntheticEpisodeSpec(duration_s=3.0, cameras=(), joint_jump_at_s=None,
joint_freeze_segment=(0.0, 3.0))`.)

So the key set this check emits depends on the data, nothing in the docstring says so, and a
curation query joining on `final_pose_unsettled_ratio` silently loses exactly the episodes where
the arm never moved. That may be the group someone is looking for.

### What to build

Delete the dead half, keep the real one, and say what it does.

The condition that matters is a zero denominator. Write it so a reader can see that is the whole
reason, and record the choice in the docstring: `final_pose_unsettled_ratio` is emitted only for
an episode whose mean velocity is above zero, because the ratio is undefined otherwise.

`trajectory_metrics` is not one of the six defaults, so it has no entry in `_DEFAULT_KEY_FACTS`
and there is no key-prediction fact to keep in sync. Do not add one; that machinery is for the
automatic defaults.

[...]

Issue #11, opened 2026-08-19
Docs: docstrings for `CanonicalMcapWriter` public methods and `DiagnosticLevel`


kstonekuan:
## What's missing

`CanonicalMcapWriter` is exported in `hflow.__all__`, but its public methods have no docstrings, the only exported class in the API with that gap:

- `register_schema` (`src/hflow/mcap_writer.py:206`)
- `register_channel` (`:217`)
- `write_message` (`:251`)
- `add_metadata` (`:287`)
- `add_attachment` (`:298`)
- `finish` (`:332`)

Also: `DiagnosticLevel` (`src/hflow/doctor.py:39`) is the only exported class with no class docstring (its members are commented inline instead).

## Proposed change

Add short docstrings stating each method's **contract**, not its implementation:

- what `register_schema` / `register_channel` return and how the IDs are used;
- what `write_message` requires (registered channel, timestamp semantics);
- the ordering rule around `finish()`: it must be called last, and the writer raises on use afterwards (see the guard near `src/hflow/mcap_writer.py:495`);
- a one-line class docstring for `DiagnosticLevel`.

Follow the existing docstring style in the codebase (see `ChannelData` properties in `src/hflow/episode.py` for tone: one line of contract, notable caveats only).

## Definition of done

- Every public method of `CanonicalMcapWriter` and the `DiagnosticLevel` class have docstrings.
- No behavior changes.

## Validation

```bash
uv run ruff check --fix && uv run ruff format
uv run ty check
```

PR #229, opened 2026-08-28
fix(steps): reject infinite and bool threshold values at construction

### LINKED ISSUE #222: Threshold accepts an infinite value, so a gate can be stuck open or shut
### Current behavior

`steps.py` already refuses the two ways a gate can be vacuous, and states why each time.

`Threshold.__post_init__` (`src/hflow/steps.py:230`) refuses NaN:

> every comparison against NaN is False, so this gate would reject every episode it evaluated

`Gate.__post_init__` (`src/hflow/steps.py:268`) refuses an empty threshold tuple:

> holds no thresholds, so it would accept every episode without reading a measurement, a verdict
> claiming a check passed when nothing was checked

An infinite `value` produces both of those outcomes and is accepted. Verified on current main
(`95da1c2`): `hflow.Threshold("a/b", hflow.Comparison.AT_MOST, float("inf"))` constructs without
complaint.

`Threshold.holds` is `measurement <= value` or `measurement >= value`
(`src/hflow/steps.py:237-241`), so against any finite measurement:

| Threshold | Every episode |
| --- | --- |
| `AT_MOST, +inf` | accepted |
| `AT_LEAST, -inf` | accepted |
| `AT_LEAST, +inf` | rejected |
| `AT_MOST, -inf` | rejected |

The first two are the thing `Gate(accept_when=())` exists to refuse. The last two are the thing
the NaN branch exists to refuse. The same file already argues that both are worth stopping at
construction.

`Threshold(value=True)` is also accepted, which is the `bool` subclasses `int` hole. `True`
becomes the threshold `1.0` and `False` becomes `0.0`, so a typo'd gate silently compares against
a number the author did not write.

### Why it is worth stopping at construction

A gate holding an infinite threshold produces a verdict on every episode, and that verdict is
recorded in `check_runs.status`. Nothing downstream can tell a real `passed` from one that could
not have been anything else. That is the argument the NaN message already makes, applied to the
value next to it.

### Pattern to copy

The two existing refusals in the same file. Match their shape: state what the gate would do, then
why that is not a verdict. `src/hflow/steps.py:230-235` is the closest sibling, since it is on the
same field.

For the `bool` half, use the inline form at `src/hflow/catalog.py:398-410`.

### Definition of done

1. `Threshold` refuses a non-finite `value`, positive or negative infinity, with a message saying
   which way the gate would be stuck.
2. `Threshold` refuses a `bool` `value`.
3. The existing NaN and empty-gate refusals are unchanged, including their messages. Their tests
   still pass untouched.
4. `Threshold` still accepts an `int` value: `Threshold("a/b", Comparison.AT_MOST, 0)` is a
   legitimate gate and `0` must keep working.
5. One test per new refusal, asserting the message names the field.
6. `hflow.checks`' shipped recommended gates still construct. Import them and check, since a
   bound of `0` or a large finite number is normal there.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

`packages/hflow-server/tests`. `tests/test_gates.py` holds the gate behavior tests.

### Notes

Scope is `Threshold.value` only. `Aggregation` and `Comparison` are enums and already constrained,
and `key_pattern` has its own refusal at `src/hflow/steps.py:224`.

Newer1107:
## What

Reject `float('inf')`, `float('-inf')`, `True`, and `False` as `Threshold.value` at construction time, following the existing NaN rejection pattern.

## Why

An infinite threshold accepts or rejects every episode unconditionally — the same as not checking at all. The existing NaN check already argues why a vacuous verdict is worth refusing at construction; infinity produces the same outcomes.

`bool` subclasses `int` in Python, so `Threshold(v, Comparison.AT_MOST, True)` silently compares against `1.0`, not the meaning the author wrote.

## How

Add two checks to `Threshold.__post_init__` in `src/hflow/steps.py`:

1. `isinstance(self.value, bool)` — checked before `math.isnan` because `bool` is a `float` subclass
2. `math.isinf(self.value)` — checked after NaN, matching the existing pattern

Each raises `ValueError` with a message matching the project style.

## Tests

Four new parametrized test cases in `tests/test_gates.py`:

- `test_an_infinite_threshold_is_refused_at_construction` — `+inf` and `-inf`
- `test_a_bool_threshold_is_refused_at_construction` — `True` and `False`

Full suite: **947 passed, 6 skipped**.

Closes #222

### DIFF
--- a/src/hflow/steps.py
+++ b/src/hflow/steps.py
@@ -227,11 +227,23 @@ class Threshold:
                 "threshold key_pattern must not be empty; it is a glob over "
                 "measurement keys, e.g. '*/black_frame_pct'"
             )
+        if isinstance(self.value, bool):
+            raise ValueError(
+                f"threshold value for {self.key_pattern!r} must be a number, not "
+                f"{self.value!r}: bool is accepted by float but silently compares "
+                "against 0.0 or 1.0, not the meaning the author wrote"
+            )
         if math.isnan(self.value):
             raise ValueError(
                 f"threshold value for {self.key_pattern!r} must be a number, not NaN: "
                 "every comparison against NaN is False, so this gate would reject "
                 "every episode it evaluated"
+            )
+        if math.isinf(self.value):
+            raise ValueError(
+                f"threshold value for {self.key_pattern!r} must be finite, not "
+                f"{self.value!r}: an infinite threshold accepts or rejects every "
+                "episode unconditionally, which is the same as not checking at all"
             )
 
     def holds(self, measurement: float) -> bool:
--- a/tests/test_gates.py
+++ b/tests/test_gates.py
@@ -250,3 +250,31 @@ def test_an_empty_gate_is_refused_at_construction() -> None:
 def test_a_nan_threshold_is_refused_at_construction() -> None:
     with pytest.raises(ValueError, match="not NaN"):
         hflow.Threshold("v", hflow.Comparison.AT_MOST, math.nan)
+
+
+@pytest.mark.parametrize(
+    "value,desc",
+    [
+        pytest.param(float("inf"), "+inf", id="positive_inf"),
+        pytest.param(float("-inf"), "-inf", id="negative_inf"),
+    ],
+)
+def test_an_infinite_threshold_is_refused_at_construction(
+    value: float, desc: str
+) -> None:
+    with pytest.raises(ValueError, match="must be finite"):
+        hflow.Threshold("v", hflow.Comparison.AT_MOST, value)
+
+
+@pytest.mark.parametrize(
+    "value,desc",
+    [
+        pytest.param(True, "True", id="true"),
+        pytest.param(False, "False", id="false"),
+    ],
+)
+def test_a_bool_threshold_is_refused_at_construction(
+    value: bool, desc: str
+) -> None:
+    with pytest.raises(ValueError, match="not True|not False"):
+        hflow.Threshold("v", hflow.Comparison.AT_MOST, value)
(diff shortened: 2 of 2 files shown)


kstonekuan:
Thank you for this, and I am sorry to close good work.

#222 was assigned to @victorwon2001 at 09:33 today, 26 minutes before this PR opened. They claimed it at 08:16 with the scope and their local validation numbers, and assignment is the only thing that reserves an issue here, so it is theirs. Timing, not quality.

Your implementation is right, for what it is worth. Ordering the `bool` check before the numeric ones is correct and is the part people get wrong, and both messages follow the house style of saying what the gate would do rather than just naming the rule.

Two things worth taking away:

**Check the assignee before starting.** The issue list shows it, and an assigned issue is taken. Everything without an assignee is fair game, and a linked PR alone does not reserve anything.

**Your PR body reports 947 passed.** Current main is at 1,161. That gap says your branch is behind main or the run covered a subset, and either way it means the suite you validated against is not the one CI will run. `git fetch origin && git rebase origin/main`, then `uv run pytest -q` from the repo root. `pytest tests` alone misses `packages/hflow-server/tests`, which is a common way to land short.

If @victorwon2001's PR has not appeared in the next day or so I will come back here and ask you to reopen this, rather than letting the work go to waste.

Free right now, all unassigned:

- **#223** `trajectory_metrics` has a guard that can never be false, next to one that silently omits a measurement key. Smallest of these.
- **#225** two finite string sets typed as `str`, one sitting beside an enum the repo already has.
- **#10** and **#11**, both documentation.

kstonekuan:
Closing the loop as promised: @victorwon2001 sent #232 and it is merged, so #222 is done and there is nothing to reopen here.

Their version reached something neither of us had: `math.isnan` converts to float, so a large enough int threshold raised `OverflowError` instead of constructing. `Threshold("v", AT_MOST, 2**1024)` was broken on main and works now. Worth a look if you want to see what the issue missed.

Your read of the ordering problem was right and the pool has moved on since: #237, #239, and #247 are all unassigned.

PR #230, opened 2026-08-28
feat(lerobot): generalize converter to LeRobot Dataset v3 metadata (#189)

ayam04:
Closes #189

## What changed

`examples/lerobot/prepare.py` no longer encodes PushT assumptions. It reads the v3 repository metadata (`meta/info.json` + episode index parquet) and derives everything from it:

- **Feature schema discovery**: video features found by `dtype: video`; state/action schemas derived from declared dtype + shape (fail-loud listing feature name/dtype/shape for `float16`, multi-dim, or non-float)
- **Multi-camera**: `--camera-key` accepts a comma-separated list; each camera becomes its own `foxglove.CompressedVideo` channel (topic `/{camera_key}`, frame_id = camera key)
- **Episode boundaries**: data windows from `dataset_from_index/to_index` against the parquet `index` column; video windows from `videos/<camera>/from_timestamp|to_timestamp`
- **Metadata**: immutable source commit resolved via the HF API and stamped; `source_dataset`/`source_revision`/`source_episode_index`/`task`/`embodiment` (`robot_type`) preserved on every output episode
- **CDR**: `_encode_cdr_float32_array` unchanged byte-for-byte for float32[N] (XCDR1 `00 01 00 00`); existing `lerobot_to_mcap` callable + PushT CLI defaults preserved
- No LeRobot-specific deps added to the core package; work stays in `examples/lerobot/`

## Verification (all run locally)

- `uv sync --locked --all-extras`: OK
- `uv run ruff check`: all checks passed
- `uv run ruff format`: clean
- `uv run ty check`: all checks passed
- `uv run pytest -q`: 1166 passed, 6 skipped (includes 5 new outcome-focused fixture tests)

New fixture tests (`tests/test_lerobot_converter.py`): schema discovery, multi-camera selection, non-2D numeric vectors, episode boundary windows, unsupported-shape fail-loud, camera-key validation, CDR byte shape. No network, no third-party internals asserted.

## Pinned dataset commands + doctor

PushT episode 0 (regression):

```
uv run python examples/lerobot/prepare.py --repo lerobot/pusht --revision main --output-dir ./data/lerobot_pusht --camera-key observation.image --episode-index 0
```

```
doctor: data/lerobot_pusht/landing/lerobot_episode_0001.mcap
  conforming: no findings
  verdict: CONFORMING
```

Acceptance corpus (multi-camera, 6-dim):

```
[...]

PR #231, opened 2026-08-28
fix(checks): remove dead type check in trajectory_metrics and document conditional key

### LINKED ISSUE #223: trajectory_metrics has a guard that cannot be false, hiding a conditional measurement key

NiranjanKumar001:
## What this PR does

Fixes a dead code issue in `trajectory_metrics` inside `src/hflow/checks.py` and adds a missing explanation in the docstring about a measurement key that is not always emitted.

## The problem

In `trajectory_metrics`, there was this condition:

```python
if isinstance(mean_speed, float) and mean_speed > 0:
    measurements[".../final_pose_unsettled_ratio"] = final_speed / mean_speed
```

The `isinstance(mean_speed, float)` check was **always True** — it could never be False. `mean_velocity` is computed and stored as `float(...)` just 26 lines earlier in the same function, so by the time we read it back, it is always a float. Having a dead check like this confuses anyone reading the code, and the type checker also warns because reading from the measurements dict gives a wider type than `float`.

The second part — `mean_speed > 0` — is the **real and necessary check**: when a robot arm is completely frozen throughout a recording, `mean_velocity` is `0.0`, and dividing by it would be undefined. So this part correctly prevents the `final_pose_unsettled_ratio` key from being emitted for motionless episodes.

The problem is this behavior was **completely undocumented**. If you wrote a curation SQL query joining on `final_pose_unsettled_ratio`, you would silently lose all motionless episodes from your results — exactly the group you might be looking for.

## What was changed

1. **Removed the dead `isinstance` check** — the condition now simply reads `if mean_velocity > 0:`, which is what it always meant.
2. **Stored `mean_velocity` in a named local variable** instead of reading it back from the measurements dict. This makes the type clear, fixes the type checker warning, and removes the unnecessary dict lookup.
3. **Updated the docstring** to clearly explain that `final_pose_unsettled_ratio` is only emitted when `mean_velocity > 0`, because the ratio is mathematically undefined when the arm never moved. Curation authors are now warned about this.
4. **Added two tests** that prove both cases:
   - A moving robot episode → both `final_pose_speed` and `final_pose_unsettled_ratio` are emitted
   - A fully frozen robot episode (`joint_freeze_segment=(0.0, 3.0)`) → only `final_pose_speed` is emitted, ratio is absent

## No behavior changes

The values emitted for any episode that already produced the ratio are **100% identical**. Nothing in the computation changed — only the dead code was removed and the variable was given a clearer name.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix   # All checks passed
uv run ruff format        # 175 files left unchanged
uv run ty check           # All checks passed
uv run pytest -q          # 1163 passed, 6 skipped
```

Fixes #223

### DIFF
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -1042,6 +1042,13 @@ def trajectory_metrics(
     dividing by it would inflate its own sensor jitter into apparent motion --
     inverting the metric on exactly the episodes worth catching.
 
+    ``{topic}/final_pose_unsettled_ratio`` is emitted only when
+    ``{topic}/mean_velocity`` is above zero. For a fully motionless episode the
+    ratio is undefined (zero denominator), so only ``{topic}/final_pose_speed``
+    appears. Curation queries joining on ``final_pose_unsettled_ratio`` will
+    silently exclude motionless episodes; use ``final_pose_speed`` when you need
+    all episodes regardless of mean motion.
+
     Evidence only, and this one ships no recommended gate at all: these are
     motion-smoothness metrics, which are known to invert on real defects
     (Voxel51's audit scored an early-gripper-release defect better than clean
@@ -1076,9 +1083,8 @@ def trajectory_metrics(
     valid_velocity_s = float(np.sum(measured_durations_s))
     measurements[f"{topic}/valid_velocity_s"] = valid_velocity_s
     measurements[f"{topic}/peak_velocity"] = float(np.max(measured_speeds))
-    measurements[f"{topic}/mean_velocity"] = float(
-        np.sum(measured_speeds * measured_durations_s) / valid_velocity_s
-    )
+    mean_velocity = float(np.sum(measured_speeds * measured_durations_s) / valid_velocity_s)
+    measurements[f"{topic}/mean_velocity"] = mean_velocity
     if valid_velocity_s > 0:
         motionless_s = float(
             np.sum(measured_durations_s[measured_speeds < motionless_speed_epsilon])
@@ -1100,11 +1106,12 @@ def trajectory_metrics(
         (profile.stamps_ns[-1] - profile.stamps_ns[:-1]) / 1e9 <= final_pose_window_s
     )
     if np.any(final_window_mask):
-        mean_speed = measurements[f"{topic}/mean_velocity"]
         final_speed = float(np.mean(profile.speeds[final_window_mask]))
         measurements[f"{topic}/final_pose_speed"] = final_speed
-        if isinstance(mean_speed, float) and mean_speed > 0:
-            measurements[f"{topic}/final_pose_unsettled_ratio"] = final_speed / mean_speed
+        if mean_velocity > 0:
+            # Omitted when mean_velocity == 0.0: the ratio is undefined for a
+            # fully motionless episode.
+            measurements[f"{topic}/final_pose_unsettled_ratio"] = final_speed / mean_velocity
     return CheckResult(measurements=measurements)
 
 
--- a/tests/test_checks.py
+++ b/tests/test_checks.py
@@ -761,3 +761,44 @@ def test_trajectory_change_threshold_uses_a_true_weighted_median(tmp_path: Path)
     assert _duration_weighted_median(values, weights) == 2.0
     # Weight concentrated on the smallest value moves the median onto it.
     assert _duration_weighted_median(values, np.array([10.0, 1.0, 1.0])) == 1.0
+
+
+def test_trajectory_metrics_emits_unsettled_ratio_for_a_moving_episode(
+    tmp_path: Path,
+) -> None:
+    """A moving episode has mean_velocity > 0, so the unsettled ratio is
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging. Both halves of #223 done, and you improved on the shape I described.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1163 passed, 6 skipped.

I said delete the `isinstance` and keep the `> 0`. You did better: keeping `mean_velocity` as a local and using it directly removes the dict round-trip that made a type check look necessary in the first place. The dead guard existed because the value was written to `measurements` and then read back as a `MeasurementValue`. Take away the round-trip and there is nothing left to guard.

The docstring paragraph goes further than the issue asked, and the extra sentence is the useful one: naming `final_pose_speed` as what to use when you want every episode turns a caveat into an instruction. That is the sentence a curation query author needs.

Both tests reach the branch through a real episode rather than by calling the helper with hand-built measurements, which was the part of item 3 worth doing carefully.

One note: your branch is based on main before #230 landed, so you validated against 1,161 tests rather than the current 1,166. No overlap between the two changes, so it merges clean and I re-run the full gate on the merged result. Rebasing before opening keeps your number matching CI's.

Since this is your second contribution: the `good first issue` pool is best kept for people arriving, and you are past needing it. Higher-leverage work if you want it is running HFlow against a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face) and reporting what breaks or what is slow. @tunglambk did that this week and it produced two better issues than anything I found by reading code. The non-starter backlog is open too: #203 and #205 are both unassigned.

kstonekuan:
Approved above, but this is still marked draft so I cannot merge it. Mark it ready for review and I will land it.

Nothing is blocking on my side: full gate is green and both tests do what #223 asked. If the draft flag was deliberate and you have more coming, say what and I will hold.

NiranjanKumar001:
You guys are fast , doing great job. thanks for the instant review @kstonekuan .

PR #232, opened 2026-08-28
fix(steps): reject non-finite and boolean threshold values

### LINKED ISSUE #222: Threshold accepts an infinite value, so a gate can be stuck open or shut

victorwon2001:
## Summary

- reject boolean threshold values before numeric validation
- reject positive and negative infinity with outcome-specific errors
- keep finite integer thresholds, including `0`, working unchanged

Closes #222

## Validation

- `uv run ruff check --fix`
- `uv run ruff format --check`
- `uv run ty check`
- `uv run pytest -q` — 1171 passed, 6 skipped

### DIFF
--- a/src/hflow/steps.py
+++ b/src/hflow/steps.py
@@ -227,11 +227,27 @@ class Threshold:
                 "threshold key_pattern must not be empty; it is a glob over "
                 "measurement keys, e.g. '*/black_frame_pct'"
             )
-        if math.isnan(self.value):
+        if isinstance(self.value, bool):
+            raise ValueError(f"threshold value for {self.key_pattern!r} must be a number, not bool")
+        value_is_integer = isinstance(self.value, int)
+        if not value_is_integer and math.isnan(self.value):
             raise ValueError(
                 f"threshold value for {self.key_pattern!r} must be a number, not NaN: "
                 "every comparison against NaN is False, so this gate would reject "
                 "every episode it evaluated"
+            )
+        if not value_is_integer and math.isinf(self.value):
+            accepts_every_finite_measurement = (
+                self.comparison is Comparison.AT_MOST and self.value > 0
+            ) or (self.comparison is Comparison.AT_LEAST and self.value < 0)
+            outcome = (
+                "accept every finite measurement"
+                if accepts_every_finite_measurement
+                else "reject every finite measurement"
+            )
+            raise ValueError(
+                f"threshold value for {self.key_pattern!r} must be finite, not "
+                f"{self.value}: this comparison would {outcome}"
             )
 
     def holds(self, measurement: float) -> bool:
--- a/tests/test_gates.py
+++ b/tests/test_gates.py
@@ -250,3 +250,27 @@ def test_an_empty_gate_is_refused_at_construction() -> None:
 def test_a_nan_threshold_is_refused_at_construction() -> None:
     with pytest.raises(ValueError, match="not NaN"):
         hflow.Threshold("v", hflow.Comparison.AT_MOST, math.nan)
+
+
+@pytest.mark.parametrize(
+    ("comparison", "value", "outcome"),
+    [
+        (hflow.Comparison.AT_MOST, math.inf, "accept every finite measurement"),
+        (hflow.Comparison.AT_LEAST, -math.inf, "accept every finite measurement"),
+        (hflow.Comparison.AT_LEAST, math.inf, "reject every finite measurement"),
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging. This closes a bug neither the issue nor I knew about.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1171 passed, 6 skipped.

**The `2**1024` case.** I read `value_is_integer` as an optimization at first, since `math.isnan(5)` is perfectly happy. It is not an optimization, it is a fix, and your test says so. On current main:

```
Threshold("v", Comparison.AT_MOST, 2**1024)
-> OverflowError: int too large to convert to float
```

`math.isnan` converts to float and a large enough int cannot. So a threshold that Python can compare perfectly well was rejected with a message about float conversion, from a line meant to be checking for NaN. On your branch it constructs and `holds(5.0)` returns `True`, which is the right answer. That was pre-existing and nobody had hit it.

The other thing worth naming: deriving the message from the comparison direction, so `AT_MOST, +inf` says "accept every finite measurement" and `AT_LEAST, +inf` says "reject", is what the issue asked for and it is what makes the error worth reading. Four parametrized cases covering both signs against both comparisons is the right shape, since the mapping is the part that could be written backwards.

`Threshold("v", AT_MOST, 0).value == 0` in the same test is the assertion I asked for on the issue. A truthiness guard would have swallowed it and the test would have caught that.

Thank you for the patience on the assignment. #229 arrived while this was in flight and I closed it in favour of yours; your version is the better one on the merits as well as on the order, which made that easier.

Welcome, and thanks for a strong first contribution.

PR #233, opened 2026-08-28
refactor(app): type step and transform kinds

### LINKED ISSUE #225: step_kind and transform_kind are finite string sets typed as str, one beside an enum that already exists

Iamnotphage:
## Summary

* replace the registration-only string literals with a private `_RegistrationStepKind` enum, deliberately kept separate from the serialized `StepKind` vocabulary
* represent sync completion transform witnesses with `_TransformKind` instead of raw strings
* preserve lenient marker parsing: missing or unrecognized transform kinds remain readable but safely disable canonical reuse

This gives each finite vocabulary one owner while turning misspellings into type-checking failures rather than runtime failures inside diagnostic handling.

## Compatibility

The sync completion marker remains byte-for-byte compatible with current `main`: the persisted values are still exactly `"default"` and `"override"`.

Markers with a missing or unknown `transform_kind` remain valid for non-sync stages and continue to fail closed for reuse. `StepKind`, `PIPELINE_MANIFEST_VERSION`, and transform behavior versions are unchanged because no serialized values or canonical episode bytes change.

## Tests

* pin the exact JSON bytes written for a default transform
* verify an override is still persisted as `"override"`
* verify missing and unknown transform kinds remain readable
* verify both cases prevent canonical reuse

## Validation

* `uv sync --locked --all-extras`
* `uv run ruff check --fix`
* `uv run ruff format`
* `uv run ty check`
* `uv run pytest -q` — 1169 passed, 6 skipped
* `uv run --python 3.11 --locked pytest -q` — 1169 passed, 6 skipped
* `uv run --python 3.14 --locked pytest -q` — 1169 passed, 6 skipped

Closes #225

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -26,6 +26,7 @@ import time
 import traceback
 from collections.abc import Callable, Iterable, Mapping, Sequence
 from dataclasses import dataclass, field
+from enum import StrEnum
 from pathlib import Path
 from types import MappingProxyType, ModuleType
 from typing import TYPE_CHECKING
@@ -236,6 +237,13 @@ _MEDIA_CONTACT_SHEET_FPS = 0.5
 _SYNC_COMPLETION_MARKER_NAME = ".sync-complete.json"
 
 
+class _TransformKind(StrEnum):
+    """Transform implementations recorded in sync-reuse witnesses."""
+
+    DEFAULT = "default"
+    OVERRIDE = "override"
+
+
 @dataclass(frozen=True)
 class _SyncCompletion:
     """Proof that sync completed for one source path and canonical version.
@@ -259,11 +267,11 @@ class _SyncCompletion:
     # The transform is stamped with an ffmpeg version that does NOT reach
     # pipeline_version, and different builds genuinely encode differently.
     ffmpeg_version: str | None = None
-    # "default" or "override". An @app.transform override is contractually
-    # required to end in write_canonical_episode, so it stamps the SAME
-    # pipeline_version the default transform would: without this, REMOVING an
-    # override would reuse a canonical the current pipeline cannot produce.
-    transform_kind: str | None = None
+    # An @app.transform override is contractually required to end in
+    # write_canonical_episode, so it stamps the SAME pipeline_version the
+    # default transform would: without this, REMOVING an override would reuse
+    # a canonical the current pipeline cannot produce.
+    transform_kind: _TransformKind | None = None
 
 
 def _source_identity(source_reference: Path | str, storage_root: StorageRoot) -> str:
@@ -372,10 +380,24 @@ def _read_sync_completion_marker(marker_path: Path) -> _SyncCompletion:
     # the non-sync stages, which never needed a witness at all.
     optional_values = {
         field_name: value
-        for field_name in ("source_digest", "ffmpeg_version", "transform_kind")
+        for field_name in ("source_digest", "ffmpeg_version")
         if isinstance(value := marker_payload.get(field_name), str) and value
     }
-    return _SyncCompletion(**required_values, **optional_values)
+    raw_transform_kind = marker_payload.get("transform_kind")
+    transform_kind: _TransformKind | None = None
+    if isinstance(raw_transform_kind, str) and raw_transform_kind:
+        try:
+            transform_kind = _TransformKind(raw_transform_kind)
+        except ValueError:
+            transform_kind = None
+    return _SyncCompletion(
+        source_path=required_values["source_path"],
+        schema_version=required_values["schema_version"],
+        pipeline_version=required_values["pipeline_version"],
+        source_digest=optional_values.get("source_digest"),
+        ffmpeg_version=optional_values.get("ffmpeg_version"),
+        transform_kind=transform_kind,
+    )
 
 
 def _file_digest(path: Path) -> str:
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
Merging. Both vocabularies typed, `StepKind` untouched, and the two constraints that could have gone wrong hold.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1169 passed, 6 skipped.

I checked the marker independently of your tests, writing one on `441e1a1` and one on this branch:

```
{"ffmpeg_version": "n8.1.2", "pipeline_version": "9cd5d34f513d", "schema_version": "1",
 "source_digest": "abc123", "source_path": "episodes-in/e.mcap", "transform_kind": "override"}
```

Byte-identical. `StrEnum` serializes as its value through `json.dumps` because it is a `str`, but that is the kind of thing worth writing a file for rather than reasoning about.

The lenient parse also holds, with one behaviour difference worth naming since it is not in your description. Feeding a marker holding `"transform_kind": "some-future-transform"`:

- on main it parses to the raw string `'some-future-transform'`
- on this branch it parses to `None`

Neither raises, which is what the issue required, and both reach the same decision because `!= "default"` and `is not _TransformKind.DEFAULT` are both true, so no reuse either way. Normalizing an unrecognized value to `None` is the better of the two: the old behaviour carried a string no caller could act on. A marker with no `transform_kind` at all still parses on both.

Two separate enums was the call I made in the issue and you followed it, but the docstring you wrote for `_RegistrationStepKind` is doing the real work: "Derived channels appear here but are not manifest steps" is the sentence that stops someone merging the two in six months. That is better than the reason I gave.

Since this is your second contribution: the `good first issue` pool is best left for people arriving, and you are past needing it. Higher-leverage work if you want it is running HFlow against a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face) and reporting what breaks or what is slow. @tunglambk did that this week and it produced two issues better than anything I found by reading code. #203 and #205 in the non-starter backlog are also unassigned.

PR #235, opened 2026-08-28
Docs: docstrings for `CanonicalMcapWriter` public methods an

### LINKED ISSUE #11: Docs: docstrings for `CanonicalMcapWriter` public methods and `DiagnosticLevel`

Mr-Neutr0n:
Fixes #11

Added contract docstrings to `GroupedMcapWriter` public methods and `DiagnosticLevel` (issue #11); the issue's class/file names were stale relative to the current codebase.

Could not run the suite locally: . Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.

### DIFF
--- a/src/hflow/_grouped_mcap_writer/_writer.py
+++ b/src/hflow/_grouped_mcap_writer/_writer.py
@@ -236,6 +236,7 @@ class GroupedMcapWriter:
         self._flush()
 
     def register_schema(self, name: str, encoding: str, data: bytes) -> SchemaId:
+        """Register a schema and return its id for :meth:`register_channel`."""
         self._raise_if_not_active("register a schema")
         schema_id = SchemaId(len(self._schemas_by_id) + 1)
         schema = Schema(id=schema_id, data=data, encoding=encoding, name=name)
@@ -255,6 +256,11 @@ class GroupedMcapWriter:
         group: str,
         metadata: dict[str, str] | None = None,
     ) -> ChannelId:
+        """Register a topic on a named group and return its id for :meth:`write_message`.
+
+        ``schema_id`` must be a prior :meth:`register_schema` return value, or
+        ``NO_SCHEMA_ID`` for schemaless channels.
+        """
         self._raise_if_not_active("register a channel")
         # schema_id 0 is the MCAP spec's "no schema" sentinel.
         if schema_id != NO_SCHEMA_ID and schema_id not in self._schemas_by_id:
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
Merging, and you were right about the issue being stale. That is worth recording, because it means #11 was asking for something that no longer exists.

Validated on Linux, since your branch had no signal: `ruff check`, `ruff format --check`, `ty check`, 1176 passed, 6 skipped.

I checked the staleness claim rather than taking it: `CanonicalMcapWriter` appears nowhere in the tree, `src/hflow/mcap_writer.py` is gone, and `GroupedMcapWriter` is not exported from `hflow`. So half of #11 evaporated in a refactor after it was filed. `DiagnosticLevel` is still exported and still had no class docstring, which is the half that survived and you fixed it.

Documenting the private writer anyway is the right call. The issue's rationale was "the only exported class with that gap" and that rationale is gone, but the methods still take ids from each other and still have an ordering rule, so a contract line each is worth having.

Two contracts I verified against the code rather than reading past:

- `finish()` being idempotent is real: `if self._state is _WriterState.FINISHED: return` at the top. Worth stating, since a caller cannot tell from the signature.
- "Further writes raise `RuntimeError`" is accurate. `_raise_if_not_active` raises `RuntimeError` (`_writer.py:547`), not `ValueError`. Easy to get backwards and you did not.

One note for next time: `uv run pytest -q` from the repo root works on a fork checkout without any CI approval, and it is the signal I need most. Your PR body has "Could not run the suite locally: ." with the reason missing, so I could not tell whether something blocked you. If something does, say what and I will help.

Welcome, and thanks. Closing #11 as resolved.

PR #242, opened 2026-08-28
fix(storage): refuse empty local paths and bucketless URLs (#238)

### LINKED ISSUE #238: parse_storage_root accepts an empty data root as the current directory, and a bucket URL with no bucket
### Current behavior

`parse_storage_root` (`src/hflow/storage.py:619`) is the data-root boundary: every path or URL a user configures goes through it. Its docstring says anything unsupported is "refused loudly rather than treated as a strange directory name". Two values that name nothing get through.

Measured on current main (`4d25314`):

```
''                 ACCEPTED LocalStorageRoot path=.
'   '              ACCEPTED LocalStorageRoot path=(three spaces)
'.'                ACCEPTED LocalStorageRoot path=.
'gs://'            REFUSED: bucket URL 'gs://' names no bucket/container
'gs://b'           ACCEPTED BucketStorageRoot
'gs:///nobucket'   ACCEPTED BucketStorageRoot
'gs:////x'         ACCEPTED BucketStorageRoot
's3:///nobucket'   ACCEPTED BucketStorageRoot
```

**An empty data root becomes the current directory.** `parse_storage_root("")` and `parse_storage_root(".")` return the same thing, because `Path("")` is `Path(".")` (`storage.py:634`). So `data_root = ""` in an `hflow.toml` silently writes episodes into whatever directory the command was run from, usually the repository. A root of three spaces becomes a directory named three spaces.

**A bucket URL with no bucket is refused only without a trailing slash.** The guard is in `BucketStorageRoot.__init__` (`src/hflow/storage.py:409-413`):

```python
normalized_url = url.rstrip("/")
_scheme, separator, remainder = normalized_url.partition("://")
if not separator or not remainder:
    raise ValueError(f"bucket URL {url!r} names no bucket/container ...")
```

`gs://` strips to `gs:`, so `partition` finds no separator and it is refused. `gs:///nobucket` has nothing to strip, so `remainder` is `/nobucket`, which is non-empty and passes. The bucket name is the part of the remainder before the first `/`, and that is empty in every accepted case above. The check tests that *something* follows the scheme, where the rule it is written for is that a *bucket* follows the scheme.

### Why it matters

Both failures are silent and both are about where data lands. A mistyped bucket URL that keeps its bucket empty gets an object-store root pointed at no bucket, and the failure surfaces later from obstore, about a request rather than about the URL the user wrote. An empty data root writes into the working directory, which is the one place nobody wants episode files to appear.

### Pattern to copy

The refusal a few lines below, `src/hflow/storage.py:645-647`, sets the style: name the value, say what is supported, give an example. The existing `names no bucket/container` message is already the right message; it just needs to fire in the cases it describes.

### What to build

1. In `BucketStorageRoot.__init__`, check the bucket rather than the remainder: the segment of `remainder` before the first `/` must be non-empty. `gs://`, `gs:///x`, and `gs:////x` all refuse with the existing message; `gs://bucket` and `gs://bucket/prefix` are unaffected.
2. In `parse_storage_root`, refuse a local path that is empty or only whitespace, with a message saying to pass a directory. Do not refuse `.` or `./data`: a relative root is legitimate and `hflow.toml` uses one (`data_root = "./data"`).

### Definition of done

1. `parse_storage_root("")` and `parse_storage_root("   ")` raise `ValueError` naming the field.
2. `parse_storage_root(".")`, `parse_storage_root("./data")`, and a `Path` argument are unchanged.
3. `gs:///x`, `gs:////x`, and `s3:///x` raise the same `names no bucket/container` error that `gs://` already raises.
4. `gs://bucket`, `gs://bucket/prefix`, and every bucket URL in the tests keeps working.
5. `file:///abs/path` is unchanged, including its existing must-be-absolute refusal.
6. One test per refusal, and one asserting the relative-path forms still parse.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

`packages/hflow-server/tests`. `tests/test_storage.py` holds the existing cases.

### Notes

`is_bucket_url("")` returning `False` is correct and out of scope: it answers a different question and has no opinion about whether the value is usable.

Check whether anything in the tree calls `parse_storage_root` with a value that could be empty before you tighten it. If a caller relies on `""` meaning "here", that is worth saying on this issue rather than working around.

mohit-bhandari45:
Fixes #238

## Summary

This change ensures that `parse_storage_root` enforces stricter input validation at the data-root boundary. It explicitly refuses empty/whitespace-only local paths and safely intercepts bucket URLs that are missing an actual bucket name (e.g., `gs:///x`).

## Why

Prior to this fix, two types of erroneous input were failing silently:

1. **Empty Local Paths**: An empty data root (like `data_root = ""`) was interpreted equally as `Path(".")`, causing canonical outputs and episode data to be silently written to the working directory (often the repository itself).
2. **Missing Buckets with Trailing Slashes**: Bucket URLs mimicking `gs:///nobucket` or `gs:////x` slipped past the `obstore` config boundary. A missed bucket generated highly obscure errors downstream during fetch/publish sequences inside `obstore` rather than during configuration validation.

Adding straightforward parameter validation ensures we catch typos transparently and safely limit unintended filesystem writes.

## Validation

- Tested empty and whitespace strings: `parse_storage_root("")` and `parse_storage_root("   ")`.
- Asserted identical functionality for purely relative representations (`.` and `./data`) and existing `Path` instances.
- Ensured standard paths like `gs://bucket/prefix` continued parsing correctly.
- Ran the core test validations in the repo via terminal:

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_storage.py
```

## Checklist

### DIFF
--- a/src/hflow/storage.py
+++ b/src/hflow/storage.py
@@ -408,7 +408,7 @@ class BucketStorageRoot:
     ) -> None:
         normalized_url = url.rstrip("/")
         _scheme, separator, remainder = normalized_url.partition("://")
-        if not separator or not remainder.partition("/")[0]:
+        if not separator or (_scheme.lower() != "file" and not remainder.partition("/")[0]):
             raise ValueError(
                 f"bucket URL {url!r} names no bucket/container -- expected e.g. gs://bucket/prefix"
             )
--- a/src/hflow/storage.py
+++ b/src/hflow/storage.py
@@ -408,6 +408,11 @@ class BucketStorageRoot:
     ) -> None:
         normalized_url = url.rstrip("/")
         _scheme, separator, remainder = normalized_url.partition("://")
+        # The bucket is the remainder up to the first slash, so gs:///prefix
+        # names none and is refused the same way gs:// is. file:// is exempt
+        # because its authority is empty by construction: the tests simulate a
+        # bucket with file://<abs path> (tests/conftest.py), and requiring an
+        # authority there fails 3 tests and errors 19 more.
         if not separator or (_scheme.lower() != "file" and not remainder.partition("/")[0]):
             raise ValueError(
                 f"bucket URL {url!r} names no bucket/container -- expected e.g. gs://bucket/prefix"
--- a/src/hflow/storage.py
+++ b/src/hflow/storage.py
@@ -408,7 +408,7 @@ class BucketStorageRoot:
     ) -> None:
         normalized_url = url.rstrip("/")
         _scheme, separator, remainder = normalized_url.partition("://")
-        if not separator or not remainder:
+        if not separator or not remainder.partition("/")[0]:
             raise ValueError(
                 f"bucket URL {url!r} names no bucket/container -- expected e.g. gs://bucket/prefix"
             )
@@ -631,6 +631,8 @@ def parse_storage_root(value: "str | Path | StorageRoot") -> StorageRoot:
         return LocalStorageRoot(value)
     scheme, separator, remainder = value.partition("://")
     if not separator:
+        if not value.strip():
+            raise ValueError(f"storage root {value!r} must be a directory path or bucket URL")
         return LocalStorageRoot(Path(value))
     normalized_scheme = scheme.lower()
     if normalized_scheme == "file":
--- a/tests/test_storage.py
+++ b/tests/test_storage.py
@@ -34,6 +34,7 @@ class TestParseStorageRoot:
     def test_path_and_plain_string_are_local(self) -> None:
         assert parse_storage_root(Path("/tmp/x")) == LocalStorageRoot(Path("/tmp/x"))
         assert parse_storage_root("./data") == LocalStorageRoot(Path("./data"))
+        assert parse_storage_root(".") == LocalStorageRoot(Path("."))
 
     def test_existing_roots_pass_through(self) -> None:
         local_root = LocalStorageRoot(Path("/tmp/x"))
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging. Both cases from #238 are fixed and every form that should keep working does.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1178 passed, 6 skipped.

```
''                 REFUSED    '.'                ACCEPTED
'   '              REFUSED    './data'           ACCEPTED
'gs://'            REFUSED    'gs://b'           ACCEPTED
'gs:///nobucket'   REFUSED    'gs://b/p'         ACCEPTED
'gs:////x'         REFUSED    'file:///abs/path' ACCEPTED (local)
's3:///nobucket'   REFUSED
```

`remainder.partition("/")[0]` is the right reading: the bucket is what precedes the first slash, which is what made `gs:///nobucket` slip past a check on the whole remainder.

**The `file` exemption is the interesting part, and it needed saying.** I read it as a special case at first and went to see what breaks without it:

```
3 failed, 1155 passed, 6 skipped, 19 errors
```

`file://<abs path>` has an empty authority by construction, and every bucket test simulates its store that way (`tests/conftest.py:33` and three others). So the exemption is load-bearing, and none of those 22 failures name the URL as the cause. One fixup pushed (9f918f8): a comment at the guard saying what it protects and the number, plus a test asserting `BucketStorageRoot("file:///tmp/remote")` still constructs, so tightening it fails one clear test instead of twenty-two unclear ones.

Refusing whitespace as well as empty is beyond what the issue asked and is right: a root of three spaces was the other way to name nothing.

PR #243, opened 2026-08-28
fix(cli): add __main__.py to enable python -m hflow execution

### LINKED ISSUE #241: python -m hflow fails: there is no __main__.py beside the console script
### Current behavior

`hflow` ships a console script (`pyproject.toml:59-60`):

```toml
[project.scripts]
hflow = "hflow.cli:main"
```

There is no `src/hflow/__main__.py`, so the module form fails. On current main (`4d25314`):

```console
$ python -m hflow doctor episode.mcap
No module named hflow.__main__; 'hflow' is a package and cannot be directly executed
```

The console script works and is what the docs use, so nothing is broken. The message is just unhelpful for a form people reach for out of habit, and it names a missing module rather than pointing at the command that does exist.

### Why bother

`python -m` is the way to run a tool when the script is not on `PATH`, which happens with `pip install --user`, inside a container that calls the interpreter directly, or in a CI step that has the venv on `sys.path` but not its `bin`. It is also the muscle memory `python -m pip`, `python -m pytest`, and `python -m ruff` build.

Small, and honestly a convenience rather than a defect. It is a good first issue for the same reason it is small: the whole change is a file with two lines and a test that proves it routes to the same place as the script.

### What to build

`src/hflow/__main__.py`:

```python
from hflow.cli import main

raise SystemExit(main())
```

Nothing else. Do not duplicate argument parsing, and do not add a second entry point to `pyproject.toml`; the console script stays the one declared interface and this is a shim onto the same `main`.

### Definition of done

1. `python -m hflow doctor <file>` produces the same output and the same exit code as `hflow doctor <file>`.
2. `python -m hflow --help` works.
3. Exit codes are preserved for all three documented values (`docs/FORMAT.md:196-208`). The `0` and `2` paths are the cheap ones to test: a conforming file and a missing one.
4. A test asserts the two invocations agree, rather than asserting the module merely imports. Run both through `subprocess` and compare exit code and stdout.
5. No new dependency, no change to `pyproject.toml`.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Check how the existing CLI tests invoke the command before adding a subprocess test; `tests/test_cli.py` mostly calls `hflow.cli.main` in-process, which is faster and is the right default. This is one of the few cases that genuinely needs a subprocess, because the thing being tested is the interpreter's module resolution rather than the function.

Do not add `python -m hflow` to the documentation. The console script is the interface we teach; this only makes the fallback work when someone reaches for it.

tarun1790:
Resolves #241

### Summary of Changes
- Added \src/hflow/__main__.py\ to allow executing \hflow\ via the module syntax (\python -m hflow\).
- Directly delegates to \hflow.cli.main\ without modifying entry points in \pyproject.toml\.
- Added test coverage in \	ests/test_cli.py\ verifying module invocation.

### DIFF
--- a/src/hflow/__main__.py
+++ b/src/hflow/__main__.py
@@ -1,0 +1,3 @@
+from hflow.cli import main
+
+raise SystemExit(main())
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,4 +1,6 @@
 import json
+import subprocess
+import sys
 from pathlib import Path
 
 import pytest
@@ -6,6 +8,8 @@ from pytest import CaptureFixture
 
 from hflow import __version__
 from hflow.cli import _build_parser, main
+from hflow.testing import SyntheticEpisodeSpec, synthesize_episode
+from hflow.transform import write_canonical_episode
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging. The shim is exactly right: three lines, no second argument parser, no change to `pyproject.toml`.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1179 passed, 6 skipped.

One fixup pushed (f63f9c4), on the test rather than the code. Yours ran `python -m hflow --help` and asserted exit 0, which cannot fail in the way this can be wrong. I changed `__main__.py` to `main()` instead of `raise SystemExit(main())`, so the exit code is swallowed and every command reports 0:

```
1 failed, 13 passed
FAILED test_the_module_form_matches_the_console_script[missing-file-exit-2]
```

The `--help` case stayed green, because `--help` exits 0 either way. That is the whole reason the comparison has to include a command that exits non-zero.

The fixup runs both invocations and compares exit code and stdout, over `--help`, a missing file (exit 2), and a conforming episode (exit 0). `raise SystemExit(main())` is the line under test and now something fails if it changes.

Worth knowing generally: a test that asserts a thing starts will pass on a broken version of almost anything. Picking the input where the two paths must differ is usually the whole design of the test.

Issue #219, opened 2026-08-28
hflow.testing: SyntheticEpisodeSpec has no validation, so a zero duration writes an empty episode


kstonekuan:
### Current behavior

`hflow.testing` exports two episode specs. One validates its fields and one does not.

`VideoEpisodeSpec` (`src/hflow/testing.py:134`) is checked by `_validate_video_episode_spec`
(`src/hflow/testing.py:338`), called from `write_video_episode` at `src/hflow/testing.py:517`. It
refuses a non-positive duration, a non-positive rate, non-positive or odd image dimensions, an
empty camera name, and a fault segment outside the episode (`_validate_video_fault_segment`,
`src/hflow/testing.py:323`).

`SyntheticEpisodeSpec` (`src/hflow/testing.py:105`) has no equivalent. `synthesize_episode`
(`src/hflow/testing.py:646`) takes it as given. Run against current main (`6d80961`):

| Spec | Result |
| --- | --- |
| `duration_s=0.0` | writes a 1530-byte episode with no messages |
| `duration_s=-1.0` | writes a 1530-byte episode with no messages |
| `joint_hz=0.0` | writes a 1530-byte episode with no messages |
| `duration_s=float("nan")` | `ValueError: cannot convert float NaN to integer` |
| `image_hz=0.0` | `RuntimeError: ffmpeg failed (...)` with the whole command line |
| `joint_count=True` | writes an episode with one joint |
| `black_segment=(5.0, 9.0)` with `duration_s=1.0` | accepted, segment lies past the end |

The last row is the one `_validate_video_fault_segment` already refuses for the other spec, with
a message naming the field and the duration.

The empty-episode rows are the ones worth fixing. A test that builds a spec with a typo'd
duration gets a syntactically valid episode with nothing in it, and then fails somewhere further
down with a message about whatever ran out of data.

### Pattern to copy

`_validate_video_episode_spec` at `src/hflow/testing.py:338`, including its sibling
`_validate_video_fault_segment` for the three segment fields.

For the type half, `src/hflow/_video_measurements/_field_guards.py` has `require_int` and
`require_float`, written for this exact problem (`bool` subclasses `int`, so `True` satisfies
every range test). Do not import them from there: that package is deliberately dependency-free
so it can be extracted (`src/hflow/_video_measurements/__init__.py:1-6`), and core reaching into
it for a helper works against that. Two four-line functions in `testing.py` are the cheaper
answer, and the module docstring there should say why the copy exists.

[...]

PR #244, opened 2026-08-29
Sharpen video-measurement boundary guard to catch all hflow import forms

### LINKED ISSUE #240: _video_measurements documents that it must not import hflow, and nothing enforces it
### Current behavior

`src/hflow/_video_measurements/__init__.py:1-6` states an architectural invariant:

> This package is private while its result model and measurement definitions settle. It deliberately has no dependency on HFlow episodes, checks, catalog keys, gates, or binary-download policy so it can become a standalone package.

The invariant holds today. Nothing under `src/hflow/_video_measurements/` imports from `hflow`, verified on current main (`4d25314`):

```
rg -n 'from hflow|import hflow' src/hflow/_video_measurements/
(no matches)
```

Nothing enforces it. Grepping the suite for the claim finds only test files describing their own fixtures as dependency-free, which is a different thing.

The dependency runs one way today, core into the package: `checks.py`, `mediapipe_hands.py`, and `_video_measurement_toolchain.py` all import from it. One import in the other direction would break extraction, and it would be an ordinary-looking line that passes review, because reaching for `Episode` or a catalog key from a measurement module reads as natural.

### Why it is worth a test rather than a comment

This has already come up twice this week. #219 and #220 both told contributors not to import `_field_guards` from that package and to copy the two functions instead, and both had to explain why. #226 landed a copy with a module-docstring note (`src/hflow/testing.py:34-36`). So the rule is being applied by hand, in review, one contributor at a time. A test states it once.

### What to build

A test that walks every `.py` file under `src/hflow/_video_measurements/`, parses it, and asserts that no `import` or `from ... import` names `hflow` or any `hflow.` submodule. Relative imports inside the package are the normal case and must keep working.

Parse with `ast` rather than grepping text: a grep matches the word in a docstring or a comment, which is exactly what the module docstring above contains, and a test that fails on its own explanation is worse than none.

The failure message should say which file and which import, and repeat why the rule exists, since whoever trips it will be seeing the constraint for the first time.

### Definition of done

1. A test fails if any module under `src/hflow/_video_measurements/` imports `hflow` or `hflow.<anything>`.
2. Relative imports within the package (`from ._frame_statistics import ...`) pass.
3. The word `hflow` appearing in a docstring, comment, or string literal does not fail the test. Prove it: the package's own `__init__.py` docstring says "HFlow" and the test must be green on current main.
4. The message names the offending file and import.
5. Core importing *from* the package is untouched and untested here. That direction is fine and is not what the invariant is about.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Prove the test is a real tripwire before you finish: add `import hflow` to one module under the package, watch the test go red, take it back out, and put the result in the PR body. A guard nobody has seen fail is a guard nobody knows works.

`ast.walk` over `ast.Import` and `ast.ImportFrom` is the whole implementation. `ImportFrom` with `level > 0` is a relative import and is the case to allow.

mohit-bhandari45:
fixes #240 

# Video Measurement Boundary Guard

## What this is

`tests/test_video_measurement_boundary.py` enforces an architectural invariant for
`src/hflow/_video_measurements/`: this package must have **zero dependency** on
`hflow` core (episodes, checks, catalog keys, gates, or binary-download policy),
so that it can eventually be split out as a standalone package. The package's own
`__init__.py` docstring already makes this claim — this test is what actually
checks it.

There is no user-visible runtime change. This PR only sharpens the existing guard.

## What changed

Previously the guard only caught `ImportFrom`-style imports (e.g.
`from hflow import Ep`). It now parses every file in the package with `ast` and
catches **any** form of an `hflow` import from inside the package, including:

- `import hflow`
- `import hflow.episode`
- `import hflow.episode as ep`
- `from hflow import Ep`
- `from hflow.episode import Ep`

### What stays green (by design)

- Docstring/comment/string mentions of `HFlow` (e.g. the package's own
  `__init__.py` docstring) — the check is AST-based, not text-based, so prose
  never trips it.
- Relative imports within the package, e.g. `from ._frame_statistics import ...`
  — this is the normal, expected pattern and remains allowed.
- Recursive discovery via `rglob` covers nested modules, so new files added
  under the package are automatically checked.

### Failure message

When the guard trips, the failure names the offending file and the exact
import statement, and restates why the rule exists — since whoever hits this
is likely seeing the constraint for the first time.

## Why

The one-way dependency (core → package, never package → core) has been a
recurring review point:

- [[#219](https://github.com/Hebbian-Robotics/hflow/issues/219)](https://github.com/Hebbian-Robotics/hflow/issues/219)
- [[#220](https://github.com/Hebbian-Robotics/hflow/issues/220)](https://github.com/Hebbian-Robotics/hflow/issues/220)
- [[#226](https://github.com/Hebbian-Robotics/hflow/pull/226)](https://github.com/Hebbian-Robotics/hflow/pull/226)

A single import in the wrong direction would quietly break the extractability
of this package and reads as an innocuous line in review. Encoding the
invariant as a test means it's enforced automatically instead of re-checked by
hand on every PR.

Closes #240.

## Scope

Limited to the guard test itself. Core importing from the package is untouched
— that direction is not what this invariant is about.

## Validation

Run from the repo root on the feature branch:

```bash
uv run ruff check --fix        # passed — All checks passed!
uv run ruff format --check     # 1 file already formatted
uv run ty check tests/test_video_measurement_boundary.py   # All checks passed!
```

**Tripwire proof:**
1. Temporarily added `import hflow` to `src/hflow/_video_measurements/_toolchain.py`
   → guard failed, naming `_toolchain.py: import hflow`.
2. Removed it → guard passed again.
3. Probed all five forbidden import forms (`import hflow`, `import hflow.foo`,
   `import hflow.foo as a`, `from hflow import Ep`, `from hflow.episode import Ep`)
   — all caught.
4. Confirmed `from ._frame_statistics import ...` stays allowed and the
   `__init__.py` docstring mention of "HFlow" stays green.

**Full suite:** could not be run natively on this checkout (Windows). `import hflow`
fails there due to the missing `fcntl` module (per `CONTRIBUTING.md`), so
`uv run pytest -q` errors at collection. The guard itself is a pure-`ast` check
that never imports `hflow`, and was validated with a standalone harness that
reads the package files directly. CI (Linux) runs the full suite, including
this guard.

## Checklist

- [x] Added/updated outcome-focused tests for changed business logic.
- [x] Updated documentation for changed behavior, flags, formats, or requirements.
- [x] Ran `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check`.
- [x] Ran the relevant pytest suite.
- [x] No recordings, generated media, credentials, private URLs, or runtime artifacts added.
- [x] Preserved stored-data compatibility (test-only change; no stored-data impact).

### DIFF
--- a/tests/test_video_measurement_boundary.py
+++ b/tests/test_video_measurement_boundary.py
@@ -6,23 +6,51 @@ from pathlib import Path
 import hflow._video_measurements as video_measurements
 import hflow.ffmpeg as hflow_ffmpeg
 
+_PACKAGE_MUST_STAY_STANDALONE = (
+    "The video-measurements package must not import hflow or any hflow.* "
+    "submodule. It is designed to be extracted into a standalone package, and "
+    "one import in the other direction -- reaching for an Episode or a catalog "
+    "key from a measurement module -- would break that extraction. Import the "
+    "HFlow domain from core into this package, never from this package back "
+    "into core."
+)
+
+
+def _forbidden_imports(source_path: Path) -> list[str]:
+    syntax_tree = ast.parse(source_path.read_text(), filename=str(source_path))
+    forbidden: list[str] = []
+    for syntax_node in ast.walk(syntax_tree):
+        if isinstance(syntax_node, ast.Import):
+            for imported_alias in syntax_node.names:
+                if imported_alias.name == "hflow" or imported_alias.name.startswith("hflow."):
+                    forbidden.append(f"import {imported_alias.name}")
+        elif isinstance(syntax_node, ast.ImportFrom):
+            if syntax_node.level > 0:
+                continue
+            imported_module = syntax_node.module
+            if imported_module is not None and (
+                imported_module == "hflow" or imported_module.startswith("hflow.")
+            ):
+                forbidden.append(f"from {imported_module} import ...")
+    return forbidden
+
 
 def test_video_measurements_do_not_import_hflow_domain_modules() -> None:
     package_directory = Path(video_measurements.__file__).parent
     forbidden_imports_by_source: dict[str, list[str]] = {}
-    for source_path in package_directory.glob("*.py"):
-        syntax_tree = ast.parse(source_path.read_text(), filename=str(source_path))
-        forbidden_imports = [
-            imported_module
-            for syntax_node in ast.walk(syntax_tree)
-            if isinstance(syntax_node, ast.ImportFrom)
-            and (imported_module := syntax_node.module) is not None
-            and (imported_module == "hflow" or imported_module.startswith("hflow."))
-        ]
-        if forbidden_imports:
-            forbidden_imports_by_source[source_path.name] = forbidden_imports
+    for source_path in package_directory.rglob("*.py"):
+        forbidden = _forbidden_imports(source_path)
+        if forbidden:
+            forbidden_imports_by_source[source_path.name] = forbidden
 
-    assert forbidden_imports_by_source == {}
+    assert forbidden_imports_by_source == {}, (
+        f"{_PACKAGE_MUST_STAY_STANDALONE}\n"
+        "Offending imports: "
+        + ", ".join(
+            f"{source}: {', '.join(imports)}"
+            for source, imports in forbidden_imports_by_source.items()
+        )
+    )
 
 
 def test_ffmpeg_namespace_no_longer_exposes_the_incubating_measurements() -> None:


kstonekuan (approved):
Merging, and first: **#240 was wrong and that is my fault.** I filed it claiming nothing enforced the boundary. `tests/test_video_measurement_boundary.py` has existed the whole time. My check grepped the suite for the words in the docstring rather than for a test doing the work, found nothing, and I wrote the issue on that. You clearly went and read the code, found the guard, and improved it instead of writing a second one, which is the right response to a wrong issue.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1176 passed, 6 skipped.

**The gap you closed is real, and I checked it.** I put `import hflow` at the top of `_field_guards.py` and ran both versions of the test against it:

```
PR version of the test     1 failed   <- catches it
main's version             2 passed   <- does not
```

`ast.ImportFrom` alone misses the plain `import hflow` and `import hflow.episode as ep` forms, which are the ones someone reaches for when they want the top-level namespace. That is the whole finding and it stands on its own without my issue.

Two smaller things, stated accurately so the record is right:

- `glob` to `rglob` is future-proofing, not a live gap. The package has no subdirectories today, so both find the same six files. Worth doing anyway, since the first subdirectory would silently leave itself unscanned.
- The failure message is the part I would keep if I could keep only one. `assert ... == {}` tells whoever trips it that a dict is not empty. Yours tells them which import, in which file, and why the rule exists.

Skipping relative imports explicitly (`level > 0`) is belt-and-braces: `from ._frame_statistics import x` gives module `_frame_statistics`, which never matched anyway. No harm and it states the intent.

On #245: the bot closed it because of the one-open-PR cap, not because anything was wrong with it. Reopen it now that this one is merged, and thank you for taking it in good humour.

Since this is your second contribution: the `good first issue` pool is best kept for people arriving. Higher-leverage work if you want it is running HFlow against a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face) and reporting what breaks. #203 and #205 in the non-starter backlog are also unassigned.

PR #245, opened 2026-08-29
fix(doctor): report chunks out of time order within a group (#239)

### LINKED ISSUE #239: FORMAT.md item 3 also requires each group's chunks to be time-ordered, and nothing checks it
### Current behavior

`docs/FORMAT.md:152` states convention item 3 as two properties:

> **Chunk purity**: for every `ChunkIndex`, the channel IDs in `message_index_offsets` belong to exactly one group; **each group's chunk sequence is time-ordered**.

Nothing checks the second half. The doctor's chunk loop (`src/hflow/doctor.py:224`) reads each `ChunkIndex` for its channel set and never compares `message_start_time` between chunks. `topic-time-order` is a different property: it checks that one channel's messages ascend, which they can while the chunks holding them do not.

Verified on current main (`4d25314`). Two channels in one group, written so that the late messages of `/alpha` land in the first chunk and the early messages of `/beta` in the second. Each channel's own log times still ascend:

```
chunks in time order       chunk starts=[0, 2]   ordered=True   conforming=True  codes=[]
chunks OUT of time order   chunk starts=[10, 0]  ordered=False  conforming=True  codes=[]
```

The second file has chunk start times `[10, 0]` and is reported conforming with no findings.

### Why the property exists

A reader seeking to a time range uses the chunk index to skip chunks it does not need. Out-of-order chunks within a group do not corrupt anything, and every message is still readable, so this is a layout deviation rather than data loss. It is the same category as `chunk-mixes-video-and-state`: legal MCAP, not the layout HFlow writes, and worth reporting to whoever has to consume the file.

### What to build

In the chunk loop at `src/hflow/doctor.py:224`, group the chunk indexes the same way the purity check does, and report when a group's chunks are not ascending by `message_start_time`.

Level `warning`, matching `chunk-mixes-video-and-state` and for the same reason: nothing is broken, the layout just is not what HFlow produces.

Report once per group rather than once per out-of-order pair. A file written in a different order can have every chunk out of place, and the finding cap (`_MAX_FINDINGS_PER_CODE`, three per code) would otherwise hide which groups are affected behind a suppression count.

### Definition of done

1. A file whose chunks within one group descend by `message_start_time` produces a finding naming the group.
2. A file whose chunks ascend produces none, and every canonical episode this repo writes still reports conforming. Run `hflow doctor` over a synthesized multi-camera episode before and after and diff.
3. Chunks in different groups may interleave in any order, and that is not reported. Only order within a group is the rule.
4. A file with no `group/` keys in `provenance/v1` is not reported either way, since without the map there are no groups to check.
5. `docs/FORMAT.md`'s finding-code table gains the new code with its level and meaning, and the paragraph above it stops listing this as unchecked.
6. Tests for each of 1 through 4.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Related to #237, which covers the first half of the same convention item. They can land in either order and touch the same loop, so whoever takes the second one should rebase rather than resolve a conflict by hand. Say on the issue if you want both.

The fixture has to be built with the stock `mcap.writer.Writer` and a small `chunk_size`, because HFlow's own writer cannot produce the violation. `tests/test_doctor.py` is where the doctor's cases live.

mohit-bhandari45:
Fixes #239 

## Summary

`hflow doctor` now enforces the second half of FORMAT.md convention item 3: a single group's chunks must be time-ordered by `message_start_time`. A file whose chunks descend within one group produces a new `warning` finding (`chunk-group-out-of-time-order`) naming the group, so readers that seek a time range and skip chunks on the assumption that later chunks start later get a heads-up. Previously the doctor reported such files as conforming with no findings.

## Why

Issue #239: the chunk-layout convention states two properties — chunk purity and per-group time ordering — but only the first was checked. In a group with two channels, the late messages of one channel could land in the first chunk and the early messages of the other in the second; each channel's own log times still ascended, so `topic-time-order` stayed silent. This is a layout deviation (the sort of file HFlow itself never writes) rather than data loss, so it is a `warning`, same level and rationale as `chunk-mixes-video-and-state`.

The finding is reported once per group, not per out-of-order pair, so a file written in a different order — where every chunk is out of place — cannot hide affected groups behind `_MAX_FINDINGS_PER_CODE` suppression. Chunks in different groups may interleave in any order and are not reported; only order within a group is the rule, and a file with no `group/` keys in `provenance/v1` is not checked (no map, no groups).

## Validation

- `docs/FORMAT.md` finding-code table gains `chunk-group-out-of-time-order` (warning) and the paragraph above it no longer lists per-group chunk time order as unchecked.
- New tests in `tests/test_doctor.py`:
  - descending chunk sequence within one group → exactly one warning naming the group (file remains `conforming`, layout deviation is not an error);
  - ascending sequence → no finding;
  - chunks interleaved across different groups → no finding;
  - no `group/` keys in provenance → check skipped.

The fixtures use the stock `mcap.writer.Writer` with a small `chunk_size`, because HFlow's own writer cannot produce the violation.

I was unable to run `uv run pytest -q`, `uv run ruff check --fix`, `uv run ruff format`, or `uv run ty check` locally (this Windows machine cannot run the project), so these still need to be verified in CI or on a compatible host.

## Checklist

### DIFF
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -156,13 +156,14 @@ A file claiming this convention must satisfy, in increasing strictness:
 `hflow doctor <file.mcap> [more.mcap ...]` checks every file given and prints
 one report each in argument order; its aggregate result follows the
 [exit code rules](#exit-codes). It validates the container, summary, indexes,
-stamps, per-topic time order, and the H.264 access-unit properties listed
-below. It does not currently classify H.264 picture coding types to detect
-B-frames, so a clean report is not proof of the unchecked no-B-frame
-constraint. The doctor also does not reject non-VCL NAL units before the first
-AUD, so a clean report does not prove the canonical AUD-first constraint. An
-unreadable or unparseable path is reported in place, in the same per-file shape
-(`[error] unreadable: ...`), and the run continues with the remaining files.
+per-group chunk time order, stamps, per-topic time order, and the H.264
+access-unit properties listed below. It does not currently classify H.264
+picture coding types to detect B-frames, so a clean report is not proof of the
+unchecked no-B-frame constraint. The doctor also does not reject non-VCL NAL
+units before the first AUD, so a clean report does not prove the canonical
+AUD-first constraint. An unreadable or unparseable path is reported in place,
+in the same per-file shape (`[error] unreadable: ...`), and the run continues
+with the remaining files.
 
 ### Doctor finding codes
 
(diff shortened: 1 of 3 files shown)


mohit-bhandari45:
Ok, I forgot.

PR #236, opened 2026-08-28
docs: document doctor finding codes

victorwon2001:
## Summary

- document every finding code currently emitted by `hflow doctor`, including its severity and trigger
- clarify that a clean report does not currently prove the no-B-frame or strict AUD-first constraints
- link the doctor command description to the shared exit-code rules

## Why

`docs/FORMAT.md` currently implies that `doctor` checks every listed H.264 constraint and does not document its stable codes for automation. This closes #10 without changing runtime behavior or the canonical format.

## Validation

- `uv run ruff check --fix` — passed
- `uv run ruff format` — 176 files left unchanged
- `uv run ty check` — passed
- `uv run pytest -q` — 1176 passed, 6 skipped (with the repository's pinned ffmpeg/ffprobe prerequisites configured)
- `lychee --no-progress --include-fragments --exclude '^https://github\.com/Hebbian-Robotics/hflow/(issues|security/advisories/new)$' --exclude-path references/mcap-spec.md --exclude-path references/foxglove-CompressedVideo.proto .` — 337 checked, 333 OK, 4 excluded, 0 errors
- `git diff --check FETCH_HEAD..HEAD` — passed

## Checklist

- [x] Outcome-focused tests are not applicable because this changes documentation only.
- [x] I updated documentation for the described format and command behavior.
- [x] I ran `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check`.
- [x] I ran the relevant pytest suite.
- [x] I did not add recordings, generated media, credentials, private URLs, or runtime artifacts.
- [x] Stored-data compatibility is preserved; this changes no runtime behavior or format bytes.

PR #251, opened 2026-08-29
fix(lerobot): disable B-frames so the converter output decodes in full

### LINKED ISSUE #250: camera_frame_stats undercounts by two on B-frame streams: the raw-to-MP4 remux drops the reorder-buffer tail
## Summary

`decoded_frame_count` reads 301 for every 303-message camera channel on the pinned svla corpus, while the source material is fully decodable. The two-frame shortfall is an artifact of `write_access_units_to_mp4` (raw Annex B -> MP4 `-c copy` remux) when the stream contains B-frames: the remuxed container's trailing reorder-buffer frames cannot be decoded. `frame_deficit_pct` (100 * (expected - message_count) / expected, checks.py:528) is blind to it because no messages are missing.

## Evidence (svla_so101_pickplace @ f641879, ep0, both cameras)

Measured through the real pipeline: `hflow.App('svla-ep0', data_root=...).process(landing/lerobot_episode_0001.mcap)`, then the catalog measurements parquet:

| channel | message_count | expected | decoded | frame_deficit_pct |
|---|---|---|---|---|
| /observation.images.up | 303 | 303 | 301 | 0.0 |
| /observation.images.side | 303 | 303 | 301 | 0.0 |

Keyframe scan sees all 303 frames on both cameras (scanned_frame_count=303, first_frame_is_keyframe=1, 11 keyframes @ 1 s).

## Layer-by-layer attribution (all with plain ffmpeg, no hflow code)

1. Source chunk: full decode 11939/11939 frames, zero errors; the exact ep0 slice (`ffmpeg -ss 0 -to 10.1 -c copy`) decodes 303/303 (`ffprobe -count_frames`).
2. Converter output: `_transcode_mp4_to_h264` emits 303 access units; decoding the raw Annex B stream directly yields 303/303.
3. hflow remux (exact `write_access_units_to_mp4` command: `ffmpeg -r 30 -f h264 -i - -c:v copy -movflags +faststart -f mp4`): container holds 303 samples, decoder emits 301. Variants all reproduce 301: `-muxdelay 0 -muxpreload 0`, `-fflags +genpts -avoid_negative_ts disabled`, `-max_interleave_delta 0`, and a two-step raw->mp4->mp4 remux. Decoded PTS are contiguous (0..300), so the loss is exactly the last two frames: the B-frame reorder-buffer tail.
4. Control: re-encoding the slice with `-x264-params aud=1:bframes=0` (same GOP/CRF) decodes 303/303 through the identical remux.

## Root cause

libx264's default (medium preset) enables B-frames. The raw-stream demuxer hands the mp4 muxer packets whose reorder delays the decoder's output; the muxed container ends before the decoder can emit the trailing reorder frames, so two decodable frames are silently lost at decode. Any converter or recorder producing B-frame H.264 (ffmpeg default, most hardware encoders) hits this on every episode.

## Verified fix (one line)

In `examples/lerobot/prepare.py` `_transcode_mp4_to_h264`: `-x264-params aud=1:bframes=0`. Verified: the bframes=0 stream remuxes and decodes 303/303.

Longer-term (separate work): `write_access_units_to_mp4` could remux in a way that preserves the reorder tail for arbitrary B-frame sources (no flag combination tried restores it), or refuse streams whose reorder depth it cannot represent. Either way the measurement should surface the shortfall: worth noting decoded vs message_count in the evidence, e.g. a `decode_deficit` companion to `frame_deficit_pct`.

ayam04:
## What

One-line converter change: `_transcode_mp4_to_h264` now encodes with `bframes=0` (was libx264 medium default, which enables B-frames).

## Why

Fixes #250. The raw Annex B -> MP4 `-c copy` remux in `write_access_units_to_mp4` drops the B-frame reorder-buffer tail at decode: the muxed file carries all 303 samples but only 301 decode, so `camera_frame_stats` undercounts `decoded_frame_count` on every healthy episode converted from a B-frame stream. `frame_deficit_pct` cannot see it (no messages are missing).

Evidence chain (svla_so101_pickplace @ f641879, ep0, plain ffmpeg):
- source slices decode 303/303, full chunks 11939/11939, zero errors
- converter emits 303 access units; raw Annex B decodes 303/303
- remuxed via the exact `write_access_units_to_mp4` command: 303 samples, 301 decoded (last two frames; reproduced under -muxdelay 0, genpts, max_interleave_delta 0, and a two-step remux)
- control: same encode with `bframes=0` decodes 303/303 through the identical remux

## Validation

- Regression test added: converter output remuxed through `write_access_units_to_mp4` must decode every source frame (fails on the old encode, passes now). `pytest tests/test_lerobot_converter.py` -> 6/6 passed.
- Full pipeline on the pinned corpus with the fixed converter: `hflow.App(...).process(...)`, then the catalog measurements parquet reads `decoded_frame_count` 303/303 on both `/observation.images.up` and `/observation.images.side` (was 301/303).
- `ruff check` + `ruff format --check` + `ty check` on both changed files: clean.

Fixes #250

### DIFF
--- a/examples/lerobot/prepare.py
+++ b/examples/lerobot/prepare.py
@@ -161,7 +161,9 @@ def _transcode_mp4_to_h264(mp4_path: Path, gop_seconds: float, fps: float) -> li
         "-sc_threshold",
         "0",
         "-x264-params",
-        "aud=1",
+        # bframes=0: B-frame streams lose their reorder-buffer tail through
+        # the raw Annex B -> MP4 remux, undercounting decoded_frame_count (#250).
+        "aud=1:bframes=0",
         "-f",
         "h264",
         "pipe:1",
--- a/tests/test_lerobot_converter.py
+++ b/tests/test_lerobot_converter.py
@@ -7,6 +7,9 @@ implementation details or touching the network.
 
 import importlib.util
 import json
+import os
+import shutil
+import subprocess
 from pathlib import Path
 
 import pytest
@@ -18,6 +21,14 @@ prep = importlib.util.module_from_spec(_spec)
 _spec.loader.exec_module(prep)
 _DERIVE = prep._derive_numeric_schema
 _ENCODE = prep._encode_cdr_float32_array
+
+_FFMPEG = shutil.which("ffmpeg")
+_FFPROBE = shutil.which("ffprobe")
+
+pytestmark = pytest.mark.skipif(
+    _FFMPEG is None or _FFPROBE is None,
+    reason="system ffmpeg/ffprobe required",
+)
 
 
 def _build_fake_corpus(tmp_path: Path) -> dict:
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging. The diagnosis on #250 is the best piece of debugging this repo has had, and the fix is one line because you did the work to earn it.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1187 passed, 6 skipped.

I reverted `bframes=0` back to `aud=1` and the new test failed, so it catches the thing it was written for rather than merely exercising the path.

**The layer-by-layer attribution is what makes this reviewable.** Source decodes 303/303, converter output decodes 303/303 as raw Annex B, the remux container holds 303 samples and the decoder emits 301, and re-encoding with `bframes=0` restores 303 through the identical remux. Four measurements, each isolating one layer, and the conclusion follows from them rather than from a plausible story. Trying `-muxdelay 0`, `genpts`, `max_interleave_delta 0`, and a two-step remux before concluding no flag recovers it is the part most people skip.

**One thing worth adding to the record: the converter was emitting non-conforming output.** `docs/FORMAT.md:152` item 4 requires no B-frames, so a canonical stream should never have had them. Nothing caught it because the doctor does not classify picture types, which #236 documented three hours before you hit this. So the chain is: a documented unchecked constraint, a converter quietly violating it, and a silent two-frame loss downstream. That is the first real consequence of that gap, and it is a good argument for closing it.

One fixup pushed (09adc9d): the `pytestmark` skip was module-level, so on a machine without system ffmpeg all six tests in the file skipped, including the five metadata tests that never touch ffmpeg. Scoped it to the one test that shells out. Without ffmpeg on `PATH` it is now 5 passed, 1 skipped instead of 6 skipped.

Your two longer-term notes on #250 are both right and both want separate issues. I would take the measurement one first: `decoded_frame_count` being below `message_count` on a healthy episode is exactly the blind spot, and a companion to `frame_deficit_pct` would have surfaced this without anyone running ffmpeg by hand. #250 stays open and assigned to you for whichever you want to carry.

PR #252, opened 2026-08-29
Fix doctor chunk purity proxy to check provenance group map (#237)

### LINKED ISSUE #237: The doctor approximates chunk purity as video-versus-state, so two non-video groups sharing a chunk pass
### Current behavior

`docs/FORMAT.md:152` states convention item 3:

> **Chunk purity**: for every `ChunkIndex`, the channel IDs in `message_index_offsets` belong to exactly one group; each group's chunk sequence is time-ordered.

The doctor checks a proxy for the first half. At `src/hflow/doctor.py:233-234` it classifies each chunk's channels as video or not:

```python
has_video = any(channel_id in video_channel_ids for channel_id in chunk_channel_ids)
has_state = any(channel_id not in video_channel_ids for channel_id in chunk_channel_ids)
```

and warns `chunk-mixes-video-and-state` only when both are true. So a chunk mixing two non-video groups, `state` and `bulk` for example, is all non-video: `has_video` is false and nothing is reported.

Verified on current main (`4d25314`) with a file written by the stock `mcap` writer, declaring `/joint_states` in group `state` and `/lidar_points` in group `bulk`, both in one chunk:

```
one group per chunk          conforming=True codes=[]
state + bulk in one chunk    conforming=True codes=[]
```

The second file violates item 3 and the doctor calls it conforming.

### The information is already in the file

This is not a case where the doctor lacks what it needs. `provenance/v1` records the resolved topic-to-group map on every canonical episode, one key per topic (`transform.py:877`, prefix `group/` from `format.py:68`). A real episode's provenance holds:

```
group//cam/compressed = cameras
group//joint_states   = state
chunk-target/cameras  = 800000
chunk-target/state    = 800000
```

The doctor already reads that same record a few lines further down, at `src/hflow/doctor.py:249`, to check the version stamps. So the exact map item 3 is written in terms of is in hand at the point the chunk loop runs.

### Why it matters more than it looks

The doctor's job is to check files claiming the convention, including files HFlow did not write. Our own transform assigns chunks by group and is correct by construction, which is why nothing has caught this. A converter or a re-chunking tool has no such guarantee, and item 3 is the rule it would be told to satisfy.

### What to build

Read the `group/` keys from `provenance/v1` and check chunk purity against the map the file itself declares: for each `ChunkIndex`, resolve every channel's topic to its group and report when a chunk spans more than one.

Keep `chunk-mixes-video-and-state` working for files with no group map. A pre-provenance file, or one whose provenance lacks `group/` keys, gets today's video-versus-state approximation; there is no map to check against and the current warning is better than nothing. Say which case produced which code in the message.

Level stays `warning`, for the reason the current code comment gives at `doctor.py:240`: a custom `topic_groups` assignment can legally produce a mixed chunk, so this is a deviation from what HFlow writes rather than a broken file.

### Definition of done

1. A file whose `provenance/v1` declares two topics in different groups, with both in one chunk, is reported. A new code, or the existing one with a message naming the two groups; say which you chose and why.
2. A file with no `group/` keys still gets the current video-versus-state check, unchanged.
3. Every currently conforming episode this repo writes still reports conforming. Run `hflow doctor` over a synthesized multi-camera episode before and after and diff the output.
4. `docs/FORMAT.md`'s finding-code table gains a row if you add a code, and the paragraph above it says what is now checked.
5. Tests cover: mixed groups with a map, one group per chunk with a map, and a file with no map at all.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Scope is the first half of item 3 only. The second half, "each group's chunk sequence is time-ordered", is a separate property and I have not checked whether anything verifies it. Do not fold it in; if you want it, file it separately with its own evidence.

`tests/test_doctor.py` is where the doctor's cases live. The probe above builds its fixture with the stock `mcap.writer.Writer` rather than HFlow's own, which is the only way to produce a file that violates a rule our writer cannot violate.

mhaye9545:
Resolves #237. Updates the doctor's chunk purity check to read the \group/\ namespace keys from \provenance/v1\ metadata and assert that each chunk spans no more than one group (throwing \chunk-mixes-groups\). Falls back to the legacy video-versus-state approximation for files lacking a group map.

### DIFF
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -156,8 +156,9 @@ A file claiming this convention must satisfy, in increasing strictness:
 `hflow doctor <file.mcap> [more.mcap ...]` checks every file given and prints
 one report each in argument order; its aggregate result follows the
 [exit code rules](#exit-codes). It validates the container, summary, indexes,
-stamps, chunk purity (using the file's group map or a video-versus-state approximation), per-topic time order, and the H.264 access-unit properties listed
-below. It does not currently classify H.264 picture coding types to detect
+stamps, chunk purity (against the file's own group map, or a video-versus-state
+approximation when it has none), per-topic time order, and the H.264
+access-unit properties listed below. It does not currently classify H.264 picture coding types to detect
 B-frames, so a clean report is not proof of the unchecked no-B-frame
 constraint. The doctor also does not reject non-VCL NAL units before the first
 AUD, so a clean report does not prove the canonical AUD-first constraint. An
@@ -177,8 +178,8 @@ automation. An `error` breaks the canonical convention (or the MCAP spec); a
 | `no-statistics` | error | The summary has no `Statistics` record. |
 | `no-chunk-indexes` | error | The summary has no `ChunkIndex` records. |
 | `chunk-missing-message-indexes` | error | A chunk index has no per-channel `MessageIndex` offsets. |
-| chunk-mixes-groups | warning | One chunk contains channels assigned to different groups; custom grouping can make this intentional. |
-| chunk-mixes-video-and-state | warning | One chunk contains both video and state channels (reported when the file lacks a group map); custom grouping can make this intentional. |
+| `chunk-mixes-groups` | warning | One chunk contains channels assigned to different groups; custom grouping can make this intentional. |
+| `chunk-mixes-video-and-state` | warning | One chunk contains both video and state channels, reported only when the file has no group map; custom grouping can make this intentional. |
 | `missing-provenance` | error | The `provenance/v1` metadata record is absent. |
 | `provenance-missing-key` | error | `provenance/v1` lacks `schema_version` or `pipeline_version`. |
 | `missing-episode-record` | warning | The optional `episode/v1` semantics record is absent. |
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
Merging, with a fixup that was more than cosmetic: **none of your three tests ran.**

All three called `Writer.add_metadata(name=..., metadata={...})`, but the stock MCAP writer's signature is `add_metadata(name, data)`. Every one raised `TypeError` before reaching `diagnose`, so on the branch as submitted:

```
3 failed, 1186 passed, 6 skipped
ty check: 6 diagnostics
ruff format --check: 1 file would be reformatted
```

The feature shipped unverified. Your PR body has no validation section, so nothing was claimed that was not true, but `uv run pytest -q` from the repo root would have shown this in a minute and it is what the section is for. Please run it before pushing.

The design is right and I verified it independently, with the reproduction from #237:

```
one group per chunk          conforming=True  codes=[]
state + bulk in one chunk    conforming=True  codes=['chunk-mixes-groups']
```

Before this change the second line read `codes=[]`. And the case the issue cared most about protecting, HFlow's own output, is untouched: a synthesized two-camera episode and a camera-less one both still report conforming with no findings.

Reading the group map from `provenance/v1` rather than adding a column, keeping the video-versus-state approximation for files with no map, and choosing `warning` for the same reason the old code gives, are all exactly what the issue asked for. Splitting into a new code instead of overloading the old one is the right call too: the two mean different things and an automation matching on codes can tell them apart.

Fixup (277b7a8): the three `add_metadata` calls, the backticks the two new table rows dropped, a reflow of the paragraph above the table, and `k`/`v` renamed in the group-map loop, since this repo spells variables out.

Welcome, and thanks. #239 is the sibling of this one, still open and unassigned: same convention item, the other half of the sentence.

PR #253, opened 2026-08-29
fix(examples): resolve camera explicitly in egocentric contact_sheet enrichment

### LINKED ISSUE #234: [Bug]:examples/egocentric/pipeline.py: contact_sheet enrichment crashes on multi-camera episodes
Hi maintainers,

While testing the egocentric pipeline on multi-camera recordings, I hit a crash in the `contact_sheet` enrichment.

### What happens

Image

Running `examples/egocentric/pipeline.py` on an episode with more than one camera causes `contact_sheet` to fail:

```text
! contact_sheet [error]
ValueError: episode has multiple cameras ['/top_cam/compressed', '/wrist_cam/compressed']; pass one explicitly
```

---

## Environment
Ubuntu 24.04, Python 3.11

---

## Minimal reproduction

```bash
# 1. Create a 2-camera synthetic canonical episode
uv run python -c "
from hflow.transform import write_canonical_episode
synthesize_episode('scratch/multi_cam_raw.mcap', SyntheticEpisodeSpec(duration_s=2.0, cameras=('wrist_cam', 'top_cam')))
write_canonical_episode('scratch/multi_cam_raw.mcap', 'scratch/multi_cam_canonical.mcap')
"

# 2. Run the egocentric pipeline
uv run python examples/egocentric/pipeline.py scratch/multi_cam_canonical.mcap
```

---

## Expected behavior

`contact_sheet` in `examples/egocentric/pipeline.py` should resolve `camera=episode.cameras[0]` (matching `camera_health` at line 35) so the example pipeline succeeds on multi-camera recordings.

---

## Actual behavior

Running `examples/egocentric/pipeline.py` on an episode with multiple camera streams causes `contact_sheet` enrichment to fail:

```text
! contact_sheet [error]
ValueError: episode has multiple cameras ['/top_cam/compressed', '/wrist_cam/compressed']; pass one explicitly
```

---

## Additional context

In `examples/egocentric/pipeline.py:66`, `contact_sheet` calls `episode.frames(fps=1.0)` without specifying a camera, whereas `camera_health` at line 35 explicitly resolves `camera_topic = episode.cameras[0]`.

I have verified this locally and would love to open a PR to fix it! Could you please verify and assign this issue to me?

NiranjanKumar001:
Fixes #234

just clearling it out until i come across some other issues which will be fun to work on.

## What
`contact_sheet` in `examples/egocentric/pipeline.py` called
`episode.frames(fps=1.0)` without specifying a camera. On episodes with
more than one camera track, `Episode._resolve_camera()` raises
`ValueError: episode has multiple cameras; pass one explicitly`.

`timestamp_regularity` (line 30) and `camera_health` (line 35) in the
same file already resolved `episode.cameras[0]`. This change makes
`contact_sheet` consistent with them.

## Validation
```bash
uv run ruff check --fix && uv run ruff format && uv run ty check
uv run pytest -q

### DIFF
--- a/examples/egocentric/pipeline.py
+++ b/examples/egocentric/pipeline.py
@@ -61,15 +61,23 @@ def camera_health(episode: hflow.Episode) -> hflow.CheckResult:
 
 @app.enrich(version="1")
 def contact_sheet(episode: hflow.Episode) -> hflow.EnrichmentResult:
+    # episode.cameras is sorted, so this is the alphabetically first camera and
+    # not necessarily the one the recorder considered primary. Recorded in the
+    # labels because a sheet that cannot say which camera it shows is a trap on
+    # any multi-camera corpus.
+    camera_topic = episode.cameras[0]
     artifact_path = DATA_ROOT / "artifacts" / f"{episode.path.stem}-contact-sheet.jpg"
     sheet = hflow.ffmpeg.contact_sheet(
-        episode.frames(camera=episode.cameras[0], fps=1.0),
+        episode.frames(camera=camera_topic, fps=1.0),
         artifact_path,
         columns=5,
         max_tiles=20,
     )
     return hflow.EnrichmentResult(
-        labels={"contact_sheet_frame_count": sheet.frames_sampled_from},
+        labels={
+            "contact_sheet_frame_count": sheet.frames_sampled_from,
+            "contact_sheet_camera": camera_topic,
+        },
         artifacts={"contact_sheet": sheet.path},
     )
 
--- a/examples/egocentric/pipeline.py
+++ b/examples/egocentric/pipeline.py
@@ -63,7 +63,7 @@ def camera_health(episode: hflow.Episode) -> hflow.CheckResult:
 def contact_sheet(episode: hflow.Episode) -> hflow.EnrichmentResult:
     artifact_path = DATA_ROOT / "artifacts" / f"{episode.path.stem}-contact-sheet.jpg"
     sheet = hflow.ffmpeg.contact_sheet(
-        episode.frames(fps=1.0),
+        episode.frames(camera=episode.cameras[0], fps=1.0),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by kstonekuan — APPROVED** [review]
Merging. The fix is right, and I pushed a fixup on the test because it did not guard what it ships.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1177 passed, 6 skipped.

**Your test defined its own enrichment with the same shape, so it passed either way.** I reverted `examples/egocentric/pipeline.py` back to `episode.frames(fps=1.0)` and ran your test:

```
11 passed
```

Green with the bug back in. The test proved that an enrichment calling `frames(camera=..., fps=1.0)` works, which was never in doubt; it could not see the file it was shipping alongside. That is the same shape as the `python -m hflow` test earlier this week, and it is worth naming because it is easy to write by accident: a test that mirrors the fix tests the mirror.

The fixup (082c104) loads the real pipeline through `hflow.import_pipeline_application`, the same way `tests/test_stage_planning.py` loads one. `monkeypatch.chdir(tmp_path)` keeps the example's `DATA_ROOT` (a relative path) inside the temp directory. Reverting either half of the change now turns `contact_sheet` into an ERROR:

```
revert the camera resolution   -> 1 failed
drop just the new label        -> 1 failed
```

**Item 3 of #234 was missing**, so the fixup adds it: `contact_sheet_camera` in the labels. `episode.cameras` is sorted, so `cameras[0]` is the alphabetically first camera rather than the one the recorder considered primary. On this two-camera fixture the sheet renders `/top_cam/compressed` while the spec declared `wrist_cam` first. A sheet that cannot say which camera it shows is a trap on any multi-camera corpus, and the assertion on that label is also what makes the test specific.

Thank you for finding this in the first place. It came out of pointing HFlow at real recordings, which keeps producing better material than reading the code does.

PR #256, opened 2026-08-29
Enforce per-group chunk time order in doctor (#239)

### LINKED ISSUE #239: FORMAT.md item 3 also requires each group's chunks to be time-ordered, and nothing checks it

mhaye9545:
Resolves #239. This PR implements the second half of \FORMAT.md\ item 3, checking that chunks within a single group ascend strictly by \message_start_time\. It adds the \group-chunks-out-of-order\ warning finding, updates the documentation table, and includes 4 new test cases covering all edge cases defined in the issue.

### DIFF
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -157,7 +157,7 @@ A file claiming this convention must satisfy, in increasing strictness:
 one report each in argument order; its aggregate result follows the
 [exit code rules](#exit-codes). It validates the container, summary, indexes,
 stamps, chunk purity (against the file's own group map, or a video-versus-state
-approximation when it has none), per-topic time order, and the H.264
+approximation when it has none), per-topic time order, per-group chunk time order, and the H.264
 access-unit properties listed below. It does not currently classify H.264 picture coding types to detect
 B-frames, so a clean report is not proof of the unchecked no-B-frame
 constraint. The doctor also does not reject non-VCL NAL units before the first
@@ -179,6 +179,7 @@ automation. An `error` breaks the canonical convention (or the MCAP spec); a
 | `no-chunk-indexes` | error | The summary has no `ChunkIndex` records. |
 | `chunk-missing-message-indexes` | error | A chunk index has no per-channel `MessageIndex` offsets. |
 | `chunk-mixes-groups` | warning | One chunk contains channels assigned to different groups; custom grouping can make this intentional. |
+| group-chunks-out-of-order | warning | A group's chunks are not strictly ascending by message_start_time. |
 | `chunk-mixes-video-and-state` | warning | One chunk contains both video and state channels, reported only when the file has no group map; custom grouping can make this intentional. |
 | `missing-provenance` | error | The `provenance/v1` metadata record is absent. |
 | `provenance-missing-key` | error | `provenance/v1` lacks `schema_version` or `pipeline_version`. |
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -157,8 +157,9 @@ A file claiming this convention must satisfy, in increasing strictness:
 one report each in argument order; its aggregate result follows the
 [exit code rules](#exit-codes). It validates the container, summary, indexes,
 stamps, chunk purity (against the file's own group map, or a video-versus-state
-approximation when it has none), per-topic time order, per-group chunk time order, and the H.264
-access-unit properties listed below. It does not currently classify H.264 picture coding types to detect
+approximation when it has none), per-topic time order, per-group chunk time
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
Merging, and the thing I most wanted to see is that this one passes its own gate. It does: `ruff check`, `ruff format --check`, `ty check`, 1195 passed, 6 skipped, lychee 345 OK. Last time all three of your tests raised `TypeError` before reaching `diagnose`; here every one runs. That is the difference that matters.

Checked two things rather than taking them:

**The tripwire.** I replaced the comparison with `and False` and got one failure, `test_group_chunks_out_of_order`, and nothing else. The check is load-bearing for exactly its own test.

**My reproduction from #239.** Same fixture, two channels in one group with the late burst written first:

```
chunks in time order       starts=[0, 2]   codes=[]
chunks OUT of time order   starts=[10, 0]  codes=['group-chunks-out-of-order']
```

The second line read `codes=[]` before this change.

Three design calls I would have had to make and you made correctly. Tracking a last-start per group so groups can interleave freely, which is what item 3 actually says and what your third test pins. Accumulating into a set and reporting once per group, so `_MAX_FINDINGS_PER_CODE` cannot hide which groups are affected behind a suppression count. And comparing with `<` rather than `<=`, so two chunks sharing a start time are not a violation.

Fixup (fd9c7f5): the new table row lost its backticks the same way the two rows in #252 did, so that is twice now, worth a glance before pushing a docs table. Also reflowed the paragraph above it, annotated the two new locals, and replaced the working notes in the descending-chunk test, one of which said `chunk_size=1024` where the fixture uses `1`.

That closes both halves of FORMAT.md item 3, which was unchecked in both directions until this morning.

PR #258, opened 2026-08-29
fix: constrain hand-count evaluation schema

### LINKED ISSUE #257: Build AI evaluation: the hand-count schema does not state its value set, but its sibling does
### Current behavior

The Build AI evaluation defines two response schemas next to each other (`examples/build_ai_evaluation/judgment.py:54-63`). One states its allowed values and the other does not:

```python
HAND_COUNT_RESPONSE_SCHEMA: dict[str, object] = {
    "type": "object",
    "properties": {"hand_count": {"type": "integer"}},
    "required": ["hand_count"],
}
ACTIVE_MANIPULATION_RESPONSE_SCHEMA: dict[str, object] = {
    "type": "object",
    "properties": {"answer": {"type": "string", "enum": ["yes", "no"]}},
    "required": ["answer"],
}
```

Both tasks have a closed value set. The prompt says so for each: `prompts/hand_count.txt` ends "Return only one of: 0, 1, 2. No extra words", and `prompts/active_manipulation.txt` ends 'Respond only with: "yes" or "no."'. Both parsers enforce it. Only `answer` tells the model.

Measured on current main (`cdcc8d2`):

```
hand_count 2 (in range)          -> 2
hand_count 7 (schema-valid)      -> ValueError: hand count must be 0, 1, or 2
hand_count -1 (schema-valid)     -> ValueError: hand count must be 0, 1, or 2
answer "yes"                     -> 'yes'
answer "maybe" (schema REJECTS)  -> ValueError: active manipulation must be "yes" or "no"
```

`{"hand_count": 7}` satisfies the schema exactly. `{"answer": "maybe"}` does not.

### Why the asymmetry costs something

`_response_format_payload` (`judgment.py:160`) sends the schema as `response_format: json_schema`, so a provider that supports constrained decoding will not emit a value the schema forbids. That is enforcement at the model, before the tokens exist. For `answer` it is on; for `hand_count` it is off.

When the count comes back out of range, `evaluate_image_with_model` catches the `ValueError` (`judgment.py:280-283`) and records the episode as unparsed: a `{prefix}/parse_error` measurement, a `{prefix}/unparsed` tag, and **no `{prefix}/prediction` key at all** (`judgment.py:230-245`). So the frame drops out of the evaluation, and a paid API call was spent to learn that. The schema could have prevented the response instead of the parser rejecting it afterwards.

### What to build

Give `hand_count` the same treatment `answer` already has, so the two schemas say the same kind of thing about their value sets.

`{"type": "integer", "enum": [0, 1, 2]}` is the direct parallel. `minimum`/`maximum` also expresses it and is worth a moment's thought: `enum` matches the sibling schema and matches how the prompt words it ("only one of: 0, 1, 2"). Say which you picked and why.

Then pin that the two cannot drift apart again. The real problem is not the missing three characters, it is that a schema and its parser encode the same contract in two places and nothing compares them. A test that walks each task's schema and checks the parser accepts exactly the values the schema permits, and rejects the ones it forbids, catches the next divergence too.

### Definition of done

1. `HAND_COUNT_RESPONSE_SCHEMA` constrains `hand_count` to 0, 1, and 2.
2. A test asserts, for both tasks, that the value set the schema allows and the value set the parser accepts are the same. Table-driven over the two tasks rather than two hand-written tests, so a third task added later is covered by construction or fails loudly.
3. Every existing parser test still passes untouched. The parsers are already correct and this issue does not change them.
4. A test pins that a schema-permitted value is parseable: pick the boundary values from the schema itself, not literals, so the test follows a future edit to the schema.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

`packages/hflow-server/tests` and this example's own tests under
`examples/build_ai_evaluation/tests/`.

### Notes

No network and no API key needed. Everything here is a pure function over strings and dicts, which is why it is a good first issue despite living in the largest example in the repo.

Out of scope, but worth seeing while you are in the file: both prompts ask for a bare scalar ("No extra words") while the schemas ask for a JSON object. The parsers accept both shapes on purpose (`_parse_json_or_scalar`, `judgment.py:74`), so nothing is broken. Changing the prompt wording is a judgment call about eval semantics and belongs in its own issue with its own reasoning; do not fold it in here.

xianjianlf2:
## Summary
- constrain `hand_count` structured output to the parser-supported values 0, 1, and 2
- add a table-driven contract test for both executable Build AI tasks
- fail the contract test when a future executable task is added without schema/parser coverage

Closes #257

## Validation
- `uv run --locked --project examples/build_ai_evaluation pytest -q examples/build_ai_evaluation/tests` (19 passed)
- `uv run pytest -q` (1190 passed, 11 skipped)
- `uv run ruff check ...`
- `uv run ruff format --check ...`
- `uv run ty check`

This changes an evaluation check contract, not transform output, so no transform behavior version bump is required.

### DIFF
--- a/examples/build_ai_evaluation/judgment.py
+++ b/examples/build_ai_evaluation/judgment.py
@@ -53,7 +53,7 @@ class VisionCheckOutcome:
 
 HAND_COUNT_RESPONSE_SCHEMA: dict[str, object] = {
     "type": "object",
-    "properties": {"hand_count": {"type": "integer"}},
+    "properties": {"hand_count": {"type": "integer", "enum": [0, 1, 2]}},
     "required": ["hand_count"],
 }
 ACTIVE_MANIPULATION_RESPONSE_SCHEMA: dict[str, object] = {
--- a/examples/build_ai_evaluation/tests/test_evaluation.py
+++ b/examples/build_ai_evaluation/tests/test_evaluation.py
@@ -76,14 +76,28 @@ _TASK_VALUE_CONTRACTS: dict[
 
 
 def test_task_schema_value_sets_match_their_parsers() -> None:
+    """Each task's schema must state the value set its parser enforces.
+
+    Both tasks answer from a closed set, so the schema can say so and a
+    provider doing constrained decoding will not emit anything else. When only
+    the parser knows, an out-of-set answer costs a request and lands as an
+    unparsed episode with no prediction (#257).
+    """
     executable_tasks = set(EvaluationTask) - {EvaluationTask.BOTH}
     assert set(_TASK_VALUE_CONTRACTS) == executable_tasks
 
-    for schema, property_name, parser, rejected_values in _TASK_VALUE_CONTRACTS.values():
+    for task, (schema, property_name, parser, rejected_values) in _TASK_VALUE_CONTRACTS.items():
         properties = cast(dict[str, dict[str, object]], schema["properties"])
-        permitted_values = cast(list[object], properties[property_name]["enum"])
+        property_schema = properties[property_name]
+        assert "enum" in property_schema, (
+            f"{task.value}: {property_name!r} has no 'enum', so its schema does not state the "
+            "value set its parser enforces and the model is free to answer outside it"
+        )
+        permitted_values = cast(list[object], property_schema["enum"])
         for value in permitted_values:
-            assert parser(json.dumps({property_name: value})) == value
+            assert parser(json.dumps({property_name: value})) == value, (
+                f"{task.value}: the schema permits {value!r} but the parser does not accept it"
+            )
         for value in rejected_values:
             with pytest.raises(ValueError):
                 parser(json.dumps({property_name: value}))
--- a/examples/build_ai_evaluation/tests/test_evaluation.py
+++ b/examples/build_ai_evaluation/tests/test_evaluation.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import json
 import sys
+from collections.abc import Callable
 from pathlib import Path
 from types import SimpleNamespace
 from typing import cast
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging. You did the part of the issue that mattered rather than only the three characters.

Validated on Linux: `ruff check`, `ruff format --check`, `ty check`, 1195 passed in the root suite, 19 passed in the example's own project.

**The guard against a third task is the piece I most wanted and the easiest to skip.** `assert set(_TASK_VALUE_CONTRACTS) == executable_tasks` means adding an executable task without a contract entry fails here rather than shipping unchecked. Deriving `executable_tasks` from the enum minus `BOTH`, instead of listing two, is what makes that hold.

Reading `permitted_values` out of the schema rather than restating `[0, 1, 2]` is the other half. The test follows a future edit to the schema instead of having to be edited alongside it, which is what item 4 was asking for.

I checked it catches the original bug: reverting the enum fails the test. One fixup pushed (3dae825), because of *how* it failed:

```
KeyError: 'enum'
```

That is the test breaking, not the test reporting. Whoever removes the enum in six months sees a `KeyError` from inside a `cast` and has to reconstruct the rule. It now reads:

```
AssertionError: hand-count: 'hand_count' has no 'enum', so its schema does not state
the value set its parser enforces and the model is free to answer outside it
```

Same failure, and the message carries the reason. The permitted-value loop got a message too, naming which value diverged. A test that only guards is worth less than one that also explains, and this one will outlive everyone's memory of #257.

Your PR body's "1190 passed, 11 skipped" for the root suite is 1195 passed, 6 skipped here. The five extra skips are almost certainly ffmpeg-gated tests, so your environment has no system ffmpeg on `PATH`. Nothing wrong with the PR; worth knowing so a real failure is not mistaken for the same gap.

PR #263, opened 2026-08-29
test: enforce documented HFlow module boundaries

### LINKED ISSUE #261: Two stated module boundaries have no test, while a third does
### Current behavior

This repo states three module boundaries in prose and mechanically enforces one.

**Enforced.** `src/hflow/_video_measurements/__init__.py:1-6` says the package must not depend on HFlow core so it can be extracted, and `tests/test_video_measurement_boundary.py` walks the package with `ast` and fails on any `hflow` import. That test is the pattern this issue is about copying.

**Not enforced, 1.** `src/hflow/testing.py:21-23`:

> Written with the stock `mcap.writer.Writer` (this module must not depend on the canonical writer -- the fixture is its test input, and input files come from arbitrary recorders).

The invariant holds today: `testing.py` imports only `hflow.ffmpeg` and `hflow.format`. Nothing keeps it that way. Verified on current main (`84c2921`) by adding one line to `testing.py`:

```python
from hflow.transform import write_canonical_episode  # deliberate boundary violation
```

```
1195 passed, 6 skipped
```

A direct violation of the module's own stated contract, and the suite is green.

**Not enforced, 2.** `hflow-server` is deliberately not a core dependency (`pyproject.toml:23-31` lists seven, and it is not among them), and `src/hflow/cli.py:1176-1185` imports it inside `_command_serve` with an `except ImportError` that tells the user to install it. The function-level import is the whole mechanism: move it to module scope and `import hflow.cli` raises for everyone who installed only the core package, so every `hflow` command dies, not just `serve`. The `except ImportError` handler sitting right there is the author saying they expected the package to be absent.

No test pins it. The three tests that touch `hflow_server` (`tests/test_serve_cli.py`, `tests/test_runtime_cli.py:1239`) all run with it installed, so none of them would notice.

### Why this shape of test earns its keep

A boundary written in a docstring is advice. The same boundary as a test is a fact, and it survives the person who wrote the docstring forgetting about it. Both of these were stated carefully by someone who had a reason, and both are one plausible import away from being quietly untrue.

The second one has a user-visible failure: a core-only install where `hflow --help` raises `ModuleNotFoundError: hflow_server`.

### What to build

Two guards, following `tests/test_video_measurement_boundary.py`.

For `testing.py`, that test is nearly the shape you want already. Think about where the assertion belongs: a check that names one module is easy to write and easy to forget when a fourth boundary appears. A table of (module, forbidden imports) that the existing video-measurements case also fits into is worth considering. Say which you chose.

For the server separation, the property is "importing `hflow.cli` does not import `hflow_server`", which is different in kind: it is about what an import pulls in, not what a file's source says. `ast` can answer it from the source (no module-level `hflow_server` import anywhere under `src/hflow/`), and so can `sys.modules` after a fresh import. They fail differently and catch slightly different mistakes. Pick one and say why in the PR.

### Definition of done

1. A test fails if `src/hflow/testing.py` imports the canonical writer or anything else outside the modules it is allowed. Prove it: add the import, watch it go red, take it out, and put the result in the PR body.
2. A test fails if any module under `src/hflow/` imports `hflow_server` at module level. Prove it the same way, by moving the import in `cli.py` to the top of the file.
3. `cli.py`'s existing function-level import and its `except ImportError` message are unchanged. This issue adds a guard; it does not restructure the thing being guarded.
4. Relative imports and imports inside function bodies are not flagged. A function-level import is the correct pattern here, and the test must not push anyone away from it.
5. Failure messages name the file, the import, and the rule. Whoever trips these will be meeting the constraint for the first time.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Do not add guards for boundaries not listed here. `snapshot.py` and `mediapipe_hands.py` also mention optional dependencies, but those are about what is installed rather than what imports what, and they already fail loudly in their own way. If you think one of them belongs, say so on this issue before writing it.

Item 1 and item 2 are independent. If the second turns out to be more than you want, do the first and say so; a partial fix is welcome and the remainder can stay open.

LunaMeerkats:
## Summary

- keep all three documented import contracts in one typed `_ImportBoundary` table, including source paths, restricted prefix, exception modules, and full-tree versus module-scope traversal
- rename the architecture test to `tests/test_module_boundaries.py` so future boundary additions are discoverable
- preserve file, line, rendered import, and violated-rule diagnostics while keeping relative and function-local imports allowed

This is a test-only change; existing production imports and CLI behavior are unchanged.

## Design choice

I used source AST inspection for all three rows. For the optional-server boundary, a fresh-import assertion would depend on the test environment not installing `hflow-server`; the root development environment does install it. The AST guard tests the architectural property directly. `_ImportScope.MODULE` uses a small `NodeVisitor` that skips function and async-function bodies while still finding module-execution imports nested under `if`, `try`, or class bodies.

## Red-before proof

At pinned upstream `100848f1f558d675b444e17b0018d316ae624f3c`:

1. Temporarily adding `from hflow.transform import write_canonical_episode` to `src/hflow/testing.py` made the unified guard fail with:

   `hflow/testing.py:58: from hflow.transform import write_canonical_episode -- hflow.testing may import only ...`

2. Temporarily adding top-level `import hflow_server` to `src/hflow/cli.py` made the same table-driven guard fail with:

   `hflow/cli.py:31: import hflow_server -- hflow-server is optional ...`

Both deliberate mutations were then removed; production files have no diff. The existing function-local `hflow_server` import remains allowed.

## Validation

- `uv sync --locked --all-extras` — passed; 124 packages resolved, 58 checked
- `uv run ruff check .` — passed
- `uv run ruff format --check .` — 188 files already formatted
- `uv run ty check` — passed
- `uv run pytest -q tests/test_module_boundaries.py` — 2 passed
- `uv run pytest -q` — 1,195 passed, 6 skipped
- both deliberate forbidden-import mutations — failed causally with the expected file, line, import, and rule; restored runs passed
- `git diff --check` — passed
- hosted CI run 33292170985 at exact head 77c8818 — all 7 jobs passed (Python 3.11/3.14 root checks, links, and four workspace-example jobs)

The checks ran under Ubuntu in WSL2 with the repository mounted from the Windows workspace rather than cloned onto the Linux filesystem. The project-pinned, checksum-verified FFmpeg/FFprobe pair was placed on `PATH`; this test-only change does not exercise a runtime data root.

## Provenance

Prepared and submitted autonomously by OpenAI Codex through @LunaMeerkats. No human review or authorship is claimed.

Closes #261

### DIFF
--- a/tests/test_module_boundaries.py
+++ b/tests/test_module_boundaries.py
@@ -1,24 +1,60 @@
 """Architecture tests for documented HFlow module boundaries."""
 
 import ast
+from dataclasses import dataclass
+from enum import Enum, auto
 from pathlib import Path
 
 import hflow._video_measurements as video_measurements
 import hflow.ffmpeg as hflow_ffmpeg
 
 _HFLOW_PACKAGE_DIRECTORY = Path(video_measurements.__file__).parent.parent
+
+
+class _ImportScope(Enum):
+    ALL = auto()
+    MODULE = auto()
+
+
+@dataclass(frozen=True)
+class _ImportBoundary:
+    source_paths: tuple[Path, ...]
+    restricted_module_prefix: str
+    allowed_modules: frozenset[str]
+    scope: _ImportScope
+    rule: str
+
+
 _DOCUMENTED_HFLOW_IMPORT_BOUNDARIES = (
-    (
-        tuple(sorted(Path(video_measurements.__file__).parent.rglob("*.py"))),
-        frozenset(),
-        "the video-measurements package must not import hflow or any hflow.* "
-        "module because it is designed to be extracted into a standalone package",
+    _ImportBoundary(
+        source_paths=tuple(sorted(Path(video_measurements.__file__).parent.rglob("*.py"))),
+        restricted_module_prefix="hflow",
+        allowed_modules=frozenset(),
+        scope=_ImportScope.ALL,
+        rule=(
+            "the video-measurements package must not import hflow or any hflow.* "
+            "module because it is designed to be extracted into a standalone package"
+        ),
     ),
-    (
-        (_HFLOW_PACKAGE_DIRECTORY / "testing.py",),
-        frozenset({"hflow.ffmpeg", "hflow.format"}),
-        "hflow.testing may import only hflow.ffmpeg and hflow.format; the fixture "
-        "must not depend on the canonical writer or other HFlow domain modules",
+    _ImportBoundary(
+        source_paths=(_HFLOW_PACKAGE_DIRECTORY / "testing.py",),
+        restricted_module_prefix="hflow",
+        allowed_modules=frozenset({"hflow.ffmpeg", "hflow.format"}),
+        scope=_ImportScope.ALL,
+        rule=(
+            "hflow.testing may import only hflow.ffmpeg and hflow.format; the fixture "
+            "must not depend on the canonical writer or other HFlow domain modules"
+        ),
+    ),
+    _ImportBoundary(
+        source_paths=tuple(sorted(_HFLOW_PACKAGE_DIRECTORY.rglob("*.py"))),
+        restricted_module_prefix="hflow_server",
+        allowed_modules=frozenset(),
+        scope=_ImportScope.MODULE,
+        rule=(
+            "hflow-server is optional for core installs, so hflow_server imports must stay "
+            "inside the command or function that needs them"
+        ),
     ),
 )
 
@@ -40,11 +76,9 @@ class _ModuleScopeImportCollector(ast.NodeVisitor):
         pass
 
 
-def _absolute_imports(
-    source_path: Path, *, module_scope_only: bool = False
-) -> list[tuple[int, str, str]]:
+def _absolute_imports(source_path: Path, scope: _ImportScope) -> list[tuple[int, str, str]]:
     syntax_tree = ast.parse(source_path.read_text(), filename=str(source_path))
-    if module_scope_only:
+    if scope is _ImportScope.MODULE:
         collector = _ModuleScopeImportCollector()
         collector.visit(syntax_tree)
         import_nodes = collector.imports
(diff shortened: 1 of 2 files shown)


kstonekuan (changes requested):
Thanks, this is careful work. I mutated all five ways and every guard behaved exactly as you described: canonical-writer import in \`testing.py\` red, module-level \`import hflow_server\` red, the same import nested under \`if\`/\`try\` red, a function-local one green, and a fresh \`hflow.storage\` import inside \`_video_measurements\` still red. Full suite 1,196 passed, 6 skipped. \`ruff check\`, \`ruff format --check\`, \`ty check\` all clean.

Two things before I merge.

**The file name.** It is still \`tests/test_video_measurement_boundary.py\` and it now holds three boundaries, two of which have nothing to do with video measurement. The next person adding a boundary will not find this file, which is the failure mode the table was meant to prevent.

**The table only covers two of the three.** \`_DOCUMENTED_HFLOW_IMPORT_BOUNDARIES\` is the right instinct, but the server boundary sits outside it in its own hand-written function, so it is the one case a fourth boundary would have to be written from scratch again. I can see why: it forbids a prefix rather than allowing a list, and it is module-scope-only, and that difference currently lives in a keyword argument instead of in the data. Have a look at whether those two axes belong in the table rows. If you conclude they genuinely do not compose, say so in a comment there and I will take that answer.

LunaMeerkats:
Thanks @kstonekuan — addressed both points in `77c8818`.

- Renamed the file to `tests/test_module_boundaries.py` so the next boundary is findable.
- Folded all three contracts into `_DOCUMENTED_HFLOW_IMPORT_BOUNDARIES`. Each `_ImportBoundary` row now owns its source paths, restricted prefix, exact allowlist, `_ImportScope.ALL`/`MODULE` traversal choice, and diagnostic rule; the separate server test is gone.

I repeated both causal mutations against the unified table: the canonical-writer import failed at `hflow/testing.py:58`, and the top-level server import failed at `hflow/cli.py:31`. After restoring them, the focused file passed 2/2; Ruff, format check, `ty`, and the root suite passed (`1,195 passed, 6 skipped`). All seven exact-head hosted jobs have now passed as well, including the Python 3.11 and 3.14 root checks.

kstonekuan (approved):
Both points addressed. Scope is data now, and the filename says what the file does.

Re-validated against `301c684`, not your branch base: main moved under you and now has a new `src/hflow/importers/` package. Applies cleanly, 1,206 passed, 6 skipped. Six mutations, all as expected:

- `testing.py` importing the canonical writer: red
- `cli.py` module-level `import hflow_server`: red
- same import nested under a module-level `try`: red
- function-local `hflow_server` import: green, correctly ignored
- `_video_measurements` importing `hflow.storage`: red
- `src/hflow/importers/lerobot.py` importing `hflow_server` at module level: red

`importers` didn't exist when you wrote this and the table covered it the moment it landed, no edit to the test.

Merging.

PR #264, opened 2026-08-29
test: guard against module-level hflow_server imports under src/hflow

nityanand123gupta:
## Scope

This addresses only item 2 of #261 (the server-import guard) — leaving item 1 (the `testing.py` boundary) open, per the issue's own note that "a partial fix is welcome and the remainder can stay open."

## What this adds

`hflow_server` ships as a separate package (`packages/hflow-server`) so pipeline workers never carry it. `cli.py`'s `_command_serve` already imports it correctly at function scope, guarded by `except ImportError`, but nothing enforced that staying true — a module-level import anywhere under `src/hflow/` would make `import hflow.cli` fail for anyone who installed only the core package.

Added `test_no_module_under_hflow_imports_server_at_module_level`, following the same `ast`-based pattern as `test_video_measurement_boundary.py`. I chose static `ast` analysis over a `sys.modules` post-import check because it's consistent with the existing pattern in this codebase and catches the violation from source without needing a real `hflow_server` install in the test environment.

The check only inspects each module's **top-level** statements (not `ast.walk`, which would recurse into function bodies) — this is deliberate, so the existing function-scoped import in `_command_serve` stays valid and isn't pushed toward module scope.

## Testing

I don't have a Linux/macOS environment available (this repo's `hflow.storage` imports `fcntl`, a POSIX-only module, so the package doesn't import at all on Windows — unrelated to this change). I validated the actual detection logic directly instead of via `pytest`, per the issue's "prove it" requirement:

- Ran the check against the real, current `src/hflow/cli.py` and the whole `src/hflow/` tree — zero false positives, confirming the existing function-scoped import is correctly allowed.
- Reproduced the exact bug scenario in an isolated snippet (`from hflow_server import ServerSettings` at module level) — confirmed it's caught.
- Confirmed a function-scoped equivalent of that same import is *not* flagged.

Also ran `uv run ruff check --fix` and `uv run ruff format` on the new file — both clean.

I'd appreciate a maintainer (or CI) running the actual `pytest` suite to confirm, since I could only verify the logic in isolation here.

Contributes to #261

### DIFF
--- a/tests/test_server_import_boundary.py
+++ b/tests/test_server_import_boundary.py
@@ -1,0 +1,57 @@
+"""Architecture test for the hflow_server module-level import boundary."""
+
+import ast
+from pathlib import Path
+
+import hflow
+
+_MUST_NOT_IMPORT_SERVER_AT_MODULE_LEVEL = (
+    "hflow_server ships as a separate package (packages/hflow-server) so pipeline "
+    "workers never carry it. A module-level import anywhere under src/hflow/ would "
+    "make `import hflow.cli` -- and therefore every hflow command, not just "
+    "`serve` -- fail for anyone who installed only the core package. The existing "
+    "function-level import inside `_command_serve`, guarded by an `except "
+    "ImportError`, is the correct pattern and must stay function-scoped."
+)
+
+
+def _module_level_server_imports(source_path: Path) -> list[str]:
+    syntax_tree = ast.parse(source_path.read_text(), filename=str(source_path))
+    forbidden: list[str] = []
+    # Only the module's top-level statements matter here: a function-scoped
+    # import (like the one in cli.py's _command_serve) is deliberately allowed,
+    # so this does not walk into function or class bodies.
+    for syntax_node in syntax_tree.body:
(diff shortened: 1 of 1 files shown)


kstonekuan:
Thanks for this, and the reasoning behind the `ast`-over-`sys.modules` choice and the top-level-only walk is exactly right.

#263 landed about fifteen minutes earlier on the same issue, and the author had claimed #261 on the issue before that, so I am going with first-PR-wins here and closing this one. Nothing about your work is the reason. It covers both items and I have it in review now.

Two things worth knowing for next time. Comment on an issue before you start, so you get assigned and nobody duplicates you. And `hflow.storage` importing `fcntl` is a real Windows wall for this repo, so WSL2 is worth the setup: several contributors run it that way and the full suite passes there.

Plenty left, and the pool is open: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

nityanand123gupta:
Thanks for the feedback! I’ll make sure to claim an issue before starting next time. Is there another good first issue you’d recommend I work on?

Issue #255, opened 2026-08-29
write_access_units_to_mp4 silently drops the reorder tail of any B-frame stream


kstonekuan:
### What happens

`write_access_units_to_mp4` remuxes raw Annex B into MP4 with `-c:v copy`. When the stream carries B-frames, the trailing reorder-buffer frames cannot be decoded from the muxed file: the container holds every sample, and the decoder emits fewer.

Measured in #250, layer by layer, with plain ffmpeg and no HFlow code:

1. Source chunk decodes 11939/11939, zero errors. The exact episode slice (`-ss 0 -to 10.1 -c copy`) decodes 303/303.
2. The converter's H.264 output is 303 access units, and decoding the raw Annex B directly yields 303/303.
3. Through `write_access_units_to_mp4`'s exact command, the container holds 303 samples and the decoder emits **301**. Decoded PTS are contiguous `0..300`, so the loss is precisely the reorder tail.
4. Re-encoding the same slice with `bframes=0` decodes 303/303 through the identical remux.

`-muxdelay 0 -muxpreload 0`, `-fflags +genpts -avoid_negative_ts disabled`, `-max_interleave_delta 0`, and a two-step raw to MP4 to MP4 remux were all tried. Every one reproduces 301.

### Why this is still open after #251

#251 set `bframes=0` in the LeRobot converter, so HFlow's own converter stops producing streams that hit this. That fixes the corpus in front of us and nothing else. `write_access_units_to_mp4` is reached by any path that hands it H.264, and B-frames are libx264's default and standard on hardware encoders, so a third-party converter or a direct recording still loses its tail silently.

There is a related fact worth holding alongside: `docs/FORMAT.md:152` item 4 requires no B-frames in canonical video, and `hflow doctor` does not check picture coding types (documented in #236). So a B-frame stream is already non-conforming, already undetected, and now also silently lossy. Three gaps in a line.

### The decision this needs

Two defensible shapes, and the choice is a design call rather than a patch:

1. **Remux losslessly for arbitrary B-frame input.** Preserves whatever the caller hands over. No flag combination found so far does it, so this likely means not using `-c copy` from a raw demuxer, which is a real change to a path on the canonical write route.
2. **Refuse a stream whose reorder depth cannot be represented**, and say so. Consistent with how the transform treats every other canonical violation: it refuses rather than silently transforming. Cheap, and it converts a silent loss into an error naming the cause.

[...]

PR #265, opened 2026-08-29
fix(checks): report camera decode deficit

### LINKED ISSUE #254: camera_frame_stats reports no deficit when frames arrive but do not decode
### Current behavior

`camera_frame_stats` emits both `decoded_frame_count` and `frame_deficit_pct`, and the second cannot see what the first reports. `frame_deficit_pct` is computed at `src/hflow/checks.py:528`:

```python
100.0 * (expected_frame_count - inter.stamps_ns.size) / expected_frame_count
```

`stamps_ns.size` is the message count, so the deficit measures **messages that never arrived**. It is blind to messages that arrived and then failed to decode.

That gap is not hypothetical. It shipped, on a clean corpus, and nobody noticed until it was chased by hand. From #250, measured on `svla_so101_pickplace` @ `f641879` through the real pipeline:

| channel | message_count | expected | decoded | frame_deficit_pct |
| --- | --- | --- | --- | --- |
| `/observation.images.up` | 303 | 303 | 301 | 0.0 |
| `/observation.images.side` | 303 | 303 | 301 | 0.0 |

Two frames per camera were undecodable and the only deficit measurement in the check read `0.0`. The cause is fixed for our own converter (#251), but the measurement is what should have surfaced it, and it still would not.

### Why it matters beyond that one bug

`decoded_frame_count` and `message_count` are both already emitted, so a query *can* compare them. Nothing names the comparison, so nobody makes it: the evidence is present and the question is not asked. Any recorder or converter producing H.264 that HFlow cannot fully decode hits this, and B-frames are the ffmpeg default and near-universal on hardware encoders, so the population is not small.

### What to build

A per-topic measurement that states the shortfall between messages present and frames decoded, alongside the existing key. Something like `{topic}/decode_deficit_pct`, `100 * (message_count - decoded_frame_count) / message_count`.

Emit it under the same condition `decoded_frame_count` is emitted, which is the decoded-evidence branch of `_camera_frame_stats_keys` (`src/hflow/checks.py:397`). The key set is declared in one place and the body iterates it, so add it there and the check body follows.

`camera_frame_stats` is one of the six auto-registered defaults, so adding a key changes what every episode records.

### Definition of done

1. A new per-camera key reports the message-to-decoded shortfall, `0.0` on a healthy episode.
2. It is declared in `_camera_frame_stats_keys` and emitted by the body, with no second list anywhere. `tests/test_default_checks.py` already pins that declared and emitted key sets match; that test must still pass.
3. **`_DEFAULT_CHECK_VERSION_BY_FUNCTION[camera_frame_stats]` is bumped.** The check emits a key it did not before, so results are not comparable across the change. Say in the PR body what the old and new versions are.
4. A test builds an episode whose decode falls short of its message count and asserts the new key is non-zero, plus one asserting a healthy episode reads `0.0`. For the first, the cheapest known source of a shortfall is a B-frame stream through `write_access_units_to_mp4` (#250 has the exact ffmpeg invocation).
5. `docs/CATALOG.md` gains the key wherever the other `camera_frame_stats` keys are described.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Do not change `frame_deficit_pct`. It measures a real and different thing, missing messages, and its definition is correct; the problem is that it was the only deficit on offer. Two keys answering two questions is the fix, not one key answering both.

Found by @ayam04 while closing an acceptance item on #189. The full layer-by-layer attribution is on #250 and is worth reading before starting.

Zer0codestuff:
## Summary

Adds `<topic>/decode_deficit_pct` to the built-in `camera_frame_stats` check so recordings expose messages whose video frames did not decode. The catalog documentation distinguishes it from the existing rate-based deficit.

## Why

`frame_deficit_pct` compares received messages with the expected message rate. A stream can receive every message and still decode fewer frames, as #254 demonstrates. The new key preserves both questions instead of changing the existing measurement.

Closes #254.

The automatic `camera_frame_stats` version changes from `1` to `2` because its result rows gain a key.

## Regression coverage

The shortfall fixture uses an H.264 stream encoded with `bframes=0`, then models a faulty recorder by replacing its final frame payload with an access-unit delimiter that contains no coded picture. This gives 90 messages and exactly 89 decoded frames without depending on the open B-frame remux bug in #255. A healthy canonical episode separately asserts `decode_deficit_pct == 0.0`.

An empty camera topic has no valid percentage denominator, so the dispatcher now raises a named `ValueError` instead of dividing by zero. The normal `camera_frame_stats` path already rejects fewer than two timestamps while constructing the video.

## Validation

- `uv sync --locked --all-extras`: passed
- `uv run ruff check --fix`: passed
- `uv run ruff format`: passed
- `uv run ty check`: passed
- `uv run pytest -q`: 1,191 passed, 11 skipped
- `lychee --no-progress --include-fragments --exclude '^https://github\.com/Hebbian-Robotics/hflow/(issues|security/advisories/new)$' --exclude-path references/mcap-spec.md --exclude-path references/foxglove-CompressedVideo.proto .`: 354 links checked, 0 errors

## Checklist

### DIFF
--- a/.changeset/report-camera-decode-deficit.md
+++ b/.changeset/report-camera-decode-deficit.md
@@ -1,0 +1,5 @@
+---
+"hunkdiff": patch
+---
+
+Report camera frames that arrive but fail to decode.
--- a/.changeset/report-camera-decode-deficit.md
+++ b/.changeset/report-camera-decode-deficit.md
@@ -1,5 +1,0 @@
----
-"hunkdiff": patch
----
-
-Report camera frames that arrive but fail to decode.
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -183,9 +183,15 @@ as media, so a user key there would ship as media the pipeline never produced.
 Keys containing `/` need double quotes in SQL:
 
 ```sql
-SELECT episode_id, "/wrist_cam/compressed/black_frame_pct" AS black_pct
-FROM episodes WHERE black_pct < 5.0
+SELECT episode_id,
+       "/wrist_cam/compressed/black_frame_pct" AS black_pct,
+       "/wrist_cam/compressed/decode_deficit_pct" AS decode_deficit_pct
+FROM episodes WHERE black_pct < 5.0 AND decode_deficit_pct = 0.0
 ```
+
+The built-in `camera_frame_stats` check keeps two deficits separate:
+`frame_deficit_pct` measures missing messages against the expected rate, while
+`decode_deficit_pct` measures messages present whose frames did not decode.
 
 **Omit the key rather than measuring NaN or infinity.** A non-finite float is
 refused at append time, naming the check and the key. NaN compares false against
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -410,7 +410,7 @@ def _camera_frame_stats_keys(episode: Episode, *, cameras: Sequence[str] | None 
 
     Per topic: ``message_count`` always; ``expected_frame_count`` and
     ``frame_deficit_pct`` only when the topic carries at least two messages;
-    the seven decoded-evidence keys always. ``camera_instrument`` and
+    the eight decoded-evidence keys always. ``camera_instrument`` and
     ``camera_measurement_definition`` are the two non-topic keys; they ride
     inside the per-topic selection, so an episode with no cameras emits
     nothing at all.
@@ -426,6 +426,7 @@ def _camera_frame_stats_keys(episode: Episode, *, cameras: Sequence[str] | None 
             f"{topic}/{name}"
             for name in (
                 "decoded_frame_count",
+                "decode_deficit_pct",
                 "black_frame_pct",
                 "overexposed_frame_pct",
                 "freeze_total_s",
@@ -529,6 +530,11 @@ def _camera_value(
                 )
             case "decoded_frame_count":
                 return inter.stats.decoded_frame_count
+            case "decode_deficit_pct":
+                message_count = inter.stamps_ns.size
+                return float(
+                    100.0 * (message_count - inter.stats.decoded_frame_count) / message_count
+                )
             case "black_frame_pct":
                 return inter.stats.black_frame_percent
             case "overexposed_frame_pct":
@@ -571,8 +577,8 @@ def camera_frame_stats(
     (blackframe + freezedetect + signalstats in one filter graph, one shared
     frame denominator) over each camera's lossless MP4 remux, and compares
     the stored frame count against the rate the stream claims (declared
-    ``expected_hz`` when given, else the stream's median delta) -- the
-    LeRobot-style frame-count-vs-rate question. Freeze spans become labeled
+    ``expected_hz`` when given, else the stream's median delta), and compares
+    messages present with frames decoded. Freeze spans become labeled
     ``freeze:<topic>`` intervals in log time. Requires a canonical episode
     (``Episode.video`` remuxes in-band H.264 only).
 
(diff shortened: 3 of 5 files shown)


kstonekuan (changes requested):
Thanks, and the split is the right one. `frame_deficit_pct` and `decode_deficit_pct` answer two different questions and you kept both. The version bump to `2` is correct, the eight-key docstring count matches the tuple, and the CATALOG paragraph says the distinction plainly. Gate is clean here and the full suite is 1,196 passed, 6 skipped.

Two things.

**The new test pins an open bug as its premise.** I ran your fixture and got `message_count == 90`, `decoded_frame_count == 88`, `decode_deficit_pct == 2.22`. Those two missing frames are the B-frame reorder tail from #255, which is open and being fixed. So `assert decoded_frame_count < message_count` is currently asserting that #255 is unfixed, and the day it lands this test goes red for a reason that has nothing to do with the measurement you added.

The measurement is right. The problem is the only proof you have of it is a defect somebody is actively removing. Think about where else a decode shortfall can come from that HFlow is not going to fix, and about what this test should assert on a stream where every frame does decode. Say what you picked.

**`.changeset/report-camera-decode-deficit.md`.** This repo has no `.changeset` directory, no changeset tooling, and no package called `hunkdiff`. Drop the file. Version policy for this repo is in CONTRIBUTING.md under the identity and behavior-version section, and the `camera_frame_stats` bump you already made is the part that matters.

Worth checking while you are in there: `decode_deficit_pct` divides by `inter.stamps_ns.size` with nothing in front of it. Convince yourself that denominator can never be zero for a topic that reaches this branch, and if it can, say what the key should do rather than dividing.

Zer0codestuff:
Addressed in d342ae0. I replaced the B-frame premise with a stable faulty-recorder case: the fixture encodes with bframes=0, then replaces the final message payload with an access-unit delimiter containing no coded picture. It now proves exactly 90 messages and 89 decoded frames, while a healthy canonical episode explicitly asserts a 0.0 decode deficit. I also removed the unsupported Changeset file. For the denominator, the public path rejects fewer than two stamps while constructing the video, and the dispatcher now additionally raises a named ValueError for an empty topic instead of allowing an implicit division by zero. Local Ruff, formatting, ty, and the full suite pass: 1,191 passed, 11 skipped. CI is rerunning.

kstonekuan (approved):
Both points addressed.

The old fixture measured the B-frame reorder tail from #255, so it asserted an open bug stayed unfixed. You switched to `bframes=0` plus an injected message carrying only an AUD and no coded picture, which is a recorder-side defect we're never going to repair. Isolated where the deficit comes from now:

```
bframes=0, no injection            messages= 90  decoded= 90  deficit=0.00%
bframes=0, one AUD-only message    messages= 91  decoded= 90  deficit=1.10%
B-frames allowed, no injection     messages= 90  decoded= 88  deficit=2.22%
```

Clean stream reports nothing, and the whole deficit comes from the message you injected. #255 landing won't touch this test.

Zero denominator raises now, with a message saying what's undefined and why, and you covered it by handing the dispatcher an empty `stamps_ns`. Stray `.changeset` is gone.

Validated on `05fae79`, gate clean, 1,236 passed, 6 skipped. One mutation: making `decode_deficit_pct` return a flat `0.0` fails both `test_dispatcher_message_count_and_decoded_frame_count_read_distinct_sources` and `test_camera_frame_stats_reports_frames_present_but_not_decoded`, so the arithmetic is pinned from two directions.

Merging.

PR #268, opened 2026-08-30
test(egosuite): cover projection validation guards

### LINKED ISSUE #267: Every rejection in the EgoSuite projection geometry is untested
### Current behavior

`examples/egosuite_evaluation/geometry.py` refuses six kinds of bad input, each with its own message. On current main (`100848f`), no test reaches any of them.

The six (`geometry.py`):

| Line | Guard | Message |
| --- | --- | --- |
| `:73-74` | quaternion norm not finite, or zero | `camera quaternion must have a finite, nonzero norm` |
| `:99-100` | width or height not positive | `camera width and height must be positive` |
| `:101-102` | either focal length not positive | `camera focal lengths must be positive` |
| `:109-110` | any calibration value not finite | `camera calibration values must be finite` |
| `:125-126` | empty joint sequence | `at least one hand joint is required` |
| `:149-150` | a joint's camera-space coordinates not finite | `hand joint coordinates must be finite` |

`project_world_joints` is called from exactly two places outside the module, `evaluate.py:353` and `evaluate.py:358`, both of which run against real MCAP data and are not unit-tested.

The two tests that do call it (`tests/test_evaluation.py:40` and `:59`) share one module-level `CALIBRATION` with `width=200`, `height=100`, both focal lengths `100.0`, and both use unit quaternions and non-empty finite joint lists. Neither passes a value any guard would reject. Nothing else in the repository imports `geometry`.

So every one of those `raise ValueError` lines could be deleted and the suite would stay green.

### Why this is worth pinning

The projection is the ground truth the whole evaluation is scored against. If a bad calibration silently produces a plausible-looking pixel coordinate instead of an error, every downstream number is wrong in a way nothing surfaces, because `is_in_frame` still returns a clean `True` or `False` and the label report still writes a clean integer count.

The guards are the only thing standing between a malformed EgoSuite recording and a confidently wrong accuracy figure, and right now they are load-bearing code with no test holding them in place.

Note what is **already** covered, so you do not duplicate it: a joint behind the camera is exercised by `test_projection_counts_a_hand_when_any_labeled_joint_is_in_frame`, whose third joint sits at `z = -1.0` and is asserted out of frame. That path is fine. This issue is about the six raises above.

### What to build

A test for each guard, asserting the refusal happens and that the message identifies which rule was broken.

Six near-identical tests is one way and probably not the best one. The guards fall into two groups with different call shapes: the calibration and quaternion ones are properties of the camera, and the joint ones are properties of the input points. Parametrizing over a table of (bad input, expected message fragment) is worth considering, and would mean a seventh guard added later is one row rather than one function. Say which you chose.

Prove each test bites. Delete or neuter the guard it covers, watch that one test go red and the others stay green, put the guard back. Put the result in the PR body. A test that passes whether or not the guard exists is the thing this issue is trying to prevent, so it is the one thing worth demonstrating.

### Definition of done

1. Each of the six guards has a test that fails when that guard is removed.
2. The two existing projection tests pass untouched. Nothing in `geometry.py` changes: this issue adds tests, it does not change behavior.
3. The assertions check the message, not just the exception type. All six raise `ValueError`, so type alone does not tell you which rule fired.
4. Boundary values are picked deliberately, not arbitrarily. `width=0` and `width=-1` are different cases from `width=1`; say which side of each boundary you tested and why.

### Validation

```bash
uv run --locked --project examples/egosuite_evaluation ruff check examples/egosuite_evaluation
uv run --locked --project examples/egosuite_evaluation ruff format examples/egosuite_evaluation
uv run --locked --project examples/egosuite_evaluation ty check --project examples/egosuite_evaluation --extra-search-path . examples/egosuite_evaluation
```

The example has its own project and its own CI job, so its tests do not run in the root suite.

### Notes

No network, no API key, no MCAP file. Everything here is a pure function over floats, which is why it is a good first issue despite sitting in the newest example in the repo.

If while writing these you find a case where a guard is wrong rather than untested, say so on this issue before changing `geometry.py`. Fixing the projection is a different piece of work from pinning it.

lKolabrodl:
Closes #267.

## Summary

Add table-driven rejection coverage for every validation guard in the EgoSuite projection geometry:

- zero and non-finite quaternion norms
- zero image width and height boundaries
- zero horizontal and vertical focal lengths
- non-finite focal lengths and principal points
- an empty joint sequence
- non-finite x, y, and z joint coordinates

This is test-only and does not change production behavior.

## Why

The existing tests covered valid projections but did not prove that malformed camera or joint data is rejected. A future regression could therefore remove any of the guards without failing the suite.

## Verification

- `uv run --project examples/egosuite_evaluation ruff check examples/egosuite_evaluation`
- `uv run --project examples/egosuite_evaluation ruff format --check examples/egosuite_evaluation`
- `uv run --project examples/egosuite_evaluation ty check examples/egosuite_evaluation`
- `uv run --project examples/egosuite_evaluation pytest -q examples/egosuite_evaluation/tests` (`21 passed`)
- Mutation-checked all six guards independently: disabling each guard made its corresponding new cases fail while the other tests stayed green.

The native Windows test run used a local no-op `fcntl` import shim solely for test collection, consistent with the project's documented Windows limitation. These geometry tests do not exercise file locking or storage.

## AI assistance

AI assistance was used to implement and validate this change. The final diff and test results were reviewed before submission.

### DIFF
--- a/examples/egosuite_evaluation/tests/test_evaluation.py
+++ b/examples/egosuite_evaluation/tests/test_evaluation.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import math
 import sys
+from dataclasses import replace
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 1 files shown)


kstonekuan:
Reads complete from here. Mark it ready for review when you want me to look and I will take it from there.

One housekeeping thing: drop a comment on #267 so I can assign it to you. GitHub silently refuses to assign someone who has not participated on the issue itself, and a linked pull request does not count, so right now the issue still looks unclaimed to anyone browsing.

kstonekuan (approved):
LGTM, merging. Welcome.

Reran the mutation experiment independently. Every guard has exactly the coverage it should, and no test fires for a guard it does not own:

```
guard removed             tests that went red
quaternion norm           zero-quaternion-norm, nonfinite-quaternion-norm
width/height positive     zero-width, zero-height
focal lengths positive    zero-horizontal-focal-length, zero-vertical-focal-length
calibration finite        all four nonfinite-focal / nonfinite-principal-point cases
nonempty joints           empty-joint-sequence
joint coords finite        nonfinite-joint-x, nonfinite-joint-y, nonfinite-joint-z
```

On DoD 4, the boundary answer is stronger than the PR body claims. Zero is the only value that pins these three guards. Weakening them instead of deleting them:

```
width/height:     <= 0   becomes < 0     -> zero-width, zero-height fail
focal lengths:    <= 0.0 becomes < 0.0   -> both zero-focal cases fail
quaternion norm:  <= 0.0 becomes < 0.0   -> zero-quaternion-norm fails
```

A `width=-1` fixture would have caught none of those. An off-by-one in the comparison is the realistic regression here, so zero is the correct side, not just the convenient one.

Two more choices that are easy to get wrong. `math.nan` for the finiteness cases rather than `-math.inf`: the positivity check runs first and `-inf <= 0.0` is True, so an infinite focal length lands on "must be positive" and never reaches the guard you meant to cover. And matching the message rather than the type. With the quaternion guard removed:

```
Quaternion(inf, 0, 0, 1)  ->  ValueError: hand joint coordinates must be finite
Quaternion(0, 0, 0, 0)    ->  ZeroDivisionError: division by zero
```

The first one still raises `ValueError`, just for a different reason two guards later, so a bare `pytest.raises(ValueError)` would have passed on the wrong refusal.

Gate clean on the example project: `ruff check`, `ruff format --check`, `ty check`, 21 passed.

The `fcntl` shim you mentioned is correctly not in the diff.

PR #269, opened 2026-08-30
fix(server): warn on non-loopback binds

### LINKED ISSUE #259: hflow serve prints the loopback URL even when it bound every interface
### Current behavior

`hflow serve` prints the same line whether it bound loopback or every interface. Measured on current main (`84c2921`), same data root, same port:

```
$ hflow serve --data-root /tmp/servedata --host 127.0.0.1 --port 45601 --no-browser
hflow serve: serving /tmp/servedata at http://127.0.0.1:45601/

$ hflow serve --data-root /tmp/servedata --host 0.0.0.0 --port 45601 --no-browser
hflow serve: serving /tmp/servedata at http://127.0.0.1:45601/
```

Byte-identical. The second launch is reachable by anyone who can route to the machine, with no login, no token, and no session.

The line comes from `packages/hflow-server/src/hflow_server/server.py:453-455`:

```python
url_host = "127.0.0.1" if settings.host == "0.0.0.0" else settings.host
workspace_url = f"http://{url_host}:{chosen_port}/"
print(f"hflow serve: serving {settings.data_root} at {workspace_url}", flush=True)
```

The rewrite itself is right: `http://0.0.0.0:4356/` is not dialable, so printing a usable URL is the correct behaviour and should stay. The effect is that the one line HFlow writes for the operator shows the safe address in the unsafe case.

### What already warns, so this is not a hole nobody documented

- `--help` says "bind address (default 127.0.0.1; widening past loopback exposes your corpus)" (`src/hflow/cli.py:600-603`).
- `docs/SERVE.md:96` has a full Trust posture section, and `:106` calls passing `--host` past loopback "a deliberate exposure ... the only flag that changes who can reach the data".
- Uvicorn logs `INFO: Uvicorn running on http://0.0.0.0:45602` **four lines below** HFlow's own line, in a third-party voice among other INFO output.

So the facts are documented and the information is technically on screen. What is missing is HFlow saying anything at the moment the exposure happens, in its own voice, on the line an operator actually reads.

### Why the launch line is the right place

A flag's help text is read once, when someone is deciding. `SERVE.md` is read by people who go looking. Neither is in front of the person who typed `--host 0.0.0.0` six months ago in a shell script that still runs every morning. The launch line is the one moment the tool has their attention and knows the answer.

This is worth doing because the server authenticates nobody: reaching it is the whole of access control (`_settings.py`, `ServerSettings` docstring). A launch that widened that and looks identical to one that did not is a bad trade for one line of output.

### What to build

When the bound host is not loopback, have `serve` say so, once, at startup, in HFlow's voice.

The shape is yours. Some things worth deciding rather than assuming: whether it goes to stdout beside the existing line or to stderr; whether it names what an operator should do instead (`SERVE.md` recommends SSH port-forwarding, and `--read-only` narrows what a reacher can do but not what they can read); and whether it belongs in `serve()` next to the print, or earlier in `_command_serve` where the settings are built. Say which you chose and why in the PR.

"Not loopback" is more than a `== "0.0.0.0"` string test. `::`, `0.0.0.0`, and a concrete LAN address all expose; `127.0.0.1`, anything else in `127.0.0.0/8`, `::1`, and `localhost` do not. The `ipaddress` module in the standard library answers this properly, and a hostname that is not an IP literal is its own case worth thinking about.

### Definition of done

1. A launch past loopback prints something that a reader can act on, and a loopback launch prints exactly what it prints today.
2. The dialable URL is unchanged. Nobody should be told to open `http://0.0.0.0:4356/`.
3. Loopback is decided by what the address *is*, not by matching one string. `::1` and `127.0.0.5` are loopback; `0.0.0.0`, `::`, and a LAN address are not.
4. Tests cover both sides, including at least one IPv6 form.
5. Nothing about who can reach the server changes. This issue is about what the operator is told, not about refusing the bind, and `--host` must keep working exactly as it does.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

`packages/hflow-server/tests`, which is where the server's own tests live.
Markdown changes want the lychee command from CONTRIBUTING.md.

### Notes

Do not add authentication, do not refuse the bind, and do not change the default. Every one of those is a larger decision than this issue, and `SERVE.md` already explains why the server is unauthenticated on purpose.

If you find yourself wanting to change `SERVE.md`, say what and why on this issue first. That page is the considered statement of the posture and it is not wrong; this is about the tool matching it at runtime.

sambai-dev:
## Summary

- print one actionable warning on stderr when `hflow serve` binds beyond loopback
- classify IPv4 and IPv6 literals with `ipaddress`, while treating case-insensitive `localhost` as the only safe hostname
- preserve the existing bind behavior and dialable URL on stdout

Closes #259

## Why

The existing launch line rewrites `0.0.0.0` to a usable `127.0.0.1` URL, which is correct, but it leaves an unauthenticated network-wide bind looking identical to a loopback bind. The warning now lives in `serve()` beside that launch output, where both the configured host and the operator-visible URL are available.

Stderr keeps the established stdout line unchanged for scripts and supervisors. For IP literals, `ipaddress.ip_address(...).is_loopback` handles the full `127.0.0.0/8` range and `::1`. Exact `localhost`, matched case-insensitively, is the only hostname treated as loopback; other hostnames warn conservatively rather than relying on environment-dependent DNS resolution.

This does not refuse or alter any bind, change the default, or change the displayed dialable URL. No documentation change is included because `docs/SERVE.md` already states the trust posture and recommends loopback plus SSH port forwarding.

## Validation

- TDD RED: the wildcard launch regression failed against unchanged `main` because stderr was empty
- focused WSL2 suite: `uv run pytest -q packages/hflow-server/tests tests/test_runtime_cli.py` (`281 passed`)
- Python 3.11.16: `uv sync --locked --python 3.11`; `uv run ruff check`; `uv run ruff format --check`; `uv run ty check`; `uv run pytest -q` (`1204 passed, 6 skipped`)
- Python 3.14.7: `UV_MANAGED_PYTHON=1 uv sync --locked --python 3.14`; `uv run ruff check`; `uv run ruff format --check`; `uv run ty check`; `uv run pytest -q` (`1204 passed, 6 skipped`)
- repository write-mode checks: `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check` (passed; formatter changed no files)
- real WSL2 launch: loopback printed no warning; `0.0.0.0` printed exactly one warning while stdout still advertised `http://127.0.0.1:<port>/`

## Checklist

- [x] I added outcome-focused tests for changed business logic.
- [x] Documentation remains current for this behavior; `docs/SERVE.md` already documents the trust posture and SSH forwarding alternative.
- [x] I ran the relevant pytest suite and the complete suite on Python 3.11 and 3.14.

### DIFF
--- a/packages/hflow-server/src/hflow_server/server.py
+++ b/packages/hflow-server/src/hflow_server/server.py
@@ -12,8 +12,10 @@ identity.
 """
 
 import importlib.resources
+import ipaddress
 import os
 import socket
+import sys
 import threading
 import webbrowser
 from pathlib import Path
@@ -439,6 +441,16 @@ class ServerStartupError(RuntimeError):
     """
 
 
+def _is_loopback_host(host: str) -> bool:
+    """Return whether a host is a loopback IP literal or the localhost name."""
+    if host.casefold() == "localhost":
+        return True
+    try:
+        return ipaddress.ip_address(host).is_loopback
+    except ValueError:
+        return False
+
+
 def serve(settings: ServerSettings) -> None:
     """Run the workspace server: free port, printed URL, browser, uvicorn."""
     application = create_app(settings)
@@ -453,6 +465,14 @@ def serve(settings: ServerSettings) -> None:
     url_host = "127.0.0.1" if settings.host == "0.0.0.0" else settings.host
     workspace_url = f"http://{url_host}:{chosen_port}/"
     print(f"hflow serve: serving {settings.data_root} at {workspace_url}", flush=True)
+    if not _is_loopback_host(settings.host):
+        print(
+            f"hflow serve: warning: binding to {settings.host} exposes this unauthenticated "
+            "workspace to reachable networks; use a loopback host and SSH port forwarding "
+            "for remote access",
+            file=sys.stderr,
+            flush=True,
+        )
     if settings.open_browser:
         # uvicorn.run blocks this thread; a short timer opens the browser
         # once the server has had time to bind.
--- a/packages/hflow-server/tests/test_server_launch.py
+++ b/packages/hflow-server/tests/test_server_launch.py
@@ -1,0 +1,89 @@
+"""Observable startup output from ``hflow serve``."""
+
+from pathlib import Path
+
+import pytest
+from hflow_server import ServerSettings
+from hflow_server import server as server_module
+
+
+def _launch_without_starting_server(
+    host: str,
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> Path:
+    def fake_create_app(_settings: ServerSettings) -> object:
+        return object()
+
+    def choose_requested_port(_host: str, requested_port: int) -> int:
+        return requested_port
+
+    def do_not_start_uvicorn(*_args: object, **_kwargs: object) -> None:
+        pass
+
+    monkeypatch.setattr(server_module, "create_app", fake_create_app)
+    monkeypatch.setattr(server_module, "_first_free_port", choose_requested_port)
+    monkeypatch.setattr(server_module.uvicorn, "run", do_not_start_uvicorn)
+
+    data_root = tmp_path / "workspace"
+    server_module.serve(
+        ServerSettings(
+            data_root=str(data_root),
+            host=host,
+            port=4356,
+            open_browser=False,
+        )
+    )
+    return data_root
+
+
+def test_loopback_launch_output_is_unchanged(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+    capsys: pytest.CaptureFixture[str],
+) -> None:
+    data_root = _launch_without_starting_server("127.0.0.1", tmp_path, monkeypatch)
+
+    captured = capsys.readouterr()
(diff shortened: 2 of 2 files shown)


kstonekuan:
Reads complete from here. Mark it ready for review when you want me to look. #259 is already assigned to you, so nothing else to do.

kstonekuan (approved):
LGTM. `is_loopback` covers all of `127.0.0.0/8` and `::1`, and an unresolvable hostname warns instead of depending on DNS.

Validated on `4f05492`: ruff, format, ty clean, 1,204 passed, 6 skipped. Two mutations, both behaved. Collapsing `_is_loopback_host` to `host == "127.0.0.1"` failed exactly `127.0.0.42`, `::1`, `localhost`, `LOCALHOST`. Sending the warning to stdout failed the three warn cases plus the wildcard test.

The loopback test asserts stdout stays byte-identical, which covers the half of this that breaks silently.

Known and fine, not a change request: `::ffff:127.0.0.1` warns even though it's effectively loopback, since `is_loopback` only matches `::1`. Warning when safe is the right direction to be wrong in.

PR #271, opened 2026-08-30
Fix SQL statement injection in constrained curate() connections

Raghav7-tech:
Fixes the SQL injection vulnerability reported in the private security advisory
submitted earlier (credited).

_stage_manifest_and_count() built its COPY statement via raw f-string
interpolation of tenant-supplied SQL:

    connection.execute(
        f"COPY ({sql}) TO {_quote_sql_string(str(staged_manifest))} (FORMAT PARQUET)"
    )

A crafted `sql` value could close the COPY statement early and append
arbitrary semicolon-separated statements, which then executed on the same
connection even under constrained=True — bypassing the intended tenant-SQL
sandbox despite the existing file-access lockdown (enable_external_access=false,
lock_configuration=true, etc.).

## Fix
Replaces the interpolated COPY with DuckDB's relation API:

    connection.sql(sql).write_parquet(str(staged_manifest))

connection.sql() parses `sql` as a single query expression and raises a
parser error on multi-statement input instead of executing it.

## Verification
- Added a regression test (`test_constrained_curate_rejects_a_second_statement_smuggled_in_sql`)
  that drives `_stage_manifest_and_count` against a constrained connection
  with a known writable directory, using the same multi-statement injection
  technique from the original report.
- Confirmed the test **fails** against the original code (injection succeeds)
  and **passes** against this fix.
- Confirmed legitimate single-statement curation queries still work correctly.
- Full `curation`-related test suite passes with no regressions
  (pre-existing, unrelated ffmpeg/Windows-path failures aside).

### DIFF
--- a/src/hflow/curation.py
+++ b/src/hflow/curation.py
@@ -429,13 +429,51 @@ def _open_connection_over_root(
     return connection
 
 
+def _reject_non_single_select(sql: str) -> None:
+    """Raise ValueError if *sql* is not exactly one SELECT statement.
+
+    ``connection.execute()`` and ``connection.sql()`` both run **every**
+    semicolon-separated statement in the input and return only the last
+    result, so a tenant-supplied string can smuggle a second statement
+    (``SELECT ...; CREATE TABLE pwned AS SELECT ...``) and it will execute
+    silently.  ``extract_statements`` parses the text **without executing
+    it** and lets us count statements and check their types before anything
+    touches the connection.  A syntactically invalid string raises
+    ``duckdb.Error`` here — treat that as "not exactly one SELECT" too.
+    """
+    parser_connection = duckdb.connect()
+    try:
+        statements = parser_connection.extract_statements(sql)
+    except duckdb.Error as exc:
+        raise ValueError("sql must be exactly one SELECT statement") from exc
+    finally:
+        parser_connection.close()
+    if len(statements) != 1 or statements[0].type != duckdb.StatementType.SELECT:
+        raise ValueError("sql must be exactly one SELECT statement")
+
+
 def _stage_manifest_and_count(
     connection: duckdb.DuckDBPyConnection, sql: str, staged_manifest: Path
 ) -> int:
-    """COPY the query's result to the staged manifest; return its row count."""
-    connection.execute(
-        f"COPY ({sql}) TO {_quote_sql_string(str(staged_manifest))} (FORMAT PARQUET)"
-    )
+    """COPY the query's result to the staged manifest; return its row count.
+
+    ``sql`` is tenant-supplied on constrained connections.
+    ``_reject_non_single_select`` calls ``connection.extract_statements``
+    to parse the input **without executing it** and raises ``ValueError``
+    on any payload that is not exactly one SELECT statement — including
+    multi-statement injections separated by semicolons (e.g.
+    ``SELECT ...; CREATE TABLE pwned AS SELECT ...``), DDL-only input, or
+    syntactically invalid strings.  ``connection.sql()`` is **not** the
+    guard: on duckdb 1.5.5 it silently executes all statements and returns
+    ``None`` for multi-statement input.  ``write_parquet`` then materialises
+    the single confirmed-safe relation to the staged path.
+
+    The row-count query only ever receives an internally-generated path,
+    so it keeps the existing f-string form.
+    """
+    _reject_non_single_select(sql)
+    connection.sql(sql).write_parquet(str(staged_manifest))
+
     (row_count,) = connection.execute(
         f"SELECT count(*) FROM read_parquet({_quote_sql_string(str(staged_manifest))})"
     ).fetchone() or (0,)
--- a/tests/test_catalog_curation.py
+++ b/tests/test_catalog_curation.py
@@ -813,6 +813,147 @@ def test_constrained_connection_confines_sql_to_the_catalog(tmp_path: Path) -> N
         connection.close()
 
 
+_MULTI_STATEMENT_PAYLOADS = [
+    # (id, sql_template, expected_refused)
+    # The sql_template may include {decoy} which is replaced at runtime with
+    # a path inside the constrained connection's known writable directory.
+    pytest.param(
+        "SELECT 1) TO {decoy} ...; CREATE TABLE p(x TEXT); --",
+        True,
+        id="copy-escape-old-pr-shape",
+    ),
+    pytest.param(
+        "SELECT 1; CREATE TABLE pwned AS SELECT 1 AS x",
+        True,
+        id="direct-multistatement-select-plus-create",
+    ),
+    pytest.param(
+        "CREATE TABLE t(x INT)",
+        True,
+        id="ddl-only-no-select",
+    ),
+    pytest.param(
+        "SELECT episode_id FROM episodes",
+        False,
+        id="legitimate-single-select",
+    ),
+]
+
+
+@pytest.mark.parametrize("sql_template,expected_refused", _MULTI_STATEMENT_PAYLOADS)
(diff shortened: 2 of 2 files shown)


kstonekuan (changes requested):
Thanks for the report. Not merging this version, it doesn't close the hole.

The premise is wrong on the duckdb we pin (`duckdb>=1.5.5`, resolved to 1.5.5). `connection.sql()` doesn't reject a multi-statement batch. It runs every statement and returns `None` for the relation:

```
>>> c.sql("SELECT 1; CREATE TABLE pwned(x TEXT)")
None
>>> c.execute("SELECT table_name FROM duckdb_tables()").fetchall()
[("pwned",)]
```

So on the patched code a tenant doesn't have to escape a `COPY (` wrapper at all. A plain two-statement string runs both, creates the table, then fails on `None.write_parquet` with an `AttributeError`. I ran that against your branch applied to current main and read the injected row back out of the connection. The escape survives in a simpler form than the one you reported, and what the caller sees is an unhandled `AttributeError`.

Your regression test passes because its payload is a syntax error once the `COPY (` prefix is gone: `SELECT 1) TO ...` can't parse on its own. It pins the old payload, not the property. The property you want is "exactly one statement, and it's a query", stated so it doesn't depend on the shape of the wrapper that used to be there.

`duckdb.extract_statements()` gives you the count directly and it's in the pinned version. I'd normally leave the approach to you, but this is a security fix and guessing is the wrong game, so take that or something you can argue is stronger.

Two other things:

- Your branch is based on `871800b`, well behind main. Rebase before the next push. Your actual change is only the two files, so nothing's at risk, but I had to isolate it to see that.
- Check whether `write_parquet` and `COPY ... (FORMAT PARQUET)` produce the same file. If the manifest bytes change, say so and say what depends on them.

Keep specifics on the advisory where practical.

kstonekuan:
The guard you need already exists in this repo.

`_reject_non_single_select` in `packages/hflow-server/src/hflow_server/_curation.py:154` does the right check: parses with `extract_statements`, requires exactly one statement whose type is `SELECT`. Its docstring describes the same multi-statement behaviour you reported, so this was already understood one layer up and the library never got the same treatment.

Attack shapes through it:

| Payload | Result |
| --- | --- |
| `SELECT 1) TO ... (FORMAT PARQUET); CREATE TABLE p(x TEXT); --` | refused at parse |
| `SELECT ...; CREATE TABLE pwned AS SELECT ...` | refused |
| `CREATE TABLE t(x INT)` | refused |
| `SELECT episode_id FROM episodes` | allowed |

Note the third row. A lone `CREATE TABLE` is one statement, so counting isn't enough on its own and the type check is doing real work. The old `COPY (` wrapper turned that case into a syntax error by accident, and a fix that only counts would open it.

Lift that rule into `curation.py` so `curate()` refuses on its own. The server keeps its copy for now since it maps the refusal to a 400 with a usable message, and I'll open a separate issue to dedupe the two once this lands.

Raghav7-tech:
@kstonekuan 
Thanks for the thorough review — genuinely appreciate the depth here, especially catching that connection.sql() doesn't reject multi-statement input on duckdb==1.5.5. I verified it myself and you're right:

python
>>> c.sql("SELECT 1; CREATE TABLE pwned(x TEXT)")
None
>>> c.execute("SELECT table_name FROM duckdb_tables()").fetchall()
[("pwned",)]

I mistakenly assumed connection.sql()'s single-expression parsing would reject a second statement the way it does in some other contexts — it doesn't on this version, it just silently runs both and returns None for the relation. My fix only changed the failure mode, not the actual vulnerability, and my regression test only passed because the payload was shaped around the old COPY(...) wrapper syntax rather than testing the real property.

Reworking it now using extract_statements() per your pointer to _reject_non_single_select — lifting that same guard into curation.py so curate() rejects on its own, matching your suggestion. Rebasing onto current main first, then re-verifying against all four of your attack-shape cases specifically (including running each through extract_statements() directly before wiring them into the test, this time). Will push an update once I've confirmed all four behave correctly and the full suite passes clean.

Raghav7-tech:
Thanks again for the detailed review — you were right on all counts. Updated with the corrected fix:

- Replaced `connection.sql()` as the guard with `extract_statements()`, lifted from `_reject_non_single_select` in `hflow-server`, into `curation.py` directly. Verified on the actual pinned `duckdb==1.5.5` that `connection.sql()` silently executes multi-statement input and returns `None` rather than rejecting it — confirmed your repro exactly.
- Replaced the old regression test with a parametrized version covering all four of your attack-shape cases (COPY-escape, direct multi-statement, DDL-only, legitimate single SELECT). Verified it fails against the previous (broken) fix and passes against this one, for all four cases.
- Added a dedicated test asserting `write_parquet()` and the old `COPY ... (FORMAT PARQUET)` produce byte-identical output (SHA-256 match) — that assertion is now permanent, not a one-off check.
- Rebased onto current `main`.
- Full `curation` suite passes with no regressions beyond the same pre-existing, unrelated ffmpeg/Windows-path issues.

Appreciate you catching this — the first version would have shipped a real bypass with a false sense of security from a passing (but wrongly-shaped) test.

kstonekuan (approved):
This closes it.

Verified on `07d8ac9`, gate clean, 1,211 passed, 6 skipped.

Every attack shape refused with zero injected tables: the advisory payload, the plain `SELECT ...; CREATE TABLE pwned AS ...`, a lone `CREATE TABLE`, and `CREATE ...; SELECT 1` all raise `ValueError` before anything is written.

Every legitimate shape still works, including the ones a narrow guard tends to break: a CTE, a trailing semicolon, a leading comment, and a semicolon inside a string literal.

Both halves of the check are load-bearing and your parametrized table catches each independently. Dropping the count check fails one case, dropping the type check a different one, dropping the guard call three.

I closed the manifest question I left you, since it was mine: `write_parquet` and `COPY ... (FORMAT PARQUET)` produce byte-identical files here. Same size, same sha256, same row groups, same compression, same `created_by`. Nothing downstream changes.

Two notes, neither blocking, both mine to carry:

The guard changes what `curate()` accepts at the edges. `PIVOT` worked before and is refused now; `DESCRIBE` / `SUMMARIZE` were parser errors before and pass now. That falls out of classifying by statement type, and it lines the library up with what the server has always enforced, so I think it's the right direction. Still needs writing down and pinning, and I'm filing that separately instead of holding this.

Your first parametrized payload is labelled `copy-escape-old-pr-shape` but the string isn't that shape, so it's refused as a parse error, not as a second statement. No harm, and probably unavoidable now the `COPY (` wrapper is gone, which is what makes the old payload indistinguishable from garbage at this layer. The two rows below it are the ones doing the work.

Merging.

kstonekuan:
Shipped in [v0.2.4](https://github.com/Hebbian-Robotics/hflow/releases/tag/v0.2.4), and the advisory is published: GHSA-57x4-5jwm-2xg8.

Final severity is Low (3.1, `CVSS:3.1/AV:N/AC:H/PR:L/UI:N/S:U/C:N/I:L/A:N`), down from the High it was filed at. Two things settled it. `hflow serve` was never affected, since its curation routes have always applied the same one-SELECT rule before any interpolation, so reaching the vulnerable path meant calling `curate()` directly from an integration without that check. And the lockdown held everywhere it mattered: reading host files, writing outside the allowlist, `ATTACH`, `INSTALL`, and lifting `enable_external_access` all still refuse, so what an attacker got was arbitrary statements against a throwaway in-memory copy of data they could already query.

None of that takes anything away from the report. The interpolation was real, it broke a contract the docs state in three places, and you found it and fixed it. You are credited as reporter on the advisory.

Verified the published wheel carries the fix: `pip install hflow==0.2.4`, then a two-statement string, a lone `CREATE TABLE`, and the original payload all raise, and `SELECT 1` still works.

Issue #260, opened 2026-08-29
Build AI evaluation: compare crashes with a traceback on a missing file while run refuses cleanly


kstonekuan:
### Current behavior

The Build AI evaluation CLI has two subcommands. Given the same kind of bad input, a path that is not there, they behave differently. Measured on current main (`84c2921`):

```console
$ evaluate.py run --dataset 10k --model x --base-url http://127.0.0.1:1 \
      --hand-count-prompt /tmp/nope.txt --allow-missing-api-key --limit 1
usage: evaluate.py [-h] {run,compare} ...
evaluate.py: error: [Errno 2] No such file or directory: '/tmp/nope.txt'
exit=2, no traceback

$ evaluate.py compare /tmp/nope.json
Traceback (most recent call last):
  ...
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/nope.json'
exit=1, 17 lines of stderr
```

`run` turns a missing file into a clean error and exit 2. `compare` lets it escape as a traceback and exit 1.

The reason is visible in `main` (`examples/build_ai_evaluation/evaluate.py:933-946`):

```python
match arguments.command:
    case "run":
        try:
            configuration = _run_configuration_from_arguments(arguments)
            run_evaluation(configuration, download=arguments.download)
        except (FileNotFoundError, RuntimeError, ValueError) as error:
            parser.error(str(error))
    case "compare":
        compare_summaries(arguments.summaries)
```

Two adjacent arms. One is wrapped and one is not.

### Why the exit code matters as much as the traceback

`docs/FORMAT.md:196-208` defines the three-value convention this repo uses: `0` success, `1` ran and found something to report, `2` bad input and nothing useful happened. A path that does not exist is `2`. `compare` currently reports `1`, which in this project's vocabulary means the comparison ran and had something to say about the results.

So a script driving this in CI cannot distinguish "the two runs disagreed" from "you typed the filename wrong", and the traceback is the only clue.

### What to build

Make `compare` refuse bad input the way `run` already does.

Copying the `try`/`except` onto the second arm is the obvious move and worth thinking past before you write it: a third subcommand added later would be equally easy to forget, in exactly the way this one was. Consider whether the handling belongs around the `match` rather than inside one arm of it, and say in the PR which you chose and why. If you keep it per-arm, say what would stop the next subcommand from missing it.

[...]

PR #272, opened 2026-08-30
fix(egosuite): refuse invalid compare summaries

### LINKED ISSUE #266: EgoSuite evaluation: compare crashes with a traceback on a missing or malformed summary while labels and run refuse cleanly
### Current behavior

The EgoSuite evaluation CLI has three subcommands. Given a path that is not there, two of them refuse cleanly and one crashes. Measured on current main (`100848f`):

```console
$ evaluate.py compare /tmp/nope.json
  ...
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/nope.json'
exit=1, 17 lines of stderr

$ evaluate.py compare /tmp/bad.json          # a real file holding "not json"
  ...
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
exit=1, 18 lines of stderr
```

The reason is visible in `main` (`examples/egosuite_evaluation/evaluate.py:1362-1386`):

```python
match arguments.command:
    case "labels":
        try:
            ...
        except (FileNotFoundError, RuntimeError, ValueError) as error:
            parser.error(str(error))
    case "run":
        try:
            run_evaluation(_run_configuration_from_arguments(arguments))
        except (FileNotFoundError, RuntimeError, ValueError) as error:
            parser.error(str(error))
    case "compare":
        compare_summaries(arguments.summaries)
```

Three arms. Two wrapped, one not, and the wrapping is copied verbatim between the two that have it.

### Why the exit code matters as much as the traceback

`docs/FORMAT.md:196-208` defines the convention this repo uses: `0` success, `1` ran and found something to report, `2` bad input and nothing useful happened. A path that does not exist is `2`. `compare` reports `1`, which in this project's vocabulary means the comparison ran and had something to say about the results.

A script driving this in CI cannot tell "the two runs disagreed" from "you typed the filename wrong".

### What to build

Make `compare` refuse bad input the way `labels` and `run` already do.

Copying the `try`/`except` onto the third arm is the obvious move and worth thinking past before you write it. This example already has three subcommands and the handler is already duplicated across two of them, so the pattern of forgetting it has happened once and the duplication is there to be read. Consider whether the handling belongs around the `match` rather than inside each arm, and say in the PR which you chose and why. If you keep it per-arm, say what would stop the fourth subcommand from missing it.

### Definition of done

1. `evaluate.py compare <missing path>` prints a message naming the path and exits 2, with no traceback.
2. A summary file that exists but does not hold the expected JSON is also refused rather than escaping as a decoder traceback. Say what exit code you chose for it and why.
3. `labels` and `run` behave exactly as they do today, including their exit codes and messages for the same class of error.
4. A successful `compare` over real summaries still exits 0 and prints what it prints today.
5. Tests cover the refusals. `examples/egosuite_evaluation/tests/test_evaluation.py` already imports from `evaluate.py` directly, so there is a pattern to follow.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

The example has its own project, so its tests do not run in the root suite. Run both.

### Notes

No network and no API key needed: every path in scope fails before a request is made.

Do not change what `compare` prints on success or how it computes anything. This issue is entirely about what happens when the input is wrong.

#260 is the same defect in `examples/build_ai_evaluation`, and is assigned. The two are separate files in separate projects, so they do not conflict, but whoever finishes second should read what the first one chose and say whether they matched it.

Iamnotphage:
Closes #266.

## Summary

* report missing and malformed comparison summaries as command-line input errors
* centralize the existing `FileNotFoundError`, `RuntimeError`, and `ValueError` boundary around subcommand dispatch
* add outcome-focused CLI coverage for both refusal paths and the successful comparison table

## Why

`compare` previously allowed filesystem and JSON decoder failures to escape as Python tracebacks, returning exit code 1 even though no comparison had taken place. Both conditions now exit 2 because the input could not be used, matching the convention documented in `docs/FORMAT.md`.

The error boundary now wraps the command dispatch itself instead of being copied into a third match arm. This preserves the existing behavior of `labels` and `run` while making the same protection the default for future subcommands. Successful comparisons retain their existing computation and output unchanged.

The related #260 was assigned but did not yet have a PR when this change was prepared, so there was no submitted implementation to align with.

## Validation

* `uv sync --locked --all-extras` — passed
* targeted regression test before the implementation — 2 failed, 1 passed
* targeted regression test after the implementation — 3 passed
* `uv run ruff check --fix` — passed
* `uv run ruff format` — passed
* `uv run ty check` — passed
* `uv run pytest -q` — 1,195 passed, 6 skipped
* `uv run --locked --project examples/egosuite_evaluation pytest -q examples/egosuite_evaluation/tests` — 11 passed
* `git diff --check` — passed

## Checklist

* [x] I added outcome-focused tests for the changed behavior.
* [x] Documentation remains current; this change brings `compare` into the existing documented exit-code convention.
* [x] I ran `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check`.
* [x] I ran both the root and EgoSuite-specific pytest suites.
* [x] I did not add recordings, generated media, credentials, private URLs, or runtime artifacts.
* [x] Stored-data compatibility is unchanged; no transform behavior version bump is required.

### DIFF
--- a/examples/egosuite_evaluation/evaluate.py
+++ b/examples/egosuite_evaluation/evaluate.py
@@ -1482,9 +1482,9 @@ def _argument_parser() -> argparse.ArgumentParser:
 def main() -> None:
     parser = _argument_parser()
     arguments = parser.parse_args()
-    match arguments.command:
-        case "labels":
-            try:
+    try:
+        match arguments.command:
+            case "labels":
                 labels_by_source = _labels_by_source_from_arguments(arguments)
                 _print_reference_summary(labels_by_source)
                 write_label_report(
@@ -1499,17 +1499,14 @@ def main() -> None:
                     output_path=arguments.output,
                 )
                 print(f"\nlabels: {arguments.output}")
-            except (FileNotFoundError, RuntimeError, ValueError) as error:
-                parser.error(str(error))
-        case "run":
-            try:
+            case "run":
                 run_evaluation(_run_configuration_from_arguments(arguments))
-            except (FileNotFoundError, RuntimeError, ValueError) as error:
-                parser.error(str(error))
-        case "compare":
-            compare_summaries(arguments.summaries)
-        case unknown_command:
-            raise AssertionError(f"unhandled command: {unknown_command}")
+            case "compare":
+                compare_summaries(arguments.summaries)
+            case unknown_command:
+                raise AssertionError(f"unhandled command: {unknown_command}")
+    except (FileNotFoundError, RuntimeError, ValueError) as error:
+        parser.error(str(error))
 
 
 if __name__ == "__main__":
--- a/examples/egosuite_evaluation/tests/test_evaluation.py
+++ b/examples/egosuite_evaluation/tests/test_evaluation.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 import math
 import sys
 from pathlib import Path
@@ -13,6 +14,7 @@ from examples.egosuite_evaluation.evaluate import (
     ProjectedHandFrameLabel,
     _evaluation_result_summary,
     _selected_frame_indices,
+    main,
     parse_hand_count_response,
     select_episode_paths,
     select_stratified_labels,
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM. Hoisting the handler around the `match` covers the next subcommand too.

Validated on `69de54a`: root suite 1,195 passed, 6 skipped; example suite 11 passed; ruff, format, ty clean on both projects. Moved `compare` back outside the `try` and confirmed exactly your two refusal tests go red, other nine stay green.

One ask, same as always and not a rule: leave the `good first issue` pool for newcomers where you can. You've got three merges now, and the starter queue is what keeps new people arriving. Higher-leverage work if you want it: run HFlow against a real corpus, Egocentric-10K or Egocentric-100K on HF, and report what breaks, what's slow, what's awkward. See #287.

PR #273, opened 2026-08-30
feat(lerobot): export an HFlow-curated selection as a loadable LeRobot Dataset v3 (fixes #190)

### LINKED ISSUE #190: LeRobot: export an HFlow-curated selection as a loadable v3 dataset

ayam04:
Closes #190

## What this adds

A standalone export callable and command under `examples/lerobot/` that turns an HFlow curation selection into a **local LeRobot Dataset v3 repository** loadable by existing LeRobot tooling:

- `examples/lerobot/export.py` — public `export()` callable + `--manifest` / `--sql` CLI
- `examples/README.md` — documented usage (manifest and SQL forms)
- `tests/test_lerobot_export.py` — 7 outcome-focused tests

## How it works

1. **Selection**: reads a `hflow curate` manifest parquet or a SQL query (exactly one required).
2. **Provenance**: each row's `metadata_json` is resolved to `source_dataset`, `source_revision`, `source_episode_index`, `task`, `embodiment` — the exact #189 contract keys the converter stamps in `episode/v1` (no contract extension: nothing new is added to the converter's stamps).
3. **Immutability**: the selection must reference exactly one repository and one immutable 40-hex commit; branch names and mixed sources fail loudly before any output exists.
4. **Materialization**: source video chunks are copied byte-for-byte (no re-encode, frame timing preserved) and the selected episode data rows are sliced from their source chunk parquets, so camera content, feature schema, dtypes, shapes, and frame timing match the source exactly. Episode indexes are renumbered sequentially in selection order; non-contiguous selections come out in deterministic selection order.
5. **Validation**: structural v3 checks run always (info.json coherence, per-episode windows == length, video files present per referenced camera); when the official `lerobot` package is installed, the staged directory is additionally loaded through `LeRobotDataset`.
6. **Atomic publish**: everything is staged under a temp dir; the destination is replaced only after creation + validation succeed. A failure never leaves a partial or destroyed destination.
7. **Provenance out**: `export-provenance.json` (source repo, source commit, exporter version, selection SQL/manifest name, selected source episode indexes, video + data parquet sha256s) + a dataset-card draft `README.md` with enough to reproduce the selection.

## Failure behavior (all tested)

- missing provenance (`metadata_json` absent) → abort, no destination
- mixed source repositories or revisions → abort
- non-immutable revision (e.g. `main`) → abort
- duplicate source episode indexes → abort
- selected episode not present in the source → abort
- staged dataset failing structural/API validation → abort

Temporary output never replaces a previously valid destination until creation and validation succeed.

## Tests

`tests/test_lerobot_export.py` (synthetic v3 corpus, network helpers monkeypatched, mirroring the converter tests):

- non-contiguous selection (episodes 0 + 2 of 4): output contains exactly {0, 2} in order, episode/data parquets renumbered (windows 0-60, 60-130), frame_index preserved per source episode, videos byte-identical, provenance + card written
- mixed repositories, non-immutable revision, missing provenance, missing source episode, duplicate selection → all fail with no destination created
- SQL selection path produces the same outcome as the manifest path

## Verification

- `ruff check`: clean (both files)
- `ty check`: clean (both files)
- `pytest tests/test_lerobot_export.py`: 7 passed
- full suite: 985 passed / 6 skipped; the only failures are 3 pre-existing `tests/test_cli.py` cases that fail identically on a clean tree (verified via stash)
- the official-API validation pass runs in CI where `lerobot` is installed; the structural pass runs everywhere

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -280,7 +280,8 @@ LeRobot Dataset v3 repository containing exactly the selected episodes.
 Episode provenance stamped by the converter (`source_dataset`,
 `source_revision`, `source_episode_index` in `episode/v1` metadata) is
 resolved back to the source; the exporter copies the source video chunks
-byte-for-byte and slices the selected data rows from the source chunk
+byte-for-byte (each source (chunk, file) keeping its identity in the
+destination path) and slices the selected data rows from the source chunk
 parquets, so cameras, feature schema, dtypes, shapes, and frame timing
 match the source. Episode indexes are renumbered sequentially in selection
 order. The exporter drives the public `hflow import lerobot` entry point to
--- a/examples/README.md
+++ b/examples/README.md
@@ -272,6 +272,50 @@ Output: one canonical MCAP per episode at
 - Public API: [`hflow.importers.lerobot`](../src/hflow/importers/lerobot.py)
 - Compatibility wrapper: [`lerobot/prepare.py`](./lerobot/prepare.py)
 
+### Exporting a curated selection back to LeRobot Dataset v3
+
+`lerobot/export.py` turns an HFlow curation selection (a `hflow curate`
+manifest parquet, or a SQL query over the catalog) into a loadable local
+LeRobot Dataset v3 repository containing exactly the selected episodes.
+Episode provenance stamped by the converter (`source_dataset`,
+`source_revision`, `source_episode_index` in `episode/v1` metadata) is
+resolved back to the source; the exporter copies the source video chunks
+byte-for-byte and slices the selected data rows from the source chunk
+parquets, so cameras, feature schema, dtypes, shapes, and frame timing
+match the source. Episode indexes are renumbered sequentially in selection
+order. The exporter drives the public `hflow import lerobot` entry point to
+materialize the source archive (meta, data chunks, video chunks) and copies
+those chunks byte-for-byte. A dataset-card draft (`README.md`) and
+`export-provenance.json` record the source repository, source commit,
+exporter version, selection SQL/manifest, and selected source episode
+indexes.
+
+From a curation manifest:
+
+```
+uv run python examples/lerobot/export.py \
+  --manifest ./manifest.parquet \
+  --destination ./data/curated_lerobot \
+  --camera-keys observation.images.up,observation.images.side
+```
+
+From a SQL query over the catalog:
+
+```
+uv run python examples/lerobot/export.py \
+  --sql "SELECT episode_id, metadata_json FROM episodes_latest WHERE status = 'ok'" \
+  --destination ./data/curated_lerobot \
+  --camera-keys observation.images.up,observation.images.side
+```
+
+Failures abort before any output is published: mixed source repositories
+or revisions, non-immutable revisions, missing provenance, missing source
+episodes, duplicate selections, and a staged dataset that fails structural
+or (when `lerobot` is installed) official-API validation never replace a
+previously valid destination.
+
+Code: [`lerobot/export.py`](./lerobot/export.py)
+
 
 ## Example requirements
 
--- a/examples/lerobot/export.py
+++ b/examples/lerobot/export.py
@@ -12,6 +12,10 @@ episode data rows are sliced from their source chunk parquets, so camera
 content, feature schema, dtypes, shapes, and frame timing match the source
 exactly. Episode indexes are renumbered sequentially in selection order.
 
+The source archive is materialized through the public
+``hflow.import_lerobot_dataset`` entry point (``hflow import lerobot``) and
+consumed from the ``_lerobot_cache`` artifacts that entry point writes.
+
 Failures fail before any publishable output is written: mixed source
 repositories or revisions, missing provenance, missing source episodes,
 duplicate selections, and incompatible feature schemas all abort without
@@ -35,21 +39,11 @@ import duckdb
 import pyarrow as pa
 import pyarrow.parquet as pq
 
-try:
-    import examples.lerobot.prepare as prepare
-except ModuleNotFoundError:
-    # Direct `python examples/lerobot/export.py` runs: repository root is
-    # not on sys.path, so add it (examples/ is not a package).
-    import sys as _sys
-    from pathlib import Path as _Path
-    _sys.path.insert(0, str(_Path(__file__).resolve().parents[2]))
-    import examples.lerobot.prepare as prepare
-
+import hflow
 
 EXPORT_VERSION = "lerobot-export-v1"
 
 _COMMIT_SHA_RE = re.compile(r"^[0-9a-f]{40}$")
-_CAMERA_KEY_RE = re.compile(r"^observation\.images\.[A-Za-z0-9_]+$")
 
 
 @dataclass(frozen=True)
@@ -151,6 +145,100 @@ def _sha256(path: Path) -> str:
     return h.hexdigest()
 
 
+def _read_corpus_from_cache(cache_dir: Path) -> dict:
+    """Reconstruct the source corpus from the importer's materialized archive.
+
+    ``hflow.import_lerobot_dataset`` writes meta/info.json, the
+    meta/episodes parquets, data chunks, and video chunks under
+    ``<output_dir>/_lerobot_cache``. Export consumes exactly those public
+    artifacts; the dictionary mirrors the archive shape the converter used.
+    """
+    info_path = cache_dir / "meta" / "info.json"
+    if not info_path.exists():
+        raise FileNotFoundError(
+            f"materialized archive has no {info_path}; the public importer writes it during export"
+        )
+    info = json.loads(info_path.read_text())
+
+    episodes_dir = cache_dir / "meta" / "episodes"
+    ep_files = sorted(episodes_dir.rglob("*.parquet"))
+    if not ep_files:
+        raise FileNotFoundError(
+            f"materialized archive has no episode parquets under {episodes_dir}"
+        )
+
+    conn = duckdb.connect()
+    try:
+        rows: list[dict] = []
+        windows: dict[int, dict[str, dict]] = {}
+        for ep_file in ep_files:
+            quoted = str(ep_file).replace("'", "''")
+            cols = [
+                d[0]
+                for d in conn.execute(f"SELECT * FROM read_parquet('{quoted}') LIMIT 0").description
+            ]
+            data = conn.execute(f"SELECT * FROM read_parquet('{quoted}')").fetchall()
+            for row in data:
+                d = dict(zip(cols, row, strict=True))
+                ep_idx = int(d["episode_index"])
+                tasks = d.get("tasks")
+                if isinstance(tasks, list):
+                    task = str(tasks[0]) if tasks else ""
+                else:
+                    task = str(tasks or "")
+                rows.append(
+                    {
+                        "episode_index": ep_idx,
+                        "task": task,
+                        "length": int(d["length"]),
+                        "data_chunk": str(d["data/chunk_index"]).split("/")[-1],
+                        "data_file": str(d["data/file_index"]).split("/")[-1],
+                        "data_from": int(d["dataset_from_index"]),
+                        "data_to": int(d["dataset_to_index"]),
+                    }
+                )
+                cam_windows: dict[str, dict] = {}
+                for col in cols:
+                    if col.startswith("videos/") and col.endswith("/from_timestamp"):
+                        cam = col.split("/")[1]
+                        cam_windows[cam] = {
+                            "chunk_index": str(d.get(f"videos/{cam}/chunk_index") or ""),
+                            "file_index": str(d.get(f"videos/{cam}/file_index") or ""),
+                            "from_timestamp": float(d.get(f"videos/{cam}/from_timestamp") or 0.0),
+                            "to_timestamp": float(d.get(f"videos/{cam}/to_timestamp") or 0.0),
+                        }
+                windows.setdefault(ep_idx, {}).update(cam_windows)
+    finally:
+        conn.close()
+
+    rows.sort(key=lambda e: e["episode_index"])
+    for e in rows:
+        e["video_windows"] = dict(windows.get(e["episode_index"], {}))
+    return {"info": info, "episodes": rows, "cache_dir": cache_dir}
+
+
+def _materialize_source_archive(
+    src_ds: str, src_rev: str, cache_dir: Path, camera_keys: tuple[str, ...]
+) -> dict:
+    """Materialize the source archive through the public importer.
+
+    ``hflow.import_lerobot_dataset`` (exported from ``hflow/__init__.py``;
+    ``hflow import lerobot`` on the CLI) resolves the immutable revision,
+    validates the camera keys, and downloads the corpus metadata, data
+    chunks, and video chunks under ``<output_dir>/_lerobot_cache``. Export
+    consumes that archive; the canonical landing MCAPs the importer also
+    writes are a side effect export does not use.
+    """
+    hflow.import_lerobot_dataset(
+        dataset_repo=src_ds,
+        revision=src_rev,
+        output_dir=cache_dir.parent,
+        episode_index=None,
+        camera_keys=camera_keys,
+    )
+    return _read_corpus_from_cache(cache_dir)
+
+
 def _episode_rows(table: pa.Table) -> list[dict]:
     cols = table.column_names
     out: list[dict] = []
@@ -172,16 +260,16 @@ def _write_v3_repository(
     cache_dir: Path = corpus["cache_dir"]
     info = corpus["info"]
     src_by_index = {e["episode_index"]: e for e in corpus["episodes"]}
-    base = f"https://huggingface.co/datasets/{src_ds}/resolve/{src_rev}"
 
-    # download data + video chunks (same access path the converter uses)
+    # Data and video chunks come from the archive the public importer
+    # materialized; a missing file is a broken materialization, not a
+    # download to retry quietly here.
     def _fetch_data(ep: dict) -> Path:
         chunk = int(ep["data_chunk"])
         file = int(ep["data_file"])
-        rel = corpus["data_path"].format(chunk_index=chunk, file_index=file)
         local = cache_dir / "data" / f"chunk-{chunk:06d}-file-{file:06d}.parquet"
         if not local.exists():
-            prepare._download_file(f"{base}/{rel}", local)
+            raise FileNotFoundError(f"source data chunk missing: {local}")
         return local
 
     def _fetch_video(cam: str, vw: dict, ep: dict) -> Path:
@@ -190,17 +278,9 @@ def _write_v3_repository(
             if str(vw.get("chunk_index", "")).isdigit()
             else int(ep["data_chunk"])
         )
-        file = (
-            int(vw["file_index"])
-            if str(vw.get("file_index", "")).isdigit()
-            else int(ep["data_file"])
-        )
-        rel = corpus["video_path"].format(
-            chunk_index=chunk, file_index=file, video_key=cam, camera_key=cam
-        )
         local = cache_dir / "videos" / f"{cam.replace('/', '_').replace('.', '_')}-chunk{chunk}.mp4"
         if not local.exists():
-            prepare._download_file(f"{base}/{rel}", local)
+            raise FileNotFoundError(f"source video chunk missing: {local}")
         return local
 
     episodes_dir = destination / "meta" / "episodes" / "chunk-000"
@@ -447,8 +527,7 @@ def export(
     *,
     manifest: Path | None = None,
     sql: str | None = None,
-    catalog_root: Path | None = None,
-    camera_keys: tuple[str, ...] | None = None,
+    camera_keys: tuple[str, ...],
 ) -> Path:
     """Export the curated selection into a new local LeRobot v3 repository."""
     rows = _read_selection(manifest, sql)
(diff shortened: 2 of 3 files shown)


kstonekuan (changes requested):
Three things, and the first one matters most.

**The tests have never run.** `tests/test_lerobot_export.py:21` is `pytest.importorskip("examples.lerobot.prepare")` at module scope. `examples/` has no `__init__.py`, so that import fails from the root suite and the whole file collapses into one silent skip. On your own branch at `a8ed43f`:

```
$ uv run pytest -q tests/test_lerobot_export.py -rs
SKIPPED [1] tests/test_lerobot_export.py:21: could not import
  examples.lerobot.prepare: No module named examples
1 skipped in 0.06s
```

Not seven tests. One skip. Every failure behaviour the description lists as tested (missing provenance, mixed repos, a branch name instead of a commit, duplicate indexes) is unverified, and 389 lines of test code have never executed on any tree.

`importorskip` is the wrong tool here. It's for a genuinely optional dependency, and it turned a hard import error into a green run. Whatever you do about the import path, make the failure loud: if the module can't be imported, that's a broken test file, not a skipped one. `examples/egosuite_evaluation/tests/test_evaluation.py` shows the sys.path approach we already use, `examples/build_ai_evaluation` shows the separate-project approach. Either's fine, say which you picked.

**`export.py` no longer imports on main.** `301c684` landed four hours after you opened this and made LeRobot import first-class: the converter moved out of `examples/lerobot/prepare.py` into `src/hflow/importers/lerobot.py`, and `prepare.py` is now nine lines forwarding to the CLI. Your `prepare._download_file`, `prepare._ensure_source_archive`, and `prepare.DatasetSource` are gone from where you call them. Not your fault, and the skip is why nothing told you.

Repoint at the public surface, not the new private location. `hflow.import_lerobot_dataset` is exported from `hflow/__init__.py` and `hflow import lerobot` is a CLI command. If you still need something only the private helpers give you, say what and why on the issue and we'll look at making it public.

**Placement, so you're not guessing:** export stays under `examples/` for now. Import going first-class doesn't automatically drag export with it, and I'd rather see this land and get used before deciding it belongs in the shipped API. Separate question, I'll open it separately. Don't expand this PR to chase it.

Rebase onto current main before the next push. Once the tests run I'll review the export logic properly. I haven't yet, since reviewing behaviour no test has exercised isn't worth either of our time.

ayam04:
All three points addressed in 88be75b, rebased onto current main:

**1. The tests run now.** Dropped the module-scope pytest.importorskip (it turned a hard import error into one silent skip). The suite uses the sys.path namespace-import approach from examples/egosuite_evaluation/tests — I picked that one over the separate-project approach since tests/test_lerobot_export.py is colocated with the root suite and the exporter is a module, not its own project. `uv run pytest -q tests/test_lerobot_export.py -rs` now runs 8 tests, 0 skips.

**2. Repointed at the public surface.** examples/lerobot/export.py no longer imports examples.lerobot.prepare at all. The source archive is materialized through hflow.import_lerobot_dataset (exported from hflow/__init__.py, hflow import lerobot on the CLI): export calls it once per run and consumes exactly the _lerobot_cache artifacts it writes (meta/info.json, meta/episodes parquets, data chunks, video chunks). Chunk "fetching" is now an existence check against that archive with a loud FileNotFoundError. Two consequences made explicit: --camera-keys is now required (the importer needs valid camera keys up front; the old "default: all source cameras" branch could not resolve them without the private helpers), and the importer's canonical-landing MCAPs are a side effect export does not use. If you would rather have a public materialize-only entry point, say the word and I will wire export to it. A new regression test pins the repoint: it asserts export drives hflow.import_lerobot_dataset with the resolved repo/revision, all episodes, and the requested camera keys.

**3. Placement unchanged.** Export stays under examples/; this PR only broadens to the new archive shape.

Validation on the pushed head: pytest tests/test_lerobot_export.py 8 passed (0 skipped); ruff check + ruff format --check clean; ty check clean; full suite 1256 passed / 6 skipped / 2 failed — both failures reproduced on clean main without this branch (catalog_ui flake re-passed on rerun; end_to_end needs filter_script:v which this WSL ffmpeg build lacks), so the export tests are the only new-signal tests here.

kstonekuan (changes requested):
Both blockers are fixed. The tests run (8 passed, 0 skips), and `export.py` no longer imports `examples.lerobot.prepare` at all: it drives `hflow.import_lerobot_dataset` and consumes the cache artifacts. The sys.path choice matches `examples/egosuite_evaluation/tests`, which is the right precedent for a module colocated with the root suite. Validated on `fbfd553`: gate clean, 1,259 passed, 6 skipped.

So now the review I said I'd do, and there are two more of the same shape as the first one.

**`_validate_v3` can be deleted and no test notices.** It's called on the tested path, at `export.py:580`, right before the atomic publish. I replaced its entire body with `return` and all 8 tests still pass. The description sells it as one of the seven guarantees:

> **Validation**: structural v3 checks run always (info.json coherence, per-episode windows == length, video files present per referenced camera)

Nothing exercises any of those three. No test stages a dataset with an inconsistent `info.json`, a window that disagrees with `length`, or a camera referenced without its video file, so the pass has never demonstrably rejected anything. That's the same defect as the `importorskip`: a guarantee whose test can't tell whether it holds.

**`_sha256` can return a constant and no test notices.** Same method: `return "0" * 64` and all 8 pass. It feeds `video_sha256` and `data_parquet_sha256` in `export-provenance.json` (`:422-423`). `test_export_noncontiguous_selection` reads that file back, so it's on the path, but it asserts nothing about the digests. A provenance record whose whole purpose is reproducibility has no test that its digests correspond to the bytes.

Neither is hard to close. For the digests, assert one equals the file's actual hash, and assert two different files give different values, so a constant fails. For the validation, build a staged dataset that breaks each of the three properties and assert the refusal, then prove it by neutering the check.

What is well covered, so don't touch it: the immutability guard. Neutering `_check_immutable` fails `test_export_mixed_repositories_fail` and `test_export_nonimmutable_revision_fails`, both for the right reason. The provenance-absent, missing-episode, and duplicate-index refusals are all real too.

I haven't reviewed the byte-for-byte chunk copying or the episode renumbering yet. Same reason as before: I'd be reading code that no test constrains, and after two rounds of exactly that I'd rather the tests come first.

ayam04:
Both findings closed, rebased onto current main (616494d). Branch head c4a9f5a, mergeable.

**Digests are pinned now.** New test `test_export_provenance_digests_match_bytes` asserts each `video_sha256` and `data_parquet_sha256` equals the sha256 of the file it names, and that distinct files get distinct digests. A constant fails on the first assert.

**Validation is pinned now.** One rejection test per structural property, each staging a valid export then corrupting a single property: `test_validate_v3_rejects_incoherent_info` (code rewritten to v2), `test_validate_v3_rejects_window_length_mismatch` (episode length bumped so window != length), `test_validate_v3_rejects_missing_video` (camera video file deleted). Each asserts the exact refusal message.

**Neutering proof, your method:** I replaced `_sha256` with `return "0" * 64` and `_validate_v3` with an immediate `return`, ran the file: 4 failed (the three validation tests with "DID NOT RAISE", the digest test on its first assert), 8 passed. Restored both, 12/12 pass.

**Rebase caught a real break.** `8116404` (#328) namespaced the importer cache as `_lerobot_cache/<resolved sha>`, and export read the flat `_lerobot_cache` path, so a fresh materialization would have failed at runtime with no test noticing (the fake importer masked it). `_materialize_source_archive` now resolves the single namespaced subdirectory after the import and fails loudly on an unexpected layout; the test fake mirrors main's namespaced layout. Commit dca04c5.

**Gate evidence:** `tests/test_lerobot_export.py` 12/12; ruff check + format clean on both files; ty clean; full suite 1352 passed, 6 skipped, 1 failed (the `filter_script:v` end-to-end failure, the same pre-existing WSL env failure reproduced on clean main in the previous round). README conflict from the rebase resolved by taking main's EgoSuite and import sections and re-layering only the curated-export subsection on top.

For the parts you haven't reviewed yet: byte-for-byte chunk copying is already asserted (`v.read_bytes() == f"fake-mp4-{cam}".encode()` in `test_export_noncontiguous_selection`), and episode renumbering is pinned by the frame-index reads in the same test (ep2 restarts at 0 after ep0's 60 frames). Happy to expand either if you want them stronger before reviewing those paths.

kstonekuan (changes requested):
Both findings from last round are closed, and I checked them the same way you did:

```
_sha256 returns a constant          -> test_export_provenance_digests_match_bytes fails
_validate_v3 body becomes a no-op   -> all three new validation tests fail
```

The digest test asserting that two different files give two different values is what makes the constant impossible, so that one is properly nailed down.

Now the export logic, which is what I owed you. Two of these are wrong output rather than missing tests.

**A selection spanning more than one source video chunk publishes the wrong video bytes.** `export.py:362` writes every camera to `chunk-000/file-000.mp4` and `:363` skips the copy when that path exists, so the second source chunk never reaches the output while the episode row still carries its own window offsets. I put episode 2's video in source chunk 1 with distinguishable bytes and changed nothing else:

```
source cache:
  observation_images_up-chunk0.mp4: b'fake-mp4-observation.images.up'
  observation_images_up-chunk1.mp4: b'fake-mp4-CHUNK-ONE-observation.images.up'

exported:
  videos/observation.images.up/chunk-000/file-000.mp4: b'fake-mp4-observation.images.up'

episode rows:
  episode 0: chunk-000/file-000  to_timestamp=2.0
  episode 1: chunk-000/file-000  to_timestamp=2.4

_validate_v3 accepted it
```

The chunk-1 bytes are absent from the output entirely, and episode 1 now says its frames are at 0.0 to 2.4 seconds of a file holding episode 0's chunk. `_sha256` then stamps a digest over the wrong video in `export-provenance.json`, so the provenance record certifies it. "Byte-faithful: source video chunks are copied unchanged" is the headline claim of the module docstring, and it holds only while the whole selection sits in one chunk, which is what the fixture does: all four episodes are `chunk_index: "0"`.

Either carry the source `(chunk, file)` identity into the destination path so distinct sources land in distinct files, or refuse a multi-chunk selection outright. Refusing is a fine answer for an example, and a loud refusal beats a quiet substitution. Whichever you pick, the fixture needs a second video chunk, because right now nothing in the suite can tell the two behaviours apart.

**The exported episode parquet cannot be read back.** `:377-378` writes `data/chunk_index` as `"chunk-000"` and `data/file_index` as `f"file-{new_idx:03d}.parquet"`. The source stores those as plain indexes, and your own `_read_corpus_from_cache` plus `_fetch_data` expect that. Feeding the exporter its own output:

```
source episode 0:    data_chunk='0'          data_file='0'
exported episode 0:  data_chunk='chunk-000'  data_file='file-000.parquet'
exported episode 1:  data_chunk='chunk-000'  data_file='file-001.parquet'

int(data_chunk) -> ValueError: invalid literal for int() with base 10: 'chunk-000'
```

Two separate problems in there. The format does not match the source, and you write `info["data_path"]` yourself at `:418` as a template over integers:

```
chunk_index=0, file_index=0                       -> data/chunk-000/file-000.parquet
chunk_index='chunk-000', file_index='file-000...' -> ValueError: Unknown format code 'd' for object of type 'str'
```

So the dataset carries a path template its own episode rows cannot be substituted into. Separately, `file-001.parquet` names a file that does not exist: every row goes into the single `data/chunk-000/file-000.parquet` at `:394`, so episode 1's reference dangles. The comment at `:302` says one data parquet per episode, which is what the references describe but not what the code writes. Pick one and make both agree.

`test_export_noncontiguous_selection` already selects `data/file_index` in its query at `:270` and then never asserts on it. Asserting there would have caught this.

**The official-API validation pass has never run anywhere.** `lerobot` is not in `pyproject.toml` and no workflow installs it, so `:496-499` takes the `ImportError` branch on every machine including CI. The PR body says the pass "runs in CI where `lerobot` is installed"; that is not true today, which is why neither of the two problems above was caught by a load. Either add it as a dev extra so the pass is real, or drop the claim and let the structural checks own the validation on their own terms. If they own it, they need to check what actually breaks loadability: that each episode's referenced data file exists, and that the index fields are the type `data_path` can format.

One smaller thing while you are in there. The call to `_validate_v3` at `:591` can be deleted and all 12 tests still pass. The body is pinned now, the call is not. A test that stages a dataset the validator rejects and asserts `export()` propagates it and leaves no destination would close that, and it also pins the atomic-publish guarantee at `:596-603`, which nothing currently exercises on the failure path.

And one more in the renumbering at `:344-346`:

```python
d["frame_index"] = local_frame
d[index_col] = len(data_frames)
```

`index_col` is `"index"` if the source has that column and `"frame_index"` otherwise, so on a source without an `index` column the second line overwrites the first and `frame_index` comes out global instead of per-episode. The fixture has an `index` column, so the suite only ever takes the safe branch, and `assert frames[60][0] == 0` passes for that reason. Either drop the fallback and require `index`, or keep the two writes from colliding.

The immutability, provenance-absent, missing-episode, and duplicate-index refusals are all still real, and on the `index` branch the renumbering does what the description says: `frame_index` restarts per episode and the global index stays contiguous across the selection.

kstonekuan:
Heads-up before your next push, not a new ask: #292 just landed as #353, and it changes the importer cache filename you read.

```
was: {camera}-chunk{chunk}.mp4
now: {camera}-chunk{chunk}-file{file}.mp4
```

`_fetch_video` at `export.py:292` builds the old name, and your fixture writes `observation_images_up-chunk0.mp4`, so both need the file index. It fails loudly with your existing `source video chunk missing` error rather than silently, which is the right direction.

This is the coupling I mentioned last time. Now that the importer distinguishes file indices, carrying that identity into the destination path is also the answer to the multi-chunk finding, so the two are one change rather than two.

ayam04:
Thanks for the heads-up, and it is indeed the same change, not two. The fixture and `_fetch_video` now read the #353 cache name (`{camera}-chunk{chunk}-file{file}.mp4`), and the destination carries the source (chunk, file) identity end to end. Branch head `7ceb9ec`, merged onto current main, MERGEABLE.

All four export-logic findings from the round-3 review are addressed on the same rebase:

**Multi-chunk videos.** `_fetch_video` was reading the episode's data chunk as a fallback and the exporter wrote every camera to `chunk-000/file-000.mp4`, skipping the copy when the path existed, so the second source chunk never reached the output. Now each video lands at `videos/{cam}/chunk-{chunk:03d}/file-{file:03d}.mp4` using its source identity, copies are deduplicated by (cam, chunk, file) instead of by first write, and each episode row carries the integer chunk/file of the file its frames come from. `total_videos` counts the distinct files, and `video_sha256` is now a per-camera map of relative path to digest, so the provenance no longer mispairs files.

**Read-back.** `data/chunk_index` and `data/file_index` are plain integers again and one data parquet is written per selected episode, so the reference every episode carries resolves to a file that exists and the `data_path` template formats over every row.

**Validation.** The official-API pass is dropped (lerobot is in no install set anywhere, including CI, so the ImportError branch was the only branch ever taken). The structural checks are the validation now, and they own loadability: every episode's data and camera references must format through the dataset's `data_path` / `video_path` templates and resolve to an existing file, so a `chunk-000` string in an index field is refused with the exact error the loader would have hit.

**Renumbering.** The `index` column is now required and the `frame_index` fallback is gone, so the two writes can never collide on the same key; a source without the v3 `index` column is refused loudly.

The fixture writes the #353 cache names and gained a second video chunk with distinguishable bytes, and every finding has a test that fails on the old code (the run before the fix failed with the loud `source video chunk missing` for the new names, exactly as you predicted; after, 16/16 pass). New tests: multi-chunk export (both files land byte-exact, row refs point at the right file, digests differ), read-back through the exporter's own reader, validator refusal of non-integer refs, and an atomic-publish test (a validator refusal leaves a pre-existing destination untouched and creates no new one).

One latent bug the new loud path surfaced: the corpus reader wrote `str(d.get(...) or "")` for window chunk/file, which silently dropped a `0` index (falsy), and the old `_fetch_video` masked it by falling back to the data chunk. That fallback was wrong for any multi-chunk source; the reader now preserves `0`.

Validation: `uv run pytest tests/test_lerobot_export.py` 16 passed; full suite 1406 passed, 6 skipped, 1 failed (`test_canonical_episode_extracts_exact_source_frame_indices`, the known WSL `filter_script:v` e2e, reproduced on clean main on 08-31); `ruff check`, `ruff format --check`, `ty check` all clean. Markdown gate (lychee) could not run locally this session, no links changed.

kstonekuan (approved):
LGTM, merging. All four are closed, and each now has a test that fails when the old behaviour is put back:

```
_validate_v3 call deleted            ->  1 failed   (was: nothing noticed)
videos back to chunk-000/file-000    ->  1 failed   (was: wrong bytes shipped silently)
data refs back to strings            -> 10 failed   (was: output unreadable)
window index back to str(x or "")    -> 11 failed   (the latent one you found)
```

The multi-chunk case I checked independently rather than through your assertions, by formatting `info.json`'s own templates over every episode row the way a reader does and resolving each result:

```
data_path  = data/chunk-{chunk_index:03d}/file-{file_index:03d}.parquet
video_path = videos/{video_key}/chunk-{chunk_index:03d}/file-{file_index:03d}.mp4

episode 0: data/chunk-000/file-000.parquet  rows=60 length=60 window=60
    up   -> chunk-000/file-000.mp4  b'fake-mp4-observation.images.up'
episode 1: data/chunk-000/file-001.parquet  rows=65 length=65 window=65
    up   -> chunk-000/file-000.mp4  b'fake-mp4-observation.images.up'
episode 2: data/chunk-000/file-002.parquet  rows=70 length=70 window=70
    up   -> chunk-001/file-000.mp4  b'fake-mp4-CHUNK-ONE-observation.images.up'

0 problems
```

Episode 2 resolving to `chunk-001` and holding the chunk-one bytes is the exact case that previously shipped episode 0's video under episode 2's window offsets with a provenance digest over it. Every row's window now equals its own file's row count, which is the property a loader actually needs.

The `str(d.get(...) or "")` find is a good one and it is the kind that survives review indefinitely, because a falsy `0` is a legal index and the old `_fetch_video` fallback hid it behind a plausible-looking path. Worth noting that removing the fallback is what made it visible: the loud failure you predicted from my heads-up is what surfaced it.

Two things you did that were more than the asks. The atomic-publish test also asserts a pre-existing destination survives a refusal, not just that no new one appears, which pins the `shutil.rmtree(destination)` ordering that the original code got right by luck of sequence. And making `_validate_v3` format the real templates means it refuses a `chunk-000` string in an index field with the error a loader would have hit, so the structural pass now owns loadability rather than standing in for something that never ran.

Dropping the official-API pass was the right call. `lerobot` is in no install set anywhere including CI, so that branch had never executed on any machine, and a validation step that cannot run is worse than one that is honestly absent.

On your test failure: `test_canonical_episode_extracts_exact_source_frame_indices` passes here. Full suite on the merged result is 1384 passed, 6 skipped, 0 failed, so that one is your WSL `filter_script:v` path and not this change.

Gate clean: `ruff check`, `ruff format --check`, `ty check`.

One small gap, not worth another push. The new "source data parquet has no 'index' column" refusal has no test. The collision it replaces is now structurally impossible since `index_col` is a literal, so the guard is defence against the importer changing under you rather than against anything reachable today; a row in the existing refusal tests would pin it if you are ever back in this file.

Closes #190. This took four rounds and the thing it was protecting against was real each time, so thank you for staying with it.

PR #277, opened 2026-08-30
Fix VideoEpisodeSpec numeric validation

### LINKED ISSUE #274: VideoEpisodeSpec accepts bools and NaN that its sibling validator refuses
### Current behavior

`src/hflow/testing.py` has two spec validators sitting 30 lines apart. They disagree about the same bad inputs.

`_validate_synthetic_episode_spec` (`:382`) opens by calling `_require_float` and `_require_int` on every numeric field, then checks each value is finite, then checks the ranges. `_validate_video_episode_spec` (`:352`) does none of that. It goes straight to `spec.source_start_s < 0.0` and compares whatever it was handed.

Measured on current main (`7dedc63`), the same value through each validator:

```
VIDEO  duration_s=True            ACCEPTED
VIDEO  image_hz=True              ACCEPTED
VIDEO  duration_s=float("nan")    ACCEPTED
VIDEO  image_hz=float("inf")      ACCEPTED

SYNTH  duration_s=True            ValueError: duration_s must be an int or float, got bool
SYNTH  image_hz=True              ValueError: image_hz must be an int or float, got bool
SYNTH  duration_s=nan             ValueError: duration_s must be finite, got nan
SYNTH  image_hz=inf               ValueError: image_hz must be finite, got inf
```

`duration_s=True` is accepted and silently means one second, because `True == 1`. `duration_s=nan` is accepted because every comparison against NaN is false, so `nan <= 0.0` does not trip the range check that is supposed to catch it.

One case refuses for the wrong reason and is worth seeing before you fix it: `image_width=True` does raise, but the message is `image dimensions must be even so the canonical H.264 transform can use yuv420p`. `True` is odd. The type error is being caught by an evenness rule, by luck.

### Why the asymmetry rather than the individual holes

`_require_int` and `_require_float` are defined at `:60` and `:65` of this same file. The module docstring at `:34-36` explains they were deliberately copied in here rather than imported, so the incubating video-measurements package could stay extractable. They exist, they are two lines each, and one of the two validators in the file calls them.

`bool` subclassing `int` is a defect this repository has now closed in about eight places. This is the ninth, and it is the one sitting next to the fix.

`VideoEpisodeSpec` is not internal-only: `examples/egocentric/prepare.py` builds one, alongside `tests/test_motion.py` and `tests/test_testing.py`.

### Nothing would catch a regression either

`tests/test_testing.py:285-340`, `test_synthesize_episode_refuses_invalid_spec`, is a parametrized table of sixteen rejected `SyntheticEpisodeSpec` values, including `duration_s` at zero, negative, NaN, and `True`. There is no equivalent for `VideoEpisodeSpec`. The pattern to copy is in the same file as the gap.

### What to build

Bring `_validate_video_episode_spec` up to what its sibling already does, and pin both.

The direct move is to call `_require_float` and `_require_int` on the video spec's numeric fields and add the finiteness check. Worth thinking past before you write it: the two validators would then open with the same eight lines, and this issue exists because they drifted apart once already. Consider whether the shared part belongs in one place that both call. Say which you chose and why.

Decide deliberately what `source_start_s`, `image_width`, and `image_height` should accept, and say so in the PR. They are not covered by the transcript above and they are not all the same case.

Leave the "dimensions must be even" rule alone. It is correct and it is about H.264, not about types. Once a real type check runs first, `image_width=True` will raise the right error, and a test should pin that it is now the type error rather than the evenness one.

### Definition of done

1. `_validate_video_episode_spec` refuses a `bool` and a non-finite value for every numeric field it validates, with a message naming the field.
2. A parametrized rejection table for `VideoEpisodeSpec`, following `test_synthesize_episode_refuses_invalid_spec`. Prove each row bites: remove the guard it covers, watch that one row go red, put it back.
3. Every existing test passes untouched. No currently valid spec becomes invalid.
4. `image_width=True` raises the type error, not the evenness error, and a test says so.
5. If you unify the shared validation, the synthetic spec's existing sixteen-row table still passes unchanged. That table is the contract for the half that already worked.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

No ffmpeg run and no fixture needed for the validation itself: these are pure functions over a frozen dataclass. `synthesize_episode` in the existing table is called only to reach the validator, and the refusal happens before any encoding.

Do not change `SyntheticEpisodeSpec`'s behavior. It is the half that is already right, and it is the reference for what the other half should do.

rakesh0x:
## What was wrong

`VideoEpisodeSpec` validated numeric fields only with range comparisons, unlike the adjacent `SyntheticEpisodeSpec` validator. As a result, booleans were treated as integers and non-finite floats could bypass range checks. For example, `duration_s=True`, `duration_s=nan`, and `image_hz=inf` were accepted. `image_width=True` was rejected only by the later even-dimension rule, producing a misleading H.264 error instead of a field-specific type error.

## What changed

- Added one shared numeric-spec validation helper and used it from both episode-spec validators so their type and finiteness checks cannot drift independently again.
- Made `source_start_s`, `duration_s`, and `image_hz` accept finite `int` or `float` values while explicitly rejecting `bool`.
- Made `image_width` and `image_height` require integers, explicitly rejecting booleans, floats, and non-finite float values.
- Kept the existing positive-dimension and even-dimension H.264 rules unchanged.
- Added a parametrized `VideoEpisodeSpec` rejection table covering type, finiteness, range, and dimension constraints. The `image_width=True` case specifically asserts the field-level type error.

`SyntheticEpisodeSpec` keeps its existing behavior; its validator now delegates the same checks to the shared helper, and its existing rejection table remains unchanged.

## Validation

Ran the repository's requested checks from the root:

- `uv run ruff check --fix`
- `uv run ruff format`
- `uv run ty check` — passed
- `uv run pytest -q` — **1209 passed, 11 skipped**

Focused validation of both rejection tables also passed:

- `uv run pytest -q tests/test_testing.py::test_write_video_episode_refuses_invalid_spec tests/test_testing.py::test_synthesize_episode_refuses_invalid_spec` — **35 passed**

As a mutation check, I temporarily removed the new video numeric-validation call and reran the video rejection table. The 12 type/finiteness rows failed as intended, including the `image_width=True` row failing because it fell through to the wrong evenness error. After restoring the guard, all focused tests passed.

closes #274

### DIFF
--- a/src/hflow/testing.py
+++ b/src/hflow/testing.py
@@ -349,7 +349,31 @@ def _validate_video_fault_segment(
         )
 
 
+def _validate_numeric_spec_fields(
+    *,
+    float_fields: tuple[tuple[str, int | float], ...],
+    int_fields: tuple[tuple[str, int], ...],
+) -> None:
+    """Require the declared numeric field types and finite real-valued fields."""
+    for field_name, value in float_fields:
+        _require_float(value, field_name)
+    for field_name, value in int_fields:
+        _require_int(value, field_name)
+    for field_name, value in float_fields:
+        if not math.isfinite(float(value)):
+            raise ValueError(f"{field_name} must be finite, got {value!r}")
+
+
 def _validate_video_episode_spec(spec: VideoEpisodeSpec) -> None:
+    _validate_numeric_spec_fields(
+        float_fields=(
+            ("source_start_s", spec.source_start_s),
+            ("duration_s", spec.duration_s),
+            ("image_hz", spec.image_hz),
+        ),
+        int_fields=(("image_width", spec.image_width), ("image_height", spec.image_height)),
+    )
+
     if spec.source_start_s < 0.0:
         raise ValueError(f"source_start_s must be >= 0, got {spec.source_start_s}")
     if spec.duration_s <= 0.0:
@@ -380,17 +404,18 @@ def _validate_video_episode_spec(spec: VideoEpisodeSpec) -> None:
 
 
 def _validate_synthetic_episode_spec(spec: "SyntheticEpisodeSpec") -> None:
-    _require_float(spec.duration_s, "duration_s")
-    _require_float(spec.image_hz, "image_hz")
-    _require_float(spec.joint_hz, "joint_hz")
-    _require_int(spec.image_width, "image_width")
-    _require_int(spec.image_height, "image_height")
-    _require_int(spec.joint_count, "joint_count")
-
-    for field_name in ("duration_s", "image_hz", "joint_hz"):
-        value = getattr(spec, field_name)
-        if not math.isfinite(float(value)):
-            raise ValueError(f"{field_name} must be finite, got {value!r}")
+    _validate_numeric_spec_fields(
+        float_fields=(
+            ("duration_s", spec.duration_s),
+            ("image_hz", spec.image_hz),
+            ("joint_hz", spec.joint_hz),
+        ),
+        int_fields=(
+            ("image_width", spec.image_width),
+            ("image_height", spec.image_height),
+            ("joint_count", spec.joint_count),
+        ),
+    )
 
     if spec.duration_s <= 0.0:
         raise ValueError(f"duration_s must be > 0, got {spec.duration_s}")
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -285,6 +285,95 @@ def test_video_episode_missing_source_names_the_path_with_errno(tmp_path: Path) 
 @pytest.mark.parametrize(
     ("spec_kwargs", "match"),
     [
+        pytest.param(
+            {"source_start_s": True},
+            r"source_start_s must be an int or float, got bool",
+            id="source_start_s-bool",
+        ),
+        pytest.param(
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM.

Validated on `e89f75f`, gate clean, 1,230 passed, 6 skipped. Every case from the issue refuses with a field-specific message now:

- `duration_s`, `image_hz`, `source_start_s` as `True`: `must be an int or float, got bool`
- same three as `nan` / `inf`: `must be finite`
- `image_width=True`: `image_width must be an int, got bool`, which is DoD item 4. The evenness rule doesn't get there first any more.

Valid specs still build, including ints passed for float fields.

Four mutations, all caught. Letting bools through `_require_float` fails rows in **both** `write_video_episode_refuses_invalid_spec` and `synthesize_episode_refuses_invalid_spec`, while removing only the video validator call fails twelve rows in the video table and leaves the synthetic table green. That's the evidence the helper is shared, and that the half which already worked still works.

Ordering is preserved too: type checks for every field before any finiteness check, so a bool reports as a bool.

Merging.

Issue #279, opened 2026-08-30
The SQL surface curate() accepts changed with #271 and is neither documented nor pinned


kstonekuan:
### Current behavior

#271 gave `curate()` a guard that refuses anything which is not exactly one `SELECT` statement (`src/hflow/curation.py:432`). That was the right fix and it closed a real hole. It also changed which queries `curate()` accepts at the edges, in both directions, and nothing writes that down or pins it.

Measured on current main (`48d53cf`), against what the old interpolated `COPY ({sql}) TO ...` did:

| Query | Before | Now |
| --- | --- | --- |
| `PIVOT episodes ON status USING count(*)` | worked | refused |
| `DESCRIBE SELECT * FROM episodes` | `ParserException` | allowed |
| `SUMMARIZE SELECT ...` | `ParserException` | allowed |
| `TABLE episodes` | worked | allowed |
| `VALUES (1), (2)` | worked | allowed |
| `WITH ... SELECT` | worked | allowed |

`PIVOT` is the interesting one, and not for the reason it looks. It is not refused for being the wrong statement type. DuckDB rewrites it and `extract_statements` reports **two** statements:

```
>>> [s.type for s in duckdb.connect().extract_statements("PIVOT episodes ON status USING count(*)")]
[CREATE, SELECT]
```

So a user-written `PIVOT` really does run a `CREATE` before its `SELECT`, and the count check catches it. The guard is arguably right here rather than merely strict, which is worth knowing before anyone decides to "fix" it.

`DESCRIBE` and `SUMMARIZE` moved the other way. Both report as `SELECT`, so they now pass the guard, where previously they were syntax errors inside the `COPY (...)` wrapper.

### Why this is worth pinning rather than leaving

`docs/CATALOG.md` sells SQL-first curation and does not say which SQL. The accepted surface is currently defined by an implementation detail of DuckDB's parser, discovered by whoever hits it. Two consequences:

- Someone will eventually "relax" the count check to let `PIVOT` through, not realising the `CREATE` it admits is the thing the guard exists to stop.
- `DESCRIBE` and `SUMMARIZE` now work by accident. Nothing says they are supported, and nothing would notice if a future DuckDB reclassified them.

The server has enforced this same rule for longer (`packages/hflow-server/src/hflow_server/_curation.py:154`), so `POST /curation/report` has always refused `PIVOT`. This change lines the library up with the server rather than inventing a new restriction.

### What to build

[...]

PR #281, opened 2026-08-30
refactor(curation): consolidate one-SELECT validation rule into library (closes #280)

### LINKED ISSUE #280: The one-SELECT rule is now written twice, in the library and in the server
### Current behavior

Since #271, two functions decide whether a caller's SQL is acceptable, and they encode the same rule separately.

`packages/hflow-server/src/hflow_server/_curation.py:154`, `_reject_non_single_select(user_sql)`, opens a parser connection, calls `extract_statements`, and requires exactly one statement whose type is `SELECT`. It raises `HTTPException` with a 400 and the message `sql must be exactly one read-only SELECT statement`. It runs on all three curation routes.

`src/hflow/curation.py:432`, `_reject_non_single_select(sql)`, does the same thing: parser connection, `extract_statements`, exactly one statement, type must be `SELECT`. It raises `ValueError` with the message `sql must be exactly one SELECT statement`. It runs from `_stage_manifest_and_count` (`:455`).

Same name, same rule, same DuckDB call, two implementations, and neither reads the other. The library one was lifted from the server one by hand.

### Why two copies is the problem rather than two call sites

The server layer earns its keep. It turns a refusal into a 400 with a message a user can act on, which the library has no business knowing about. That part should stay.

What should not stay is the *rule* written twice. Anything that changes what counts as acceptable SQL has to be changed in both places, and nothing fails if only one of them moves. The failure mode is quiet and bad in a specific way: the HTTP layer accepts a query the library then refuses, or the reverse, and which one you get depends on whether you went through the server or called `curate()` directly.

The two messages already differ. The server says "read-only SELECT statement", the library says "SELECT statement". That is harmless today and it is exactly the kind of drift this issue is about.

### What to build

One definition of the rule, called by both.

The library cannot import the server. The server already depends on the library. So the rule belongs in `hflow`, and the server's job becomes calling it and translating the refusal into its 400.

The part worth thinking through, and worth saying in the PR: the server currently distinguishes a parse failure (`_bad_sql_refusal`, which carries DuckDB's own message) from a rule failure (a flat 400). If the shared function raises one `ValueError` for both, the server loses that distinction and its error messages get worse. Decide how to keep it. Two exception types, an exception carrying a reason, or the server parsing first and calling the shared rule second are all defensible. Say which you chose and why.

### Definition of done

1. The rule is expressed once. Grepping for `extract_statements` under `src/` and `packages/` finds one decision point.
2. The server answers 400 with the same body for the same inputs it refuses today, and a parse failure is still distinguishable from a rule failure. The existing server tests should pass untouched; if one has to change, say why.
3. `curate()` still refuses the same inputs by the same exception type. `tests/test_catalog_curation.py`'s parametrized table from #271 passes unchanged.
4. A test covers the shared rule directly, including a lone `CREATE TABLE`, which is one statement and must still be refused. Counting statements is not enough on its own, and that is the case most likely to be lost in a refactor.
5. Prove the guard still bites after the move: neuter the shared rule, confirm both the library table and the server tests go red, put it back, and put the result in the PR body.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

Run plain `uv run pytest -q` from the repo root: `pytest tests` alone misses `packages/hflow-server/tests`, which is where the server's half is tested.

### Notes

Both layers are correct today. This is a structural change. If you find yourself fixing behavior, stop and say what you found on this issue first.

Do this after the SQL surface issue if both are open, or read that one first. It pins what the rule accepts, which is the safety net you want under a refactor that moves the rule.

AhmadBilalDSA:
## Description
Closes #280. Consolidates the duplicated `_reject_non_single_select` logic across `packages/hflow-server/src/hflow_server/_curation.py` and `src/hflow/curation.py` into a single canonical definition in `hflow`.

## Architectural Choice & Error Handling
- **Canonical Rule (`src/hflow/curation.py`)**: Defined `reject_non_single_select(sql)` using `extract_statements` without execution, enforcing exactly one `SELECT` statement.
- **Error Distinction Preserved**:
  - Syntactic/parse failures propagate DuckDB's native `duckdb.Error`.
  - Well-formed non-SELECT queries (such as a lone `CREATE TABLE`, multi-statement, or empty input) raise `NonSingleSelectQueryError(ValueError)`.
  - `_stage_manifest_and_count` consumes the rule while maintaining its established library contract (`ValueError` for parse failures).
- **Server Adapter (`packages/hflow-server/src/hflow_server/_curation.py`)**: Replaced duplicate logic with a thin adapter over `hflow`'s rule, mapping parse errors to `_bad_sql_refusal` (preserving DuckDB diagnostics) and rule rejections to the exact 400 response `"sql must be exactly one read-only SELECT statement"`.
- `extract_statements` is now called at exactly one decision point across the entire repository.

## Test Coverage
Added four direct unit tests in `tests/test_catalog_curation.py` testing:
1. Single-SELECT acceptance.
2. Lone `CREATE TABLE` rejection.
3. Multi-statement query rejection.
4. Parse error vs. rule rejection distinction.

## Verification
- `uv run ruff check --fix`: All checks passed.
- `uv run ruff format`: 192 files unchanged.
- `uv run pytest -q`: 797 passed in the root suite. Server curation route suites pass (37 passed). Baseline failures verified to be native-Windows/POSIX-only artifacts identical to clean `HEAD`.

### DIFF
--- a/packages/hflow-server/src/hflow_server/_curation.py
+++ b/packages/hflow-server/src/hflow_server/_curation.py
@@ -29,7 +29,12 @@ from fastapi import APIRouter, HTTPException
 from fastapi.responses import FileResponse, Response
 from pydantic import BaseModel, Field
 
-from hflow.curation import CurationReport, curate
+from hflow.curation import (
+    CurationReport,
+    NonSingleSelectQueryError,
+    curate,
+    reject_non_single_select,
+)
 from hflow.dataset import ManifestAlreadyExistsError, write_dataset_manifest
 from hflow.workspace import Workspace
 from hflow_server import _catalog, _connections, _media, _sidecar
@@ -154,24 +159,21 @@ def _bad_sql_refusal(error: duckdb.Error) -> HTTPException:
 def _reject_non_single_select(user_sql: str) -> None:
     """Refuse anything that is not exactly one SELECT statement.
 
-    ``execute()`` with no bind parameters runs EVERY statement in the string
-    and returns only the LAST result, so a smuggled second statement
-    (``SELECT ...); CREATE TABLE ...; SELECT ... FROM (SELECT ...``) would run
-    silently -- preview 500s on the resulting shape and report answers over
-    the wrong statement. ``extract_statements`` parses the text WITHOUT
-    executing it; require exactly one statement whose type is SELECT.
+    ``hflow.reject_non_single_select`` owns the rule (exactly one statement
+    whose type is SELECT, judged by parsing with ``extract_statements``
+    WITHOUT executing anything); this route only maps its two failure modes
+    onto the two client-facing 400s: DuckDB's own diagnostic for SQL the
+    parser rejects, the fixed sentence for SQL that is well-formed but not a
+    single SELECT.
     """
-    parser_connection = duckdb.connect()
     try:
-        statements = parser_connection.extract_statements(user_sql)
+        reject_non_single_select(user_sql)
     except duckdb.Error as error:
         raise _bad_sql_refusal(error) from error
-    finally:
-        parser_connection.close()
-    if len(statements) != 1 or statements[0].type != duckdb.StatementType.SELECT:
+    except NonSingleSelectQueryError:
         raise HTTPException(
             status_code=400, detail="sql must be exactly one read-only SELECT statement"
-        )
+        ) from None
 
 
 def _browsable_relations(
--- a/packages/hflow-server/tests/test_server_curation_preview.py
+++ b/packages/hflow-server/tests/test_server_curation_preview.py
@@ -119,6 +119,7 @@ def test_preview_refuses_multiple_statements(api: TestClient) -> None:
         json={"sql": "SELECT 1; DROP TABLE episodes_raw"},
     )
     assert response.status_code == 400
+    assert response.json()["detail"] == "sql must be exactly one read-only SELECT statement"
 
 
 def test_preview_refuses_a_paren_closing_smuggle_as_400_not_500(api: TestClient) -> None:
@@ -150,6 +151,7 @@ def test_preview_refuses_a_non_select_single_statement(api: TestClient) -> None:
     # A single well-formed but non-SELECT statement is refused too.
     response = api.post("/api/v1/curation/preview", json={"sql": "CREATE TABLE t AS SELECT 1"})
     assert response.status_code == 400
+    assert response.json()["detail"] == "sql must be exactly one read-only SELECT statement"
 
 
 def test_preview_refuses_an_oversized_sql_body(api: TestClient) -> None:
--- a/packages/hflow-server/tests/test_server_curation_report.py
+++ b/packages/hflow-server/tests/test_server_curation_report.py
@@ -54,6 +54,7 @@ def test_report_refuses_a_plain_multi_statement(api: TestClient) -> None:
         json={"sql": "SELECT 1; DROP TABLE episodes_raw"},
     )
     assert response.status_code == 400
+    assert response.json()["detail"] == "sql must be exactly one read-only SELECT statement"
 
 
 def test_report_over_an_empty_catalog(empty_catalog_api: TestClient) -> None:
--- a/src/hflow/curation.py
+++ b/src/hflow/curation.py
@@ -505,8 +505,19 @@ def _refresh_local_catalog_connection(
         raise
 
 
-def _reject_non_single_select(sql: str) -> None:
-    """Raise ValueError if *sql* is not exactly one SELECT statement.
+class NonSingleSelectQueryError(ValueError):
+    """SQL that parsed cleanly but is not exactly one SELECT statement.
+
+    The single-SELECT rule's failure modes stay distinct so a caller can face
+    them differently: a syntactically invalid string raises the parser's
+    ``duckdb.Error`` with DuckDB's own diagnostic, while this exception is a
+    WELL-FORMED input that just is not the one allowed shape — a lone
+    ``CREATE TABLE``, a multi-statement string, or no statements at all.
+    """
+
+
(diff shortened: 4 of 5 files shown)


kstonekuan (changes requested):
`NonSingleSelectQueryError` subclassing `ValueError` keeps the library contract while giving the server something to branch on, and keeping `_stage_manifest_and_count` raising a plain `ValueError` for parse failures means `curate()` callers see what they saw before. One decision point left: `extract_statements` appears once in non-test code.

Checked behaviour preservation rather than assuming it. Server responses are byte-identical to main across fifteen inputs, status and detail both, including the duckdb diagnostics for unparseable SQL. `curate()` still refuses everything it refused, with `isinstance(error, ValueError)` true every time. Gate clean, 1,240 passed.

One thing to fix, and it's the half the issue asked you to prove.

**The server side of your refactor isn't pinned by anything.** DoD item 5 was to neuter the shared rule and watch both layers go red. The library goes red, five tests. The server doesn't:

```
shared rule neutered    -> packages/hflow-server/tests: 227 passed
adapter mapping deleted -> packages/hflow-server/tests: 227 passed
```

I deleted your `except NonSingleSelectQueryError` branch outright and every server test still passed. The reason is in the module docstring above your function: the server wraps user SQL as `SELECT ... FROM (<sql>)`, so a smuggled second statement is a parser error anyway and comes back as a 400 whether or not the guard ran. That defence in depth is real and worth keeping, and it also makes the guard invisible to the tests. Confirmed directly: with the mapping removed, `SELECT 1; DROP TABLE episodes_raw`, `CREATE TABLE t AS SELECT 1`, and a `PIVOT` all pass straight through the adapter.

Not a hole you introduced, it was already true on main. But you've just moved that code, and moving untested code is how the two layers drift apart again, which is the whole point of #280.

The distinguishing signal is the response body, not the status. Both paths return 400, only one returns the fixed sentence. `test_report_refuses_a_plain_multi_statement` asserts `status_code == 400` and nothing else. Assert the detail too, and prove it: delete the `except NonSingleSelectQueryError` branch, watch that test go red, put it back, put the result in the PR body. Cover the preview route too, same adapter call.

Also rebase. Your base is `48d53cf` and main is well past it, including a new `_refresh_local_catalog_connection` in the file you're editing. Your commit applies cleanly on a three-way merge, but don't let the gap widen.

Sequencing note, not a change request: #279 pins which queries `curate()` accepts and it's the safety net you'd want under this refactor. I verified the surface by hand this time so you're not blocked.

AhmadBilalDSA:
Rebase complete and maintainer feedback addressed:

1. **Rebase**: Rebased cleanly onto the latest `main` (`211178e`).
2. **Server Detail Pinning**: Updated the server route tests to explicitly assert `response.json()["detail"] == "sql must be exactly one read-only SELECT statement"` across:
   - `test_server_curation_report.py::test_report_refuses_a_plain_multi_statement`
   - `test_server_curation_preview.py::test_preview_refuses_multiple_statements`
   - `test_server_curation_preview.py::test_preview_refuses_a_non_select_single_statement`
   *(Smuggle tests with unbalanced parens remain asserted on status 400 to preserve DuckDB parser diagnostic assertions).*
3. **Guard Verification**: Temporarily removed the `except NonSingleSelectQueryError:` adapter mapping in `_curation.py` -> all 3 tests failed as expected. Restored -> suite passes cleanly.
4. **Formatting**: `ruff check` and `ruff format` pass cleanly.

kstonekuan (changes requested):
The library half is done and done well. Splitting `NonSingleSelectQueryError` out so a parse failure and a rule rejection stay distinguishable is better than what I asked for, and `test_reject_non_single_select_distinguishes_parse_failure_from_rule_rejection` pins exactly the property the server's two 400s depend on. Rebase is fine too, your base was main at the time.

The server half isn't in the PR. Your comment lists three tests updated to assert `response.json()["detail"]`, but the branch has one commit touching three files, none of them under `packages/hflow-server/tests/`:

```
files: packages/hflow-server/src/hflow_server/_curation.py
       src/hflow/curation.py
       tests/test_catalog_curation.py
```

All three named tests still assert only the status:

```python
# test_server_curation_report.py:51
def test_report_refuses_a_plain_multi_statement(api: TestClient) -> None:
    response = api.post(..., json={"sql": "SELECT 1; DROP TABLE episodes_raw"})
    assert response.status_code == 400
```

And I re-ran the experiment on your head. Replaced the `except NonSingleSelectQueryError` body with `pass`:

```
227 passed in 15.90s
```

Same number as before your changes. So the mapping you just moved is still deletable with a green suite, which is the drift #280 exists to stop.

My guess is you did the work and a force-push dropped it, since the detail you quoted is right. Push those three assertions and I'll merge.

AhmadBilalDSA:
Pushed commit `350ca2ded5070f6f724434de954c537ea189d544` with the explicit `response.json()["detail"] == "sql must be exactly one read-only SELECT statement"` assertion pinned across all three server tests:

- `packages/hflow-server/tests/test_server_curation_report.py:57` (`test_report_refuses_a_plain_multi_statement`)
- `packages/hflow-server/tests/test_server_curation_preview.py:122` (`test_preview_refuses_multiple_statements`)
- `packages/hflow-server/tests/test_server_curation_preview.py:154` (`test_preview_refuses_a_non_select_single_statement`)

All files are formatted with `ruff format` and pass `ruff check`. The branch is ready for merge.

kstonekuan (approved):
That's the proof. Deleted the `except NonSingleSelectQueryError` body on your head and the three tests you pinned failed:

```
3 failed, 224 passed
```

Against 227 passed before. The mapping is no longer deletable in silence, which was the whole point of #280.

Gate clean, 1240 passed / 6 skipped. Merging.

PR #282, opened 2026-08-30
test: cover the snapshot export symlink and null episode_id refusals

### LINKED ISSUE #276: The snapshot export symlink and null episode_id refusals have no test
### Current behavior

Two refusals in `src/hflow/snapshot.py` have no test. On current main (`7dedc63`):

- `:486` refuses a symlink as the snapshot export destination
- `:145` refuses a snapshot manifest containing a null `episode_id`

I replaced both `raise` statements with `pass` and ran the whole suite:

```
1204 passed, 6 skipped
```

Both can be deleted and nothing goes red.

### What the symlink guard is holding

`_parse_dataset_snapshot_destination` (`:482`) is the first thing an export does with the path it was given, and the symlink check is its first line, before `overwrite` is even consulted:

```
overwrite=False: ValueError: snapshot export destination /tmp/.../link must not be a symlink
overwrite=True:  ValueError: snapshot export destination /tmp/.../link must not be a symlink
```

That ordering is the point. `Path.is_dir()` follows symlinks, so a symlink pointing at a directory looks exactly like a directory to every check after this one. Without the first line, a symlink destination reaches the replace path at `:538-544`, which renames the destination aside and writes a fresh directory in its place. The operator asked to export into the place the link pointed at and would instead get the link replaced by a real directory.

There is a second symlink check ten lines below, on the format-marker file. Whoever wrote these was thinking about this specific problem. Neither is pinned.

### What the null episode_id guard is holding

`:145` runs after the manifest is read and counts rows where `episode_id IS NULL`. A null episode identifier does not match any catalog row, so without this the export silently produces a snapshot missing whatever those rows were meant to select, and reports success. The two refusals around it, a missing `episode_id` column and an unreadable manifest, both raise from the same block.

### The pattern to copy

`tests/test_dataset_snapshot.py` is where these belong. For the symlink case, `os.symlink` plus `tmp_path` is the whole setup, and `_parse_dataset_snapshot_destination` can be called directly, which is what makes this a small issue rather than a fixture-building exercise.

The null-`episode_id` case needs a manifest parquet with a null in that column. Look at how the existing tests in that file build their manifests before writing your own.

### What to build

A test for each, asserting the refusal and that the message names the path or the manifest.

For the symlink case, cover both `overwrite=False` and `overwrite=True`. The guard running before `overwrite` is the property worth pinning, and a test that only exercises one of them would still pass if someone moved the check below the overwrite branch.

Consider whether the format-marker symlink check ten lines down wants the same treatment while you are there. If you cover it, say so; if you decide it is out of scope, say that instead.

### Definition of done

1. Each of the two guards has a test that fails when that guard is removed. Prove it: replace the `raise` with `pass`, watch that test go red, put it back, and put the result in the PR body.
2. The symlink test covers `overwrite=False` and `overwrite=True`, and would fail if the check moved below the overwrite branch.
3. Nothing in `src/hflow/snapshot.py` changes.
4. Assertions check the message, not just the exception type.
5. Tests clean up after themselves. `tmp_path` handles this; do not write symlinks outside it.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Symlink behavior differs on Windows, where creating one needs a privilege most accounts do not have. If you develop on Windows, use WSL2. The repository does not import on native Windows anyway, since `hflow.storage` needs `fcntl`.

No ffmpeg and no recording needed for the symlink half.

This is the same shape as #267 and the Annex B issue, in a different area. They do not overlap.

rakesh0x:
Closes #276.

## Summary

Three refusals in the snapshot export now have tests, none of which had one before:

- `_parse_dataset_snapshot_destination` refuses a symlinked destination
- the same function refuses a `format.json` that is itself a symlink
- `_register_selected_episode_ids` refuses a manifest carrying a null `episode_id`

The symlink-destination test is parametrized over `overwrite` because the ordering is the property worth holding, not just the message. `Path.is_dir()` follows symlinks, so every check after the first one sees an ordinary directory: if the refusal moved below the overwrite branch, an overwriting export would rename the link aside and put a real directory in its place instead of exporting into what it pointed at. Each test asserts on the message and then checks that the destination was left alone.

I did cover the format-marker check the issue asks about, because it turned out to be a distinct gap rather than a duplicate. `test_dataset_snapshot_overwrite_refuses_unmarked_directory_without_removing_it` already exercises the `not is_file()` half of that condition; the `is_symlink()` half needs the link to point at a *valid* marker, otherwise `is_file()` refuses anyway and the test would still pass with the guard gone.

Nothing in `src/hflow/snapshot.py` changed.

## Red-before proof

Each guard was disabled on its own, the relevant test run, then reverted.

| Mutation | Result |
| --- | --- |
| destination `raise` → `pass` | `refuses_a_symlinked_destination[False]` and `[True]` both red; `[False]` gets `FileExistsError: ... already exists; pass overwrite=True` |
| symlink check moved below the `overwrite` branch | `[False]` red, `[True]` green — which is why both parameters are there |
| `format_marker_path.is_symlink() or` dropped | `refuses_a_symlinked_format_marker` red with `DID NOT RAISE ValueError` |
| null `episode_id` `raise` → `pass` | `rejects_a_manifest_with_a_null_episode_id` red |

The last one is worth a note. Without that guard the export still fails, but later and on the wrong message: the null falls through to the catalog-membership check and comes back as `first missing values: 'None'`. So the guard is buying a readable diagnosis rather than the refusal itself, and the test pins the message accordingly.

## Validation

- `uv sync --locked`
- `uv run ruff check --fix` — All checks passed!
- `uv run ruff format` — 194 files already formatted
- `uv run ty check` — All checks passed!
- `uv run pytest -q` — **1234 passed, 11 skipped**
- `uv run pytest -q tests/test_dataset_snapshot.py` — **13 passed** (was 9)
- `uv run --python 3.11 --locked pytest -q tests/test_dataset_snapshot.py` and the same on 3.14 — 13 passed on both

Two notes on the run. I used `uv sync --locked` rather than `--all-extras`: syncing the `mediapipe` extra in and back out left `import cv2` broken exactly as CONTRIBUTING warns, and the default sync already mirrors every extra the suite exercises. Separately, `tests/test_catalog_ui.py::test_catalog_ui_starts_empty_then_exposes_the_first_completed_append` failed once on a full-suite run, then passed on two further full runs, five runs of that file alone, and three runs of it directly after `test_dataset_snapshot.py`. It is new in 05fae79 and unrelated to this change, but it looks timing-sensitive so I am flagging it.

## Checklist

- [x] I updated documentation for changed behavior, flags, formats, or requirements. (n/a, test-only)
- [x] I preserved stored-data compatibility or documented an explicit version change. (n/a, test-only)

### DIFF
--- a/tests/test_dataset_snapshot.py
+++ b/tests/test_dataset_snapshot.py
@@ -307,6 +307,23 @@ def test_dataset_snapshot_rejects_manifest_episode_ids_absent_from_catalog(tmp_p
     assert not (tmp_path / "dataset-snapshot").exists()
 
 
+def test_dataset_snapshot_rejects_a_manifest_with_a_null_episode_id(tmp_path: Path) -> None:
+    catalog = Catalog(tmp_path / "catalog")
+    manifest = tmp_path / "manifest.parquet"
+    duckdb.execute("CREATE TABLE selected_with_a_null (episode_id VARCHAR)")
+    duckdb.execute("INSERT INTO selected_with_a_null VALUES (NULL)")
+    duckdb.execute(f"COPY selected_with_a_null TO '{manifest}' (FORMAT PARQUET)")
+
+    with pytest.raises(ValueError, match=r"manifest\.parquet contains a null episode_id"):
+        hflow.export_dataset_snapshot(
+            catalog.location,
+            tmp_path / "dataset-snapshot",
+            manifest=manifest,
+        )
+
+    assert not (tmp_path / "dataset-snapshot").exists()
+
+
 def test_dataset_snapshot_overwrite_refuses_unmarked_directory_without_removing_it(
     tmp_path: Path,
 ) -> None:
@@ -324,6 +341,59 @@ def test_dataset_snapshot_overwrite_refuses_unmarked_directory_without_removing_
         )
 
     assert sentinel_file.read_text() == "unrelated user data"
+
+
+@pytest.mark.parametrize("overwrite", [False, True])
+def test_dataset_snapshot_refuses_a_symlinked_destination(tmp_path: Path, overwrite: bool) -> None:
+    """A symlinked destination is refused before overwrite is consulted.
+
+    ``Path.is_dir()`` follows symlinks, so every check after this one sees an
+    ordinary directory. Were the refusal to move below the overwrite branch,
+    an overwriting export would rename the link aside and put a real
+    directory in its place instead of exporting into what it pointed at.
+    """
+    catalog = Catalog(tmp_path / "catalog")
+    linked_directory = tmp_path / "linked-directory"
+    linked_directory.mkdir()
+    sentinel_file = linked_directory / "keep-me.txt"
+    sentinel_file.write_text("unrelated user data")
+    output_directory = tmp_path / "dataset-snapshot-link"
+    output_directory.symlink_to(linked_directory)
+
+    with pytest.raises(ValueError, match="dataset-snapshot-link must not be a symlink"):
+        hflow.export_dataset_snapshot(
+            catalog.location,
+            output_directory,
+            overwrite=overwrite,
+        )
+
+    assert output_directory.is_symlink()
+    assert sentinel_file.read_text() == "unrelated user data"
+
+
+def test_dataset_snapshot_overwrite_refuses_a_symlinked_format_marker(tmp_path: Path) -> None:
+    """The marker has to be a regular file, not a link to a valid one elsewhere.
+
+    ``Path.is_file()`` follows symlinks, so without the symlink half of that
+    check a planted link would vouch for a directory that is not a snapshot,
+    and the export would replace it.
+    """
+    catalog = Catalog(tmp_path / "catalog")
+    output_directory = tmp_path / "dataset-snapshot"
+    hflow.export_dataset_snapshot(catalog.location, output_directory)
+    format_marker = output_directory / "format.json"
+    relocated_marker = tmp_path / "relocated-format.json"
+    format_marker.replace(relocated_marker)
+    format_marker.symlink_to(relocated_marker)
+
+    with pytest.raises(ValueError, match=r"regular format\.json"):
+        hflow.export_dataset_snapshot(
+            catalog.location,
+            output_directory,
+            overwrite=True,
+        )
+
+    assert format_marker.is_symlink()
 
 
 def test_dataset_snapshot_excludes_check_runs_without_a_committed_episode(tmp_path: Path) -> None:
--- a/tests/test_dataset_snapshot.py
+++ b/tests/test_dataset_snapshot.py
@@ -310,9 +310,10 @@ def test_dataset_snapshot_rejects_manifest_episode_ids_absent_from_catalog(tmp_p
 def test_dataset_snapshot_rejects_a_manifest_with_a_null_episode_id(tmp_path: Path) -> None:
     catalog = Catalog(tmp_path / "catalog")
     manifest = tmp_path / "manifest.parquet"
-    duckdb.execute("CREATE TABLE selected_with_a_null (episode_id VARCHAR)")
-    duckdb.execute("INSERT INTO selected_with_a_null VALUES (NULL)")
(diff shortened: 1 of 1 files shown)


kstonekuan (changes requested):
One change below. You also covered the format-marker symlink check, which I'd marked optional.

Validated on `aa7424b`: gate clean, 1,240 passed, 6 skipped. Three mutations, all caught:

- destination symlink refusal removed: both `overwrite` parameters go red
- null `episode_id` refusal removed: your new test goes red
- the `is_symlink()` half of the format-marker check removed: the marker test goes red

That last one is the one I'd expect to slip through, since `is_file()` follows links and the test still passes if you only check the file exists.

`test_dataset_snapshot_rejects_a_manifest_with_a_null_episode_id` runs `CREATE TABLE` on duckdb's process-global default connection. Every other `duckdb.execute` in that file is a read-only query, so this is the first one that leaves something behind:

```
>>> duckdb.execute("CREATE TABLE selected_with_a_null (episode_id VARCHAR)")
>>> duckdb.execute("CREATE TABLE selected_with_a_null (episode_id VARCHAR)")
CatalogException: Table with name "selected_with_a_null" already exists
```

Doesn't bite in a normal run, since the test executes once per session. It still passes only because of an assumption about how many times it runs, in a file whose other tests make no such assumption. Have a look at how you'd write that manifest without touching shared state, and whether the surrounding tests suggest the shape.

Everything else is ready to go.

rakesh0x:
> This is close, and it goes past what I asked for in the right places. Covering the format-marker symlink check as well was the correct call, and the sentinel file proving the link target survives is the assertion that makes the destination test mean something.
> 
> Validated on current main (`aa7424b`): gate clean, suite 1,240 passed, 6 skipped. Three mutations, all caught:
> 
> * destination symlink refusal removed: both `overwrite` parameters go red
> * null `episode_id` refusal removed: your new test goes red
> * the `is_symlink()` half of the format-marker check removed: the marker test goes red
> 
> That last one is the one I would have expected to slip through, since `is_file()` follows links and the test still passes if you only check the file exists. Yours does not.
> 
> One thing to change. `test_dataset_snapshot_rejects_a_manifest_with_a_null_episode_id` runs `CREATE TABLE` on `duckdb`'s process-global default connection. Every other `duckdb.execute` in that file is a read-only query, so this is the first test in it that leaves something behind:
> 
> ```
> >>> duckdb.execute("CREATE TABLE selected_with_a_null (episode_id VARCHAR)")
> >>> duckdb.execute("CREATE TABLE selected_with_a_null (episode_id VARCHAR)")
> CatalogException: Table with name "selected_with_a_null" already exists
> ```
> 
> It does not bite in a normal run, because the test executes once per session. It is still a test that passes only because of an assumption about how many times it runs, in a file whose other tests make no such assumption. Have a look at how you would write that manifest without touching shared state, and at whether the surrounding tests already suggest the shape.
> 
> Everything else is ready to go.

got it @kstonekuan

rakesh0x:
Fixed in 78a6b1f. The manifest is now built on a dedicated `duckdb.connect()` — the shape the orphan-connection read in the export round-trip test already uses — so nothing lands on the process-global default connection.

Reproduced your failure first: running the test body twice in one process hits `CatalogException: Table with name "selected_with_a_null" already exists` on the second run. After the change both runs pass and `duckdb_tables()` on the default connection shows zero tables afterwards. Re-ran the mutation too: dropping the null `episode_id` refusal still turns exactly this test red.

One thing I noticed while in there: `test_dataset_snapshot_rejects_manifest_episode_ids_absent_from_catalog` does the same `CREATE TABLE` on the default connection. If you want the file uniformly free of that assumption, happy to move that one onto a dedicated connection as well — say the word. 

@kstonekuan

kstonekuan (approved):
Fixed, merging.

Re-validated on `5b66902`: gate clean, 1,244 passed, 6 skipped. All three guards still bite after the change, and the `CREATE TABLE` now succeeds twice in one process where it used to raise on the second.

PR #283, opened 2026-08-30
test: cover malformed video input guards

### LINKED ISSUE #275: Four malformed-bitstream refusals in video.py have no test
### Current behavior

`src/hflow/video.py` refuses four kinds of malformed input. On current main (`7dedc63`), no test in the repository reaches any of them.

- `:111` `cannot map sampled frames onto an empty source stream`, in `source_log_times_for_sampled_frames` (`:87`)
- `:238` `invalid Annex B start code at byte {nal_start_offset}`
- `:264` `slice header has no complete first_mb_in_slice value`
- `:269` `slice header truncates its first_mb_in_slice value`

I replaced all four `raise` statements with `pass` at once and ran the whole suite:

```
1204 passed, 6 skipped
```

Every one of them can be deleted and nothing goes red.

### Why these four and not every raise in the file

The neighbouring guard at `:316`, `cannot insert an access-unit delimiter: no VCL NAL found`, **is** covered, by `test_ensure_aud_rejects_payload_without_coded_slice_data` (`tests/test_video.py:182`). So this is not a file nobody tested. It is four specific refusals that were missed while the ones around them were pinned, which is exactly the shape that stays missed.

These parse an H.264 bitstream that arrives from an arbitrary recorder. They are the code that decides a stream is malformed rather than quietly reading past the end of it, and the two `first_mb_in_slice` guards sit in the path that decides where one picture ends and the next begins. A truncated slice header that stops raising does not produce an error, it produces a wrong access-unit boundary.

### The pattern to copy

`tests/test_video.py` already builds malformed NAL byte strings by hand and asserts the refusal. Two examples:

- `test_split_rejects_garbage_without_aud` (`:128`) hand-assembles bytes and checks a refusal. Note it does not reach the `:238` start-code guard, so read it for the technique rather than assuming it covers this.
- `test_ensure_aud_rejects_payload_without_coded_slice_data` (`:182`) is the tightest example: build the bytes, `pytest.raises(ValueError, match="no VCL NAL")`.

The fixtures at the top of that file produce real encoded frames from ffmpeg when you need a valid stream to corrupt.

### What to build

A test for each of the four, asserting the refusal happens and that the message identifies which rule was broken.

The two `first_mb_in_slice` cases are the interesting ones. "Has no complete value" and "truncates its value" are different states of the same field, and constructing a byte string that reaches one but not the other is most of the work. Do that deliberately rather than by trial and error: read the surrounding code and say in the PR what distinguishes the two inputs you built.

`:111` is unrelated to bitstream parsing and much simpler. It is fine to do that one first.

Prove each test bites. Replace the guard it covers with `pass`, watch that one test go red and the others stay green, put it back. Put the result in the PR body. A test that passes whether or not the guard exists is the thing this issue exists to prevent.

### Definition of done

1. Each of the four guards has a test that fails when that guard is removed.
2. Nothing in `src/hflow/video.py` changes. This issue adds tests.
3. Assertions check the message, not just `ValueError`. All four raise the same type, so the type alone does not say which one fired.
4. The two `first_mb_in_slice` tests reach different guards. Show it: with `:264` removed, the `:264` test fails and the `:269` test still passes, and the reverse.
5. Byte strings are built from named constants or a helper that says what each part is, not bare hex literals. Whoever reads these next will be meeting Annex B for the first time.

### Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

ffmpeg is needed only if you choose to start from a real encoded stream. The refusals themselves are reachable with hand-built bytes and no toolchain, which is how the existing tests do it.

If you find a case where a guard is wrong rather than untested, say so on this issue before changing `video.py`. Fixing the parser is different work from pinning it.

This is the same shape as #267 but a different area of the codebase. They do not overlap.

harshitethic:
## Summary

- add focused regression tests for the four malformed-input guards in `hflow.video`
- cover the empty source-stream mapping failure and invalid Annex B start code
- distinguish the two `first_mb_in_slice` failure paths with deliberately different RBSP inputs
- assert the specific error messages
- leave `src/hflow/video.py` unchanged

Fixes #275

The two `first_mb_in_slice` inputs are intentional:
- `0x00` contains no terminating `1` bit, reaching the "no complete" guard.
- `0x04` has five leading zero bits followed by `1`, but not enough suffix bits, reaching the truncation guard.

### DIFF
--- a/docs/contrib-note.md
+++ b/docs/contrib-note.md
@@ -1,0 +1,1 @@
+Temporary note
--- a/docs/contrib-note.md
+++ b/docs/contrib-note.md
@@ -1,1 +1,0 @@
-Temporary note
--- a/tests/test_video_malformed_guards.py
+++ b/tests/test_video_malformed_guards.py
@@ -1,0 +1,48 @@
+"""Regression tests for malformed H.264 inputs rejected by hflow.video."""
+
+import pytest
+
+from hflow.video import (
+    _nal_header_offset,
+    ensure_access_unit_delimiter,
+    source_log_times_for_sampled_frames,
+)
+
+_ANNEX_B_4_BYTE_START_CODE = b"\x00\x00\x00\x01"
+_IDR_NAL_HEADER = b"\x65"  # NAL type 5: IDR coded slice.
+
+
+def _single_slice_idr(rbsp: bytes) -> bytes:
+    """Build one Annex B IDR NAL with an explicitly supplied slice-header RBSP."""
+    return _ANNEX_B_4_BYTE_START_CODE + _IDR_NAL_HEADER + rbsp
+
+
+def test_sampled_frame_mapping_rejects_empty_source_stream() -> None:
+    with pytest.raises(ValueError, match="cannot map sampled frames onto an empty source stream"):
+        source_log_times_for_sampled_frames(
+            [], source_fps=30.0, sample_fps=10.0, start_s=0.0, frame_count=1
+        )
+
+
+def test_nal_header_rejects_invalid_annex_b_start_code() -> None:
+    with pytest.raises(ValueError, match="invalid Annex B start code at byte 3"):
+        _nal_header_offset(b"\x00\x00\x00\xff", 3)
+
+
+def test_ensure_aud_rejects_slice_header_without_complete_first_mb_value() -> None:
+    # first_mb_in_slice is unsigned Exp-Golomb. An all-zero RBSP has no
+    # terminating one bit, so the decoder must reject it as incomplete.
+    malformed_slice = _single_slice_idr(b"\x00")
+
+    with pytest.raises(ValueError, match="slice header has no complete first_mb_in_slice value"):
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Validated on `aa7424b`: gate clean, 1,240 passed, 6 skipped. Four mutations, each guard failing exactly one test:

- empty source stream -> `test_sampled_frame_mapping_rejects_empty_source_stream`
- invalid start code -> `test_nal_header_rejects_invalid_annex_b_start_code`
- no complete `first_mb_in_slice` -> `test_ensure_aud_rejects_slice_header_without_complete_first_mb_value`
- truncated `first_mb_in_slice` -> `test_ensure_aud_rejects_truncated_first_mb_value`

The two `first_mb_in_slice` guards are one line apart and your tests reach different ones, which is DoD item 4. The comments explain why `\x00` has no terminating one bit while `\x04` has the one bit but runs out of suffix.

PR #285, opened 2026-08-30
test: pin SQL surface for curate and document exactly one SELECT rule

### LINKED ISSUE #279: The SQL surface curate() accepts changed with #271 and is neither documented nor pinned

Prateekathub:
## Summary
Closes #279
Added the required documentation to CATALOG.md pinning the exactly-one-SELECT rule.
Added parameterized tests for PIVOT, DESCRIBE, and SUMMARIZE. I have decided to officially support DESCRIBE and SUMMARIZE since DuckDB parses them as SELECT, so they are marked as allowed in the tests.

## Why
To fulfill the documentation and testing requirements outlined in issue #279 and ensure future developers don't accidentally remove the security guard.

## Validation
Ran `uv run ruff check --fix` and `uv run ruff format` successfully.
Local `pytest` execution failed with `ModuleNotFoundError: No module named 'fcntl'` because I am on a Windows machine, so I am relying on the GitHub Actions CI to verify the tests run successfully on Linux.

## Checklist

- [ ] I ran the relevant pytest suite. *(Skipped due to Windows fcntl error)*

### DIFF
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -155,6 +155,7 @@ hflow curate --sql-file query.sql \
 
 Pass exactly one of the positional SQL string or `--sql-file`; passing both or
 neither is an error.
+The curate() function accepts exactly one SELECT statement
 
 The manifest is written **manifest-last**: to a temp file, renamed into place
 only after the query completed, so a partial manifest is unreachable.
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -127,6 +127,16 @@ REST API and does not ship a browser client.
 
 ### Query from Python or the CLI
 
+Curation takes exactly one read-only `SELECT` statement. Common table
+expressions count, and so do the SELECTs that do not start with the word:
+`FROM episodes WHERE status = 'ok'`, a parenthesized `(SELECT ...)`, and
+`VALUES` are all accepted. Anything that is not one read-only SELECT is
+refused before it reaches the catalog, including multiple statements, DDL,
+`PIVOT` (DuckDB rewrites it into a `CREATE` followed by a `SELECT`), and the
+introspection forms `DESCRIBE`, `SUMMARIZE`, `SHOW` and `PRAGMA`. Those last
+four parse as SELECT but describe columns rather than selecting episodes, so
+a manifest built from one has no `episode_id` and every consumer refuses it.
+
 ```python
 import hflow
 
@@ -155,7 +165,6 @@ hflow curate --sql-file query.sql \
 
 Pass exactly one of the positional SQL string or `--sql-file`; passing both or
 neither is an error.
-The curate() function accepts exactly one SELECT statement
 
 The manifest is written **manifest-last**: to a temp file, renamed into place
 only after the query completed, so a partial manifest is unreachable.
--- a/tests/test_catalog_curation.py
+++ b/tests/test_catalog_curation.py
@@ -843,12 +843,12 @@ _MULTI_STATEMENT_PAYLOADS = [
         id="pivot",  # DuckDB secretly expands this to CREATE then SELECT
     ),
     pytest.param(
-        "DESCRIBE SELECT * FROM episodes",
+        "DESCRIBE SELECT episode_id FROM episodes",
         False,
         id="describe",
     ),
     pytest.param(
-        "SUMMARIZE SELECT ...",
+        "SUMMARIZE SELECT episode_id FROM episodes",
         False,
         id="summarize",
     ),
--- a/tests/test_catalog_curation.py
+++ b/tests/test_catalog_curation.py
@@ -837,6 +837,21 @@ _MULTI_STATEMENT_PAYLOADS = [
         False,
         id="legitimate-single-select",
     ),
+    pytest.param(
+        "PIVOT episodes ON status USING count(*)",
+        True,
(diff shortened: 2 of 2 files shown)


kstonekuan (changes requested):
`DESCRIBE` and `SUMMARIZE` pass the guard but do not produce a manifest: their output has no `episode_id`, and the next step refuses it. Pin them as incidental rather than supported.

`TABLE episodes` and `VALUES (1), (2)` are missing from the table and both work.

The docs line has no full stop, says `curate()` in a CLI section, and sits in the `--sql-file` paragraph.

`_MULTI_STATEMENT_PAYLOADS` now describes three of its rows.

kstonekuan:
#453 merged and settled the `DESCRIBE`/`SUMMARIZE` question your way: both are refused now, along with `PRAGMA` and `SHOW`.

Rebase first. Two of your rows fail on current main, and the surface also grew: FROM-first, parenthesized SELECTs and `VALUES` are accepted and none is in your table, plus `TABLE episodes`.

kstonekuan (approved):
LGTM, merging.

Your `PIVOT` row and comment went in untouched; that is the one that reads like a false positive and is not.

#453 settled `DESCRIBE`/`SUMMARIZE` your way, so both flipped to refused and `PRAGMA`/`SHOW` joined them. Added the four rows the surface grew (`TABLE`, `VALUES`, FROM-first, parenthesized), gave each row an expected row count, renamed the list, and moved the docs line into the curation section.

PR #286, opened 2026-08-30
Pin the SQL surface curate() accepts and document the one-SELECT rule

### LINKED ISSUE #279: The SQL surface curate() accepts changed with #271 and is neither documented nor pinned

RinZ27:
## Summary

Pins which SQL forms `_reject_non_single_select` accepts and refuses, and documents the one-SELECT rule in `docs/CATALOG.md`.

## Why

After #271 landed the guard, the accepted SQL surface changed in both directions but nothing wrote it down. `PIVOT` is now refused (DuckDB rewrites it to `[CREATE, SELECT]`, so the count check catches it — arguably correct rather than merely strict). `DESCRIBE SELECT ...` and `SUMMARIZE SELECT ...` moved the other way: both report as `SELECT` and now pass the guard, where previously they were syntax errors inside the old `COPY (...)` wrapper.

Without a pin, someone will eventually "relax" the count check to let `PIVOT` through without realising the `CREATE` it admits is the thing the guard exists to stop. And `DESCRIBE`/`SUMMARIZE` working by accident means nothing would notice if a future DuckDB reclassified them.

**Decision on DESCRIBE and SUMMARIZE:** both are accepted incidentally. The test records that fact and names them as not a promised surface, matching what the docs paragraph now says. If a future DuckDB version reclassifies either one, the test goes red and forces a deliberate choice rather than silent breakage.

## Validation

```
$ uv run pytest tests/test_catalog_curation.py::test_reject_non_single_select_sql_surface -v
11 passed in 0.78s

$ uv run pytest tests/test_catalog_curation.py -q
80 passed, 5 errors in 2.98s
# (errors are pre-existing ffmpeg._binary — no ffmpeg on this machine)

$ uv run ruff check --fix && uv run ruff format && uv run ty check
All checks passed
```

## Checklist

Closes #279

### DIFF
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -186,6 +186,15 @@ posture a service uses for tenant-supplied SQL
 ([docs/HOSTING.md](./HOSTING.md#trust-model)). The default stays
 unrestricted for your own exploration.
 
+Curation SQL must be exactly one `SELECT` statement. `curate()` parses the
+input with DuckDB's `extract_statements` before executing anything, and
+refuses multi-statement strings, DDL, DML, and queries that DuckDB expands
+internally (notably `PIVOT`, which DuckDB rewrites to a `CREATE` followed by
+a `SELECT`). `WITH ... SELECT`, `TABLE`, and `VALUES` all classify as a
+single `SELECT` and are accepted. `DESCRIBE SELECT ...` and
+`SUMMARIZE SELECT ...` also pass the guard because DuckDB reports them as
+`SELECT`; they work today but are not a promised surface.
+
 ### The view surface
 
 `hflow.open_catalog_connection(catalog_root)` returns a DuckDB connection with
--- a/tests/test_catalog_curation.py
+++ b/tests/test_catalog_curation.py
@@ -909,6 +909,111 @@ def test_stage_manifest_and_count_rejects_non_single_select(
         connection.close()
 
 
+# -- Pin the SQL surface that _reject_non_single_select accepts (#279) --------
+#
+# DuckDB's extract_statements classifies each input and the guard requires
+# exactly one statement whose type is SELECT.  Several query forms land in
+# surprising buckets:
+#
+#   PIVOT  → DuckDB rewrites it to [CREATE, SELECT], so the *count* check
+#            catches it.  That is arguably correct rather than merely strict.
+#   DESCRIBE / SUMMARIZE → both report as SELECT, so they pass the guard.
+#            Before #271 they were syntax errors inside the COPY (...) wrapper.
+#   TABLE / VALUES → report as SELECT, so they pass.
+#
+# Whether DESCRIBE and SUMMARIZE *should* be accepted is a policy decision.
+# As of this writing (main @ 48d53cf, #271 merged) they are accepted, and
+# this table records that fact so the next reader sees a deliberate pin
+# rather than an accident.
+
+_SQL_SURFACE_CASES = [
+    # (sql, expected_refused)
+    pytest.param(
+        "SELECT 1",
+        False,
+        id="plain-select",
+    ),
+    pytest.param(
+        "WITH cte AS (SELECT 1 AS x) SELECT * FROM cte",
+        False,
+        id="with-select",
+    ),
+    pytest.param(
+        "TABLE episodes",
+        False,
+        id="table-shorthand",
+    ),
+    pytest.param(
+        "VALUES (1), (2)",
+        False,
+        id="values-expression",
+    ),
+    pytest.param(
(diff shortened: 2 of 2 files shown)


kstonekuan:
Sorry, closing this one. #279 is assigned and #285 has been open on it since 10:20, so first-PR-wins applies. Timing, nothing about your work.

For next time: comment on an issue before you start and I'll assign it, and check whether it already has an assignee or a linked PR. An issue with an assignee is taken, everything else is fair game.

The starter pool is picked clean this morning, more go up regularly. If you'd rather not wait, the most useful thing anyone can do here is run HFlow against a real corpus, Egocentric-10K or Egocentric-100K on HF, and report what breaks, what's slow, what's awkward. See #287.

<https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

PR #288, opened 2026-08-30
fix: classify transform content refusals as source-unsupported, not infrastructure

aleks-drozy:
## Summary

`classify_ingest_failure()` now files a transform's content refusal (non-conforming pass-through video, a raw-image schema, an unsupported compressed-image format, or a derived-topic collision) under a new `IngestFailureKind.SOURCE_UNSUPPORTED`, instead of the catch-all `infrastructure` bucket. `SourceNotFound` still classifies as `source-missing` and any `McapError` still classifies as `source-unreadable`.

## Why

Per #203: `infrastructure` is what a retry is for, but a content refusal fails identically on every retry, so an operator draining the ledger by kind retries a batch that cannot succeed, and a converter author debugging their own output is told the fault is on our side. This implements shape (1)+(2) from the issue, which the issue itself notes compose: a dedicated `SourceNotConforming(ValueError)` exception, raised at the transform's content-refusal sites (previously plain `ValueError`/`NotImplementedError` with the same messages), classified into the new `source-unsupported` kind. `docs/CATALOG.md` and `docs/how-to/write-a-converter.md` are updated to document the new kind; the write-a-converter guidance that told converter authors to bypass `failure_kind` entirely is rewritten now that the kind is actionable.

Scope was kept to exactly the four refusal categories the issue names. Left untouched (still plain `ValueError`, still `infrastructure`): transform config validation (`crf`/`gop_seconds`/`chunk_size`), decoder/encoder-resolution failures, "mixes image formats", and "derived topics must be unique" (that one is caller-config, not source content).

## Validation

Ran inside WSL2/Ubuntu on the Linux filesystem, per CONTRIBUTING.md:

```
uv sync --locked --all-extras
uv run ruff check          # All checks passed!
uv run ty check            # All checks passed!
uv run pytest -q tests/test_ingest_ledger.py tests/test_processing_regressions.py
# 50 passed in 1.68s
uv run pytest -q
# 1133 passed, 7 skipped, 39 failed, 79 errors
```

Every failure/error in the full run traces to `PinnedDownloadError`/`FfmpegNotFoundError` — the pinned ffmpeg build's release URL 404s in this sandbox with no network path to it — confirmed unrelated to this change by running one such failing test in isolation and inspecting the traceback (`src/hflow/ffmpeg/_binary.py:184`). No failure touches a file this PR modifies.

## Checklist

---
Investigated and implemented with AI assistance (Claude Code — Claude Fable 5 for the initial repo/issue investigation, Claude Sonnet 5 for the implementation and verification), under my direction and review; every change and the WSL2 test run above were reviewed by me before opening this PR.

### DIFF
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -55,6 +55,18 @@ next to the verbatim `error_type` and `message` rather than in place of them;
 anything unrecognized is `infrastructure`, never an accusation against the
 recording.
 
+`source-unsupported` was added after `ingest_failures` already existed in
+deployed catalogs. Rows written before an ingest binary carrying this change
+keep whatever kind that binary assigned at write time -- a content refusal
+recorded as `infrastructure` by an older binary is not retroactively
+reclassified, because the ledger is append-only and nothing rewrites past
+rows. An operator draining an old ledger by kind should expect
+`infrastructure` rows from before the cutover to include content refusals
+alongside genuine infrastructure trouble; `error_type` and `message` remain
+the ground truth for any row, which is exactly why they are kept verbatim
+next to the heuristic. Only rows written by an ingest binary that already
+raises `SourceNotConforming` (see below) can carry `source-unsupported`.
+
 Three durability rules govern writes:
 
 - **Content-addressed**: `episode_id` is a sha256 of the canonical file's
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -44,7 +44,7 @@ append at all:
 | `measurements` | measurement key | `key`, typed value columns (`value_double`, `value_text`, `value_bool`), the producing `check_name`/`check_version` |
 | `tags` | tag | `check_name`, `tag` |
 | `intervals` | labeled time span | `label`, `start_ns`, `end_ns` |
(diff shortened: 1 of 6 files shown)


kstonekuan (approved):
LGTM, merging. You answered both questions before I asked them again, and the "only way in" argument is the right one: one `isinstance` check, no string matching, bare `ValueError` falls back to `infrastructure`. The cutover note in CATALOG.md is what an operator draining an old ledger actually needs.

Validated on your branch: `ruff check`, `ruff format --check`, `ty check` clean, 1252 passed / 6 skipped. Your 39 failures were the pinned ffmpeg download 404ing in your sandbox, nothing to do with the change.

I also checked the tests can detect a regression, since a type-based classifier is easy to break silently. Reverted the raw-image site to `NotImplementedError` and `test_raw_image_schema_classifies_as_source_unsupported` failed. Good, that's the property that matters.

One gap, not blocking. `transform.py:510` refuses a camera topic that mixes jpeg and png with a bare `ValueError`, so it still lands in `infrastructure`, while the unsupported-format refusal one line later is now `source-unsupported`:

```
mixes image formats      -> infrastructure
unsupported single format -> source-unsupported
```

Same channel, same kind of complaint about its contents, two different kinds. Not in the issue's list of four so you were right to scope it out, but it's the one place the new vocabulary is inconsistent. I'll file it.

Issue #287, opened 2026-08-30
Real-corpus friction (Egocentric-10K/100K): no WebDataset ingest path; the video adapter's JPEG detour costs 2.3x landing size and a double lossy transcode for H.265 sources


rakesh0x:
## Context

I ran HFlow against the shape of `builddotai/Egocentric-10K` and `builddotai/Egocentric-100K` (the largest real egocentric factory corpora on Hugging Face, and the same source the `examples/egocentric` quickstart shard comes from). Both corpora are WebDataset-style tars of ~180 s H.265/MP4 clips (1080p30 for 10K, 456x256 for 100K) plus per-clip JSON sidecars and per-worker `intrinsics.json`.

The finding: **there is no first-party ingestion path for this corpus shape, and the only available video-to-MCAP adapter detours through JPEG, which costs roughly 2.3x landing size and a double lossy transcode for H.265 sources.**

## What exists today

- `src/hflow/importers/` contains only the LeRobot v3 converter. WebDataset-style tars have no importer; users must write their own converter (the documented escape hatch) or reach for `hflow.testing.write_video_episode`.
- `write_video_episode` is explicitly a demo/test helper, and it decodes the source video and re-encodes **every frame as JPEG** `sensor_msgs/CompressedImage` (`_build_video_camera_messages` in `src/hflow/testing.py`). The canonical transform then re-encodes that JPEG into in-band H.264.

So for an H.265 corpus the media is transcoded **HEVC -> JPEG -> H.264** (two lossy steps, one avoidable) even though H.264 is the canonical target and ffmpeg decodes HEVC directly.

## Measurements

Profile-matched HEVC media (ffmpeg `testsrc2`, same resolution/fps/codec as the corpora; the real footage is gated so I could not download it for this test). Apple Silicon macOS, ffmpeg 8.1.1 from Homebrew, hflow at `78a6b1f`.

30 s @ 1920x1080 @ 30 fps (Egocentric-10K profile):

| Step | Result |
| --- | --- |
| Source HEVC bytes (same span) | ~7 MB |
| `write_video_episode` landing MCAP (JPEG) | **55.3 MB** (~8x source bitrate) |
| Canonical episode H.264 media bytes | 24.5 MB (landing is 2.3x this) |
| Adapter wall time | 1.8 s |
| Cold full pipeline (`app.test`) | **12.0 s (~2.5x slower than realtime)** |
| of which `camera_frame_stats` decode | 8.1 s |

[...]

PR #289, opened 2026-08-30
Pin tests for the snapshot export refusal guards (issue #276)

k0shir0:
## Pin tests for the snapshot export refusal guards (issue #276)

`src/hflow/snapshot.py` has three refusal guards in the export path that were not pinned by any test. On current `main`, replacing each `raise` with `pass` keeps the whole suite green. This change pins all three.

### What's covered

1. **`_parse_dataset_snapshot_destination` refuses a symlink as the export destination.** Called directly with both `overwrite=False` and `overwrite=True`. The issue calls this out specifically: the guard runs *before* `overwrite` is consulted, and a test that only exercised one branch would still pass if the check were moved below the overwrite branch.
2. **`export_dataset_snapshot` refuses a manifest whose `episode_id` column contains NULL.** Mirrors the structure of the existing `test_dataset_snapshot_rejects_manifest_episode_ids_absent_from_catalog`. Builds a parquet manifest with `INSERT INTO selected VALUES (NULL)`.
3. **`_parse_dataset_snapshot_destination` refuses a symlink at `format.json` inside an existing destination.** The issue asks me to consider whether this second symlink check ten lines below the first wants the same treatment. It does — the pattern is identical (symlink whose `is_file()` follows it to a real file would otherwise be indistinguishable from a regular marker file), so I added it. Both the destination-level and the format-marker-level symlink checks were written together and should be pinned together.

### User-visible outcome

Each guard now has a regression test that goes red the moment the `raise` is replaced with `pass`. A future refactor that drops or reorders any of them breaks the suite instead of silently shipping.

### Validation

```bash
uv run pytest tests/test_dataset_snapshot.py -q
# 12 passed
```

```bash
uv run ruff check tests/test_dataset_snapshot.py
# All checks passed!

uv run ruff format --check tests/test_dataset_snapshot.py
# 1 file already formatted
```

Full suite from the repo root (excluding the opt-in ffmpeg and storage suites that need `HFLOW_NETWORK_TESTS=1` / a writable object-store prefix):

```bash
uv run pytest --ignore=tests/test_ffmpeg.py -q
# 1146 passed, 6 skipped
```

### Sabotage proof (Definition of Done from #276)

Replacing each guard's `raise` with `pass` (then restoring the file):

| Guard disabled | Test that goes red |
| --- | --- |
| Destination symlink check (`is_symlink()` at line 486) | `test_parse_dataset_snapshot_destination_refuses_symlink_before_overwrite_check` |
| Null `episode_id` in manifest (line 145) | `test_dataset_snapshot_rejects_manifest_with_null_episode_id` |
| Format-marker symlink check (`is_symlink()` at line 500) | `test_parse_dataset_snapshot_destination_refuses_symlinked_format_marker` |

In each case the named test fails, and restoring the guard returns the file to green.

### DIFF
--- a/tests/test_dataset_snapshot.py
+++ b/tests/test_dataset_snapshot.py
@@ -557,3 +557,51 @@ def test_snapshot_samples_report_unverified_for_a_crashed_critical_check(
         [str(output_directory / "samples.parquet")],
     ).fetchone()
     assert sample_row == (append_result.episode_id, "unverified")
+
+
+def test_parse_dataset_snapshot_destination_refuses_symlink_before_overwrite_check(
+    tmp_path: Path,
+) -> None:
+    target = tmp_path / "real-target"
+    target.mkdir()
+    link = tmp_path / "link-to-target"
+    link.symlink_to(target)
+
+    with pytest.raises(ValueError, match=r"must not be a symlink"):
+        snapshot_module._parse_dataset_snapshot_destination(link, overwrite=False)
+
+    with pytest.raises(ValueError, match=r"must not be a symlink"):
+        snapshot_module._parse_dataset_snapshot_destination(link, overwrite=True)
+
+
+def test_dataset_snapshot_rejects_manifest_with_null_episode_id(tmp_path: Path) -> None:
+    catalog = Catalog(tmp_path / "catalog")
+    manifest = tmp_path / "manifest.parquet"
+    duckdb.execute("DROP TABLE IF EXISTS selected")
+    duckdb.execute("CREATE TABLE selected (episode_id VARCHAR)")
+    duckdb.execute("INSERT INTO selected VALUES (NULL)")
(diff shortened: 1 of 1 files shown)


kstonekuan:
Sorry, the timing went against you. #276 was closed a few hours before you opened this, by #282, which pins the same three guards.

Your opening claim is the clearest way to see it: replacing each `raise` with `pass` turns tests red on main now, it doesn't leave the suite green. For next time, a closed issue is done, and the `good first issue` list filters to open ones.

You and #282 both reached for the same shape, calling `_parse_dataset_snapshot_destination` directly with both `overwrite` values.

Pool is picked clean at the moment, more go up regularly. The highest-value thing anyone can do here is run HFlow against a real corpus, Egocentric-10K or Egocentric-100K on HF, and report what breaks, what's slow, what's awkward. See #287.

<https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

PR #290, opened 2026-08-30
fix: handle invalid compare inputs

### LINKED ISSUE #260: Build AI evaluation: compare crashes with a traceback on a missing file while run refuses cleanly

raghulajithn:
## Summary

Fixes the `compare` command so missing or malformed summary files are reported as CLI errors instead of escaping as unhandled tracebacks.

## Why

The `compare` command previously allowed errors from missing or malformed summary files to escape as unhandled tracebacks.

The exception handling is intentionally placed around the match statement rather than inside the compare arm. This keeps the error handling at the CLI boundary and ensures that the same handling applies consistently to all current subcommands and future subcommands without duplicating the error handling logic.

## Validation

Ran all of the following validations successfully:

```bash
uv sync --locked --all-extras

Resolved 124 packages in 1ms
Uninstalled 61 packages in 142ms

uv run ruff check --fix

All checks passed!

uv run ruff format

195 files left unchanged

uv run ty check

All checks passed!

uv run pytest -q

1244 passed, 6 skipped in 88.32s (0:01:28)

23 passed in 4.70s
```

Closes #260

## Checklist

### DIFF
--- a/examples/build_ai_evaluation/evaluate.py
+++ b/examples/build_ai_evaluation/evaluate.py
@@ -800,7 +800,13 @@ def run_evaluation(configuration: EvaluationConfiguration, *, download: bool) ->
 
 
 def compare_summaries(summary_paths: Sequence[Path]) -> None:
-    summaries = [json.loads(path.read_text()) for path in summary_paths]
+    summaries = []
+    for path in summary_paths:
+        content = path.read_text()
+        try:
+            summaries.append(json.loads(content))
+        except json.JSONDecodeError as error:
+            raise ValueError(f"invalid JSON in {path}: {error}") from error
     print(
         "| run | dataset | source | hand n | 1+ hands | 2 hands | active n | "
         "active yes | hand agreement | active agreement |"
@@ -933,17 +939,17 @@ def _argument_parser() -> argparse.ArgumentParser:
 def main() -> None:
     parser = _argument_parser()
     arguments = parser.parse_args()
-    match arguments.command:
-        case "run":
-            try:
+    try:
+        match arguments.command:
+            case "run":
                 configuration = _run_configuration_from_arguments(arguments)
                 run_evaluation(configuration, download=arguments.download)
-            except (FileNotFoundError, RuntimeError, ValueError) as error:
-                parser.error(str(error))
-        case "compare":
-            compare_summaries(arguments.summaries)
-        case unknown_command:
-            raise AssertionError(f"unhandled command: {unknown_command}")
+            case "compare":
+                compare_summaries(arguments.summaries)
+            case unknown_command:
+                raise AssertionError(f"unhandled command: {unknown_command}")
+    except (FileNotFoundError, ValueError, RuntimeError) as error:
+        parser.error(str(error))
 
 
 if __name__ == "__main__":
--- a/examples/build_ai_evaluation/evaluate.py
+++ b/examples/build_ai_evaluation/evaluate.py
@@ -932,7 +932,12 @@ def _argument_parser() -> argparse.ArgumentParser:
     )
 
     compare_parser = subparsers.add_parser("compare", help="compare completed run summaries")
-    compare_parser.add_argument("summaries", nargs="+", type=Path)
+    compare_parser.add_argument(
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM.

Validated on `8840f0d`: gate clean, root suite 1,244 passed 6 skipped, example suite 23 passed. Observed:

```
compare /tmp/nope.json -> exit=2, 2 lines, "No such file or directory: /tmp/nope.json"
compare /tmp/bad.json  -> exit=2, 2 lines, "invalid JSON in /tmp/bad.json: Expecting value..."
```

Wrapping the `JSONDecodeError` earns its keep even though it is already a `ValueError` and would have been caught anyway: the raw one does not name the file, so a multi-file compare would tell you something was malformed without saying which.

Moved `compare` back outside the shared `try` and exactly your two refusal tests go red, other 21 stay green.

Merging.

PR #302, opened 2026-08-31
fix: reject Windows drive-letter paths in storage key validation

Raghav7-tech:
_validated_relative_key() checked absoluteness using PurePosixPath, which
has no concept of Windows drive letters, so a key shaped like
"C:/Windows/win.ini" passed validation as "relative". LocalStorageRoot
later joins a validated key onto the real root with self.path / key using
the native platform's pathlib semantics: on native Windows, joining a
drive-letter path onto an existing path discards the left operand entirely
(documented pathlib behavior), so the join resolves to the drive-letter
path outright, escaping the intended storage root.

## Summary
Adds a check for PureWindowsPath(...).is_absolute() alongside the existing
PurePosixPath check, so drive-letter-anchored keys are rejected regardless
of which OS validates the key.

## Why
Found while reviewing storage.py's path-validation logic after the #271
review. This repo's own docs state Windows is supported only via WSL2,
where Python's pathlib resolves as POSIX and this escape doesn't occur --
so I have NOT confirmed a fully wired, attacker-reachable path to
LocalStorageRoot with externally-influenced key input on native Windows
specifically. I'm flagging this as a real logic bug in the guard itself,
independent of confirmed end-to-end reachability, since the fix is small
and costs nothing. Happy to be told this isn't practically reachable --
would appreciate the correction if so.

## Validation
- Added a regression test confirming drive-letter keys ("C:/...", "D:/...",
  backslash form) are rejected, while non-drive-letter colon-containing
  keys remain valid (gated to skip the write on native Windows, since
  colons in filenames are illegal there regardless of this check).
- Confirmed the test fails against the original code (raises
  FileNotFoundError instead of ValueError -- proving the guard wrongly
  accepted the key) and passes against the fix.
- Full test suite: 1024 passed, 6 skipped, 0 regressions.
- ruff check / format: clean.


kstonekuan (changes requested):
The gap is real and the guard is right. `_validated_relative_key('C:/Windows/win.ini')` returns it unchanged on main, and your test goes red when I drop the `PureWindowsPath` clause. Gate clean, 1,251 passed, 6 skipped.

One change: the docstring overstates the impact, and it's the part future readers will trust.

"escaping the intended root" can't happen on any platform HFlow runs on. `storage.py:36` imports `fcntl` at module scope, so hflow doesn't import at all on native Windows, which is the only place `Path / 'C:/...'` discards the left operand. On Linux the join is contained:

```
PurePosixPath('C:/Windows/win.ini').is_absolute()   False
PureWindowsPath('C:/Windows/win.ini').is_absolute() True
Path('/data/root') / 'C:/Windows/win.ini'  ->  /data/root/C:/Windows/win.ini
```

Contained, not an escape. So this is hardening a validator against a platform that can't currently reach it, which is worth doing (the validator's job is to be right about keys, wherever they came from, and a key can arrive from a Windows client and be validated on Linux). It isn't a live path traversal, and a docstring that says it is will send the next person hunting for a vulnerability that isn't there.

Rewrite that paragraph to say what's true: the check is platform-independent because keys cross platforms even though the process doesn't. Keep the `pathlib` explanation, drop "escaping the intended root".

While you're in there: no ` -- ` in new docstrings. Period, comma, colon, or parentheses.

Nothing else. The error message listing all three refusals reads well, and covering the backslash form as well as the forward-slash one is the right pair to test.

Raghav7-tech:
> The gap is real and the guard is right. `_validated_relative_key('C:/Windows/win.ini')` returns it unchanged on main, and your test goes red when I drop the `PureWindowsPath` clause. Gate clean, 1,251 passed, 6 skipped.
> 
> One change: the docstring overstates the impact, and it's the part future readers will trust.
> 
> "escaping the intended root" can't happen on any platform HFlow runs on. `storage.py:36` imports `fcntl` at module scope, so hflow doesn't import at all on native Windows, which is the only place `Path / 'C:/...'` discards the left operand. On Linux the join is contained:
> 
> ```
> PurePosixPath('C:/Windows/win.ini').is_absolute()   False
> PureWindowsPath('C:/Windows/win.ini').is_absolute() True
> Path('/data/root') / 'C:/Windows/win.ini'  ->  /data/root/C:/Windows/win.ini
> ```
> 
> Contained, not an escape. So this is hardening a validator against a platform that can't currently reach it, which is worth doing (the validator's job is to be right about keys, wherever they came from, and a key can arrive from a Windows client and be validated on Linux). It isn't a live path traversal, and a docstring that says it is will send the next person hunting for a vulnerability that isn't there.
> 
> Rewrite that paragraph to say what's true: the check is platform-independent because keys cross platforms even though the process doesn't. Keep the `pathlib` explanation, drop "escaping the intended root".
> 
> While you're in there: no `--` in new docstrings. Period, comma, colon, or parentheses.
> 
> Nothing else. The error message listing all three refusals reads well, and covering the backslash form as well as the forward-slash one is the right pair to test.

You're right, and thanks for catching that — I overstated it. storage.py imports fcntl at module scope, so hflow doesn't import at all on native Windows, meaning the join scenario I described could never actually execute on any platform this code currently runs on. Good catch on the exact reachability.

Reworded the docstring to say what's actually true: the check is platform-independent because storage keys can be generated by a client on one OS and validated by a process on a different OS, so the validator should recognize a path as absolute under either convention regardless of which OS is running the check. Kept the concrete pathlib join example since it's accurate and useful context. Also removed the em-dashes.

Pushed as b9878a5. Suite still clean (1024 passed, 6 skipped).

kstonekuan (changes requested):
Function docstring is right now. Two things left, and the second is a real hole.

**The claim I asked you to drop moved into the test docstring.** `test_windows_drive_letter_keys_are_refused` still says pathlib "resolves to the drive-letter path outright, escaping the root entirely". Same sentence, same problem: `storage.py:36` imports `fcntl`, so nothing reaches that join on native Windows. Say it the way you said it in the function. Also no ` -- ` in new prose.

**`is_absolute()` is the wrong predicate.** A drive-*relative* key has a drive and no root, so it isn't absolute, and it slips through:

```
refused   'C:/Windows/win.ini'
refused   'D:/secrets.txt'
refused   '//server/share/x'
ACCEPTED  'C:foo.txt'          <- has a drive
```

The join discards the root for that shape too:

```
PureWindowsPath('D:/data/root') / 'C:foo.txt'          ->  C:foo.txt
PureWindowsPath('D:/data/root') / 'C:/Windows/win.ini' ->  C:\Windows\win.ini
```

So the guard catches the form that looks dangerous and misses the form with identical behavior. Key on the drive instead:

```python
if key.is_absolute() or PureWindowsPath(str(key)).drive or ".." in key.parts:
```

I ran that: it refuses all six drive and UNC shapes including `C:foo.txt` and bare `C:`, still accepts `weird/C:notdrive.txt`, `file:with:colons.txt` and every ordinary key, and the suite stays at 1251 passed / 6 skipped. Your non-drive-colon case was the right thing to test, it just needs the version that also covers `C:foo.txt`.

Raghav7-tech:
> Function docstring is right now. Two things left, and the second is a real hole.
> 
> **The claim I asked you to drop moved into the test docstring.** `test_windows_drive_letter_keys_are_refused` still says pathlib "resolves to the drive-letter path outright, escaping the root entirely". Same sentence, same problem: `storage.py:36` imports `fcntl`, so nothing reaches that join on native Windows. Say it the way you said it in the function. Also no `--` in new prose.
> 
> **`is_absolute()` is the wrong predicate.** A drive-_relative_ key has a drive and no root, so it isn't absolute, and it slips through:
> 
> ```
> refused   'C:/Windows/win.ini'
> refused   'D:/secrets.txt'
> refused   '//server/share/x'
> ACCEPTED  'C:foo.txt'          <- has a drive
> ```
> 
> The join discards the root for that shape too:
> 
> ```
> PureWindowsPath('D:/data/root') / 'C:foo.txt'          ->  C:foo.txt
> PureWindowsPath('D:/data/root') / 'C:/Windows/win.ini' ->  C:\Windows\win.ini
> ```
> 
> So the guard catches the form that looks dangerous and misses the form with identical behavior. Key on the drive instead:
> 
> ```python
> if key.is_absolute() or PureWindowsPath(str(key)).drive or ".." in key.parts:
> ```
> 
> I ran that: it refuses all six drive and UNC shapes including `C:foo.txt` and bare `C:`, still accepts `weird/C:notdrive.txt`, `file:with:colons.txt` and every ordinary key, and the suite stays at 1251 passed / 6 skipped. Your non-drive-colon case was the right thing to test, it just needs the version that also covers `C:foo.txt`.

Thanks — all three points addressed in the latest push (8a27719):

Guard: switched from PureWindowsPath(...).is_absolute() to PureWindowsPath(...).drive. You're right that is_absolute() is wrong here — a drive-relative key like C:foo.txt has .drive == "C:" but .is_absolute() == False, so it slipped through the old check. .drive is non-empty for every shape that needs to be refused: drive-absolute (C:/foo), drive-relative (C:foo), and UNC (//server/share).

Test: added explicit cases for C:foo.txt and bare C: alongside the existing drive-absolute cases. The weird/C:notdrive.txt acceptance case is still there (gated to non-Windows where colons are legal at the OS level), which now also covers the C:foo.txt shape you called out.

Docstrings: rewrote both the function docstring and the test docstring — no em-dashes, no "escaping the root entirely" claim in either. Both now say what is actually true: the check is platform-independent because a key from a Windows client can be validated on a Linux server, so the validator must recognize Windows-anchored paths under either convention.

kstonekuan (approved):
LGTM, merging. All three done, and the docstrings now say the true thing in both places.

Re-ran the acceptance probe against your version, refusals and acceptances together:

```
refused      'C:/Windows/win.ini'      accepted     'weird/C:notdrive.txt'
refused      'C:foo.txt'               accepted     'file:with:colons.txt'
refused      'C:'                      accepted     'CC:/notadrive.txt'
refused      '//server/share/x'        accepted     'episodes/abc123.mcap'
refused      '/etc/passwd'             accepted     'a/b/c.parquet'
refused      'a/../../b'

0 problems
```

`CC:/notadrive.txt` is the one I added on top of yours: two letters before the colon is not a drive, and `.drive` correctly leaves it alone. Worth knowing the predicate is that precise, because a regex over `[A-Za-z]:` would have got that wrong.

Gate clean, 1251 passed / 6 skipped.

Pushed one fixup, `35e2065`: hoisted `import sys` out of the test body to module scope. House style, nothing wrong with it.

PR #306, opened 2026-08-31
fix(lerobot): reject boolean vector dimensions

### LINKED ISSUE #300: LeRobot treats a boolean vector dimension as one
## Current behavior

`_derive_numeric_schema` accepts a one-dimensional positive integer shape at [`src/hflow/importers/lerobot.py:478-493`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L478-L493). The check uses `isinstance(declared_shape[0], int)`, and Python booleans are integers.

Measured on `main` at `1f5c7c8`:

```python
_derive_numeric_schema(
    "action",
    {"dtype": "float32", "shape": [True]},
)
```

returns:

```text
_NumericSchema(name="action", dim=1)
```

A malformed JSON feature schema can therefore register a one-element ROS array from a boolean declaration. The existing unsupported-schema test covers float64, multidimensional, and empty shapes at [`tests/test_lerobot_converter.py:145-157`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_lerobot_converter.py#L145-L157), but not booleans.

## Pattern to copy

The public `episode_index` boundary explicitly excludes `bool` before accepting integers at [`src/hflow/importers/lerobot.py:523-526`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L523-L526).

`TransformConfig` applies the same integer-but-not-boolean rule to `crf` and `chunk_size_bytes` at [`src/hflow/transform.py:137-160`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/transform.py#L137-L160).

## What to build

Exclude booleans from LeRobot numeric vector dimensions. Extend `test_derive_numeric_schema_rejects_unsupported` with boolean shapes and keep the existing contextual unsupported-feature message.

## Definition of done

1. `shape: [True]` and `shape: [False]` are refused as unsupported numeric feature schemas.
2. A positive integer dimension such as `[1]` or `[6]` remains accepted.
3. Zero, negative, floating-point, empty, and multidimensional shapes remain refused.
4. The error names the feature, dtype, and declared shape.
5. The regression fails against current `main` and passes with the fix.
6. No output is written for a dataset whose required numeric feature has a boolean dimension.
7. No new dependency or converter-version bump is needed.

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

rootsec1:
Fixes #300.

Python treats `bool` as a subclass of `int`, so LeRobot metadata such as `shape: [true]` passed the integer-dimension check and became a one-dimensional numeric feature. This change rejects booleans explicitly while leaving valid integer dimensions alone.

The tests cover both `true` and `false` dimensions and check that importer validation stops before writing a landing dataset or prepared manifest.

Tested with:

- `uv run pytest -q tests/test_lerobot_converter.py`
- `uv run pytest -q` (1251 passed, 6 skipped)
- `uv run ruff check --fix`
- `uv run ruff format`
- `uv run ty check`

### DIFF
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -484,7 +484,12 @@ def _derive_numeric_schema(feature_name: str, feature_specification: dict) -> _N
             f"dtype={declared_dtype}, shape={declared_shape} "
             "(only float32 fixed-width numeric vectors are supported)"
         )
-    if len(declared_shape) != 1 or not isinstance(declared_shape[0], int) or declared_shape[0] < 1:
+    if (
+        len(declared_shape) != 1
+        or isinstance(declared_shape[0], bool)
+        or not isinstance(declared_shape[0], int)
+        or declared_shape[0] < 1
+    ):
         raise ValueError(
             f"unsupported feature {feature_name}: "
             f"dtype={declared_dtype}, shape={declared_shape} "
--- a/tests/test_lerobot_converter.py
+++ b/tests/test_lerobot_converter.py
@@ -155,6 +155,46 @@ def test_derive_numeric_schema_rejects_unsupported() -> None:
         _DERIVE("observation.state", {"dtype": "float32", "shape": [2, 3]})
     with pytest.raises(ValueError, match="unsupported feature"):
         _DERIVE("observation.state", {"dtype": "float32", "shape": []})
+    for shape in ([True], [False]):
+        with pytest.raises(
+            ValueError,
+            match=rf"unsupported feature action: dtype=float32, shape=\[{shape[0]}\]",
+        ):
+            _DERIVE("action", {"dtype": "float32", "shape": shape})
+
+
+def test_import_rejects_required_boolean_dimension_without_dataset_output(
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM.

Validated on `1f5c7c8`: gate clean, 1,251 passed, 6 skipped. Dropped the `isinstance(declared_shape[0], bool)` clause and both `test_derive_numeric_schema_rejects_unsupported` and `test_import_rejects_required_boolean_dimension_without_dataset_output` go red, so the guard is pinned from the unit and the importer level.

Ordering matches the rest of the repo: `bool` checked before `int`, same as `_require_int` in `hflow.testing`. That is now the tenth place this family has been closed.

The second test is the one that earns its keep. Asserting no `landing/` and no `prepared-manifest.json` pins that the refusal happens before anything is written, which is the part a future refactor could quietly lose.

Merging.

PR #313, opened 2026-08-31
fix(lerobot): refuse non-finite FPS values in meta/info.json

### LINKED ISSUE #299: LeRobot metadata accepts non-finite FPS values
## Current behavior

The LeRobot metadata boundary checks that `fps` is numeric and greater than zero at [`src/hflow/importers/lerobot.py:312-321`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L312-L321). It does not check finiteness.

Measured on `main` at `1f5c7c8`, both `float("nan")` and `float("inf")` pass this validation and continue into episode metadata discovery. With an empty fake tree, both fail later as:

```text
RuntimeError: no meta/episodes parquet files found
```

The reported failure has nothing to do with the invalid FPS. If discovery succeeds, the value is converted with `int(source_archive["fps"])` at [`src/hflow/importers/lerobot.py:584-594`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L584-L594), where non-finite values can fail again with a raw conversion error.

## Pattern to copy

`TransformConfig` validates a positive duration with `math.isfinite` at [`src/hflow/transform.py:144-150`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/transform.py#L144-L150). [`tests/test_transform.py:54-63`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_transform.py#L54-L63) covers both `NaN` and infinity.

The existing LeRobot synthetic metadata and discovery test live in [`tests/test_lerobot_converter.py:32-215`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_lerobot_converter.py#L32-L215).

## What to build

Require `fps` from `meta/info.json` to be finite as well as numeric and positive. Keep the existing contextual error and update it to describe the complete rule.

Add metadata-boundary regressions for `NaN`, positive infinity, and negative infinity. Assert the error names `fps` and occurs before episode files are downloaded or any landing output is written.

## Definition of done

1. `NaN`, positive infinity, and negative infinity are refused as invalid LeRobot FPS metadata.
2. Boolean, missing, nonnumeric, zero, and negative FPS values keep their current refusal.
3. A normal positive FPS such as 30 remains accepted.
4. The error names the bad value and says FPS must be finite and positive.
5. No episode metadata, source video, Parquet, or canonical output is written after the invalid info document is read.
7. No new dependency or converter-version bump is needed.

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

rakesh0x:
## Summary

Fixes #299.

The LeRobot metadata boundary in `_ensure_source_archive` validated that `fps` from `meta/info.json` is numeric and positive, but not finite — so `float("nan")` and `float("inf")` passed validation and only failed later with misleading errors (`no meta/episodes parquet files found`, or a raw `int()` conversion failure in episode conversion).

This change mirrors the existing `TransformConfig.gop_seconds` pattern (`math.isfinite`) at the metadata boundary:

- `fps` must now be finite as well as numeric and positive.
- The contextual `ValueError` names the offending value and states the complete rule: `FPS must be finite and positive`.
- The refusal happens before episode metadata discovery, so no episode metadata, source video, Parquet, or canonical output is written after an invalid info document is read.

## Tests

Parametrized metadata-boundary regressions in `tests/test_lerobot_converter.py`:

- `NaN`, `+inf`, and `-inf` are refused as invalid FPS (new behavior).
- Boolean, missing, nonnumeric, zero, and negative FPS keep their existing refusal.
- A normal positive FPS (`30`) is still accepted through discovery.
- Each refusal case asserts the error names the bad value, mentions "finite and positive", and that `_hf_tree` episode discovery never runs and no `meta/episodes` output is created.
- The regressions were verified to fail against unpatched main and pass with the fix.

No new dependency and no converter-version bump.

## Validation

- `uv run ruff check --fix` — clean
- `uv run ruff format` — clean
- `uv run ty check` — clean
- `uv run pytest -q tests/test_lerobot_converter.py` — 18 passed
- Full suite from repo root (`tests` + `packages/hflow-server/tests` per `testpaths`): 1255 passed, 11 skipped, 0 failed

### DIFF
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -16,6 +16,7 @@ Hugging Face Hub, and uses HFlow's managed FFmpeg build.
 
 import json
 import logging
+import math
 import os
 import struct
 import subprocess
@@ -313,11 +314,12 @@ def _ensure_source_archive(dataset_source: DatasetSource, cache_dir: Path) -> _S
     if (
         isinstance(frames_per_second, bool)
         or not isinstance(frames_per_second, int | float)
+        or not math.isfinite(frames_per_second)
         or frames_per_second <= 0
     ):
         raise ValueError(
             f"LeRobot meta/info.json has invalid fps={frames_per_second!r}; "
-            "expected a positive number"
+            "FPS must be finite and positive"
         )
     data_path_template = dataset_information.get("data_path")
     if not isinstance(data_path_template, str) or not data_path_template.strip():
--- a/tests/test_lerobot_converter.py
+++ b/tests/test_lerobot_converter.py
@@ -395,3 +395,94 @@ def test_converter_output_remuxes_without_tail_loss(
         check=True,
     )
     assert "nb_read_frames=90" in probe.stdout, probe.stdout
+
+
+@pytest.mark.parametrize(
+    "bad_fps",
+    [
+        float("nan"),
+        float("inf"),
+        float("-inf"),
+        True,
+        None,
+        "30",
+        0,
+        -30,
+    ],
+    ids=[
+        "nan",
+        "positive-infinity",
+        "negative-infinity",
+        "boolean",
+        "missing",
+        "nonnumeric",
+        "zero",
+        "negative",
+    ],
+)
+def test_info_json_refuses_non_finite_or_non_positive_fps(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch, bad_fps: object
+) -> None:
+    """Invalid fps metadata must be refused before any episode discovery runs."""
+    corpus = _build_fake_corpus(tmp_path)
+    info = dict(corpus["info"])
+    if bad_fps is None:
+        info.pop("fps")
+    else:
+        info["fps"] = bad_fps
+    expected_value = info.get("fps")
+    monkeypatch.setattr(
+        prep, "_hf_repo_info", lambda repo, rev: {"sha": rev, "license": "apache-2.0"}
+    )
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Validated on your branch: `ruff check`, `ruff format --check`, `ty check` clean, 1260 passed / 6 skipped.

Also checked the tests can detect the fix being removed. Dropped the `math.isfinite` clause and the `nan` and `positive-infinity` cases failed, the other six still passed. That's the right split: `-inf` was already caught by `<= 0`, so `nan` and `+inf` are the actual new coverage and your parametrization makes that visible instead of hiding it behind one green test.

"finite and positive" matches what `_camera_motion.py:59` and `_frame_statistics.py:98` already say, so the message reads like the rest of the repo.

One thing, and it's an invitation not a rule: the `good first issue` pool is where newcomers land, and you're three PRs in now, so it's worth leaving those for people who need somewhere to start. #320 (parse LeRobot metadata into a typed archive model) is the natural next step from this one, and it's where the `_SourceArchive` TypedDict you just validated a field of gets turned into something that can't hold a bad fps in the first place.

Or the higher-leverage version: you already filed #287 off Egocentric-10K, and that report was more useful than any of these fixes. More of that, pointing HFlow at real corpora and telling us what breaks, is worth more to us than the starter queue.

Issue #315, opened 2026-08-31
Parse Airflow DAG runs and task instances into typed client responses


kstonekuan:
## Current behavior

`AirflowClient` returns raw dictionaries for triggered runs, one run, run lists, and task-instance lists at [`src/hflow/runtime/_client.py:251-307`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/runtime/_client.py#L251-L307).

The workspace server parses those dictionaries again:

- Run summaries at [`packages/hflow-server/src/hflow_server/_runtime.py:282-306`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/packages/hflow-server/src/hflow_server/_runtime.py#L282-L306)
- Task instances at [`packages/hflow-server/src/hflow_server/_graph.py:246-313`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/packages/hflow-server/src/hflow_server/_graph.py#L246-L313)
- Stage-run matching at [`packages/hflow-server/src/hflow_server/_graph.py:348-426`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/packages/hflow-server/src/hflow_server/_graph.py#L348-L426)

`dag_runs()` and `task_instances()` verify only that the container is a list. A non-object member reaches server code typed as `Any` and fails on `.get()`, outside the client error boundary.

Verified against `main` at `fbfd553`.

## Pattern to copy

The same client already parses health responses into `AirflowHealth` at [`src/hflow/runtime/_client.py:63-77`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/runtime/_client.py#L63-L77) and returns that type from [`health()`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/runtime/_client.py#L207-L215). Authentication is also an explicit `PasswordCredentials | BearerToken` union.

## What to build

[...]

Issue #312, opened 2026-08-31
Preserve non-404 Airflow failures on stage run surfaces


kstonekuan:
## Current behavior

The server intentionally lets an unregistered stage DAG or an expired stage run render as an empty lane. Three handlers currently catch every AirflowClientError:

- stage task instances in the run graph at [packages/hflow-server/src/hflow_server/_graph.py:451-461](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/packages/hflow-server/src/hflow_server/_graph.py#L451-L461)
- stage run discovery in the run graph at [packages/hflow-server/src/hflow_server/_graph.py:519-529](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/packages/hflow-server/src/hflow_server/_graph.py#L519-L529)
- recent stage runs at [packages/hflow-server/src/hflow_server/_runtime.py:391-402](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/packages/hflow-server/src/hflow_server/_runtime.py#L391-L402)

AirflowClientError also carries authentication failures, rate limits, and upstream 5xx responses. Those failures are converted to empty tasks, an unmatched stage, or an empty recent-run strip. The same routes already map failures from the master DAG to the shared browser-safe 502 response at [packages/hflow-server/src/hflow_server/_graph.py:501-516](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/packages/hflow-server/src/hflow_server/_graph.py#L501-L516) and [packages/hflow-server/src/hflow_server/_runtime.py:384-390](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/packages/hflow-server/src/hflow_server/_runtime.py#L384-L390).

[...]

Issue #311, opened 2026-08-31
Paginate Airflow task instances before building run graphs


kstonekuan:
## Current behavior

[AirflowClient.task_instances()](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/runtime/_client.py#L295-L307) says it returns every task instance for one run, including every dynamically mapped process_batch instance. It makes one GET request with no limit, offset, or cursor.

Airflow's Get Task Instances endpoint is paginated. The generated API client documents limit=50 and offset=0 as its defaults and returns total_entries: https://github.com/apache/airflow-client-python/blob/main/docs/TaskInstanceApi.md#get_task_instances

The workspace run graph consumes that one returned list and calculates mapped totals and per-state counts from it at [packages/hflow-server/src/hflow_server/_graph.py:513-540](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/packages/hflow-server/src/hflow_server/_graph.py#L513-L540). A fan-out larger than the API page reports only the first page, so mapped_summary.total and by_state omit later batches.

The neighboring dag_runs() API makes its truncation explicit in both its signature and docstring at [src/hflow/runtime/_client.py:281-293](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/runtime/_client.py#L281-L293). task_instances() has no equivalent caller-selected cap because its contract is complete run detail.

Verified against main at fbfd553.

## Definition of done

1. AirflowClient.task_instances() returns all task instances for the selected run across every API page.
2. Dynamically mapped instances retain every map_index exactly once.
3. A failure on any later page raises AirflowClientError. The client doesn't return a partial list as complete.
4. Pagination terminates safely from the documented response metadata and can't loop forever on an inconsistent response.
5. Existing run-id URL encoding and token refresh behavior remain unchanged on every page.
6. The workspace run graph reports complete mapped totals and state counts for a fan-out larger than one Airflow page.
7. Add a local HTTP-boundary regression with more than one page. Assert the returned business result and requested pagination values, not Airflow internals.
8. Keep dag_runs() caller-capped behavior unchanged.
9. Update the task_instances() docstring with the pagination contract.

## Non-goals

- A generic pagination framework for every Airflow endpoint
- Removing the deliberate limits on recent DAG-run lists
[...]

PR #322, opened 2026-08-31
fix: wrap Airflow response decode errors

### LINKED ISSUE #295: Airflow client leaks decoder exceptions for malformed successful responses
## Current behavior

`AirflowClient._http_json` converts HTTP, URL, socket, and HTTP protocol failures into `AirflowClientError` at [`src/hflow/runtime/_client.py:136-157`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/runtime/_client.py#L136-L157). It also reports a valid JSON value with the wrong shape as `AirflowClientError` at [`src/hflow/runtime/_client.py:158-163`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/runtime/_client.py#L158-L163).

Two response-decoding failures escape that boundary:

1. `response.read().decode()` can raise a raw `UnicodeDecodeError` for invalid UTF-8.
2. `json.loads(response_text)` can raise a raw `JSONDecodeError` for malformed JSON returned with an HTTP 2xx status.

Measured on `main` at `1f5c7c8` with a successful `data:application/json,not-json` response:

```text
JSONDecodeError Expecting value: line 1 column 1 (char 0)
```

Callers such as `hflow status` expect request failures to arrive through the typed client error so they can report the endpoint failure without an internal traceback.

## Pattern to copy

The HTTP error branch at [`src/hflow/runtime/_client.py:139-149`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/runtime/_client.py#L139-L149) includes the method, URL, status, and a bounded one-line body excerpt.

[`tests/test_runtime_client.py:23-137`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_runtime_client.py#L23-L137) owns a local stub Airflow server. Extend that handler with a small raw-response path so the regression exercises the public client over HTTP without Docker or external network access.

## What to build

Convert successful-response UTF-8 and JSON decoding failures into `AirflowClientError`. Include the HTTP method and URL in the message plus a short, bounded description or body excerpt that helps identify an HTML proxy response or malformed API payload.

Add outcome-focused tests for malformed JSON and invalid UTF-8 returned with HTTP 200. Assert the public client error type and useful message. Do not assert `json` or `urllib` implementation details.

## Definition of done

1. A 2xx response containing malformed JSON raises `AirflowClientError`, not `JSONDecodeError`.
2. A 2xx response containing invalid UTF-8 raises `AirflowClientError`, not `UnicodeDecodeError`.
3. The error names the HTTP method and URL and keeps any response excerpt bounded.
4. Empty successful bodies still return `{}`.
5. Valid JSON objects and the existing non-object JSON refusal are unchanged.
6. Existing HTTP status, connection, and bearer-token behavior is unchanged.
7. No new runtime dependency is added.

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_runtime_client.py
uv run pytest -q
HFLOW_DOCKER_TESTS=1 uv run pytest tests/test_runtime_integration.py -q
```

Run the full suite from the repo root. `pytest tests` alone misses `packages/hflow-server/tests`. The Docker integration command is part of the documented gate for REST client changes.

Iamnotphage:
## Summary

`AirflowClient` currently converts HTTP and connection failures into
`AirflowClientError`, but successful HTTP responses can still leak raw
`UnicodeDecodeError` or `JSONDecodeError` exceptions when their bodies contain
invalid UTF-8 or malformed JSON.

This change keeps those failures inside the client's typed error boundary:

- convert invalid UTF-8 in HTTP 2xx responses into `AirflowClientError`
- convert malformed JSON in HTTP 2xx responses into `AirflowClientError`
- include the HTTP method, URL, and a bounded one-line response excerpt
- preserve the existing behavior for empty bodies, valid JSON objects,
  non-object JSON, HTTP status errors, connection failures, and bearer tokens

This prevents callers such as `hflow status` from exposing internal decoder
tracebacks when Airflow or an intermediate proxy returns a malformed response.

## Regression coverage

The local Airflow stub server can now return arbitrary successful response
bytes. The new outcome-focused tests exercise the public client over HTTP and
verify that:

- malformed JSON returned with HTTP 200 raises `AirflowClientError`
- invalid UTF-8 returned with HTTP 200 raises `AirflowClientError`
- error messages contain the HTTP method, URL, and a bounded response excerpt
- empty successful responses still behave as an empty JSON object
- the existing non-object JSON refusal remains unchanged

## Validation

- `UV_OFFLINE=1 uv run --locked ruff check --fix`
  - passed
- `UV_OFFLINE=1 uv run --locked ruff format`
  - passed
- `UV_OFFLINE=1 uv run --locked ty check`
  - passed
- `UV_OFFLINE=1 uv run --locked pytest -q tests/test_runtime_client.py`
  - 24 passed
- `UV_OFFLINE=1 uv run --locked pytest -q`
  - 1255 passed, 6 skipped

The Docker integration gate was not run because Docker is unavailable in the
current environment:

`HFLOW_DOCKER_TESTS=1 uv run pytest tests/test_runtime_integration.py -q`

## Documentation and versioning

No documentation update is needed because this normalizes the existing client
error boundary without introducing a new API, flag, format, or operational
requirement.

`TRANSFORM_BEHAVIOR_VERSION` is unchanged because this change does not alter
episode processing or canonical output bytes.

No runtime dependency was added.

Closes #295

### DIFF
--- a/src/hflow/runtime/_client.py
+++ b/src/hflow/runtime/_client.py
@@ -60,6 +60,11 @@ class BearerToken:
 AirflowAuth = PasswordCredentials | BearerToken
 
 
+def _body_excerpt(body: str) -> str:
+    """Collapse a response body to the bounded excerpt exposed to callers."""
+    return " ".join(body.split())[:200]
+
+
 @dataclass(frozen=True)
 class AirflowHealth:
     """Parsed ``/api/v2/monitor/health`` body."""
@@ -135,12 +140,12 @@ class AirflowClient:
             request.add_header("Authorization", f"Bearer {bearer_token}")
         try:
             with urllib.request.urlopen(request, timeout=self._request_timeout_s) as response:
-                response_text = response.read().decode()
+                response_body = response.read()
         except urllib.error.HTTPError as error:
             error_body = error.read().decode(errors="replace")
             # Airflow puts the useful part ("detail") in the body; surface a
             # truncated copy so failures are diagnosable from the message.
-            body_excerpt = " ".join(error_body.split())[:200]
+            body_excerpt = _body_excerpt(error_body)
             raise AirflowClientError(
                 f"{method} {url} failed with HTTP {error.code}"
                 + (f": {body_excerpt}" if body_excerpt else ""),
@@ -155,9 +160,27 @@ class AirflowClient:
             # published port before the api-server listens) must also become
             # AirflowClientError, or wait_until_healthy cannot retry them.
             raise AirflowClientError(f"{method} {url} failed: {error!r}") from error
+        try:
+            response_text = response_body.decode("utf-8")
+        except UnicodeDecodeError as error:
+            response_text = response_body.decode("utf-8", errors="replace")
+            body_excerpt = _body_excerpt(response_text)
+            raise AirflowClientError(
+                f"{method} {url} returned invalid UTF-8"
+                + (f": {body_excerpt}" if body_excerpt else ""),
+                body=response_text,
+            ) from error
         if not response_text:
             return {}
-        parsed = json.loads(response_text)
+        try:
+            parsed = json.loads(response_text)
+        except json.JSONDecodeError as error:
+            body_excerpt = _body_excerpt(response_text)
+            raise AirflowClientError(
+                f"{method} {url} returned malformed JSON"
+                + (f": {body_excerpt}" if body_excerpt else ""),
+                body=response_text,
+            ) from error
         if not isinstance(parsed, dict):
             raise AirflowClientError(f"{method} {url} returned non-object JSON: {parsed!r}")
         return parsed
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Validated on your branch: `ruff check`, `ruff format --check`, `ty check` clean, 1255 passed / 6 skipped. Docker gate not needed here, this doesn't touch anything the integration test exercises.

Two things I liked. Pulling `_body_excerpt` out instead of adding a fourth copy of `" ".join(body.split())[:200]` was not asked for and is the right call. And bounding is actually asserted: `end-of-response` sits past 200 chars and you assert it's absent, so the test fails if the excerpt ever stops being bounded. Most tests of a truncation just check the prefix and pass forever.

Checked the tests can detect the fix being removed. Dropped both try/excepts and `test_malformed_success_response_raises_typed_client_error` and `test_invalid_utf8_success_response_raises_typed_client_error` both failed. Also grepped `src/hflow/runtime/` for `.decode(`/`json.loads` and `_client.py` is now fully wrapped, so the boundary is closed rather than closed at one site.

One invitation, not a rule: you're four PRs in, so the `good first issue` pool is best left for people who need a first landing spot. The same file has better work in it. #315 parses DAG runs and task instances into typed responses instead of handing callers a `dict[str, Any]`, which is the next layer up from the boundary you just fixed. #311 (paginate task instances) and #312 (preserve non-404 failures) are also open and in `runtime/`.

PR #324, opened 2026-08-31
fix: accept NumPy integer frame indices

### LINKED ISSUE #308: Let Episode.frames_at_indices accept NumPy integer indices
## Current behavior

[Episode.frames_at_indices()](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/episode.py#L612-L645) copies the requested indices, excludes Python booleans, then requires every value to satisfy isinstance(value, int).

NumPy integer scalars don't satisfy that check:

~~~python
>>> isinstance(np.int64(3), int)
False
~~~

A list or array of frame indices obtained from NumPy is rejected with frame indices must be integers, even though the values are integral and HFlow already uses NumPy arrays for channel data. The existing end-to-end coverage passes only Python integers at [tests/test_end_to_end.py:166-188](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/tests/test_end_to_end.py#L166-L188).

## Pattern to copy

Catalog timestamp normalization converts NumPy scalars to Python scalars before applying the integer and boolean rules at [src/hflow/catalog.py:413-458](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/catalog.py#L413-L458). Keep the same normalization order here so np.bool_ doesn't become an accepted integer.

## What to build

Normalize supported NumPy integer scalars before validating, sorting, hashing, and indexing the requested frames.

## Definition of done

1. Python int, np.int32, np.int64, and unsigned NumPy integer indices are accepted when their values are valid.
2. bool and np.bool_ remain refused.
3. Floating-point values, including integral-looking values such as 3.0, remain refused.
4. Negative, duplicate, descending, and out-of-range indices keep their current errors.
5. Normalized NumPy indices produce the same extracted paths and log timestamps as equivalent Python indices.
6. Extend the existing exact-frame behavioral test. Don't add tests for NumPy's own conversion behavior.
7. No cache-key or extracted-frame layout changes.
8. No new dependency or transform behavior-version bump is needed.

## Non-goals

- Accepting fractional indices
- Sorting or deduplicating caller input
- Changing ffmpeg selection behavior
- Changing the public return type

## Validation

~~~bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_end_to_end.py
uv run pytest -q
~~~

ravik453:
## Summary

* Normalize NumPy integer scalars before validating frame indices.
* Preserve existing boolean, ordering, bounds, and indexing behavior.
* Extend the exact-frame end-to-end test to cover signed and unsigned NumPy integer types.

Fixes #308.

## Validation

* `uv run ruff check --fix`
* `uv run ruff format`
* `uv run ty check`
* `uv run pytest -q tests/test_end_to_end.py::test_canonical_episode_extracts_exact_source_frame_indices`
* `uv run pytest -q tests/test_end_to_end.py` — 16 passed
* `uv run pytest -q` — 1251 passed, 6 skipped

### DIFF
--- a/src/hflow/episode.py
+++ b/src/hflow/episode.py
@@ -613,7 +613,7 @@ class Episode:
         self,
         camera: str | None = None,
         *,
-        frame_indices: Sequence[int],
+        frame_indices: Sequence[int | np.integer[Any]],
     ) -> list[ExtractedFrame]:
         """Extract JPEGs at exact, ascending source-message frame indices.
 
--- a/src/hflow/episode.py
+++ b/src/hflow/episode.py
@@ -623,7 +623,10 @@ class Episode:
         unambiguous one-to-one order.
         """
 
-        selected_frame_indices = list(frame_indices)
+        selected_frame_indices = [
+            frame_index.item() if isinstance(frame_index, np.generic) else frame_index
+            for frame_index in frame_indices
+        ]
         if any(isinstance(frame_index, bool) for frame_index in selected_frame_indices):
             raise ValueError("frame indices must be integers, not booleans")
         if any(not isinstance(frame_index, int) for frame_index in selected_frame_indices):
--- a/tests/test_end_to_end.py
+++ b/tests/test_end_to_end.py
@@ -2,7 +2,7 @@
 infrastructure. Mirrors the README design-target example."""
 
 from pathlib import Path
-from typing import Any
+from typing import Any, cast
 
 import numpy as np
 import pytest
@@ -177,7 +177,25 @@ def test_canonical_episode_extracts_exact_source_frame_indices(
             frame_indices=selected_frame_indices,
         )
 
+        numpy_frame_indices = [
+            cast(int, np.int32(0)),
+            cast(int, np.int64(1)),
+            cast(int, np.uint32(2)),
+            cast(int, np.uint64(7)),
+            cast(int, np.int64(8)),
+            cast(int, np.uint32(29)),
+        ]
+        numpy_frames = episode.frames_at_indices(
+            camera_topic,
+            frame_indices=numpy_frame_indices,
+        )
+
         assert [frame.log_time_ns for frame in selected_frames] == expected_log_times.tolist()
+        assert [frame.log_time_ns for frame in numpy_frames] == expected_log_times.tolist()
+        assert [frame.path.read_bytes() for frame in numpy_frames] == [
+            frame.path.read_bytes() for frame in selected_frames
+        ]
+        assert numpy_frames == selected_frames
         assert all(frame.path.read_bytes()[:2] == b"\xff\xd8" for frame in selected_frames)
         assert (
             episode.frames_at_indices(
--- a/tests/test_end_to_end.py
+++ b/tests/test_end_to_end.py
@@ -2,7 +2,7 @@
 infrastructure. Mirrors the README design-target example."""
 
 from pathlib import Path
-from typing import Any, cast
+from typing import Any
 
 import numpy as np
 import pytest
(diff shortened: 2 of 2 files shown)


kstonekuan (changes requested):
Thanks, and welcome. The normalization itself is right, including the part that's easy to get wrong: `.item()` before the bool check, so `np.bool_` collapses to a Python `bool` and gets caught instead of sneaking through as an integer. That was the reason the issue pointed at `catalog.py`.

Two things before merge.

**1. The signature still refuses what the code now accepts.** `frame_indices: Sequence[int]` is unchanged, so a caller doing the obvious thing gets a type error even though it works at runtime:

```
error[invalid-argument-type]: Argument to bound method `Episode.frames_at_indices` is incorrect
   |     episode.frames_at_indices("/cam", frame_indices=[np.int64(0), np.int64(1)])
   |                                       ^^^^ Expected `Sequence[int]`, found `list[signedinteger[_64Bit]]`
```

The six `cast(int, np.int32(0))` calls in your test are the same problem: the test has to lie to `ty` to exercise the feature, which means nobody can use it from typed code. Widen it to `Sequence[int | np.integer[Any]]` and the casts come out of the test on their own.

**2. Nothing tests the refusals, before your change or after.** I deleted both guards entirely:

```python
# if any(isinstance(frame_index, bool) ...): raise
# if any(not isinstance(frame_index, int) ...): raise
```

and the full suite passed, 1251 passed / 6 skipped. So DoD 2 and 3 aren't met, and more to the point the ordering you got right is unpinned: someone could move `.item()` after the bool check and no test would object. Add cases for `bool`, `np.bool_`, `3.0`, and `np.float64(3.0)`. `pytest.mark.parametrize` over the four with `pytest.raises(ValueError)` is enough, and it goes next to the accepting test you already extended.

The rest is fine. Ordering, bounds, dedup, and the path/timestamp equality assertions all hold, and `numpy_frames == selected_frames` is a good way to say "same answer" in one line.

Validation commands are in CONTRIBUTING.md; `uv run pytest -q` from the repo root, not `pytest tests`. Ping me when you push and I'll take another look.

ravik453:
Updated based on your feedback. The public type annotation now accepts NumPy integer scalars, the test casts are removed, and I added coverage for bool, np.bool_, 3.0, and `np.float64(3.0).

Validation is green: 1255 passed, 6 skipped.

Please take another look. Thanks!

kstonekuan (approved):
LGTM, merging. Both done.

The signature change is confirmed from the caller's side. `ty` on the three shapes a user would actually write, all clean now:

```python
episode.frames_at_indices("/cam", frame_indices=[np.int64(0), np.int64(1)])
episode.frames_at_indices("/cam", frame_indices=list(np.array([0, 1], dtype=np.int64)))
episode.frames_at_indices("/cam", frame_indices=[0, 1])
```

And the `cast(int, ...)` calls came out of your test on their own, which is the tell that the type now matches the behavior.

The refusal tests are load-bearing. Disabled both guards and all four cases failed:

```
4 failed, 16 deselected
```

Gate clean, 1255 passed / 6 skipped.

One thing for next time, not worth another push: `pytest.raises(ValueError)` with no `match=` passes on any `ValueError`, including one thrown for an unrelated reason further down. Here it happens to work because with the guards gone the bad values reach an `IndexError` instead, so the test still fails. But `match="not booleans"` versus `match="must be integers"` would pin which refusal fired, and it's the difference between testing that something broke and testing that the right thing refused. The rest of the repo mostly does this, `test_catalog_curation.py` being the closest example.

PR #325, opened 2026-08-31
fix(catalog-ui): reject invalid numeric settings

### LINKED ISSUE #297: Catalog UI settings accept invalid numeric values
## Current behavior

`CatalogUiSettings.__post_init__` checks only numeric ranges at [`src/hflow/catalog_ui.py:25-38`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/catalog_ui.py#L25-L38).

Measured on `main` at `1f5c7c8`:

```text
port=True                         ACCEPTED
catalog_poll_interval_seconds=nan ACCEPTED
catalog_poll_interval_seconds=inf ACCEPTED
```

`bool` subclasses `int`, so `True` passes the port range as port 1. Comparisons with `NaN` are false, so it bypasses the nonpositive interval check. Positive infinity also passes. The accepted interval is used by `Event.wait` at [`src/hflow/catalog_ui.py:132-148`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/catalog_ui.py#L132-L148); `Event().wait(float("inf"))` raises `OverflowError: timestamp out of range for platform time_t` on this host.

The existing tests cover only ports 0 and 65536 at [`tests/test_catalog_ui.py:106-109`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_catalog_ui.py#L106-L109).

## Pattern to copy

`RuntimeConfig` checks type before range and explicitly refuses `bool` at [`src/hflow/runtime/_bundle.py:222-246`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/runtime/_bundle.py#L222-L246). Its tests in [`tests/test_runtime_bundle.py:913-981`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_runtime_bundle.py#L913-L981) cover booleans, wrong numeric types, and legal boundary values.

Finite positive float validation also exists in [`src/hflow/transform.py:144-150`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/transform.py#L144-L150).

## What to build

Validate both numeric settings at construction:

- `port` must be an `int`, excluding `bool`, in 1 through 65535.
- `catalog_poll_interval_seconds` must be an `int` or `float`, excluding `bool`, finite, and greater than zero.

Extend `tests/test_catalog_ui.py`. Assert each invalid setting is refused before DuckDB, socket, browser, or wait-loop work begins.

## Definition of done

1. Boolean, floating-point, string, and out-of-range ports raise a contextual `ValueError`.
2. Boolean, nonnumeric, zero, negative, `NaN`, and infinite poll intervals raise a contextual `ValueError`.
3. Port values 1 and 65535 remain valid.
4. Finite positive integer and float intervals remain valid.
5. An `IntEnum` port remains valid, matching the runtime config convention.
6. Existing catalog UI behavior is unchanged for valid settings.

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_catalog_ui.py
uv run pytest -q
```

Ftgn-dpA:
## Summary

Reject invalid Catalog UI numeric settings at construction, before they can reach socket setup or the polling wait loop.

Closes #297

## Why

`CatalogUiSettings` previously relied only on range comparisons. Because `bool` subclasses `int`, `port=True` was accepted as port 1; floating-point ports were also accepted. For the polling interval, `NaN`, positive infinity, and `True` bypassed the existing non-positive check. Positive infinity can later raise `OverflowError` inside `Event.wait`.

This change keeps valid behavior unchanged while enforcing the field contracts described in the issue:

- `port` is an `int` other than `bool`, in the inclusive range 1-65535.
- `catalog_poll_interval_seconds` is an `int` or `float` other than `bool`, finite, and greater than zero.
- `IntEnum` port values remain accepted.

## Validation

- `python -m pytest -q tests/test_catalog_ui.py` — 20 passed
- `ruff check src/hflow/catalog_ui.py tests/test_catalog_ui.py` — passed
- `ruff format --check src/hflow/catalog_ui.py tests/test_catalog_ui.py` — 2 files already formatted
- `ty check src/hflow/catalog_ui.py tests/test_catalog_ui.py` — passed
- The new validation cases were run against unmodified `main` first and failed for the reported boundaries.

## Checklist

- [x] Documentation is unchanged because valid behavior, flags, formats, and requirements are unchanged.
- [x] I ran Ruff check/format and ty in the WSL development environment.
- [x] I preserved stored-data compatibility; this only validates settings at construction.

### DIFF
--- a/src/hflow/catalog_ui.py
+++ b/src/hflow/catalog_ui.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import math
 import sys
 import webbrowser
 from dataclasses import dataclass
@@ -32,10 +33,26 @@ class CatalogUiSettings:
     catalog_poll_interval_seconds: float = DEFAULT_CATALOG_POLL_INTERVAL_SECONDS
 
     def __post_init__(self) -> None:
+        if isinstance(self.port, bool) or not isinstance(self.port, int):
+            raise ValueError(f"port must be an int, got {type(self.port).__name__}")
         if not 1 <= self.port <= 65535:
             raise ValueError("port must be between 1 and 65535")
-        if self.catalog_poll_interval_seconds <= 0:
-            raise ValueError("catalog poll interval must be greater than zero")
+
+        if isinstance(self.catalog_poll_interval_seconds, bool) or not isinstance(
+            self.catalog_poll_interval_seconds, int | float
+        ):
+            raise ValueError(
+                "catalog_poll_interval_seconds must be an int or float, got "
+                f"{type(self.catalog_poll_interval_seconds).__name__}"
+            )
+        if (
+            not math.isfinite(self.catalog_poll_interval_seconds)
+            or self.catalog_poll_interval_seconds <= 0
+        ):
+            raise ValueError(
+                "catalog_poll_interval_seconds must be positive and finite, got "
+                f"{self.catalog_poll_interval_seconds}"
+            )
 
 
 def _raise_if_loopback_port_is_unavailable(port: int) -> None:
--- a/tests/test_catalog_ui.py
+++ b/tests/test_catalog_ui.py
@@ -1,5 +1,7 @@
 """The local DuckDB catalog browser lifecycle."""
 
+from dataclasses import replace
+from enum import IntEnum
 from pathlib import Path
 from socket import AF_INET, SOCK_STREAM, socket
 from threading import Event, Thread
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging. Welcome.

Validated on your branch: `ruff check`, `ruff format --check`, `ty check` clean, 1267 passed / 6 skipped.

Pushed one fixup, `4c494d0`: hoisted `replace` and `IntEnum` to module scope. Nothing wrong with it, just house style.

Testing the wrong-type cases through `replace()` was a good instinct. It keeps `ty` honest about the direct constructor while still exercising the guard, and since `replace()` re-runs `__post_init__`, the invariant now covers mutation too, not only construction.

Worth knowing where this actually earns its keep: the CLI already declares `--port` with `type=int` and has no flag for the poll interval, so argparse was covering the command line. The hole was SDK callers building `CatalogUiSettings` directly, which is why the boundary belongs on the dataclass.

PR #326, opened 2026-08-31
fix(lerobot): contextual ValueError on HF JSON decode failures (#301)

### LINKED ISSUE #301: LeRobot Hugging Face responses lose their source on JSON decode failures
## Current behavior

Three LeRobot HTTP boundaries decode response bytes directly:

- repository information at [`src/hflow/importers/lerobot.py:218-235`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L218-L235)
- tree listings at [`src/hflow/importers/lerobot.py:238-247`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L238-L247)
- `meta/info.json` at [`src/hflow/importers/lerobot.py:260-276`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L260-L276)

Each calls `response.read().decode()` and `json.loads(...)` without translating `UnicodeDecodeError` or `JSONDecodeError`. A proxy HTML response, truncated JSON body, or invalid UTF-8 payload reaches library callers as a raw decoder exception. The CLI catches these through the `ValueError` hierarchy, but prints a generic message such as `Expecting value: line 1 column 1` without identifying the repo endpoint, tree path, or metadata file.

Verified against `main` at `1f5c7c8`. Existing LeRobot tests replace these helpers with already-parsed values, so none exercise the response boundary.

## Pattern to copy

Snapshot marker parsing catches both decoding failures and adds file context at [`src/hflow/snapshot.py:502-519`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/snapshot.py#L502-L519).

The local HTTP stub pattern in [`tests/test_runtime_client.py:23-137`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_runtime_client.py#L23-L137) can return controlled response bytes without external network access. A smaller response test double at the `urlopen` boundary is also fine here. Assert what the caller sees, not `json` or `urllib` internals.

## What to build

Translate invalid UTF-8 and malformed JSON into contextual `ValueError` messages at all three Hugging Face JSON boundaries. Name the source being decoded:

- the requested repo and revision for repository information
- the requested repo tree path for tree responses
- `meta/info.json` plus the repo and revision for dataset metadata

Keep the original exception as the cause.

## Definition of done

1. Invalid UTF-8 and malformed JSON from each of the three response types raise contextual `ValueError`.
2. The message identifies the repo plus revision and distinguishes repository info, a tree path, and `meta/info.json`.
3. The CLI reports the contextual message with exit code 2 and no traceback.
4. Valid JSON objects and the existing wrong-shape refusals are unchanged.
5. HTTP status and download-stream behavior are unchanged.
6. Tests use local response bytes and make no external network request.

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

shoaibyazdani:
Fixes #301

## Problem

Three LeRobot Hugging Face HTTP boundaries decoded response bytes with a bare `json.loads(response.read().decode())`:

- repository info (`_hf_repo_info`)
- tree listings (`_hf_tree`)
- `meta/info.json` (`_fetch_info_json`)

Invalid UTF-8 or malformed JSON (proxy HTML, truncated body) leaked out as raw `UnicodeDecodeError`/`JSONDecodeError`. The CLI catches `ValueError` and prints exit code 2, but the message was generic (`Expecting value: line 1 column 1`) with no indication of which repo, tree path, or metadata file failed.

## Fix

Apply the existing snapshot-marker pattern (`src/hflow/snapshot.py:502-519`) at all three boundaries: catch `(UnicodeDecodeError, json.JSONDecodeError)` and re-raise a contextual `ValueError` naming the source, chaining the original exception as the cause:

- `Hugging Face repository response for {repo}@{revision} is not valid JSON: ...`
- `Hugging Face tree response for {repo}@{revision} path {path!r} is not valid JSON: ...`
- `LeRobot meta/info.json for {repo}@{revision} is not valid JSON: ...`

The CLI already reports these through `import lerobot: {error}` with exit code 2 and no traceback.

## Behavior preserved

- Valid JSON objects and the existing wrong-shape refusals are unchanged
- HTTP status and download-stream behavior untouched
- No new dependency

## Tests

10 new tests in `tests/test_lerobot_converter.py` using a local `urlopen` stub that returns controlled bytes — no external network. Each boundary is covered for both invalid UTF-8 and malformed JSON (asserting the contextual message and the chained cause), plus regression tests that valid JSON and wrong-shape refusals are unchanged.

## Validation

- `uv run ruff check --fix` ✅
- `uv run ruff format` ✅
- `uv run ty check` ✅
- `uv run pytest -q tests/test_lerobot_converter.py` ✅ (18 passed)
- `uv run pytest -q` ✅ (1259 passed, 6 skipped; one unrelated flaky `test_catalog_ui` timing test passed on re-run)


kstonekuan (approved):
LGTM, merging. Welcome.

Validated on your branch: `ruff check`, `ruff format --check`, `ty check` clean, 1260 passed / 6 skipped.

DoD 3 was the one item nothing in the suite covers, so I ran it through the CLI with a stubbed `urlopen` returning proxy HTML:

```
exit code: 2
stderr: import lerobot: Hugging Face repository response for lerobot/pusht@main is not valid JSON: Expecting value: line 1 column 1 (char 0)
traceback in stderr: False
```

Which is the whole point of the issue: that used to be `Expecting value: line 1 column 1` with nothing saying which repo or which of the three endpoints.

Asserting `excinfo.value.__cause__` is the right call. It pins the chaining rather than just the message, so the `from error` can't be dropped silently.

One note, no action needed. The message says "is not valid JSON" even when the cause is a `UnicodeDecodeError`, which is a different failure. #322 landed this morning and splits them ("returned invalid UTF-8" vs "returned malformed JSON") on the Airflow client. Not worth a round trip here since your chained cause already distinguishes them, but if the two boundaries ever get read side by side, that's the wording to converge on.

Issue #316, opened 2026-08-31
Model Build AI evaluation artifacts at the JSON boundary


kstonekuan:
## Current behavior

The Build AI evaluation adapter creates `run.json` as a `dict[str, object]` at [`examples/build_ai_evaluation/evaluate.py:437-473`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/build_ai_evaluation/evaluate.py#L437-L473). A saved file is loaded with `json.loads()` and immediately used through `.get()` and indexed keys at [`evaluate.py:482-494`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/build_ai_evaluation/evaluate.py#L482-L494).

Per-sample outcomes use one dictionary shape controlled by a free-form status string at [`evaluate.py:641-672`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/build_ai_evaluation/evaluate.py#L641-L672):

- `ok` requires `predicted_value` and no error.
- `invalid` requires a parse error and no prediction.
- `error` requires an execution error and no prediction.

The summarizer reconstructs those invariants through string comparisons, key lookups, and casts at [`evaluate.py:543-633`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/build_ai_evaluation/evaluate.py#L543-L633). A valid JSON value with the wrong shape can currently escape as `AttributeError`, `KeyError`, or a misleading summary.

Verified against `main` at `fbfd553`.

## Pattern to copy

The direct model boundary in [`examples/build_ai_evaluation/judgment.py:43-69`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/build_ai_evaluation/judgment.py#L43-L69) already uses explicit parsed and unparsed variants. [`model_output_check_result()`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/build_ai_evaluation/judgment.py#L237-L292) consumes them with an exhaustive match.

## What to build

[...]

PR #327, opened 2026-08-31
refactor(egosuite): make judgment outcomes exclusive

### LINKED ISSUE #318: Make EgoSuite hand-count judgment outcomes exclusive
## Current behavior

The EgoSuite example represents a model judgment with two independent optional fields at [`examples/egosuite_evaluation/judgment.py:33-39`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/egosuite_evaluation/judgment.py#L33-L39):

```python
predicted_hand_count: int | None
parse_error: str | None
```

That type permits a prediction and a parse error together, and it also permits neither. `evaluate_image_with_model()` follows the intended convention when it constructs values at [`judgment.py:129-189`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/egosuite_evaluation/judgment.py#L129-L189), but the pipeline has to recover the state repeatedly through `is not None` checks at [`pipeline.py:205-312`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/egosuite_evaluation/pipeline.py#L205-L312).

## Pattern to copy

The neighboring Build AI example already models this exact boundary as `ParsedVisionModelOutcome | UnparsedVisionModelOutcome` at [`examples/build_ai_evaluation/judgment.py:43-69`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/build_ai_evaluation/judgment.py#L43-L69). Its HFlow adapter uses an exhaustive match at [`judgment.py:237-292`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/build_ai_evaluation/judgment.py#L237-L292).

## What to build

Replace `HandCountJudgment` with parsed and unparsed variants sharing one response-metadata value. Make `evaluate_image_with_model()` return the union, then update `hand_visibility_check_result()` to match on the variants.

Keep the HFlow evidence contract unchanged: the same measurements, observations, tags, and intervals must be emitted for the same three frames.

## Definition of done

1. A parsed outcome always carries `predicted_hand_count` and cannot carry a parse error.
2. An unparsed outcome always carries `parse_error` and cannot carry a prediction.
3. Response model and usage metadata remain available on both variants.
4. The pipeline handles the union with an exhaustive match.
5. Missing completion content remains an unparsed outcome with the current error text.
6. Aggregate counts, agreement, observations, intervals, and tags in the existing pipeline test stay unchanged.
7. Tests construct both variants directly and assert their observable HFlow results.

## Non-goals

- Sharing implementation code between the EgoSuite and Build AI workspace projects
- Changing prompts, response parsing, or endpoint requests
- Changing the persisted evaluation summary format
- Adding a validation framework or dependency

## Validation

```bash
uv sync --locked --project examples/egosuite_evaluation
  examples/egosuite_evaluation
```

xianjianlf2:
## Summary

- replace the optional-field `HandCountJudgment` with parsed and unparsed outcome variants
- keep response model and usage data in shared `ModelResponseMetadata`
- handle every outcome exhaustively while preserving the existing HFlow measurements, observations, intervals, and tags

Closes #318

## Why

The previous shape allowed a prediction and parse error together, or neither. Typed variants make those invalid states unrepresentable and let the pipeline prove that every outcome is handled, following the neighboring Build AI example.

This changes check/example code only. It does not change canonical episode bytes, so no transform behavior-version bump is needed.

## Validation

- `uv run --locked --project examples/egosuite_evaluation pytest -q examples/egosuite_evaluation/tests` — 15 passed
- `uv run --locked --project examples/egosuite_evaluation ruff check examples/egosuite_evaluation` — passed
- `uv run --locked --project examples/egosuite_evaluation ruff format --check examples/egosuite_evaluation` — passed
- `uv run ruff check --fix && uv run ruff format` — passed
- `uv run ty check` — passed
- `uv run pytest -q` — 1245 passed, 11 skipped; one unrelated catalog UI thread-shutdown timeout failed
- `uv run pytest -q tests/test_catalog_ui.py::test_catalog_ui_starts_empty_then_exposes_the_first_completed_append` — passed on immediate rerun

## Checklist

- [x] Documentation is unchanged because the emitted evidence contract and user workflow are unchanged.
- [x] Stored-data compatibility is preserved; this check-only refactor does not change canonical output bytes.

### DIFF
--- a/examples/egosuite_evaluation/judgment.py
+++ b/examples/egosuite_evaluation/judgment.py
@@ -31,12 +31,32 @@ class ResponseFormat(StrEnum):
 
 
 @dataclass(frozen=True)
-class HandCountJudgment:
-    raw_response: str
-    predicted_hand_count: int | None
+class ModelResponseMetadata:
+    """Provider response fields retained for every hand-count outcome."""
+
     response_model: str | None
     usage: dict[str, object]
-    parse_error: str | None = None
+
+
+@dataclass(frozen=True)
+class ParsedHandCountOutcome:
+    """A model response successfully parsed as a supported hand count."""
+
+    raw_response: str
+    response_metadata: ModelResponseMetadata
+    predicted_hand_count: int
+
+
+@dataclass(frozen=True)
+class UnparsedHandCountOutcome:
+    """A completed model response outside the supported hand-count vocabulary."""
+
+    raw_response: str
+    response_metadata: ModelResponseMetadata
+    parse_error: str
+
+
+HandCountOutcome = ParsedHandCountOutcome | UnparsedHandCountOutcome
 
 
 def image_file_data_url(image_path: Path) -> str:
@@ -118,12 +138,19 @@ def _chat_completion_response_text(response: object) -> str:
     raise ValueError("endpoint returned no text completion content")
 
 
-def _response_usage(response: object) -> dict[str, object]:
+def _response_metadata(response: object) -> ModelResponseMetadata:
+    response_model = getattr(response, "model", None)
+    normalized_response_model = response_model if isinstance(response_model, str) else None
     usage = getattr(response, "usage", None)
-    if usage is None or not callable(getattr(usage, "model_dump", None)):
-        return {}
-    dumped_usage = usage.model_dump(exclude_none=True)
-    return dumped_usage if isinstance(dumped_usage, dict) else {}
+    parsed_usage: dict[str, object] = {}
+    if usage is not None and callable(getattr(usage, "model_dump", None)):
+        dumped_usage = usage.model_dump(exclude_none=True)
+        if isinstance(dumped_usage, dict):
+            parsed_usage = dumped_usage
+    return ModelResponseMetadata(
+        response_model=normalized_response_model,
+        usage=parsed_usage,
+    )
 
 
 def evaluate_image_with_model(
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
LGTM, merging.

Validated on your branch. Example project: `ruff check`, `ruff format --check`, `ty check` clean, 15 passed. Root: clean, 1251 passed / 6 skipped.

I checked the exhaustiveness is load-bearing rather than decorative, since `assert_never` only earns its place if the checker actually enforces it. Added a third variant to `HandCountOutcome`:

```
error[type-assertion-failure]: Argument does not have asserted type `Never`
   --> pipeline.py:231:17  |  assert_never(unexpected_judgment)
error[type-assertion-failure]: Argument does not have asserted type `Never`
   --> pipeline.py:317:17  |  assert_never(unexpected_judgment)
```

Both sites, so a future variant can't be silently dropped at either.

Folding the interval loop into the observation loop was the right call and I checked it preserves order and content: same `zip`, parsed-and-equal still appends nothing, unparsed always appends. And `agreement` for an unparsed outcome is now an explicit `False` where it used to fall out of `None == int`, which is the same value arrived at on purpose instead of by accident. The pipeline test's assertions are untouched in the diff and still pass, which is the evidence for DoD 6.

`ModelResponseMetadata` as a shared value on both variants is nicer than the issue asked for. It means the usage-totals and response-models loops don't need to know which variant they have.

One invitation, not a rule: this is your second PR and the `good first issue` pool is best kept for people who need a first landing spot. #316 is the natural next one for you, unassigned and `advanced`: it's the same JSON-boundary modelling on the Build AI example, the one you just mirrored the outcome union from. Say the word and I'll assign it.

PR #332, opened 2026-08-31
fix(catalog): reject timestamps outside BIGINT range

### LINKED ISSUE #309: Reject catalog evidence timestamps outside DuckDB BIGINT range
## Current behavior

The catalog stores observation timestamps and interval bounds as signed BIGINT columns at [src/hflow/catalog.py:82-95](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/catalog.py#L82-L95).

[_normalized_intervals()](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/catalog.py#L413-L458) and [_normalized_observations()](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/catalog.py#L461-L525) normalize NumPy scalars, check the integer type, and reject negative values. They don't enforce the signed 64-bit upper bound.

Measured on main at fbfd553:

~~~text
_normalized_observations(... timestamp_ns=9223372036854775808) -> accepted
_normalized_intervals(... start_ns=9223372036854775808) -> accepted
DuckDB INSERT -> ConversionException: value is out of range for INT64
~~~

The catalog append reaches a backend conversion error after HFlow's evidence boundary has accepted the value. The error doesn't name the check, observation, interval, or timestamp field.

## Pattern to copy

[_normalized_measurements()](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/catalog.py#L340-L410) rejects values the storage/query contract can't represent before any table write and includes the owning check and key in the error. The duplicate-observation regression also proves a boundary refusal leaves no committed episode file at [tests/test_catalog_curation.py:199-222](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/tests/test_catalog_curation.py#L199-L222).

## What to build

Require every stored nanosecond timestamp to fit the catalog's nonnegative signed BIGINT domain: 0 through 2**63 - 1.

## Definition of done

1. Observation timestamp_ns, interval start_ns, and interval end_ns accept 0 and 2**63 - 1.
2. Each field refuses 2**63 and larger values with a contextual ValueError.
3. Python and NumPy integer scalars follow the same range rule after normalization.
4. Existing boolean, float, negative, and inverted-interval refusals stay unchanged.
5. An out-of-range value is rejected before any episode or dependent Parquet file is committed.
6. The maximum accepted value round-trips through observations_latest or intervals.
7. Add outcome-focused coverage to the existing catalog validation tests.
8. No catalog format change or new dependency is needed.

## Non-goals

- Changing the catalog timestamp column type
- Supporting negative relative timestamps
- Adding range rules to unrelated numeric measurements

## Validation

~~~bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_catalog_curation.py
uv run pytest -q
~~~

askarshabdanov1-collab:
## Summary

Reject observation timestamps and interval bounds above DuckDB's signed 64-bit `BIGINT` maximum before a catalog append writes any Parquet files.

Closes #309.

## Why

Catalog evidence timestamps are normalized before the run fingerprint is computed, but values above `2**63 - 1` previously remained valid Python integers. They then failed later at the DuckDB insert boundary with a context-free `ConversionException`; for an oversized interval start, the existing inverted-interval check could mask the storage constraint instead.

The validation now runs immediately after Python/NumPy integer normalization and reports the check, evidence identifier, offending field/value, and `BIGINT` constraint. Existing handling of booleans, floats, negative values, and inverted intervals is unchanged.

## Validation

- Before the fix: the focused regression selection produced 6 failures; oversized Python and NumPy integers reached DuckDB (or the inverted-interval check).
- After the fix: `uv run pytest -q tests/test_catalog_curation.py` — 85 passed.
- After rebasing onto current `main`: `uv run pytest -q` — 1294 passed, 6 skipped.
- `uv run ruff check --fix` — passed.
- `uv run ruff format` — passed.
- `uv run ty check` — passed.
- `git diff --check` — passed.

## Checklist

- [x] Documentation is not required: this enforces the existing catalog `BIGINT` schema and does not change flags, formats, or requirements.
- [x] I preserved stored-data compatibility; no format version change is required.

### DIFF
--- a/src/hflow/catalog.py
+++ b/src/hflow/catalog.py
@@ -111,6 +111,11 @@ _reconciled_append_stems: set[tuple[str, str]] = set()
 
 _FORMAT_MARKER_NAME = "format_version"
 
+# DuckDB BIGINT columns store signed 64-bit integers. Validate timestamp
+# evidence before any catalog files are committed so callers get contextual
+# errors instead of a late database conversion failure.
+_MAX_CATALOG_TIMESTAMP_NS = 2**63 - 1
+
 # The quarantine state rendered as a queryable column by the curation
 # views (curation.py builds both episodes views with it). Declared here
 # because measurement keys are validated against the episodes view's full
@@ -437,6 +442,11 @@ def _normalized_intervals(check_name: str, intervals: list[Interval]) -> list[In
                     f"check {check_name!r} set interval {bound_name} as "
                     f"{type(value).__name__}: interval bounds are int nanoseconds"
                 )
+            if value > _MAX_CATALOG_TIMESTAMP_NS:
+                raise ValueError(
+                    f"check {check_name!r} interval {interval.label!r} set "
+                    f"{bound_name}={value}: interval bounds must fit a DuckDB BIGINT"
+                )
             bounds[bound_name] = value
         start_ns = bounds["start_ns"]
         end_ns = bounds["end_ns"]
@@ -498,6 +508,11 @@ def _normalized_observations(check_name: str, observations: list[Observation]) -
             raise ValueError(
                 f"check {check_name!r} observation {observation_id!r} set timestamp_ns "
                 f"as {type(timestamp_ns).__name__}: observation timestamps are int nanoseconds"
+            )
+        if timestamp_ns > _MAX_CATALOG_TIMESTAMP_NS:
+            raise ValueError(
+                f"check {check_name!r} observation {observation_id!r} set "
+                f"timestamp_ns={timestamp_ns}: timestamps must fit a DuckDB BIGINT"
             )
         if timestamp_ns < 0:
             raise ValueError(
--- a/tests/test_catalog_curation.py
+++ b/tests/test_catalog_curation.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import cast
 
 import duckdb
+import numpy as np
 import pytest
 
 import hflow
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging. Welcome.

Validated on your branch: `ruff check`, `ruff format --check`, `ty check` clean, 1294 passed / 6 skipped.

Checked the tests catch the guards being removed. Deleted both range checks and got exactly the failure the issue described, six of them:

```
src/hflow/catalog.py:701: ConversionException
6 failed, 79 passed
```

The part worth calling out is where you put the interval check. Inside the per-bound loop, before `start_ns`/`end_ns` are compared, so an oversized `start_ns` reports the BIGINT constraint instead of tripping the inverted-interval check and blaming the wrong thing. Your `[9223372036854775808-start_ns]` case is `start=2**63, end=10`, which is inverted, and it asserts the BIGINT message. That pins the masking fix specifically rather than by accident, and it wasn't spelled out in the issue.

Ordering in `_normalized_observations` is right too: numpy normalize, then type, then range, then negative. `np.uint64(2**63)` reaching the range check as a Python int is what makes DoD 3 real rather than assumed.

`rglob("*.parquet")` returning nothing is the right way to assert DoD 5. It fails if the refusal ever moves after a write, which a `pytest.raises` alone would not.

PR #337, opened 2026-08-31
fix(transform): classify mixed image formats as source-unsupported

### LINKED ISSUE #331: A camera topic mixing image formats is still filed as an infrastructure failure
### What happens

`_decode_camera_images` (`src/hflow/transform.py:509`) refuses a camera topic whose messages
carry more than one image format:

```python
if len(formats) != 1:
    raise ValueError(f"camera topic {topic!r} mixes image formats {sorted(formats)}")
return images, frame_ids, _input_codec_for_image_format(formats.pop(), topic)
```

The very next line calls `_input_codec_for_image_format`, which raises `SourceNotConforming`
after #288. So two refusals about the same channel's contents, one line apart, land in
different `ingest_failures.failure_kind` values:

```
camera topic '/cam' mixes image formats ['jpeg', 'png']  -> infrastructure
camera topic '/cam' carries unsupported format 'bogus'   -> source-unsupported
```

### Why it matters

`infrastructure` is the kind an operator retries. A recording whose camera topic mixes jpeg and
png fails identically on every retry, which is exactly the reasoning in #203 that produced
`source-unsupported` in the first place. #288 scoped itself to the four refusals #203 listed,
and this one was not on that list, so nothing here is a regression. It is the one place the new
vocabulary is inconsistent with itself.

### Definition of done

`_decode_camera_images` raises `SourceNotConforming` for the mixed-format case, and a test
triggers the real raise site (not a synthetic exception) and asserts
`classify_ingest_failure(error) == IngestFailureKind.SOURCE_UNSUPPORTED`. `tests/test_ingest_ledger.py`
has four of these to copy. The message text does not need to change.

Check whether the docstring list in `SourceNotConforming` (`transform.py:114`) and the paragraph
in `docs/how-to/write-a-converter.md` enumerate the refusal categories, and add this one if they
do.

### Validation

```
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

### Notes

Found while reviewing #288.

aadetya:
## Summary

- Classify compressed-image channels that mix formats as `source-unsupported`.
- Add a regression test using an MCAP camera channel containing both JPEG and PNG messages.
- Update the exhaustive refusal lists in the exception docstring and converter guide.

Closes #331

## Why

#203 established that transform failures caused by a recording’s contents should not be classified as `infrastructure`, since retrying the same recording cannot resolve them. #288 implemented that distinction by introducing `SourceNotConforming` and mapping it to `source-unsupported` for the four refusal cases in scope.

During review of #288, the mixed-image-format path was identified as the remaining inconsistency. It still raised a plain `ValueError`, causing `classify_ingest_failure` to treat it as `infrastructure`, even though the failure comes from the source recording.

This change brings that path in line with the classification introduced in #288. The existing error message remains unchanged.

Mixed-format inputs still fail before an output file is written, so no behavior version bump is needed.

## Validation

- `uv run ty check` — passed
- `uv run pytest -q tests/test_ingest_ledger.py` — 9 passed
- `uv run pytest -q` — 1315 passed, 11 skipped
- Markdown link check — 360 OK, 0 errors

## Checklist


kstonekuan (approved):
LGTM, merging. Welcome.

This closes the gap exactly as measured. When I filed #331 the two adjacent refusals disagreed:

```
mixes image formats       -> infrastructure
unsupported single format -> source-unsupported
```

On your branch:

```
mixes image formats       -> source-unsupported
unsupported single format -> source-unsupported
```

Validated: `ruff check`, `ruff format --check`, `ty check` clean, 1320 passed / 6 skipped. Reverted your raise back to a plain `ValueError` and `test_mixed_compressed_image_formats_classify_as_source_unsupported` failed, so the test holds the classification rather than just describing it.

Driving it through `write_canonical_episode` with a real two-message MCAP instead of constructing the exception directly is what the four tests next to it do, and it's the difference between pinning the classifier and pinning the raise site as well. Your comment on the payload bytes is right too: the format consistency check runs before anything decodes the image, which is why a placeholder works there.

You also updated both places that enumerate the refusal categories, the `SourceNotConforming` docstring and the converter guide. I checked `docs/CATALOG.md` and it lists kinds rather than categories, so there was correctly nothing to do there.

PR #333, opened 2026-08-31
refactor(egosuite): model evaluation artifacts at the JSON boundary

Sagar-024:
refactor(egosuite): model evaluation artifacts at the JSON boundary

Closes #317

## Summary

Models run metadata and per-sample outcomes as typed values instead of dicts.

- `RunMetadata` is parsed once when run.json loads. Task creation and summary generation read typed fields and never touch the JSON shape.
- Sample outcomes are exclusive variants: success carries `predicted_value`, invalid-response carries the parse error, execution error carries the Inspect error. A result that both predicts and errors, or does neither, is now unrepresentable.
- A non-object or malformed saved run.json raises `ValueError` naming the file and the bad field, instead of surfacing later as `AttributeError` or `KeyError` far from the cause.

run.json and summary.json schemas are unchanged, including compare support for existing summary files. Existing experiment directories with valid metadata still resume by fingerprint, and all summary numbers for the existing fixtures are unchanged.

## Scope

No reuse of build_ai judgment variants, no shared code across workspace projects, no changes to reference-label selection, prompts, or evaluation formulas.

## Validation

    uv sync --locked --project examples/egosuite_evaluation
    uv run --locked --project examples/egosuite_evaluation ruff check --fix examples/egosuite_evaluation
    uv run --locked --project examples/egosuite_evaluation ruff format examples/egosuite_evaluation
    uv run --locked --project examples/egosuite_evaluation ty check --project examples/egosuite_evaluation --extra-search-path . examples/egosuite_evaluation
    uv run --locked --project examples/egosuite_evaluation pytest -q examples/egosuite_evaluation/tests

All checks passed, 25 tests passed. sync and pytest ran in WSL2; ruff and ty ran on Windows. Native Windows cannot import `hflow.storage` (fcntl), which is the repo's documented requirement.

PR #338, opened 2026-09-01
refactor(build-ai): model evaluation artifacts at the JSON boundary

### LINKED ISSUE #316: Model Build AI evaluation artifacts at the JSON boundary

Sagar-024:
refactor(build-ai): model evaluation artifacts at the JSON boundary

Closes #316

## Summary

Models run metadata and per-sample outcomes as typed values instead of dicts, the same boundary treatment the EgoSuite adapter got in #333.

- `BuildAIRunMetadata` is parsed once when run.json loads. Task creation and summary generation read typed fields, never the JSON shape. The `cast` that reconstructed the prompts mapping is gone.
- Sample outcomes are exclusive variants: success carries `predicted_value`, invalid-response carries the parse error, execution error carries the Inspect error. `summarize_results` consumes the union through an exhaustive `match`, no status-string branching.
- A non-object or malformed saved run.json raises `ValueError` naming the file and the bad field.

run.json and summary.json schemas are unchanged, including comparison support for existing summary files. Existing experiment directories with valid metadata still resume by fingerprint, and summary values for the existing mixed-outcome fixture are unchanged. One fixture adjustment: the error sample now carries an `expected_value`, because `_sample_result` has always derived it from the sample target. The old fixture omitted it only because the dict shape allowed the impossible.

## Scope

No reuse of EgoSuite types, no shared framework across workspace projects, no changes to prompts, scoring, summary formulas, or Inspect log formats.

## Validation

All five commands ran in WSL2. Native Windows cannot import `hflow.storage` (fcntl), and the shared `.venv` was recreated as Linux-side, so the full sequence executed on the repo's required platform. The lockfile pins identical ruff, ty, and pytest versions.

### DIFF
--- a/examples/build_ai_evaluation/evaluate.py
+++ b/examples/build_ai_evaluation/evaluate.py
@@ -25,7 +25,7 @@ from dataclasses import dataclass
 from enum import StrEnum
 from itertools import islice
 from pathlib import Path
-from typing import Any, cast
+from typing import Any, assert_never, cast
 from urllib.parse import urlsplit, urlunsplit
 
 import duckdb
@@ -434,7 +434,68 @@ def _required_parquet_paths(
     return selected_paths
 
 
-def _run_metadata(configuration: EvaluationConfiguration) -> dict[str, object]:
+def _required_run_metadata_string(
+    document: Mapping[str, object], field_name: str, record_context: str
+) -> str:
+    field_value = document.get(field_name)
+    if not isinstance(field_value, str) or not field_value:
+        raise ValueError(f"{record_context} field {field_name!r} must be a string")
+    return field_value
+
+
+def _required_prompt_digests(document: Mapping[str, object], record_context: str) -> dict[str, str]:
+    raw_prompts = document.get("prompts")
+    if not isinstance(raw_prompts, dict):
+        raise ValueError(f"{record_context} field 'prompts' must be a JSON object")
+    digests: dict[str, str] = {}
+    for task_key, entry in raw_prompts.items():
+        if not isinstance(entry, dict) or not isinstance(entry.get("sha256"), str):
+            raise ValueError(
+                f"{record_context} prompts entry {task_key!r} must carry a string 'sha256'"
+            )
+        digests[str(task_key)] = entry["sha256"]
+    return digests
+
+
+@dataclass(frozen=True)
+class BuildAIRunMetadata:
+    """Saved run.json metadata, parsed once at the load boundary.
+
+    ``document`` is the full persisted schema; the other fields are the
+    consumed subset, typed so task creation and summary generation never
+    index back into the JSON shape.
+    """
+
+    label: str
+    fingerprint: str
+    dataset_variant: str
+    model: str
+    prompt_sha256s: dict[str, str]
+    document: Mapping[str, object]
+
+    def to_json_dict(self) -> dict[str, object]:
+        return dict(self.document)
+
+    @classmethod
+    def from_json_file(cls, metadata_path: Path) -> BuildAIRunMetadata:
+        try:
+            document = json.loads(metadata_path.read_text())
+        except (OSError, json.JSONDecodeError) as error:
+            raise ValueError(f"could not read run metadata {metadata_path}: {error}") from error
+        if not isinstance(document, dict):
+            raise ValueError(f"{metadata_path} must contain a JSON object")
+        context = str(metadata_path)
+        return cls(
+            label=_required_run_metadata_string(document, "label", context),
+            fingerprint=_required_run_metadata_string(document, "fingerprint", context),
+            dataset_variant=_required_run_metadata_string(document, "dataset_variant", context),
+            model=_required_run_metadata_string(document, "model", context),
+            prompt_sha256s=_required_prompt_digests(document, context),
+            document=document,
+        )
+
+
+def _run_metadata_document(configuration: EvaluationConfiguration) -> dict[str, object]:
     specification = DATASET_SPECIFICATIONS[configuration.dataset_variant]
     prompt_metadata = {
         task.value: {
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

The fixture change is the one judgment call in here and it holds up. Rather than take the reasoning on faith I drove a single scenario through `_sample_result` on both main and your branch, so both sides derive `expected_value` the way the adapter actually does, and diffed the summary:

```
scenario: 2 tasks, 6 samples (success, invalid response, execution error), latency + usage
main   -> 69 lines
branch -> 69 lines
identical
```

That is what DoD 7 needed. The `if "expected_value" in result` filter you dropped was dead code: `base_result` always set the key, so no result `_sample_result` could ever produce would miss it, and an error sample without an `expected_value` was a shape the adapter cannot emit. Dropping the filter and fixing the fixture are one correction, not two.

The `assert_never` is enforced rather than decorative. Added a fourth variant to the union and changed nothing else:

```
error[type-assertion-failure]: Argument does not have asserted type `Never`
   --> evaluate.py:652:25
    |   assert_never(result.outcome)
    |   Inferred type of argument is `TimeoutSampleOutcome & ~SuccessfulSampleOutcome & ...`
```

So a fifth outcome cannot be added without the summarizer being made to handle it, which is what DoD 8 was really asking for.

All four new guards bite:

```
run.json field type check dropped     -> test_run_metadata_names_the_file_and_the_bad_field
prompts digest validation dropped     -> that one, plus ..._writes_and_resumes_the_same_fingerprint
non-object run.json check dropped     -> test_run_metadata_refuses_a_non_object_run_json
fingerprint mismatch stops refusing   -> test_prepare_output_directory_refuses_a_different_experiment
```

`to_json_dict` plus the exact key-set assertion is a tighter pin on the run.json schema than #316 asked for, since it fails on an added key as well as a dropped one.

One small thing, no action needed. `_run_metadata` recomputes the prompt digests with the same comprehension `_run_metadata_document` already ran, so the same fact is derived twice from `task_definitions`. They agree today because the two filters are identical, which is the kind of agreement that quietly stops being true. `_required_prompt_digests` over the document you just built would leave one source. The `cast` at `:856` is over a saved summary's `sources`, a different boundary from the one #316 named, so leaving it is right.

Gate clean with main merged in: 32 passed on the example project, 1347 passed / 6 skipped at the root.

#341 is unblocked now, so reopen it whenever you want.

PR #339, opened 2026-09-01
fix(runtime): refuse hostless remote Airflow URLs at resolution boundary (fixes #298)

### LINKED ISSUE #298: Remote runtime accepts Airflow URLs with no host
## Current behavior

`resolve_remote_endpoint` parses the Airflow URL but checks only whether its scheme is `http` or `https` at [`src/hflow/runtime/_endpoint.py:43-66`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/runtime/_endpoint.py#L43-L66).

Measured on `main` at `1f5c7c8`, this is accepted:

```python
resolve_remote_endpoint(
    airflow_url="http://",
    dag_id="demo",
    environ={"HFLOW_AIRFLOW_TOKEN": "token"},
)
```

It returns a `RemoteRuntimeEndpoint` with no hostname. Client construction strips the trailing slashes, and `hflow status` later reports an unreachable `http:/api/v2/...` endpoint. A bad config reaches the request layer as a runtime outage.

The same happens for `https:///path`, which has an allowed scheme and no network location.

## Pattern to copy

[`TestResolveRemoteEndpoint`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_runtime_client.py#L293-L347) already proves that a scheme-less URL is refused at the config boundary and that valid flag and environment values resolve normally. Extend that class.

`parse_storage_root` at [`src/hflow/storage.py:632-653`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/storage.py#L632-L653) shows the same boundary rule for storage URLs: parse the components and reject a value that names no target.

## What to build

Require an HTTP(S) Airflow URL to contain a hostname. Keep optional ports and path prefixes working. Raise `ValueError` from `resolve_remote_endpoint` with a message that names `--airflow-url` and `HFLOW_AIRFLOW_URL`.

Add outcome-focused resolver and CLI coverage. Prove the invalid value is refused before any Airflow client or HTTP request is created.

## Definition of done

1. `http://`, `https://`, and `https:///path` are refused as hostless Airflow URLs.
2. Valid hostnames, IPv4 loopback addresses, explicit ports, and path-prefixed base URLs remain accepted.
3. Both the explicit argument and environment-variable paths enforce the same rule.
4. The error names `--airflow-url` and `HFLOW_AIRFLOW_URL` and reaches the CLI as exit code 2 without a traceback.
5. Existing scheme, DAG ID, and credential validation is unchanged.

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_runtime_client.py tests/test_runtime_cli.py
uv run pytest -q
HFLOW_DOCKER_TESTS=1 uv run pytest tests/test_runtime_integration.py -q
```

Run the full suite from the repo root. `pytest tests` alone misses `packages/hflow-server/tests`. The Docker integration command is part of the documented gate for runtime endpoint changes.

webdevsamran:
Fixes #298.

## Summary

esolve_remote_endpoint validated that an Airflow URL had an http or https scheme, but did not verify that a hostname was present. As a result, hostless URLs such as http://, https://, and https:///path were accepted at configuration resolution time and passed down to the client/request layers, failing downstream with unhelpful request errors or status timeouts.

This change:
1. Validates that parsed_base_url.hostname is non-empty in 
esolve_remote_endpoint.
2. Raises a ValueError naming --airflow-url and HFLOW_AIRFLOW_URL when a hostless URL is supplied via flags or environment variables.
3. Adds resolver tests in TestResolveRemoteEndpoint covering http://, https://, https:///path, and environment variable resolution.
4. Adds CLI test coverage proving that passing a hostless Airflow URL exits with code 2 and outputs a clean error to stderr without an unhandled traceback.

## Verification

- uv run ruff check (clean)
- uv run ty check --platform linux (clean)
- uv run pytest -q tests/test_runtime_client.py -k "test_hostless or test_scheme_less" (5 passed)
- uv run pytest -q tests/test_runtime_cli.py -k "test_ingest_remote" (2 passed)

### DIFF
--- a/src/hflow/runtime/_endpoint.py
+++ b/src/hflow/runtime/_endpoint.py
@@ -57,11 +57,17 @@ def resolve_remote_endpoint(
     base_url = airflow_url or environment.get(AIRFLOW_URL_ENVIRONMENT_VARIABLE)
     if not base_url:
         return None
-    if urlsplit(base_url).scheme not in ("http", "https"):
+    parsed_base_url = urlsplit(base_url)
+    if parsed_base_url.scheme not in ("http", "https"):
         # Parse at the boundary: a scheme-less URL would otherwise surface
         # as a raw urllib ValueError traceback deep inside the first request.
         raise ValueError(
             f"remote Airflow URL {base_url!r} needs an http:// or https:// scheme "
+            f"(from --airflow-url or {AIRFLOW_URL_ENVIRONMENT_VARIABLE})"
+        )
+    if not parsed_base_url.hostname:
+        raise ValueError(
+            f"remote Airflow URL {base_url!r} needs a host "
             f"(from --airflow-url or {AIRFLOW_URL_ENVIRONMENT_VARIABLE})"
         )
 
--- a/tests/test_runtime_cli.py
+++ b/tests/test_runtime_cli.py
@@ -942,6 +942,27 @@ def test_ingest_remote_without_credentials_names_the_environment_fix(
     assert "HFLOW_AIRFLOW_TOKEN" in capsys.readouterr().err
 
 
+def test_ingest_remote_with_hostless_url_names_the_fix(
+    monkeypatch: pytest.MonkeyPatch,
+    capsys: pytest.CaptureFixture[str],
+) -> None:
+    monkeypatch.setenv("HFLOW_AIRFLOW_TOKEN", "minted-token")
+    exit_code = main(
+        [
+            "ingest",
+            "a.mcap",
+            "--airflow-url",
+            "http://",
+            "--dag-id",
+            "kitchen_ingest",
+        ]
+    )
+    assert exit_code == 2
+    error_output = capsys.readouterr().err
+    assert "needs a host" in error_output
+    assert "--airflow-url" in error_output or "HFLOW_AIRFLOW_URL" in error_output
+
+
 def test_explicit_bundle_dir_stays_local_even_with_remote_environment(
     monkeypatch: pytest.MonkeyPatch,
     pipeline_file: Path,
--- a/tests/test_runtime_cli.py
+++ b/tests/test_runtime_cli.py
@@ -960,7 +960,8 @@ def test_ingest_remote_with_hostless_url_names_the_fix(
     assert exit_code == 2
     error_output = capsys.readouterr().err
     assert "needs a host" in error_output
-    assert "--airflow-url" in error_output or "HFLOW_AIRFLOW_URL" in error_output
+    assert "--airflow-url" in error_output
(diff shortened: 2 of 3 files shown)


kstonekuan (approved):
LGTM, merging. `parsed_base_url.hostname` is the right predicate and the interesting part is why.

Ran the same set of shapes against main and your branch:

```
                          main       branch
'http://'                 accepted   refused
'https://'                accepted   refused
'https:///path'           accepted   refused
'http:///'                accepted   refused
'http://:8080'            accepted   refused
'http://user@'            accepted   refused

'http://127.0.0.1:8080'            accepted   accepted
'http://127.0.0.1:8080/airflow'    accepted   accepted
'https://example.com:8443/prefix'  accepted   accepted
'http://localhost'                 accepted   accepted
'http://[::1]:8080'                accepted   accepted
'workspace.example.com:8080'       refused    refused   (scheme, unchanged)
```

Three more hostless shapes than #298 named, and no acceptance narrowed, so DoD 2 and DoD 6 both hold.

The one thing worth pushing back on is that nothing held the predicate in place. `netloc` is the tempting alternative, and swapping your line to `if not parsed_base_url.netloc:` passed the entire suite:

```
hostname swapped for netloc  ->  88 passed, nothing noticed
```

It passes because `urlsplit` gives `''` for the netloc of all three URLs you parametrized, so they cannot tell the two apart. `http://:8080` and `http://user@` are where they diverge: a port or userinfo makes `netloc` truthy while there is still no host, which is the same class of bad config #298 is about.

Pushed one fixup, `8e5c652`, doing three things:

```
host check removed entirely  ->  7 failed  (was 5)
hostname swapped for netloc  ->  2 failed  (was 0)
```

Added those two shapes to your parametrize, added an acceptance case over ports, path prefixes, and loopback addresses so DoD 2 is a test rather than an inspection, and changed the CLI assertion from `"--airflow-url" in err or "HFLOW_AIRFLOW_URL" in err` to two separate asserts. DoD 4 asks the error to name both, and an `or` passes when it names one.

Gate clean with main merged in: 1358 passed / 6 skipped.

On what to pick up next: #340, #342, and #343 are all still good work, so reopen whichever you want to see through now that this one is merged. Past that, the `good first issue` pool is best left for people arriving after you, and the higher-leverage thing is running HFlow against a real corpus. Egocentric-10K or Egocentric-100K on Hugging Face, and tell us what breaks, what is slow, or what is awkward. #287 came out of exactly that and it is worth more than any single starter fix.

PR #344, opened 2026-09-01
Use typed Airflow responses in runtime APIs

### LINKED ISSUE #315: Parse Airflow DAG runs and task instances into typed client responses

ravik453:
## Summary

- Add immutable typed Airflow DAG-run and task-instance response models.
- Parse Airflow responses inside `AirflowClient` and reject malformed non-object entries with endpoint context.
- Update runtime, graph, endpoint, and test consumers to use typed responses.
- Preserve existing trigger idempotency, run-ID encoding, token refresh, and runtime API behavior.

## Validation

- `uv run pytest -q tests/test_runtime_client.py` — 28 passed
- `uv run pytest -q packages/hflow-server/tests/test_server_run_graph.py` — 12 passed
- `uv run pytest -q packages/hflow-server/tests/test_server_runtime.py` — 19 passed
- `uv run pytest -q` — **1307 passed, 6 skipped**

Closes [#315](https://github.com/Hebbian-Robotics/hflow/issues/315)


kstonekuan (changes requested):
The typed models and the two `_parse_*` functions are the right shape, and the client-side coverage does its job: HTTP-boundary parsing for both response kinds, and a non-object member raising `AirflowClientError` with endpoint context. Three things before this can land.

**`POST /runtime/ingest` raises on the value `ingest()` now returns.** `ingest()` at `_client.py:415-455` is still annotated `-> dict[str, Any]` while returning `trigger_dag_run(...)`, which is now an `AirflowDagRun`, and `_runtime.py:457-458` still calls `.get()` on it. I stubbed `AirflowClient.ingest` with exactly what `_parse_dag_run` builds for a real 200 and changed nothing else:

```
route raised: AttributeError: 'AirflowDagRun' object has no attribute 'get'
```

`test_ingest_triggers_the_master_dag` cannot see this because its stub returns `{"dag_run_id": "manual__ui", "state": "queued"}`, a plain dict. The stub agrees with the old signature, so the test passes while the route is broken. That is the same defect #315 exists to remove, one layer up: a boundary the tests describe rather than exercise. When you fix the route, make the stub return the type the real method returns, otherwise the next signature change hides behind it again.

**`ty check` is red, 11 diagnostics.** It is not in your validation list, which is why these are still here:

```
src/hflow/runtime/_client.py:455          ingest() -> dict[str, Any] returns AirflowDagRun
packages/.../_graph.py:447                stage_task_instances -> list[dict[str, Any]] returns list[AirflowTaskInstance]
packages/.../_graph.py:529                consequence of :447
tests/test_runtime_client.py:101, 209
tests/test_runtime_integration.py:120, 167, 168, 319
```

`_graph.py:445` is the second annotation miss: `stage_task_instances` still declares dicts, so `_sorted_task_instances` is being handed a value the checker knows is the wrong type at the one call site that matters.

**The Docker integration tests were not updated.** The four `test_runtime_integration.py` hits above are `run.get("state")` and `run.get("conf")` against `dag_run()` and `dag_runs()`. Those are skipped without `HFLOW_DOCKER_TESTS=1`, which is why 1307 passed, but #315's validation section runs them and they are the only coverage that this works against a real Airflow.

Everything else reads correct. `_duration_seconds` keeps the bool guard through `_optional_float`, `queued_when or queued_at` moved across intact, and the `try_number` and `map_index` fallbacks match what `_graph.py` did before: I checked whether dropping the `int(...)` normalization was observable and it is not, because the pydantic field coerces at the model boundary.

Once `ty` is clean and the ingest route returns the right thing, this is a merge.

ravik453:
Thanks for the detailed review. I’ve addressed the requested changes:

Updated the typed AirflowDagRun consumers and ingest route.
Updated the affected tests, including the Docker integration tests.
Fixed the remaining Ruff lint issue.
Verified ty check, ruff check, formatting, and the full test suite locally.

The latest CI run is now green on both Python 3.11 and 3.14.

kstonekuan (approved):
LGTM, merging. All three, and you found a fourth I had missed.

`cli.py:1183` was doing `dag_run.get("dag_run_id", "<unknown>")` on the same value. I did not list it because `ty` could not see it either: `ingest()` claiming `dict[str, Any]` made that call legal, so the wrong annotation was hiding a second broken call site. Fixing the annotation is what surfaced it. That one had the worse failure mode of the two, since it lands after the trigger succeeds, so an operator would have watched `hflow ingest` traceback over a run that had already started.

Both sites are pinned now rather than just fixed. Reverting each to dict access:

```
ingest route back to dict access  ->  2 failed  (test_ingest_triggers_the_master_dag,
                                                 test_ingest_batch_count_rides_the_trigger_conf)
cli ingest back to dict access    ->  4 failed  (all four CLI ingest tests)
```

Both were 0 failures before this push. Converting every stub to return `AirflowDagRun` is what did it, and it is the right fix rather than the narrow one: the stubs now fail when the real signature moves, which is the property that was missing.

`ty check` clean, `ruff check` and `ruff format --check` clean, 1362 passed / 6 skipped with current main merged in. The four `test_runtime_integration.py` call sites are updated, so the `HFLOW_DOCKER_TESTS=1` command in #315 will typecheck now too.

On DoD 5, the composition holds: the client raises `AirflowClientError` for a non-object member, and both routes already map any `AirflowClientError` to 502, so a malformed entry cannot reach server parsing and cannot escape as a local `AttributeError`.

PR #345, opened 2026-09-01
fix(video): refuse B-frame streams whose reorder tail a remux drops

### LINKED ISSUE #255: write_access_units_to_mp4 silently drops the reorder tail of any B-frame stream

Sagar-024:
Supersedes #341, which the pr-limit workflow auto-closed while #338 was still open; same branch, same commit, nothing changed.

Implements option 2 from #255. A `-c:v copy` remux of a raw Annex B stream to MP4 cannot represent the B-frame reorder tail: the container holds every sample and the decoder emits fewer. #250 measured it at 303 samples in, 301 decoded, with no flag combination restoring the tail.

`scan_picture_coding_types` classifies each picture from its slice headers: `first_mb_in_slice` then `slice_type`, both Exp-Golomb. Codes 1 and 6 name B slices. It reports the B-picture count, the longest consecutive B run (the reorder depth the stream needs), and the trailing B run (the frames a remux drops). Detection reads slice headers rather than AUD primary_pic_type, because the AUD repair path deliberately writes the permissive type-7 delimiter without knowing the real picture type. An unparseable slice header fails closed.

`write_access_units_to_mp4` refuses a B-frame payload with an error naming the depth and the frames at risk. The transform refuses pass-through video carrying B-frames, alongside its other canonical-constraint refusals. docs/FORMAT.md documents both.

Streams without B-frames are untouched: the canonical fixture scans to zero B pictures and the remux path is unchanged. SPS VUI max_num_reorder_frames was deliberately not used: VUI parsing is a larger surface and the slice-header evidence is sufficient and local.

All five commands ran in WSL2 per the fcntl requirement. Lockfile-pinned ruff, ty, and pytest. Full suite: 1347 passed, 6 skipped.

Closes #255.

### DIFF
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -153,6 +153,14 @@ A file claiming this convention must satisfy, in increasing strictness:
 4. **Video constraints**: every `foxglove.CompressedVideo` message is one Annex B access unit beginning with an AUD; every IDR access unit contains SPS and PPS; no B-frames; `format="h264"`.
 5. **Stamps**: `provenance/v1` present with `schema_version` and `pipeline_version`.
 
+The no-B-frame constraint is enforced by refusal at two boundaries, because a `-c:v copy` remux
+to MP4 cannot represent the reorder tail: the trailing B frames are silently undecodable from
+the muxed file (measured in [#250](https://github.com/Hebbian-Robotics/hflow/issues/250): 303
+samples in, 301 decoded). The transform refuses pass-through video that carries B-frames, and
+the MP4 remux behind `Episode.video` refuses any B-frame payload outright, with an error naming
+the observed reorder depth and the frames at risk. `hflow doctor` still does not classify
+picture coding types (below), so a clean report does not prove this constraint; the refusals do.
+
 `hflow doctor <file.mcap> [more.mcap ...]` checks every file given and prints
 one report each in argument order; its aggregate result follows the
 [exit code rules](#exit-codes). It validates the container, summary, indexes,
--- a/src/hflow/transform.py
+++ b/src/hflow/transform.py
@@ -509,6 +509,14 @@ def _validate_passthrough_video_payload(
             f"pass-through video topic {topic!r} starts mid-GOP: its first message "
             "is not a keyframe, so the stream is not decodable from the start"
         )
+    coding_types = video_module.scan_picture_coding_types(message.data)
+    if coding_types.b_picture_count:
+        raise SourceNotConforming(
+            f"pass-through video topic {topic!r} carries B-frames (reorder depth "
+            f"{coding_types.reorder_depth}); the canonical convention requires "
+            "bframes=0 because a -c:v copy MP4 remux drops the reorder tail "
+            "(measured 301 of 303 in #250) -- re-encode upstream with bframes=0"
+        )
 
 
 def _decode_compressed_images(
--- a/src/hflow/video.py
+++ b/src/hflow/video.py
@@ -274,6 +274,34 @@ def _decode_unsigned_exp_golomb(rbsp: bytes) -> int:
     return (1 << leading_zero_bits) - 1 + suffix
 
 
+def _decode_unsigned_exp_golomb_at(rbsp: bytes, bit_offset: int) -> tuple[int, int] | None:
+    """Offset-aware form of :func:`_decode_unsigned_exp_golomb`.
+
+    Returns ``(value, next_bit_offset)``, or ``None`` when the value is
+    incomplete or truncated. Returning ``None`` instead of raising lets
+    callers count unparseable slice headers and refuse fail-closed rather
+    than guess a picture type.
+    """
+    bit_count = len(rbsp) * 8
+    leading_zero_bits = 0
+    while True:
+        if bit_offset + leading_zero_bits >= bit_count:
+            return None
+        byte = rbsp[(bit_offset + leading_zero_bits) // 8]
+        if (byte >> (7 - (bit_offset + leading_zero_bits) % 8)) & 1:
+            break
+        leading_zero_bits += 1
+    suffix_start = bit_offset + leading_zero_bits + 1
+    suffix_end = suffix_start + leading_zero_bits
+    if suffix_end > bit_count:
+        return None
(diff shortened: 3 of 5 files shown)


kstonekuan (approved):
LGTM, merging. Option 2 is the right call and #255 already leaned that way, so this is not a design decision taken here.

The thing worth reporting is that `trailing_b_pictures` is not just descriptive: it predicts the loss exactly. I rebuilt your fixture, pushed it through `write_access_units_to_mp4`'s own ffmpeg command with the guard bypassed, and counted container samples against decoded frames:

```
B-frames (bframes=3:b_adapt=0, your fixture)
  scan:  pictures=48  b=35  reorder_depth=3  trailing_b=2
  remux: container samples=48   decoder emits=46   lost=2

canonical control (bframes=0)
  scan:  pictures=48  b=0   reorder_depth=0  trailing_b=0
  remux: container samples=48   decoder emits=48   lost=0
```

That is #250 reproduced from scratch at a different frame count, and the number your error message quotes as "at risk" is the number the decoder actually drops. The control matters as much: a bframes=0 stream scans to zero and round-trips whole, so the refusal is not over-broad.

All five guards bite:

```
remux refusal removed                  -> test_remux_refuses_a_b_frame_stream_naming_the_tail
transform refusal removed              -> test_transform_rejects_passthrough_video_carrying_b_frames
scanner blind to B slices              -> the above two, plus both scan tests
reorder depth always zero              -> test_scan_measures_reorder_depth_and_trailing_tail,
                                          test_scan_reports_b_pictures_in_a_real_b_frame_stream
trailing tail always zero              -> test_scan_measures_reorder_depth_and_trailing_tail
```

Reading slice headers rather than AUD `primary_pic_type` is the correct choice for the reason you give, and `slice_type % 5 == 1` is exactly Table 7-6's B set. The `test_scan_reports_no_b_pictures_for_the_canonical_stream` assertion against the whole `PictureCodingScan` is what makes a false positive on canonical video impossible to introduce quietly.

The FORMAT.md paragraph is the right kind of honest, particularly saying outright that a clean `hflow doctor` report does not prove the constraint and the refusals are what do.

Two things for later, neither blocking.

`trailing_b_pictures` is pinned only by the synthetic test. The real-stream test asserts `b_picture_count > 0` and `reorder_depth >= 1` but says nothing about the tail, which is the one field the error message turns into a claim about frames. The measurement above is what ties them together, and asserting it would keep them tied.

The scan is now the dominant cost of the remux call:

```
 300 frames, 1.22 MB:  scan  45 ms | count_h264_pictures  45 ms | whole remux call  65 ms
3000 frames, 12.2 MB:  scan 448 ms | count_h264_pictures 445 ms | whole remux call 494 ms
```

`_remove_emulation_prevention_bytes` copies every NAL byte by byte to read two Exp-Golomb values off the front. Unescaping only the first 16 bytes gives identical results on canonical, B-frame, and multi-slice B streams, 9x to 28x faster. It is not this PR's defect, since `count_h264_pictures` has carried the same cost all along and is now doing the same walk twice, so I filed it separately as #346. It is yours if you want it.

PR #348, opened 2026-09-01
refactor(app): group sync reuse witness fields

### LINKED ISSUE #319: Group sync reuse witness fields into one typed value
## Current behavior

`_SyncCompletion` describes one optional reuse witness with three independent optional fields at [`src/hflow/app.py:247-274`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/app.py#L247-L274):

```python
source_digest: str | None
ffmpeg_version: str | None
transform_kind: _TransformKind | None
```

The current writer always supplies the complete set. The lenient marker parser can produce every partial combination at [`src/hflow/app.py:355-400`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/app.py#L355-L400). The reuse gate then checks the fields separately at [`src/hflow/app.py:1154-1206`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/app.py#L1154-L1206).

The behavior is fail-closed today, but the type does not express the rule the code relies on: a reuse witness is complete or absent.

## Pattern to copy

[`RetainedDatasetSnapshotBackup`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/snapshot.py#L59-L81) groups two values that only have meaning together, and the report carries that group as one optional value.

## What to build

Introduce a frozen `_SyncReuseWitness` holding `source_digest`, `ffmpeg_version`, and `transform_kind`. Change `_SyncCompletion` to carry `reuse_witness: _SyncReuseWitness | None`.

Parse a complete, known witness into `_SyncReuseWitness`. Treat a missing, partial, malformed, or future witness as absent so old markers remain readable for non-sync stages and sync still retranscodes on any doubt.

Keep the JSON marker shape byte-for-byte compatible. This is an internal representation change.

## Definition of done

1. `_SyncCompletion` cannot represent a partial reuse witness.
2. The current marker still contains the same six keys with the same JSON bytes and ordering.
3. A marker with all valid witness fields can reuse an unchanged default transform.
4. A legacy marker with no witness remains readable and retranscodes once.
5. Every partial witness combination remains readable and cannot trigger reuse.
6. An unknown `transform_kind` remains readable and cannot trigger reuse.
7. Transform overrides still never reuse.
8. No transform behavior version bump is needed because canonical bytes and marker bytes stay unchanged.

## Non-goals

- Migrating existing marker files
- Versioning transform overrides
- Changing the marker format
- Changing which canonical episodes qualify for reuse

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

ravik453:
## Summary

- group the sync reuse witness into a single frozen `_SyncReuseWitness`
- parse partial, malformed, missing, or future witnesses as absent while keeping legacy markers readable
- keep the existing sync completion marker JSON shape byte-for-byte compatible
- add explicit coverage for all six partial witness combinations

Closes #319

## Validation

- `uv run ruff check --fix`
- `uv run ruff format`
- `uv run ty check`
- `uv run pytest -q tests/test_sync_reuse.py`
- `uv run pytest -q`

All checks passed:
- 22 focused tests passed
- 1374 tests passed
- 6 skipped

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -253,10 +253,24 @@ class _TransformKind(StrEnum):
 
 @dataclass(frozen=True)
 class _SyncReuseWitness:
-    """Proof that a canonical can be safely reused by sync."""
+    """Enough to decide that re-running sync would rewrite byte-identical output.
 
+    Complete or absent: a partial witness cannot prove that re-running sync
+    would reproduce the existing canonical, so the parser drops the whole
+    group rather than letting the surviving fields answer for the missing one.
+    """
+
+    # Which source BYTES produced the canonical. Content, never size+mtime:
+    # this repo identifies by content everywhere, and a same-length rewrite
+    # or a preserved mtime would defeat the weaker test silently.
     source_digest: str
+    # The transform is stamped with an ffmpeg version that does NOT reach
+    # pipeline_version, and different builds genuinely encode differently.
     ffmpeg_version: str
+    # An @app.transform override is contractually required to end in
+    # write_canonical_episode, so it stamps the SAME pipeline_version the
+    # default transform would: without this, REMOVING an override would reuse
+    # a canonical the current pipeline cannot produce.
     transform_kind: _TransformKind
 
 
@@ -264,8 +278,11 @@ class _SyncReuseWitness:
 class _SyncCompletion:
     """Proof that sync completed for one source path and canonical version.
 
-    ``reuse_witness`` is complete or absent: a partial witness cannot prove
-    that re-running sync would reproduce the existing canonical.
+    ``reuse_witness`` is optional because markers written before it existed
+    are still valid proof for the non-sync stages, which is all those stages
+    ever asked of them. A witness-less marker simply never satisfies the reuse
+    gate: one re-transcode rewrites it in the current shape, and that is the
+    whole migration.
     """
 
     source_path: str
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -252,33 +252,26 @@ class _TransformKind(StrEnum):
 
 
 @dataclass(frozen=True)
+class _SyncReuseWitness:
+    """Proof that a canonical can be safely reused by sync."""
+
+    source_digest: str
+    ffmpeg_version: str
+    transform_kind: _TransformKind
+
+
+@dataclass(frozen=True)
 class _SyncCompletion:
     """Proof that sync completed for one source path and canonical version.
 
-    The last three fields are the *reuse witness*: enough to decide that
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

DoD 2 is the one with real downside if it slips, so I checked the bytes rather than the shape. Same run on main and on your branch:

```
raw bytes length: 264
key order: ['ffmpeg_version', 'pipeline_version', 'schema_version',
            'source_digest', 'source_path', 'transform_kind']
trailing byte: b'\n'
transform_kind in JSON: str
identical
```

Writing `transform_kind` as the enum object rather than `.value` is fine because `StrEnum` serializes as its string, and `sort_keys=True` means the insertion order you changed cannot show up either way.

The gate rules bite:

```
unknown transform_kind falls back to DEFAULT  -> ..._without_a_known_transform_kind_...
gate stops checking transform_kind            -> test_removing_an_override_does_not_reuse_its_canonical
gate stops comparing source_digest            -> test_a_source_replaced_in_place_is_transcoded_again
```

Pushed one fixup, `da1fc0c`, for two things.

The three field comments went missing in the move. They are the reason the fields are what they are, particularly the override one, which explains a failure mode nobody would reconstruct from the code: removing an `@app.transform` override would otherwise reuse a canonical the current pipeline cannot produce. I put all three back on `_SyncReuseWitness` and kept your all-or-nothing sentence.

The six-combination test pins the outcome but not the mechanism. I made the parse lenient again, one field at a time:

```
parse stops requiring ffmpeg_version  -> 22 passed, nothing noticed
parse stops requiring source_digest   -> 22 passed, nothing noticed
```

Both pass because a partial witness still fails a later gate comparison, so `sync_reused is False` holds for the wrong reason. `ty` does not catch it either: the values come off `json.loads` as `Any`, so nothing stops a `None` reaching a `str` field. The fixup asserts `_read_sync_completion_marker(...).reuse_witness is None` for each combination, read before the runs that rewrite the marker. Both mutations now fail.

Gate clean, 1374 passed / 6 skipped.

Two other things. #349 arrived an hour after yours against the same issue; first PR wins, so that one is closed. And while measuring the gate I found the ffmpeg comparisons at `app.py:1378-1387` are not covered by anything, because every fixture in that module is camera-less, so `ffmpeg_version` is `not-used` on both sides. Deleting the comparison leaves all 1374 tests green. That predates your change; it is #351 now.

Issue #346, opened 2026-09-01
Slice-header scans unescape whole NALs to read two Exp-Golomb values


kstonekuan:
## Current behavior

Reading a slice header needs the first two Exp-Golomb values, `first_mb_in_slice` and `slice_type`, which live in the first bytes of the NAL payload. Both slice-header walks unescape the entire NAL to reach them:

- `count_h264_pictures` at `src/hflow/video.py:323`
- `scan_picture_coding_types` at `src/hflow/video.py:391`

Both call `_remove_emulation_prevention_bytes` (`video.py:241`) over `stream[nal_header_offset + 1 : nal_end_offset]`, the whole payload. That helper appends byte by byte to a `bytearray` in pure Python, so each walk copies the entire stream once regardless of how few bytes it reads.

`scan_picture_coding_types` runs on the canonical write path: `write_access_units_to_mp4` calls it before every remux, reached from `Episode.video` at `src/hflow/episode.py:544`. Measured on `main`, remuxing canonical (bframes=0) testsrc2 video at 640x480:

```
 300 frames, 1.22 MB:  scan  45 ms | count_h264_pictures  45 ms | whole remux call  65 ms
3000 frames, 12.2 MB:  scan 448 ms | count_h264_pictures 445 ms | whole remux call 494 ms
```

The ffmpeg subprocess that does the actual remux is the remaining 20 to 46 ms. The scan is 70% to 91% of the call.

## What to build

Unescape only as much of each NAL as the slice header can occupy, instead of all of it. Reading the first 16 bytes is enough for both fields at any legal `first_mb_in_slice`, and the existing `_decode_unsigned_exp_golomb_at` and `_slice_header_fields` already return `None` on a short buffer, so a head too small is a detectable condition rather than a wrong answer.

Measured with a 16-byte head, same inputs:

```
canonical 300 frames    45 ms -> 1.6 ms   28x
canonical 3000 frames  447 ms ->  16 ms   27x
B-frames 300 frames     42 ms -> 1.5 ms   27x
multi-slice B 300       47 ms -> 5.1 ms    9x
```

All four produce a `PictureCodingScan` identical to the current code's.

[...]

PR #349, opened 2026-09-01
refactor(app): type the sync-reuse witness as one all-or-nothing value

### LINKED ISSUE #319: Group sync reuse witness fields into one typed value

Sagar-024:
refactor(app): type the sync-reuse witness as one all-or-nothing value

Closes #319.

The reuse witness (source_digest, ffmpeg_version, transform_kind) lived 
as three scattered optional fields on _SyncCompletion, and the reader 
dropped each field independently on a malformed value. A partial 
record, any one of the three missing or wrong-typed, would silently 
pass the gate on the surviving two checks.

This groups the three into a single _SyncReuseWitness dataclass 
mirroring RetainedDatasetSnapshotBackup. The reader builds the witness 
all-or-nothing: any missing or non-string field, or any unknown 
_TransformKind value, drops the whole group to None. The writer 
serializes it back into the same 6 JSON keys, so the marker is 
byte-for-byte identical to main.

The gate now matches on the None | DEFAULT | OVERRIDE union with 
assert_never on the residual, so a future _TransformKind member added 
without extending the gate is a static error, not a silent miss.

The JSON marker shape is unchanged: same 6 keys, same sort order, same 
trailing newline. Existing markers stay readable. A legacy marker with 
no witness fields still parses, reuse is refused, and the marker is 
rewritten with the full witness on the next sync pass.

All five commands ran in WSL2. Native Windows cannot import 
hflow.storage (fcntl), and the shared .venv was recreated as Linux-side, 
so the full sequence executed on the repo's required platform. The 
lockfile pins identical ruff, ty, and pytest versions.

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -30,7 +30,7 @@ from dataclasses import dataclass, field
 from enum import StrEnum
 from pathlib import Path
 from types import MappingProxyType, ModuleType
-from typing import TYPE_CHECKING, Generic, NewType, TypeVar, assert_never
+from typing import TYPE_CHECKING, Generic, Never, NewType, TypeVar, assert_never, cast
 
 if TYPE_CHECKING:
     from hflow.runtime import BundlePaths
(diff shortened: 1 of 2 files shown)


kstonekuan:
Closing this one: #348 landed against #319 an hour ahead of it, and duplicates go to whoever opened first. Nothing wrong with the work, and thank you for it.

Two things from yours worth saying out loud, since they were not just differences.

You kept the three field comments. #348 dropped them in the move, and I restored them onto `_SyncReuseWitness` as a fixup before merging, which is the call you had already made. You also wrote `transform_kind.value` explicitly rather than relying on `StrEnum` serializing itself.

The other one is worth more than the PR. The comment on the residual branch says a future `_TransformKind` member added without extending the gate "becomes a static error here, not a silent miss". It does not:

```
as submitted (cast present, 2 enum members)   ty clean
cast present, 3 enum members                  ty clean      <- the case the comment promises to catch
cast dropped, 2 enum members                  ty error at the assert_never
cast dropped, 3 enum members                  ty error at the assert_never
```

`cast(Never, x)` tells the checker the value is already `Never`, so `assert_never` can never object. Dropping the cast shows why it was added: `ty` will not prove that match exhaustive even today, because it does not narrow an enum-valued field inside a class pattern. So the cast silenced a real "cannot prove this" and the comment then claimed the opposite guarantee.

The version that works is matching on the enum itself rather than on a class pattern that contains it: bind the witness, then `match witness.transform_kind` with `assert_never(transform_kind)` on the residual. That is the shape in `summarize_results` on #338, where adding a variant does fail the check. Worth knowing generally, since `assert_never` only buys anything when the checker can actually reach the contradiction.

#346 is still unclaimed and is yours if you want it: the slice-header scan you added in #345 is 70 to 91 percent of every remux call, and a bounded head fixes it at 9x to 28x.

PR #352, opened 2026-09-01
test: cover ffmpeg_version sync-reuse gate with camera episodes

### LINKED ISSUE #351: The sync reuse gate's ffmpeg comparisons are untested: every fixture is camera-less
## Current behavior

The sync reuse gate refuses reuse on three video-related grounds in `_reusable_canonical_episode` (`src/hflow/app.py:1306`):

```python
src/hflow/app.py:1378    if canonical_stamps.ffmpeg_version != witness.ffmpeg_version:
src/hflow/app.py:1380    if canonical_stamps.ffmpeg_version != FFMPEG_VERSION_NOT_USED:
src/hflow/app.py:1387            if canonical_stamps.ffmpeg_version != ffmpeg_version():
```

None of them is exercised. Every episode in `tests/test_sync_reuse.py` is built by its `_episode` helper at `tests/test_sync_reuse.py:25`:

```python
SyntheticEpisodeSpec(duration_s=duration_s, cameras=(), seed=seed)
```

No cameras means no transcode, so `stamps.ffmpeg_version` is the literal `FFMPEG_VERSION_NOT_USED` on both sides of `:1378` and the branch at `:1380` never opens. Measured by deleting `:1378-1379` outright:

```
whole suite, marker ffmpeg_version comparison removed: 1374 passed, 6 skipped
```

Removing the `:1380` branch entirely, or the live-build comparison at `:1387`, is likewise unnoticed.

## Why this is worth pinning

`tests/test_sync_reuse.py` opens by saying every test there is about the proof, "because a wrong reuse silently publishes stale bytes under a current version stamp". The ffmpeg build is the one input that changes encoded bytes without changing `pipeline_version`, which is exactly why the witness carries it (see the field comment on `_SyncReuseWitness.ffmpeg_version`). So the half of the gate that guards the expensive, lossy, machine-dependent part of the output is the half no test holds.

The behavior is correct today. This is missing coverage, not a defect.

## Pattern to copy

`SyntheticEpisodeSpec` accepts cameras; `hflow.testing` is already imported in the module. `test_a_marker_without_a_witness_transcodes_once_and_gains_one` and the partial-witness test at `tests/test_sync_reuse.py:146` both show the idiom of writing a doctored marker back and re-running `app.process`, which is how you forge a marker recording a different `ffmpeg_version`.

For the live-build comparison at `:1387`, note the local import inside the branch: `from hflow.ffmpeg import ffmpeg_version`. Monkeypatching has to target the source module, not a name bound in `hflow.app`.

## What to build

Reuse coverage over an episode that actually has video:

1. An unchanged source with a camera is reused on the second run, so the video path can reuse at all. This is the control, and without it the rest can pass vacuously.
2. A marker recording a different `ffmpeg_version` than the canonical's stamps refuses reuse.
3. A canonical whose stamped `ffmpeg_version` no longer matches the resolved build refuses reuse.

Prove each bites: delete the comparison it covers, watch that test go red, put it back. Put the result in the pull request body.

## Definition of done

1. Each of the three comparisons above has a test that fails when that comparison is removed.
2. A camera-bearing episode still reuses normally when nothing changed.
3. Camera-less episodes keep their current behavior, and the existing tests are unchanged.
4. No test forces an ffmpeg download on a camera-less path: the `:1380` branch exists precisely to avoid that, so do not defeat it.
5. Runtime stays reasonable. One short camera-bearing episode is enough; this suite is fast today and should stay that way.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

## Notes

Found while reviewing #348, which regrouped these fields into `_SyncReuseWitness`. The gap predates that change and is not caused by it.

cestercian:
## Summary

The sync reuse gate's three `ffmpeg_version` comparisons in `_reusable_canonical_episode` were untested: every episode in `tests/test_sync_reuse.py` was camera-less, so stamps stayed `FFMPEG_VERSION_NOT_USED` and those branches never ran.

## Changes

Added a short camera-bearing episode helper and three tests:

1. Control: unchanged camera source is reused on the second run (real ffmpeg stamp)
2. Marker recording a different `ffmpeg_version` refuses reuse
3. Canonical stamps that no longer match live `hflow.ffmpeg.ffmpeg_version` refuse reuse (monkeypatch the source module, not `hflow.app`)

Camera-less tests are unchanged.

## Delete-comparison proof

- Drop `stamps != witness` → marker test fails (`sync_reused` was True)
- Drop the video/live-build branch → live-build test fails
- Drop `stamps != ffmpeg_version()` → live-build test fails
- Restore comparisons → all pass again

## Validation

```
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_sync_reuse.py   # 25 passed
uv run pytest -q                            # 1377 passed, 6 skipped
```

Fixes #351

### DIFF
--- a/tests/test_sync_reuse.py
+++ b/tests/test_sync_reuse.py
@@ -16,6 +16,7 @@ import pytest
 
 import hflow
 from hflow.app import _read_sync_completion_marker
+from hflow.format import FFMPEG_VERSION_NOT_USED
 from hflow.testing import SyntheticEpisodeSpec, synthesize_episode
 from hflow.transform import EpisodeStamps, TransformConfig, write_canonical_episode
 
@@ -25,6 +26,22 @@ SYNC_ONLY = {hflow.Stage.SYNC}
 def _episode(path: Path, *, duration_s: float = 1.0, seed: int = 0) -> Path:
     return synthesize_episode(
         path, SyntheticEpisodeSpec(duration_s=duration_s, cameras=(), seed=seed)
+    )
+
+
+def _camera_episode(path: Path, *, duration_s: float = 0.5, seed: int = 0) -> Path:
+    """One short camera-bearing source: enough for ffmpeg to stamp a real version."""
+    return synthesize_episode(
+        path,
+        SyntheticEpisodeSpec(
+            duration_s=duration_s,
+            cameras=("wrist_cam",),
+            image_hz=10.0,
+            seed=seed,
+            black_segment=None,
+            timestamp_offset_segment=None,
+            joint_jump_at_s=None,
+        ),
     )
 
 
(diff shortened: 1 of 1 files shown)


kstonekuan (approved):
LGTM, merging. Welcome, and that was fast.

Reran the delete-comparison proof independently. All three bite, each on its own test and nothing over-firing:

```
marker vs canonical stamps comparison removed  -> ..._a_marker_recording_a_different_ffmpeg_version_...
video branch skipped entirely                  -> ..._no_longer_matches_the_live_build_...
live build comparison removed                  -> ..._no_longer_matches_the_live_build_...
```

Every one of those left all 1374 tests green two hours ago.

The control is what makes the other two mean anything, and asserting `first.stamps.ffmpeg_version != FFMPEG_VERSION_NOT_USED` inside it is the detail that matters: without it, a fixture that quietly stopped producing video would make the two refusal tests pass for the wrong reason, which is the exact failure #351 was about.

The witness assertion in the marker test is the other one I would have asked for:

```python
witness = _read_sync_completion_marker(marker_path).reuse_witness
assert witness is not None
assert witness.ffmpeg_version != first.stamps.ffmpeg_version
```

#348 landed an all-or-nothing witness parse an hour before this, so a doctored marker refusing reuse is now ambiguous between "the ffmpeg comparison fired" and "the whole witness was dropped". Pinning the witness as complete before the run removes that, and you would only think to do it having read the change that created the ambiguity.

Passing `black_segment=None`, `timestamp_offset_segment=None` and `joint_jump_at_s=None` explicitly is right too. Those default to real fault injection at 2s, 6s and 5s, all outside a 0.5s episode today, so leaving them would work by accident and stop working if the duration ever grew.

Runtime is fine: the three camera tests are 0.18s each, 25 passed in 1.42s.

Gate clean with main merged in: 1381 passed / 6 skipped.

If you want another, the starter list is here: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

Issue #336, opened 2026-08-31
Catalog UI shutdown installs the ui extension from the network to call stop_ui_server()


kstonekuan:
### What happens

`serve_catalog_ui` sets `server_started = True` immediately after `_start_duckdb_ui_server`
returns (`src/hflow/catalog_ui.py:132-133`), and shutdown calls `stop_ui_server()` on that flag
alone (`catalog_ui.py:90-95`):

```python
def _stop_duckdb_ui_server(catalog_connection, server_started) -> None:
    try:
        if server_started:
            catalog_connection.execute("CALL stop_ui_server()")
```

`stop_ui_server()` is a function the `ui` extension provides. Calling it on a connection that
never loaded the extension makes DuckDB go and **install `ui` from the network**, at shutdown,
to run a no-op: there is no server to stop if the extension was never loaded.

### Why it is reachable, and where it bites

`tests/test_catalog_ui.py:49` replaces the real starter:

```python
monkeypatch.setattr(catalog_ui, "_start_duckdb_ui_server", record_server_start)
```

The stub records the call and returns. It never runs `INSTALL ui` / `LOAD ui`, but
`serve_catalog_ui` still sets `server_started = True`, so every one of these tests reaches
`CALL stop_ui_server()` with no extension loaded and DuckDB fetches the extension over the
network while the test is tearing down.

That is almost certainly the `test_catalog_ui` timing flake three separate contributors reported
on 2026-08-31 (#326, #327, #328), all describing a thread-shutdown or timing failure that passed
on a rerun. A network fetch inside a teardown that a test is waiting on with a fixed timeout is
exactly that shape, and it makes the suite depend on network access it otherwise does not need.

### What to build

Only call `stop_ui_server()` when the `ui` extension is actually loaded on that connection:

```python
def _ui_extension_is_loaded(catalog_connection) -> bool:
    row = catalog_connection.execute(
        "SELECT count(*) FROM duckdb_extensions() WHERE extension_name = 'ui' AND loaded"
    ).fetchone()
    return row is not None and int(row[0]) > 0
```

and gate the call on `server_started and _ui_extension_is_loaded(...)`.

### Definition of done

1. Shutdown does not call `stop_ui_server()` when the `ui` extension is not loaded.
2. A real UI session still stops its server on shutdown.
3. A test pins that the stubbed-starter path issues no `stop_ui_server()` call. Recording the
   statements executed on the connection is enough; asserting DuckDB's install behavior is not
   the point.
[...]

Issue #320, opened 2026-08-31
Parse LeRobot metadata into a typed internal archive model


kstonekuan:
## Current behavior

The LeRobot importer has typed top-level episode and video-window dictionaries, but `_SourceArchive` still contains untyped nested values at [`src/hflow/importers/lerobot.py:50-89`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/importers/lerobot.py#L50-L89):

```python
info: dict
numeric_features: dict[str, dict]
```

`_fetch_info_json()` returns a raw dictionary at [`src/hflow/importers/lerobot.py:260-276`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/importers/lerobot.py#L260-L276). `_ensure_source_archive()` then parses `fps`, path templates, features, and feature kinds at several later points in [`src/hflow/importers/lerobot.py:300-469`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/importers/lerobot.py#L300-L469). Conversion continues to index the archive and inspect optional dictionaries at [`src/hflow/importers/lerobot.py:639-685`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/importers/lerobot.py#L639-L685).

Recent metadata bugs such as #299 and the boolean shape-dimension fix in `fbfd553` both came from values remaining loosely typed after the external JSON boundary.

Verified against `main` at `fbfd553`.

## Pattern to copy

[`parse_project_config()`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/project.py#L96-L157) accepts an open external document, validates its supported fields once, and returns a `ProjectConfig` dataclass. `DatasetSource` in the importer already provides the same internal shape for repository identity.

## What to build

Parse supported `meta/info.json` fields into immutable internal values before metadata discovery or conversion continues. Give the source archive, episode rows, video windows, and supported numeric feature specifications attribute-based domain types.

Keep the downloaded `info.json` cache as the original JSON-compatible document. Unknown upstream fields can remain ignored. Unsupported selected feature shapes should still fail before any episode or manifest is published.

[...]

Issue #314, opened 2026-08-31
Centralize ingest URI parsing across CLI, server, and SDK


kstonekuan:
## Current behavior

The CLI validates ingest URIs inline at [`src/hflow/cli.py:1113-1135`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/cli.py#L1113-L1135). The workspace server implements the same rule again at [`packages/hflow-server/src/hflow_server/_runtime.py:412-428`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/packages/hflow-server/src/hflow_server/_runtime.py#L412-L428).

The two entry points already differ:

- The server strips surrounding whitespace and refuses an empty result.
- The CLI forwards the original string to `AirflowClient.ingest()` at [`src/hflow/cli.py:1162-1168`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/cli.py#L1162-L1168).
- The public client accepts any `list[str]` and puts it into trigger conf unchanged at [`src/hflow/runtime/_client.py:312-352`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/runtime/_client.py#L312-L352).
- The task-side resolver still accepts a bare string and strips leading slashes for bucket roots at [`src/hflow/stage_execution.py:124-128`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/stage_execution.py#L124-L128).

That leaves the validity and normalization of one runtime URI dependent on which public entry point received it.

Verified against `main` at `fbfd553`.

## Pattern to copy

[`parse_step_version()`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/steps.py#L31-L45) parses a public string once and returns a refined `StepVersion`. [`_validated_relative_key()`](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/storage.py#L111-L125) owns the analogous containment rule for storage keys.

## What to build

[...]

PR #353, opened 2026-09-01
fix(lerobot): separate video cache files by index

### LINKED ISSUE #292: LeRobot video cache aliases different file indices to one MP4
## Current behavior

The LeRobot importer computes each remote video path from both `chunk_index` and `file_index` in [`src/hflow/importers/lerobot.py:744-755`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L744-L755). The local cache name at [`src/hflow/importers/lerobot.py:756-762`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L756-L762) includes the camera and chunk, but omits `file_index`.

Two episodes that use the same camera and video chunk with different file indices resolve to one local MP4. Once the first file exists, the second download is skipped and the later episode is converted from the earlier video.

Verified on `main` at `1f5c7c8`.

## Pattern to copy

The Parquet cache path in [`src/hflow/importers/lerobot.py:656-668`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L656-L668) includes both chunk and file indices and reuses only that exact source file.

The synthetic LeRobot corpus and network boundary used by existing tests live in [`tests/test_lerobot_converter.py:32-142`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_lerobot_converter.py#L32-L142).

## What to build

Include the video file index in the local cache identity. Keep reuse for repeated access to the same camera, chunk, and file.

Add an outcome-focused regression with two episodes that use the same camera and chunk but different video file indices. Give the two source videos distinguishable content or paths and prove each episode reaches the video conversion boundary with its own source. The test can replace the network and FFmpeg boundaries, but should assert the selected source outcome, not helper call counts or third-party behavior.

## Definition of done

1. Different `file_index` values in one camera and chunk use different cached MP4 paths.
2. Repeated access to the same camera, chunk, and file still reuses the cached MP4.
3. The regression fails against current `main` and passes with the fix.
4. Existing single-camera and multi-camera imports are unchanged.
5. Bump `CONVERTER_VERSION`, since an affected import can currently complete with the wrong video bytes.
6. No new runtime dependency is added.

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

Galabavamsi:
Closes #292.

## Summary

Make LeRobot video cache paths unique by camera key, video chunk index, and video file index. This prevents different MP4 source files from sharing one cache entry while preserving reuse for the same source.

## Why

The previous cache filename omitted `file_index`. Two episodes using the same camera and chunk but different video files could therefore be converted from the wrong cached MP4. The converter version is bumped to `lerobot-converter-v4` so previously prepared outputs are distinguished.

## Validation

- `uv run --python 3.11 --locked pytest -q tests/test_lerobot_converter.py` — 35 passed
- `uv run --python 3.11 --locked pytest -q` — 1,382 passed, 6 skipped
- `uv run --python 3.14 --locked pytest -q` — 1,382 passed, 6 skipped
- `uv run --python 3.11 --locked ruff check .` — passed
- `uv run --python 3.11 --locked ruff format --check .` — passed
- `uv run --python 3.11 --locked ty check` — passed

## Checklist

- [x] Added outcome-focused regression coverage for distinct video file indices and cache reuse.
- [x] Bumped the converter version for affected cached imports.
- [x] Ran Ruff check and format validation.
- [x] Ran type checking.
- [x] Ran the focused and full test suites on Python 3.11 and 3.14.
- [x] Added no runtime dependencies or generated artifacts.

### DIFF
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -41,7 +41,7 @@ DEFAULT_REVISION = "main"
 DEFAULT_OUTPUT_DIR = Path("./data/lerobot_pusht")
 DEFAULT_CAMERA_KEY = "observation.image"
 
-CONVERTER_VERSION = "lerobot-converter-v3"
+CONVERTER_VERSION = "lerobot-converter-v4"
 PRESENTATION_TIMESTAMP_EPSILON_S = 0.050
 
 # Timestamp handling
@@ -835,7 +835,10 @@ def _convert_single_episode(
         local_video_path = (
             cache_directory
             / "videos"
-            / (f"{camera_key.replace('/', '_').replace('.', '_')}-chunk{video_chunk_index}.mp4")
+            / (
+                f"{camera_key.replace('/', '_').replace('.', '_')}"
+                f"-chunk{video_chunk_index}-file{video_file_index}.mp4"
+            )
         )
         if not local_video_path.exists():
             _download_file(f"{dataset_base_url}/{video_relative_path}", local_video_path)
--- a/tests/test_lerobot_converter.py
+++ b/tests/test_lerobot_converter.py
@@ -11,6 +11,7 @@ import shutil
 import subprocess
 import urllib.request
 from pathlib import Path
+from typing import cast
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Both halves of the cache identity are pinned, which is the part that matters here since #292 is as much about keeping reuse as about splitting the paths:

```
cache name back to main's (chunk only)   -> test_video_cache_distinguishes_file_indices_...
video cache reuse defeated (always re-download) -> the same test
```

The test design is what earns that. Writing each video's URL as its own bytes means `converted_sources` records which source every conversion actually consumed, so the assertion is the outcome #292 asked for rather than a call count. And `fake_download` raising on a repeated URL makes reuse enforced by the harness instead of asserted after the fact. The `(0, 1, 0)` pass order gets distinctness and reuse out of one run.

`CONVERTER_VERSION` bumped to v4, which is DoD 5 and the reason a stale cache from before this cannot quietly satisfy a later import.

Gate clean with current main merged in: 1387 passed / 6 skipped.

One thing now that this is your second: the `good first issue` pool is best left for people arriving after you, so treat this as an invitation rather than a rule. The higher-leverage work is running HFlow against a real corpus, Egocentric-10K or Egocentric-100K on Hugging Face, and reporting what breaks, what is slow, or what is awkward. #287 came out of exactly that and is worth more than any single starter fix. #314, #320 and #336 are also open and none of them is a starter issue.

PR #354, opened 2026-09-02
perf(video): unescape only the slice-header head, not the whole NAL

### LINKED ISSUE #346: Slice-header scans unescape whole NALs to read two Exp-Golomb values

Sagar-024:
Unescapes only the slice-header head, not the whole NAL

count_h264_pictures and scan_picture_coding_types unescaped the entire VCL NAL payload to read two Exp-Golomb values. The unescape is byte-by-byte pure Python, so the cost scaled with NAL size even though the readers only need the first 16 bytes (first_mb_in_slice and slice_type fit at any legal macroblock index). On a 300-frame libx264 bframes=3 encode at 640x480 the scan was 70-91 percent of the remux call (issue measurement).

This adds _unescape_ebsp_head(ebsp, max_bytes) and uses it in both walks with _SLICE_HEADER_HEAD_BYTES = 16. When the unescaped head is too short to finish both fields, the scan and the picture counter fall back to the full unescape so the existing fail-closed messages (no complete first_mb_in_slice value / truncates) still fire on the malformed inputs the existing tests pin.

count_h264_pictures keeps a thin raising wrapper around the same algorithm so the existing hflow doctor diagnostic text stays byte-identical.

Measured on the same shape as the issue table (300 frames, bframes=3, b_adapt=0, 640x480), 10 iterations, median of the warm 5:

| Implementation           | Median | Speedup |
|---|---|---|
| Full NAL unescape (main) | 52.36 ms | 1.0x |
| Head-only scan (branch)  |  1.67 ms | 31.3x |

All four scan runs (each branch scanning its own and the other's B-frame stream) agree on picture_count=300, b_picture_count=223, reorder_depth=3, trailing_b_pictures=2. The full hflow test suite (1372 tests) passes unchanged.

All five commands ran in WSL2. Native Windows cannot import hflow.storage (fcntl), and the shared .venv was recreated as Linux-side, so the full sequence executed on the repo's required platform. The lockfile pins identical ruff, ty, and pytest versions.

Closes #346.

### DIFF
--- a/src/hflow/video.py
+++ b/src/hflow/video.py
@@ -43,6 +43,12 @@ _VCL_NAL_TYPES = frozenset({1, 2, 3, 4, 5})
 # Types whose RBSP starts with a slice header. Data partitions B/C (3/4) are
 # VCL data but carry no ``first_mb_in_slice`` of their own.
 _SLICE_HEADER_NAL_TYPES = frozenset({1, 2, 5})
+# How many bytes of a VCL NAL's payload to unescape when reading the slice
+# header. ``first_mb_in_slice`` and ``slice_type`` are the first two Exp-Golomb
+# values; 16 bytes is enough at any legal macroblock index (``#346`` measured
+# this against 4K-class inputs). A NAL whose slice header needs more is
+# caught by the head-too-short fallback in the scan.
+_SLICE_HEADER_HEAD_BYTES = 16
 # Annex B AUD with primary_pic_type 7 (any I/P/B picture), followed by the
 # required RBSP stop bit. x264 writes the narrower ``09 10`` for the streams
 # HFlow encodes; repair deliberately uses the permissive value because it must
@@ -238,6 +244,36 @@ def _nal_header_offset(stream: bytes, nal_start_offset: int) -> int:
     raise ValueError(f"invalid Annex B start code at byte {nal_start_offset}")
 
 
+def _unescape_ebsp_head(ebsp: bytes, max_bytes: int) -> bytes:
+    """Unescape up to ``max_bytes`` of an EBSP into a bytearray prefix.
+
+    The slice-header readers only need the first handful of bytes of any VCL
+    NAL's payload: ``first_mb_in_slice`` and ``slice_type`` are the first two
+    Exp-Golomb values, and 16 bytes is enough at any legal macroblock index.
+    Walking the whole payload copies the entire NAL per VCL NAL even when the
+    scan never looks at most of it, which is the cost ``#346`` measured.
+
+    Returns the unescaped prefix as a ``bytes`` value. The emulation-prevention
+    state is honored across the boundary, so a 0x03 byte straddling the
+    ``max_bytes`` cut is consumed and does not appear in the output -- the
+    caller is expected to fall back to the full unescape when the prefix
+    is too short to parse, which is what ``scan_picture_coding_types`` does.
+    """
+    if max_bytes >= len(ebsp):
+        return _remove_emulation_prevention_bytes(ebsp)
+    rbsp = bytearray()
+    consecutive_zeros = 0
+    for index, byte in enumerate(ebsp):
+        if index >= max_bytes:
+            break
+        if consecutive_zeros >= 2 and byte == 0x03:
+            consecutive_zeros = 0
+            continue
+        rbsp.append(byte)
+        consecutive_zeros = consecutive_zeros + 1 if byte == 0 else 0
+    return bytes(rbsp)
+
+
 def _remove_emulation_prevention_bytes(ebsp: bytes) -> bytes:
     rbsp = bytearray()
     consecutive_zeros = 0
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

This is on the canonical write path, so the property I cared about was equivalence rather than speed. Ran both readers on main and on your branch over sixteen inputs and diffed:

```
real encodes:  bframes=0, bframes=3, multi-slice, 320x240 and 1280x720
synthetic:     emulation-prevention triples straddling the head boundary (0,1,2,4,6,8,16)
synthetic:     first_mb_in_slice = 0, 1, 99, 8100, 139263
identical
```

139263 is the level 6.2 macroblock maximum, so the 16-byte budget is covered at the top of the legal range, not just in practice.

The fallback is what makes that safe, and it holds much harder than it needs to. Shrinking `_SLICE_HEADER_HEAD_BYTES` from 16 down to 0:

```
budget=16 (as shipped)  pictures=5 b=3 depth=2 tail=1
budget=8,4,2,1,0        identical, every time
```

At zero the head is empty, every slice takes the fallback, and the answers are still right. So the budget is purely a speed knob and a wrong value costs time rather than correctness. That is the right shape for a constant nobody will revisit.

Confirmed the win independently, and it is larger end to end than the issue asked for, because `hflow doctor` sits on the same walk:

```
              scan            count_h264_pictures   whole remux call
 300 frames   45.1 -> 1.9 ms   44.6 -> 1.5 ms        86.5 -> 21.0 ms
3000 frames  447.8 -> 18.6 ms  457.8 -> 15.3 ms     492.1 -> 62.4 ms
```

The scan went from 91 percent of the remux call to 30 percent.

One thing to be straight about: DoD 4 did not land. The two walks still have their own loop and their own Exp-Golomb decoder, and the head-then-fallback logic is now written out in both, so the duplication grew. Your reason is sound, since `count_h264_pictures` owes two specific messages that the scan does not, and reconciling that is real work rather than a tidy-up. That is my fault for putting a "while in there" paragraph into a definition of done. It is #355 now, on its own.

Splitting `_decode_first_mb_in_slice` out with the old messages preserved verbatim was the right call for this PR either way: it keeps `hflow doctor` byte-identical on malformed input, which is the one thing a performance change here must not touch.

Gate clean with current main merged in: 1391 passed / 6 skipped.

PR #356, opened 2026-09-02
fix: escape xcom_objectstorage_url before compose YAML substitution

Raghav7-tech:
_render_compose() substituted xcom_objectstorage_url into the generated
docker-compose.yaml's %{xcom_objectstorage_path} slot unescaped, unlike
other values in the same function (host_data_root, hflow_source) which
correctly go through _compose_path_scalar()'s escaping.

## Summary
A value containing a newline followed by the environment block's own
4-space indentation is parsed as a new sibling YAML key in
x-airflow-common.environment, letting an override value overwrite an
unrelated existing key. Confirmed: AIRFLOW__CORE__LOAD_EXAMPLES can be
flipped from 'false' to 'true' this way against the unescaped form,
affecting every container that inherits x-airflow-common (scheduler,
apiserver, dag-processor, triggerer).

## Why
Traced all current callers of RuntimeConfig (cli.py's `up` command,
App.run()): neither currently passes xcom_objectstorage_url, so this is
not reachable via any exposed CLI flag or API today -- flagging that
honestly rather than overstating current severity. Fixing regardless,
since the field is documented for multi-machine deployments and
inconsistent escaping in a generated infrastructure file is worth
closing before it's wired up to any less-trusted input source.

## Fix
Wraps the template slot in single quotes (matching the convention used
elsewhere in this file) and applies the same two escapes
_compose_path_scalar uses (YAML quote-doubling, Compose $ -> $$) before
substitution.

## Validation
- Confirmed the injection succeeds against the original code (existing
  key overwritten).
- Confirmed the fix neutralizes it (payload preserved as one literal
  string value instead of a new key).
- Confirmed legitimate values still render and round-trip correctly.
- Added a regression test, confirmed to fail against the original code
  and pass against the fix.
- Full test suite: 1165 passed, 6 skipped, 0 regressions.
- ruff check / format: clean.


kstonekuan (approved):
LGTM, merging. And thank you for scoping the severity honestly rather than upward; that is worth more than the fix.

Confirmed the reachability claim: nothing outside `_bundle.py` and the tests passes `xcom_objectstorage_url`, so no CLI flag or API surface reaches it today. Fixing it anyway is right, because the escaping being inconsistent with its neighbours in the same function is the actual defect, and the field exists for the deployment shape where the value stops being the operator's own.

Ran your payload plus five others against main and against the branch:

```
                                 main       branch
s3://bucket/xcom                 intact     intact
s3://bucket/$xcom                intact     intact
newline + block indent           INJECTED   intact
single quote break-out           INJECTED   intact
doubled quote attempt            INJECTED   intact
s3://${HOME}/xcom                intact     intact
newline + merge-key anchor       truncated  intact
comment then a sibling key       INJECTED   intact
```

Four shapes flip `AIRFLOW__CORE__LOAD_EXAMPLES` on main, not one. The merge-key one is worth noting separately: it did not inject, it silently truncated the value to `s3://foo` and dropped the rest, so an operator would have got a working-looking bundle pointing at the wrong store. On the branch all eight land as one literal scalar.

Pushed one fixup, `4330a48`, for two things.

Your inline escaping was byte-identical to `_compose_path_scalar`, which meant the rule existed twice. I extracted `_compose_single_quoted_scalar(value: str)` and made `_compose_path_scalar` delegate to it, so the next single-quoted slot has an obvious place to go and there is one line to change if Compose ever changes how it interpolates. That is a security-relevant rule, and two copies of it is how one of them gets missed.

Your test covers the YAML half. The Compose half was untested, so I added the `$` case:

```python
xcom_objectstorage_url="s3://bucket/${HOME}/xcom"
-> "s3://bucket/$${HOME}/xcom"
```

Without the doubling, the store URL silently becomes whatever the launch shell had in `HOME`, or empty. That is the quieter half of the same bug and it deserved its own assertion.

Also ran `ruff format`, which rewrapped one assert in the new test.

Gate clean, 1393 passed / 6 skipped from the repo root. One note for next time: `1165 passed` is `pytest tests` alone, which skips `packages/hflow-server/tests`. Plain `uv run pytest -q` from the root picks both up.

Third security find, and each one has been a real boundary rather than a scanner hit. If you want more of that kind of thing, #314 (ingest URI parsing scattered across CLI, server, and SDK) is the shape where these keep turning up.

kstonekuan:
Resolved the conflict and pushed as `106ec92`. Main renamed `_endpoint_environment_passthrough_lines` to `_environment_passthrough_lines(passthrough_environment_variables)` and renamed the test that covers it while this was open; neither touches the xcom escaping, so both sides kept their change. Revalidated on the merged result: gate clean, 1386 passed / 6 skipped, and all eight payloads still land as one literal scalar.

Raghav7-tech:
> LGTM, merging. And thank you for scoping the severity honestly rather than upward; that is worth more than the fix.
> 
> Confirmed the reachability claim: nothing outside `_bundle.py` and the tests passes `xcom_objectstorage_url`, so no CLI flag or API surface reaches it today. Fixing it anyway is right, because the escaping being inconsistent with its neighbours in the same function is the actual defect, and the field exists for the deployment shape where the value stops being the operator's own.
> 
> Ran your payload plus five others against main and against the branch:
> 
> ```
>                                  main       branch
> s3://bucket/xcom                 intact     intact
> s3://bucket/$xcom                intact     intact
> newline + block indent           INJECTED   intact
> single quote break-out           INJECTED   intact
> doubled quote attempt            INJECTED   intact
> s3://${HOME}/xcom                intact     intact
> newline + merge-key anchor       truncated  intact
> comment then a sibling key       INJECTED   intact
> ```
> 
> Four shapes flip `AIRFLOW__CORE__LOAD_EXAMPLES` on main, not one. The merge-key one is worth noting separately: it did not inject, it silently truncated the value to `s3://foo` and dropped the rest, so an operator would have got a working-looking bundle pointing at the wrong store. On the branch all eight land as one literal scalar.
> 
> Pushed one fixup, `4330a48`, for two things.
> 
> Your inline escaping was byte-identical to `_compose_path_scalar`, which meant the rule existed twice. I extracted `_compose_single_quoted_scalar(value: str)` and made `_compose_path_scalar` delegate to it, so the next single-quoted slot has an obvious place to go and there is one line to change if Compose ever changes how it interpolates. That is a security-relevant rule, and two copies of it is how one of them gets missed.
> 
> Your test covers the YAML half. The Compose half was untested, so I added the `$` case:
> 
> ```python
> xcom_objectstorage_url="s3://bucket/${HOME}/xcom"
> -> "s3://bucket/$${HOME}/xcom"
> ```
> 
> Without the doubling, the store URL silently becomes whatever the launch shell had in `HOME`, or empty. That is the quieter half of the same bug and it deserved its own assertion.
> 
> Also ran `ruff format`, which rewrapped one assert in the new test.
> 
> Gate clean, 1393 passed / 6 skipped from the repo root. One note for next time: `1165 passed` is `pytest tests` alone, which skips `packages/hflow-server/tests`. Plain `uv run pytest -q` from the root picks both up.
> 
> Third security find, and each one has been a real boundary rather than a scanner hit. If you want more of that kind of thing, #314 (ingest URI parsing scattered across CLI, server, and SDK) is the shape where these keep turning up.

Appreciate the deep dive, especially catching the merge-key truncation. Already looking at #314.

PR #358, opened 2026-09-02
refactor(video): serve both slice-header readers from one NAL walk

### LINKED ISSUE #355: The two slice-header walks still duplicate their loop and their Exp-Golomb decoder
## Current behavior

`count_h264_pictures` (`src/hflow/video.py:347`) and `scan_picture_coding_types` (`src/hflow/video.py:419`) walk the same NALs the same way. After #354 the shared body is twelve identical lines in each:

```python
nal_offsets_and_types = _annex_b_nal_offsets_and_types(stream)
for nal_index, (nal_start_offset, nal_type) in enumerate(nal_offsets_and_types):
    if nal_type not in _SLICE_HEADER_NAL_TYPES:
        continue
    nal_end_offset = (...)
    nal_header_offset = _nal_header_offset(stream, nal_start_offset)
    slice_payload = stream[nal_header_offset + 1 : nal_end_offset]
    rbsp = _unescape_ebsp_head(slice_payload, _SLICE_HEADER_HEAD_BYTES)
    # ... decode, and on failure re-unescape the whole payload and decode again
```

They then diverge into two Exp-Golomb decoders that must agree:

- `_decode_first_mb_in_slice` (`:289`) raises, with two distinct messages that `hflow doctor` and the AUD repair path depend on.
- `_decode_unsigned_exp_golomb_at` (`:319`) plus `_slice_header_fields` (`:380`) return `None`, and the scan turns that into one message.

The head-then-fallback logic #354 introduced is written out in both, so the duplication grew rather than shrank.

`count_h264_pictures(stream)` is exactly `scan_picture_coding_types(stream).picture_count`. Verified across sixteen inputs: real encodes at 320x240 and 1280x720, bframes=0 and bframes=3, multi-slice, synthetic streams with emulation-prevention triples straddling the head boundary, and `first_mb_in_slice` from 0 to 139263. The two agree on every one.

## Why this is worth doing

Two decoders over the same bitstream field, kept in step by nothing but care. They already differ in failure style (raise versus `None`), which is the part that makes them look like separate concerns when they are not. A future change to the head budget, the fallback, or the NAL filter has to be made twice and can be made inconsistently, and the inconsistency would be silent: both still return a number, just different ones.

This was DoD 4 on #346 and did not land with #354, which is on me for putting a "while in there" item into a definition of done. It is its own change.

## What to build

One walk that yields `(first_mb_in_slice, slice_type)` per slice NAL, with the two public functions as thin consumers. `count_h264_pictures` keeps its two specific messages, so the shared piece has to let a caller choose how a short or truncated header is reported: returning an explicit outcome the callers translate is one way, passing the messages in is another. Say which you picked and why.

## Definition of done

1. One NAL walk and one Exp-Golomb decoder serve both functions.
2. `count_h264_pictures` raises byte-identical messages for a header with no complete `first_mb_in_slice` and for a truncated one. The existing tests cover both; they must pass untouched.
3. `scan_picture_coding_types` raises its existing single message for the same inputs.
4. The head-then-fallback path exists once.
5. Results are unchanged on canonical, B-frame, multi-slice, and escape-heavy streams. Adding a case to `tests/test_video.py` that asserts `scan_picture_coding_types(s).picture_count == count_h264_pictures(s)` pins the equivalence directly; #354 added one, extend it rather than duplicating it.
6. No measurable regression against the #354 timings. `hflow doctor` and the remux both sit on this path.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
  tests/test_processing_regressions.py
uv run pytest -q
```

## Notes

Not a good first issue: the error-message compatibility is the whole difficulty, and getting it wrong changes `hflow doctor` output on malformed input.

Sagar-024:
Serves both slice-header readers from one NAL walk

count_h264_pictures and scan_picture_coding_types each carried their own NAL loop, their own head-then-fallback unescape, and between them two Exp-Golomb decoders over the same bitstream field: one raising, one returning None. The duplication grew with #354, and a future change to the head budget or the fallback had to be made twice and could be made inconsistently, silently: both readers still return a number, just different ones.

This is the DoD 4 follow-up from #346 that #354 did not land.

One walk now serves both. _iter_slice_headers yields one _SliceHeader per VCL NAL with the two fields decoded per field: a field is None when its bits ran out, and malformed carries the fully unescaped payload for exactly those NALs. The head-then-fallback pair exists once, in the walk. Each consumer owns its own failure reporting, which is how the three messages stay byte-identical: count_h264_pictures translates the failure through _first_mb_failure_message into the pinned "no complete" and "truncates" strings hflow doctor displays, and the scan keeps its single classification-refusal message.

Per-field failure is the part that preserves behavior: the counter reads only first_mb_in_slice, so a header whose slice_type is truncated still counts, exactly as before. _decode_first_mb_in_slice, the second decoder, is gone.

Measured on the #354 bench shape (300 frames, 640x480, bframes=3, 10 iterations, median of the warm 5): 2.59 ms on main, 2.36 ms on this branch, so the shared walk costs nothing. Equivalence on canonical, B-frame, multi-slice, and escape-heavy streams is pinned in test_scan_and_count_agree_on_multi_slice_and_escape_heavy_streams, extending the #354 test per the issue.

Closes #355.


kstonekuan (approved):
LGTM, merging. This is the change #354 should have carried, and the hard part landed correctly.

Equivalence first, since the readers sit on the canonical write path. Same sixteen inputs I used on #354, main against this branch:

```
real encodes:  bframes=0, bframes=3, multi-slice, 320x240 and 1280x720
synthetic:     escape triples straddling the head boundary (0,1,2,4,6,8,16)
synthetic:     first_mb_in_slice = 0, 1, 99, 8100, 139263
identical
```

The three pinned messages are byte-identical too, including through `ensure_access_unit_delimiter`:

```
all-zero payload            count: no complete first_mb_in_slice value
terminating bit, no suffix  count: truncates its first_mb_in_slice value
valid first_mb, bad type    count: 1          scan: cannot be classified
empty payload               count: no complete first_mb_in_slice value
identical on both sides
```

`_first_mb_failure_message` re-deriving the kind from the RBSP rather than threading a failure enum out of the walk is the right call. The walk stays ignorant of message text, and the helper is only reachable for a payload whose first field already failed, so the two kinds really are the whole space.

Your escape-heavy case is the test I said was missing on #354. A `first_mb_in_slice` of `2**70` whose escapes push the header past a 16-byte head is only decodable through the fallback, and removing the fallback fails exactly that test:

```
fallback to the full unescape removed  ->  test_scan_and_count_agree_on_multi_slice_and_escape_heavy_streams
both failure messages collapsed        ->  test_first_mb_failure_message_names_both_failure_kinds,
                                           test_ensure_aud_rejects_slice_header_without_complete_first_mb_value
```

Pushed one fixup, `d165b99`. Per-field failure is the subtlest thing here and you flagged it yourself in the description, but nothing held it: I changed the counter to refuse a truncated `slice_type` too, and all 28 tests still passed. That is the edit someone makes while tidying two adjacent `is None` checks into one, and it would quietly start rejecting streams `hflow doctor` counts today. There is now a test for it, and that mutation fails.

One correction to the description. The shared walk is not free, and it is the counter rather than the scan that pays, because it now decodes a `slice_type` it never reads. Medians over eleven runs, warm:

```
                       main      branch
 300 frames  count     1.43 ms   1.76 ms
 300 frames  scan      1.68 ms   1.78 ms
3000 frames  count    14.54 ms  17.75 ms   (+22%)
3000 frames  scan     16.93 ms  18.08 ms   (+7%)
```

Your 300-frame scan number is the one that looks flat; the counter at 3000 frames does not. It is still the right trade: 3 ms on a 12 MB stream against the 430 ms #354 took out, for one walk, one decoder and one fallback instead of two of each. Worth stating accurately so whoever reads the next benchmark knows where it went, and if the counter's cost ever matters, deciding `slice_type` lazily is the seam.

Gate clean: 1389 passed / 6 skipped with main merged in.

Closes #355, which closes out the DoD I split badly on #346. Three PRs, but the end state is right.

Sagar-024:
@kstonekuan thanks for the merge and the detailed review. Three takeaways:

* The mutation test you added for the per-field failure is exactly the guard I should have written, and I will carry that pattern forward so any behavior subtle enough to flag in the description gets a mutation test to prevent the exact tidying edit you caught.
* Thanks for the honest perf correction showing the 22 percent counter cost at 3000 frames and noting the lazy decode seam, which taught me to be more precise about perf claims and measure both functions at both sizes to avoid overstating the cost.
* The #346 and #355 arc is closed, so let me know if you want me to specifically test something else in this module right now or if I should grab the next unclaimed issue from the queue.

PR #334, opened 2026-08-31
Transcode egocentric landing episodes to H.264 directly

rakesh0x:
Closes #291

## What and why

`examples/egocentric` landed its episodes through `hflow.testing.write_video_episode`, which decodes each source and re-encodes every frame as JPEG; the canonical transform then encoded those JPEGs into in-band H.264. For the Egocentric corpora, whose clips are H.265, the media went **HEVC → JPEG → H.264**: two lossy steps where one would do, and a landing file ~2.3x the size the canonical stage actually needs (#287's baseline).

`prepare.py` now transcodes each excerpt **once, directly to Annex B H.264**, and the canonical transform passes the media bytes through unmodified:

- One ffmpeg invocation per episode applies `fps`/`scale`/`pad`, plants faults in the same graph, and encodes `libx264 -preset medium -crf 23` with `-x264-params keyint=<fps>:min-keyint=<fps>:scenecut=0:bframes=0:repeat-headers=1:aud=1` — so every message is one AUD-delimited access unit and every keyframe carries SPS/PPS, which is exactly what `transform.py` needs to pass media through.
- The stream is cut with `hflow.video.split_annex_b_stream` and written as `foxglove.CompressedVideo` (`format="h264"`) on `/head_camera/compressed`.
- The example also now records an episode metadata record plus a `source-provenance/v1` record (converter version, ffmpeg version, `hf://` source URI) that the JPEG fixture path never carried.
- `hflow/testing.py` is untouched, and `examples/egocentric` no longer imports `hflow.testing`.

## Fault injection

Substituting frames after decoding is no longer possible without the JPEG step, so both faults are planted **inside the single transcode**:

- **Blackout**: `drawbox=x=0:y=0:w=iw:h=ih:color=black:t=fill:enable=between(n\,<first>\,<last>)` after the base `fps,scale,pad` chain.
- **Freeze**: the filtered source is `split=2` into a main and a replacement branch, and `freezeframes=first=<first>:last=<last>:replace=<first>` swaps the interval for its own first frame within the same decode/filter/encode invocation.

Both still land in `camera_frame_stats` with the same shape of result (black-frame percentage over the blackout window, freeze duration over the freeze window), verified by test.

## Evidence

`tests/test_egocentric_prepare.py` builds a moving HEVC source, lands it through the new path, and asserts that:

- the landing file carries only `foxglove.CompressedVideo`/`h264` (no `sensor_msgs/msg/CompressedImage`);
[...]

PR #359, opened 2026-09-02
Catalog UI shutdown must not auto-install the ui extension to call stop_ui_server()

### LINKED ISSUE #336: Catalog UI shutdown installs the ui extension from the network to call stop_ui_server()

rakesh0x:
Closes #336. Supersedes #347, which the 1-open-PR bot closed while #334 was still open (#334 has since merged).

## What and why

`serve_catalog_ui` sets `server_started = True` as soon as `_start_duckdb_ui_server` returns, and shutdown calls `stop_ui_server()` on that flag alone. `stop_ui_server()` is a function the `ui` extension provides — calling it on a connection that never loaded the extension makes DuckDB **auto-install the extension from the network at shutdown**, to run what is a no-op: there is no server to stop if the extension was never loaded.

Measured with a cold extension directory (`extension_directory` pointed at an empty dir):

```
before call: ('ui', False, False)
stop_ui_server ok
elapsed: 3.16s
after call: ('ui', True, True)
```

That network fetch inside a teardown that another thread waits on with a fixed timeout is almost certainly the flaky `test_catalog_ui` timing failure three contributors reported on 2026-08-31 (#326, #327, #328) — `tests/test_catalog_ui.py` stubs `_start_duckdb_ui_server`, so the extension never loads, but `server_started` is still `True` and teardown calls `stop_ui_server()` anyway.

## The fix

Guard the call on the extension actually being loaded on that connection:

```python
def _ui_extension_is_loaded(catalog_connection: duckdb.DuckDBPyConnection) -> bool:
    row = catalog_connection.execute(
        "SELECT count(*) FROM duckdb_extensions() WHERE extension_name = 'ui' AND loaded"
    ).fetchone()
    return row is not None and int(row[0]) > 0
```

and gate `stop_ui_server()` on `server_started and _ui_extension_is_loaded(...)`. Production is unchanged: the real starter runs `INSTALL ui` / `LOAD ui` before the server starts, so the loaded check passes and the stop call still happens. A connection that never loaded the extension now closes directly.

## DoD coverage

1. **Shutdown does not call `stop_ui_server()` when `ui` is not loaded** — `test_catalog_ui_shutdown_does_not_stop_a_server_the_ui_extension_never_loaded` pins it by recording every statement issued on the connection (per the issue: recording statements is enough; DuckDB's install behavior is not the point).
2. **A session whose `ui` extension is loaded still stops its server** — `test_catalog_ui_shutdown_stops_the_server_when_the_ui_extension_is_loaded` pins the stop call on the loaded branch, and that the connection still closes.
3. Existing catalog UI startup, port-conflict, and refresh behavior is untouched — the full existing suite passes unchanged.

One implementation note on the tests: the recording proxy **records but does not delegate** `CALL stop_ui_server()` to the real connection. Delegating it on a connection without the loaded extension would itself trigger the network auto-install this fix removes — I hit exactly that running the suite with a fresh `HOME` (the CI condition): the loaded-branch test failed on the 2s join timeout because DuckDB was downloading the extension mid-teardown. Not delegating keeps the tests hermetic and offline, which is the point of the issue.

## Validation

```
uv sync --locked
uv run ruff check        # All checks passed!
uv run ruff format       # (check) clean
uv run ty check          # only the pre-existing cv2 optional-import diagnostic
uv run pytest -q tests/test_catalog_ui.py   # 22 passed
```

`tests/test_catalog_ui.py` passes 5/5 runs under both a warm `HOME` and a fresh `HOME` (empty extension cache, i.e. the CI runner condition).

Found by me while working on #291; the reachability trace (stubbed starter → `server_started = True` → network fetch in teardown) is from the issue.

### DIFF
--- a/src/hflow/catalog_ui.py
+++ b/src/hflow/catalog_ui.py
@@ -86,11 +86,22 @@ def _start_duckdb_ui_server(catalog_connection: duckdb.DuckDBPyConnection, port:
         ) from error
 
 
+def _ui_extension_is_loaded(catalog_connection: duckdb.DuckDBPyConnection) -> bool:
+    row = catalog_connection.execute(
+        "SELECT count(*) FROM duckdb_extensions() WHERE extension_name = 'ui' AND loaded"
+    ).fetchone()
+    return row is not None and int(row[0]) > 0
+
+
 def _stop_duckdb_ui_server(
     catalog_connection: duckdb.DuckDBPyConnection, server_started: bool
 ) -> None:
     try:
-        if server_started:
+        # stop_ui_server only exists once the ui extension is loaded. Calling it
+        # without the extension makes DuckDB auto-install the extension from the
+        # network -- seconds on a slow connection, and pointless: there is no
+        # server to stop if the extension was never loaded.
+        if server_started and _ui_extension_is_loaded(catalog_connection):
             catalog_connection.execute("CALL stop_ui_server()")
     except duckdb.Error:
         # The process is already leaving. A server that stopped independently
--- a/tests/test_catalog_ui.py
+++ b/tests/test_catalog_ui.py
@@ -123,6 +123,109 @@ def test_catalog_ui_refuses_a_port_of_the_wrong_type(
         replace(settings, port=port)
 
 
+class _RecordingCatalogConnection:
+    """DuckDB connection proxy that records every statement it executes.
+
+    ``CALL stop_ui_server()`` is recorded but not delegated to the inner
+    connection: on a connection without the loaded ui extension, executing it
+    would make DuckDB auto-install the extension from the network -- the exact
+    behavior #336 removes. The shutdown tests assert which statements the
+    shutdown path issues, not what DuckDB does with them.
+    """
+
+    def __init__(self, inner: duckdb.DuckDBPyConnection) -> None:
+        self._inner = inner
+        self.executed_statements: list[str] = []
+        self.closed = False
+
+    def execute(self, sql: str, *parameters: object) -> object:
+        self.executed_statements.append(sql)
+        if sql == "CALL stop_ui_server()":
+            return None
+        return self._inner.execute(sql, *parameters)
+
+    def close(self) -> None:
+        self.closed = True
+        self._inner.close()
+
+    def __getattr__(self, name: str) -> object:
+        return getattr(self._inner, name)
+
+
+def _serve_until_shutdown(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+    ui_extension_loaded: bool,
+) -> _RecordingCatalogConnection:
+    shutdown_event = Event()
+    server_started_event = Event()
+    connections: list[_RecordingCatalogConnection] = []
+    background_failures: list[BaseException] = []
+
+    real_open_catalog_connection = catalog_ui.open_catalog_connection
+
+    def record_server_start(catalog_connection: duckdb.DuckDBPyConnection, port: int) -> None:
+        assert port == catalog_ui.DEFAULT_CATALOG_UI_PORT
+        server_started_event.set()
+
+    def open_recorded_connection(catalog_root: Path) -> _RecordingCatalogConnection:
+        connection = _RecordingCatalogConnection(real_open_catalog_connection(catalog_root))
+        connections.append(connection)
+        return connection
+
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging. Nice to see this land as its own PR: you found it, and it turned out to be the flake three other people were hitting.

Both directions of the guard are pinned, which is more than I expected:

```
loaded-extension guard removed (back to main)  -> ..._does_not_stop_a_server_the_ui_extension_never_loaded
guard inverted                                 -> that one AND ..._stops_the_server_when_the_ui_extension_is_loaded
```

The inverted case failing both is what says the tests describe the rule rather than one side of it.

The part I most wanted to confirm is that the flake is gone. Three runs against a completely fresh `HOME`, so an empty extension cache, which is the CI condition:

```
22 passed in 0.51s
22 passed in 0.33s
22 passed in 0.35s
```

Nothing downloaded, no 2-second join timeout. On #334 I traced the flaky `test_catalog_ui` on #326, #327 and #328 to this and could only say "almost certainly"; a third of a second on a cold cache is the answer.

Your note about the recording proxy not delegating `CALL stop_ui_server()` is the right call and worth having written down. Delegating it would have made the test harness perform the exact network install the fix removes, and you found that by running with a fresh `HOME` rather than by reasoning about it. That is the same measurement that makes the fix credible.

Gate clean with main merged in: `ruff check`, `ruff format --check`, `ty check`, 1391 passed / 6 skipped.

One note on that: your `ty check` showed a cv2 optional-import diagnostic because `uv sync --locked` without `--all-extras` leaves the vision extra out. With `--all-extras`, the documented form in CONTRIBUTING, it is clean. Worth using that so a real diagnostic does not get filed under "pre-existing".

Closes #336, and that is #291's side quest fully closed out.

PR #362, opened 2026-09-02
fix(egosuite): match saved labels by source provenance

### LINKED ISSUE #310: Match EgoSuite saved labels through HFlow source provenance
## Current behavior

The saved-label loader resolves each source_path, requires source_episode to equal that path's stem, and groups records by source_episode at [examples/egosuite_evaluation/evaluate.py:158-223](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/egosuite_evaluation/evaluate.py#L158-L223). Its duplicate key is also only (source_episode, frame_index).

The pipeline removes .canonical.mcap from the canonical filename and uses the remaining stem as its lookup key at [examples/egosuite_evaluation/pipeline.py:148-165](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/egosuite_evaluation/pipeline.py#L148-L165).

HFlow supports recordings with the same basename in different source directories. It gives them distinct source identities and run directories at [src/hflow/app.py:1319-1344](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/app.py#L1319-L1344) and [src/hflow/app.py:1833-1847](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/app.py#L1833-L1847), while each canonical filename keeps the source stem. The canonical provenance records that source identity as source_uri at [src/hflow/app.py:1898-1903](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/app.py#L1898-L1903).

Two sources such as run-a/episode.mcap and run-b/episode.mcap therefore collide in the label report. Overlapping frame indices are rejected as duplicates. Disjoint frame indices are merged and can be offered to either canonical episode. An unrelated canonical recording with the same basename can also consume the saved labels because source_path is stored but never used for the pipeline match.

Verified against main at fbfd553. Report envelope validation is tracked separately in #307.

## Definition of done

1. A saved label report carries or derives a source identity that follows HFlow's existing App.source_identity and canonical source_uri semantics.
2. Two source recordings with the same basename remain separate throughout report loading and pipeline lookup.
3. The pipeline matches labels against the canonical episode's provenance. A missing or mismatched source identity fails before frame extraction or model requests.
4. Absolute host paths aren't the sole durable identity. The report remains usable across the host/container path representations HFlow already normalizes.
5. Source-level sampling and duplicate detection use the same identity rule where a basename currently participates.
6. Existing unique-basename reports have an explicit compatibility path, or the report schema version is bumped and the error explains regeneration.
7. Add behavioral coverage for two different sources sharing one filename, plus a same-name unrelated canonical episode.
8. Update the EgoSuite how-to with the identity and portability contract.

## Non-goals

- Changing canonical MCAP bytes or adding new provenance fields
- Moving label records into the HFlow catalog
- Generalizing the report into a project-wide annotation format
- Changing projection geometry, camera selection, or model judgment

## Validation

~~~bash
  examples/egosuite_evaluation
uv run pytest -q
~~~

Galabavamsi:
## Summary

- record HFlow `source_uri` identity in new EgoSuite label reports
- match saved labels to canonical episode provenance instead of basenames
- preserve unambiguous legacy reports and reject ambiguous ones with regeneration guidance
- use source identity for deterministic episode, frame, and stratified sampling

## Why

Recordings in different directories can share a basename. Basename-keyed reports merged or rejected those labels and could apply them to an unrelated canonical episode. This keeps report matching aligned with `App.source_identity()` and canonical `source_uri` semantics.

Closes #310.

## Validation

- `uv run --locked --project examples/egosuite_evaluation ruff check --fix examples/egosuite_evaluation` — passed
- `uv run --locked --project examples/egosuite_evaluation ruff format examples/egosuite_evaluation` — 6 files unchanged
- `uv run --locked --project examples/egosuite_evaluation pytest -q examples/egosuite_evaluation/tests` — 47 passed
- `uv run pytest -q` — 1391 passed, 6 skipped
- fresh pre-reopen full suite — 1391 passed, 6 skipped

## Checklist

### DIFF
--- a/docs/how-to/run-egosuite-hand-evaluation.md
+++ b/docs/how-to/run-egosuite-hand-evaluation.md
@@ -148,10 +148,26 @@ export EGOSUITE_LABEL_MANIFEST='data/egosuite-evaluation/labels/natural-1000.jso
 ```
 
 The pipeline validates every manifest record, selects its declared source
-frame indices for each canonical episode, and includes the manifest digest in
-the HFlow check version. Missing or malformed model content is retained as a
+frame indices only when the report's `source_uri` matches the canonical
+episode's HFlow provenance, and includes the manifest digest in the HFlow
+check version. Missing or mismatched provenance fails before frame extraction
+or a model request. Missing or malformed model content is retained as a
 `valid=false` observation so provider output failures lower end-to-end
 accuracy instead of discarding the episode's evidence.
+
+Current label reports record the same source identity returned by
+`App.source_identity()`. Set `HFLOW_DATA_ROOT` consistently when creating the
+report and running the pipeline. When the dataset lives below that root, the
+identity is root-relative and remains stable across equivalent host and
+container mount paths. For the layout in this guide, use:
+
+```bash
+export HFLOW_DATA_ROOT='data/egosuite-evaluation'
+```
+
+Reports created before `source_uri` was added remain usable when each
+`source_episode` basename refers to exactly one source path. An older report
+that uses one basename for multiple paths is ambiguous and must be regenerated.
 
 Use this entrypoint when the judgment belongs in an HFlow episode pipeline.
 Use `evaluate.py` below when comparing models over a declared dataset slice;
@@ -172,8 +188,9 @@ uv run --project examples/egosuite_evaluation \
 
 The terminal table reports the number of frames labeled 0, 1, or 2. The JSON
 file records the projected joint count for each hand, the resulting hand count,
-the exact source frame index, and any pose-quality reasons. Run `labels` first
-on a new corpus to inspect class balance before interpreting model accuracy.
+the exact source frame index, HFlow's canonical `source_uri`, and any
+pose-quality reasons. Run `labels` first on a new corpus to inspect class
+balance before interpreting model accuracy.
 
 For a naturally distributed random sample, select episodes and then select
 temporally separated frames within each episode:
--- a/examples/egosuite_evaluation/evaluate.py
+++ b/examples/egosuite_evaluation/evaluate.py
@@ -68,6 +68,7 @@ SCHEMA_VERSION = 1
 EXPECTED_HAND_JOINT_COUNT = 21
 DEFAULT_RUNS_DIRECTORY = Path("data/egosuite-evaluation/runs")
 DEFAULT_LABELS_DIRECTORY = Path("data/egosuite-evaluation/labels")
+DEFAULT_HFLOW_DATA_ROOT = Path("data/egosuite-evaluation/hflow")
 INSPECT_LOGS_DIRECTORY_NAME = "logs"
 RUN_METADATA_FILE_NAME = "run.json"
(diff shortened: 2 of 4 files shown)


kstonekuan (approved):
LGTM, merging.

Every guard bites, which for an eight-item definition of done is the thing I most wanted to check:

```
pipeline matches on the basename again      -> matches_same_named_sources, rejects_missing_or_unrelated
pipeline stops requiring source_uri         -> rejects_missing_or_unrelated_canonical_provenance
legacy one-basename-many-paths accepted     -> rejects_one_basename_for_multiple_sources
duplicate key back to the episode name      -> matches_same_named_sources_by_canonical_provenance
episode sampling ranks by basename again    -> episode_sampling_ranks_same_named_sources_by_source_uri
report stops recording source_uri           -> label_report_records_hflow_source_identity
```

DoD 4 is the one I went looking at hardest, because "not the sole durable identity" is easy to claim and hard to hold. Checked what the report actually records:

```
under the data root   -> 'episodes-in/episode.mcap'
outside it            -> '/tmp/.../run-a/episode.mcap'
two same-basename sources get distinct identities: True
same relative position, different root prefix: identical
```

So the report inherits HFlow's own portability envelope rather than inventing one: root-relative and mount-independent where `source_identity` normalizes, absolute only where HFlow itself would be absolute. That is DoD 1 and DoD 4 answered by the same decision, which is why it holds. Telling the reader to `export HFLOW_DATA_ROOT='data/egosuite-evaluation'` in the how-to is the part that turns the guarantee into something a user actually gets, and keeping `DEFAULT_HFLOW_DATA_ROOT` at `<root>/hflow` matches `build_ai_evaluation`, so I would not change it.

Bumping the check contract to `-v2` is right and easy to forget: the matching semantics moved, so a cached result from `-v1` must not satisfy the new path. That is the same discipline as DoD 6's schema note.

The legacy compatibility path is better than the issue asked for. Accepting old reports whose basenames are unambiguous, and refusing exactly the ones where a basename covers two paths, means nobody regenerates a report that was never ambiguous, and nobody keeps one that silently was. Detecting a mixed report (some records with `source_uri`, some without) rather than guessing per record closes the other half of that.

Gate clean: 47 passed on the example project, 1409 passed / 6 skipped at the root with current main merged in.

One observation, no action. `_source_uri_by_path` builds a fresh `hflow.App` each time it is called, and `run_evaluation` plus both argument paths call it separately. It is cheap and side-effect-free, and I confirmed it leaves nothing on disk, so this is only worth knowing if identity resolution ever grows teeth.

Thanks also for #365 — that is the corpus-testing invitation working exactly as intended, and it is accepted.

Issue #303, opened 2026-08-31
Make multi-episode LeRobot imports resumable after interruption


kstonekuan:
## Current behavior

`import_lerobot_dataset()` builds the selected episode indexes and converts them in one loop at [`src/hflow/importers/lerobot.py:574-595`](https://github.com/Hebbian-Robotics/hflow/blob/1f5c7c8f59265110cad7c8ffe82a1951a4411bba/src/hflow/importers/lerobot.py#L574-L595). The prepared manifest is only written after the whole loop succeeds at [`src/hflow/importers/lerobot.py:597-630`](https://github.com/Hebbian-Robotics/hflow/blob/1f5c7c8f59265110cad7c8ffe82a1951a4411bba/src/hflow/importers/lerobot.py#L597-L630).

If a later episode fails, the earlier canonical MCAP files remain. Running the same import again doesn't recognize them as completed work, so it converts every selected episode again. A large import can't resume at the episode boundary.

Each canonical MCAP is already published atomically by `GroupedMcapWriter`. The missing piece is a durable definition of which output belongs to this exact import.

## Definition of done

- Define a completed episode identity from the resolved dataset commit, source episode, selected camera keys, converter version, and any canonical behavior that affects the bytes.
- Re-running an identical import reuses completed matching episodes.
- Partial, unreadable, or identity-mismatched outputs are never treated as completed work.
- A retry after a mid-batch failure finishes the remaining episodes and publishes `prepared-manifest.json` only when the selected set is complete.
- Episodes completed before the interruption keep the same bytes across the retry.
- The final manifest records enough episode-level identity to prove which selected episodes completed. Bump its schema version if its shape changes.
- Add a behavioral regression test that interrupts a multi-episode import, retries it, and observes both reuse of the first committed episode and a complete final manifest.
- Update `docs/how-to/import-lerobot-v3.md` with the retry contract.

## Non-goals

- Parallel episode conversion
- Airflow jobs or dynamic task mapping
- Following a mutable branch for new episodes
- Supporting more LeRobot feature layouts

Cache isolation across source revisions is tracked in #294.

## Validation

```bash
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest tests/test_lerobot_converter.py -q
uv run pytest -q
```

PR #363, opened 2026-09-02
fix(runtime): page task instances so a wide fan-out is complete

### LINKED ISSUE #311: Paginate Airflow task instances before building run graphs

chiruu12:
Closes #311.

`task_instances` made one GET to `/taskInstances` with no `limit` or `offset`. Airflow's endpoint defaults to 50 per page, so any run whose fan-out is wider than 50 returned the first page and the caller treated it as the whole run. `ingest_dag_topology` then built a run graph off a truncated list, and the mapped-task summary undercounted without any sign that it had.

The method now pages until the run is exhausted. `dag_runs` stays caller-capped as it already documents; this one's contract is complete run detail, so a short read is not an acceptable answer.

Three independent stops, so a server that disagrees with its own metadata cannot spin: a page that added nothing new, a page shorter than the one requested, and the reported `total_entries` being reached. Entries are keyed by `(task_id, map_index)` so overlapping pages do not double count. A failure on any page raises `AirflowClientError` rather than returning a short list as though it were complete.

Page size is 100 rather than the default 50. Asking for more per request costs nothing and halves the round trips.

## Tests

`tests/test_runtime_client.py`
- every page of a 251-instance fan-out is returned, and the mapped indices are complete
- pages are requested in turn with the expected `limit` and `offset`
- a failure on a later page raises instead of returning the pages already collected

`packages/hflow-server/tests/test_server_run_graph.py`
- the run-graph endpoint totals a fan-out wider than one Airflow page. `task_instances` is left real here and the HTTP layer under it is stubbed instead, so this fails if the client ever stops paging.

All four fail on main and pass here. Reverting only the paging loop and keeping the tests gives 4 failed, 53 passed.

Local run: 269 passed across `tests/test_runtime_client.py` and `packages/hflow-server/tests`, plus 1167 passed on the root suite. `ruff check` and `ruff format --check` are clean.

## Rebase note

This was written against `e6009fc`, before task instances became typed. It is rebased onto current main and now returns `list[AirflowTaskInstance]` through `_parse_task_instance`, so the paging loop composes with the typed model rather than bypassing it.

### DIFF
--- a/packages/hflow-server/tests/test_server_run_graph.py
+++ b/packages/hflow-server/tests/test_server_run_graph.py
@@ -5,6 +5,7 @@ Every Airflow call is stubbed at the AirflowClient method level (the idiom of
 is a really rendered one so the dag ids are the real derived ones.
 """
 
+import urllib.parse
 from pathlib import Path
 from typing import Any
 
@@ -665,3 +666,77 @@ def test_run_graph_over_a_remote_runtime_derives_the_stage_dag_ids(
         "kitchen_labels",
         "kitchen_media",
     ]
+
+
+def test_run_graph_totals_a_fan_out_wider_than_one_airflow_page(
+    bundle_api: TestClient, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """The whole point of paging the client: the graph counts every batch.
+
+    ``task_instances`` is left real here and the HTTP layer under it is
+    stubbed instead, so this fails if the client ever stops paging.
+    """
+    fan_out = 250
+    # Raw JSON, not AirflowTaskInstance: the stub sits under the client, so it
+    # returns what Airflow returns and lets the client do its own parsing.
+    every_instance: list[dict[str, Any]] = [
+        {"task_id": "plan", "state": "success", "map_index": -1, "try_number": 1}
+    ] + [
+        {
+            "task_id": "process_batch",
+            "state": "success" if index % 2 else "failed",
+            "map_index": index,
+            "try_number": 1,
+        }
+        for index in range(fan_out)
+    ]
+    monkeypatch.setattr(
+        AirflowClient,
+        "dag_run",
+        lambda self, dag_id, dag_run_id: _dag_run(
+            {
+                "dag_run_id": MASTER_RUN_ID,
+                "state": "running",
+                "start_date": MASTER_STARTED_AT,
+            }
+        ),
+    )
+    monkeypatch.setattr(
+        AirflowClient,
+        "dag_runs",
+        lambda self, dag_id, *, limit=100, order_by=None: (
+            [
+                _dag_run(
+                    {
+                        "dag_run_id": "sync__new",
+                        "state": "running",
+                        "start_date": "2026-08-21T10:00:05+00:00",
+                    }
+                )
+            ]
+            if dag_id == "demo_pipeline_sync"
+            else []
+        ),
+    )
+
+    def _paged(self: AirflowClient, method: str, path: str, payload: Any = None) -> dict[str, Any]:
+        request_path, _, request_query = path.partition("?")
+        assert request_path.endswith("/taskInstances")
+        query = urllib.parse.parse_qs(request_query)
+        limit = int(query["limit"][0])
+        offset = int(query["offset"][0])
+        instances = every_instance if "demo_pipeline_sync" in request_path else []
+        return {
+            "task_instances": instances[offset : offset + limit],
+            "total_entries": len(instances),
+        }
+
+    monkeypatch.setattr(AirflowClient, "_authenticated", _paged)
+
+    payload = bundle_api.get(f"/api/v1/runtime/runs/{MASTER_RUN_ID}/graph").json()
+    sync_stage = next(stage for stage in payload["stages"] if stage["stage"] == "sync")
+    assert sync_stage["mapped_summary"] == {
+        "task_id": "process_batch",
+        "total": fan_out,
+        "by_state": {"failed": fan_out // 2, "success": fan_out // 2},
+    }
--- a/src/hflow/runtime/_client.py
+++ b/src/hflow/runtime/_client.py
@@ -443,17 +443,25 @@ class AirflowClient:
                 seen.add(identity)
                 collected.append(parsed)
                 added += 1
-            # Three independent stops, because a server that disagrees with
-            # its own metadata must not be able to spin here: a page that
-            # added nothing new, a page shorter than the one we asked for,
-            # and the reported total being reached.
-            if added == 0 or len(page) < _TASK_INSTANCE_PAGE_SIZE:
-                break
             offset += len(page)
             total_entries = response.get("total_entries")
-            # Counted against what was kept, not what was asked for: overlapping
-            # pages advance the offset past entries we have not seen yet.
-            if isinstance(total_entries, int) and len(collected) >= total_entries:
+            # A page that added nothing new is the guard against a server that
+            # disagrees with its own metadata: it stops the loop whatever the
+            # other two say.
+            if added == 0:
+                break
+            if isinstance(total_entries, int):
+                # Counted against what was kept, not what was asked for:
+                # overlapping pages advance the offset past entries we have not
+                # seen yet.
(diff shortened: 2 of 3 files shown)


kstonekuan (changes requested):
The paging itself is right, and the three-stop reasoning is the part I would have asked for: `added == 0` is a real loop guard, keying on `(task_id, map_index)` handles overlap, counting `total_entries` against what was kept rather than the offset is a subtlety most people miss, and raising instead of returning a short list is DoD 3 exactly. One thing to fix before this lands.

**The short-page stop truncates on any deployment whose `api.maximum_page_limit` is below 100.** Airflow clamps the `limit` query parameter to that setting and does not error, so a server configured lower serves shorter pages than you asked for, and `len(page) < _TASK_INSTANCE_PAGE_SIZE` reads that as "run exhausted". A 251-instance fan-out, server clamping at the value shown:

```
maximum_page_limit=100   3 requests   returned 251 of 251   COMPLETE
maximum_page_limit= 50   1 request    returned  50 of 251   TRUNCATED
maximum_page_limit= 25   1 request    returned  25 of 251   TRUNCATED
```

One request, then a silent partial list reported as complete. That is the defect #311 exists to remove, reintroduced under a config change rather than a fan-out size, and it fails DoD 1 and DoD 3 on that deployment. The default for `maximum_page_limit` is 100, which is exactly what you request, so it passes on a stock Airflow and breaks the moment an operator lowers it. Your stub honours `limit` verbatim, which is why no test sees this.

The information to distinguish the two cases is already in the response you read. When `total_entries` is present and has not been reached, a short page means the server clamped, not that the run ended, so the offset should advance by what was served and the loop continue. The short-page rule is only a safe terminator when `total_entries` is absent. Your `added == 0` stop is what keeps that from spinning, so the loop guard does not weaken.

Worth a test with a stub that clamps below the requested limit, since that is the shape no current test can produce.

Two things I checked that are fine. An empty run still returns `[]` rather than raising, because `{"task_instances": [], "total_entries": 0}` hits the empty-page break before the missing-list refusal, so the run graph does not start answering 502 for a run with no tasks. And `dag_runs` is untouched and still caller-capped, with a test saying so.

Gate is clean on the merged result otherwise: `ruff check`, `ruff format --check`, `ty check`, 1417 passed / 6 skipped.

chiruu12:
Fixed. The short-page stop now only fires when `total_entries` is absent.

```python
if added == 0:
    break
if isinstance(total_entries, int):
    if len(collected) >= total_entries:
        break
elif len(page) < _TASK_INSTANCE_PAGE_SIZE:
    break
```

`offset` advances before the checks now, which is what lets a clamped page carry on. `added == 0` is unchanged and still unconditional, so a server contradicting its own metadata cannot spin.

The stub gained a `task_instance_max_page_limit` that clamps `limit` the way Airflow does. Your table reproduces on the old rule:

```
maximum_page_limit=25   old: 1 request, 25 of 251   new: 11 requests, 251 of 251
```

Two tests, and they are not equal. `test_task_instances_completes_when_the_server_clamps_the_page_limit` is the reproduction: 25 on the unfixed code, 251 here.

`test_task_instances_stops_on_a_short_page_when_no_total_is_reported` passes on both versions. It is a guard against this fix looping forever on a server that reports no total, not a second catch for the reported bug.

Rebased onto current main while I was here. Gates: `ruff check`, `ruff format --check`, `ty check` clean, 1451 passed / 11 skipped in the root suite and 229 in hflow-server.

kstonekuan (approved):
LGTM, merging. The clamp fix is right and the reproduction is a real one.

Each guard is the only thing holding its own failure, which is what makes the test set worth having rather than just green:

```
mutation                                          only failure
short page always ends the run                    ..._completes_when_the_server_clamps_the_page_limit
total counted against the offset                  ..._keeps_paging_when_pages_overlap
dedup removed                                     ..._keeps_paging_when_pages_overlap
a page with no list degrades instead of raising   ..._raises_when_a_page_carries_no_list
no paging at all (main's single request)          6 failures
```

Advancing `offset` before the checks is the part that makes a clamped page carry on, and counting the total against what was kept rather than the offset is what keeps an overlapping server from dropping the tail. Those are two different bugs and you pinned them separately.

I pushed one fixup, `8be0c7d`. The `added == 0` check was the one guard nothing tested: removing it left the whole suite green. It is also the only stopping condition left against a server that serves the same page at every offset while `total_entries` claims more, where the offset never reaches the total and no page is ever empty. The stub bounds itself at six pages so a client that cannot stop fails the test rather than hanging the run.

One behavior change worth putting on the record, because it is not in the issue. A response carrying no `task_instances` list used to return `[]` and now raises. Under #378 that reaches the graph route as a non-404, so a malformed Airflow response is now a 502 instead of a silently empty task graph. That is the right way round and it matches where #378 left the other three call sites, but it is a second decision riding along with the paging one.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1463 passed / 6 skipped. The one failure is `test_preview_stats_render_timestamps_in_utc_on_a_non_utc_host`, which is unrelated to this and fails on main today: it asserts `"-04" not in` the rendered timestamp, and that substring matches the day in `2026-09-04`. #385 fixes it.

Closes #311. If you want another, the advanced list is at <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3Aadvanced>, and #303 just came free. The other thing worth your time, given how much of the runtime you have now been through: point HFlow at a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face) and report what breaks, what is slow, or what is awkward. #287 came out of exactly that and it is more useful to us than any single issue we could write.

Issue #305, opened 2026-08-31
Let hflow catalog ui explore bucket-backed catalogs


kstonekuan:
## Current behavior

`hflow catalog ui` rejects every bucket URL before startup at [`src/hflow/cli.py:1287-1300`](https://github.com/Hebbian-Robotics/hflow/blob/1f5c7c8f59265110cad7c8ffe82a1951a4411bba/src/hflow/cli.py#L1287-L1300). `CatalogUiSettings` and the polling lifecycle also require a local `Path` at [`src/hflow/catalog_ui.py:25-39`](https://github.com/Hebbian-Robotics/hflow/blob/1f5c7c8f59265110cad7c8ffe82a1951a4411bba/src/hflow/catalog_ui.py#L25-L39) and [`src/hflow/catalog_ui.py:93-152`](https://github.com/Hebbian-Robotics/hflow/blob/1f5c7c8f59265110cad7c8ffe82a1951a4411bba/src/hflow/catalog_ui.py#L93-L152).

`open_catalog_connection()` already accepts bucket catalog roots and syncs their append-only Parquet tables into the local mirror at [`src/hflow/curation.py:221-233`](https://github.com/Hebbian-Robotics/hflow/blob/1f5c7c8f59265110cad7c8ffe82a1951a4411bba/src/hflow/curation.py#L221-L233) and [`src/hflow/curation.py:260-281`](https://github.com/Hebbian-Robotics/hflow/blob/1f5c7c8f59265110cad7c8ffe82a1951a4411bba/src/hflow/curation.py#L260-L281). The browser is the remaining local-only catalog surface.

## Definition of done

- `hflow catalog ui --catalog s3://...`, `gs://...`, and `az://...` start against existing HFlow catalogs when the `bucket` extra and provider credentials are available.
- Startup validates the existing remote catalog marker. Browsing never creates or changes objects in the bucket.
- Catalog Parquet files sync into the existing local mirror before the DuckDB connection opens.
- The polling lifecycle syncs newly appended catalog files and makes the first and later completed appends visible in the open UI without a restart.
- A valid empty bucket catalog starts the UI and waits for its first completed append.
- The startup output names the bucket catalog clearly and doesn't present the mirror directory as the source catalog.
- DuckDB continues listening on loopback. SQL reads the local mirror and gets no direct object-store write surface.
- Missing bucket dependencies, credentials, and remote catalog markers produce concise CLI errors with no traceback.
[...]

Issue #304, opened 2026-08-31
Allow LeRobot imports to publish into bucket-backed HFlow data roots


kstonekuan:
## Current behavior

`hflow import lerobot` parses `--output-dir` as a `Path` and describes it as a local destination at [`src/hflow/cli.py:297-302`](https://github.com/Hebbian-Robotics/hflow/blob/1f5c7c8f59265110cad7c8ffe82a1951a4411bba/src/hflow/cli.py#L297-L302). The public importer also accepts `output_dir: Path`, keeps `_lerobot_cache` below it, and returns `list[Path]` at [`src/hflow/importers/lerobot.py:500-550`](https://github.com/Hebbian-Robotics/hflow/blob/1f5c7c8f59265110cad7c8ffe82a1951a4411bba/src/hflow/importers/lerobot.py#L500-L550).

The rest of HFlow treats a data root as local storage or `s3://`, `gs://`, and `az://` through `parse_storage_root()` and the shared publish/fetch boundary at [`src/hflow/storage.py:630-655`](https://github.com/Hebbian-Robotics/hflow/blob/1f5c7c8f59265110cad7c8ffe82a1951a4411bba/src/hflow/storage.py#L630-L655). A bucket-backed workspace currently needs an extra manual upload step after LeRobot import.

## Definition of done

- The public importer and CLI accept every data-root form supported by `parse_storage_root()`.
- Existing local-path behavior and CLI usage remain compatible.
- For a bucket destination, Hugging Face downloads, video slices, and MCAP construction use local staging. Source cache files aren't uploaded into the durable data root.
- Canonical episodes publish under `landing/` through the existing `StorageRoot` boundary.
- `prepared-manifest.json` publishes after every selected episode object succeeds.
- The public result reports real remote object URIs for bucket outputs. It doesn't represent a remote object as a `Path`.
- A failed upload leaves no partial remote object or manifest claiming that the import completed.
- Missing bucket dependencies and provider authentication failures reach the CLI as concise errors with no traceback.
- Add behavioral coverage for both local and bucket-backed destinations using the existing storage test boundary.
- Update `docs/how-to/import-lerobot-v3.md` with one bucket example and explain where the local cache lives.

Batch retry and episode reuse are tracked separately in #303. The two designs should compose, but this issue doesn't add resumable imports itself.

## Non-goals

- Airflow execution
- Parallel episode conversion
- Bucket or filesystem watchers
- A new storage backend or package
[...]

PR #364, opened 2026-09-02
fix(egosuite): validate label report envelope

### LINKED ISSUE #307: Validate the EgoSuite label-report envelope before reading frames
## Current behavior

[write_label_report()](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/egosuite_evaluation/evaluate.py#L877-L905) writes schema_version and label_type alongside the frame records.

[load_projected_hand_label_report()](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/egosuite_evaluation/evaluate.py#L141-L225) checks that the payload is an object with a frames array, then starts parsing frame records. It never reads schema_version or label_type. The current happy-path test builds a report containing only frames at [examples/egosuite_evaluation/tests/test_evaluation.py:148-179](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/examples/egosuite_evaluation/tests/test_evaluation.py#L148-L179).

A report from another producer, an unrelated report type, or a future incompatible schema can therefore enter the current frame parser when its records happen to have the expected field names.

## Pattern to copy

The dataset snapshot marker parser validates both the format name and version before consuming the directory at [src/hflow/snapshot.py:502-530](https://github.com/Hebbian-Robotics/hflow/blob/fbfd5536121907c85bb9dc69b63560299d08e16b/src/hflow/snapshot.py#L502-L530). Its errors name the marker, expected format, and supported version.

## What to build

Validate the saved label report's envelope before reading any frame record. The accepted envelope is the current SCHEMA_VERSION plus label_type projected-hand-joints.

## Definition of done

1. A report written by write_label_report() still loads unchanged.
2. Missing, non-integer, older, and future schema_version values are refused before frame parsing.
3. Missing, non-string, and unsupported label_type values are refused before frame parsing.
4. Each error names the report path, the invalid field, the value found, and the supported value.
5. The existing per-frame validation and ordering behavior stays unchanged.
6. The happy-path fixture includes the real envelope.
7. Add boundary cases for every rejected envelope shape without testing json library behavior.
8. No new dependency is added.

## Non-goals

- Migrating unknown report versions
- Changing projected-hand geometry or label semantics
- Changing how a canonical episode is matched to its labels

## Validation

~~~bash
  examples/egosuite_evaluation
uv run pytest -q
~~~

WilliamK112:
## Summary

- validate the saved projected-hand label report schema version and label type before reading frame records
- use one projected-hand label type constant for the writer and reader contract
- exercise the writer-to-reader happy path and every requested missing, type, older/future, and unsupported envelope boundary

Fixes #307

## Validation

- uv run --locked --project examples/egosuite_evaluation ruff check --fix examples/egosuite_evaluation
- uv run --locked --project examples/egosuite_evaluation ruff format examples/egosuite_evaluation
- uv run --locked --project examples/egosuite_evaluation ty check --project examples/egosuite_evaluation --extra-search-path . examples/egosuite_evaluation
- uv run --locked --project examples/egosuite_evaluation pytest -q examples/egosuite_evaluation/tests (46 passed)
- uv run ruff check
- uv run ruff format --check
- uv run ty check
- uv run pytest -q (1,404 passed, 11 skipped)

## AI assistance disclosure

I used OpenAI Codex to help inspect, implement, and validate this change. I reviewed the final diff and the validation results.

### DIFF
--- a/examples/egosuite_evaluation/evaluate.py
+++ b/examples/egosuite_evaluation/evaluate.py
@@ -65,6 +65,7 @@ from examples.egosuite_evaluation.judgment import (  # noqa: E402
 )
 
 SCHEMA_VERSION = 1
+PROJECTED_HAND_LABEL_TYPE = "projected-hand-joints"
 EXPECTED_HAND_JOINT_COUNT = 21
 DEFAULT_RUNS_DIRECTORY = Path("data/egosuite-evaluation/runs")
 DEFAULT_LABELS_DIRECTORY = Path("data/egosuite-evaluation/labels")
@@ -151,6 +152,22 @@ def load_projected_hand_label_report(
         ) from error
     if not isinstance(report_payload, dict):
         raise ValueError(f"projected-hand label report {report_path} must contain a JSON object")
+    schema_version = report_payload.get("schema_version")
+    if (
+        not isinstance(schema_version, int)
+        or isinstance(schema_version, bool)
+        or schema_version != SCHEMA_VERSION
+    ):
+        raise ValueError(
+            f"projected-hand label report {report_path} field 'schema_version' has value "
+            f"{schema_version!r}; supported value is {SCHEMA_VERSION}"
+        )
+    label_type = report_payload.get("label_type")
+    if not isinstance(label_type, str) or label_type != PROJECTED_HAND_LABEL_TYPE:
+        raise ValueError(
+            f"projected-hand label report {report_path} field 'label_type' has value "
+            f"{label_type!r}; supported value is {PROJECTED_HAND_LABEL_TYPE!r}"
+        )
     raw_frame_records = report_payload.get("frames")
     if not isinstance(raw_frame_records, list):
         raise ValueError(f"projected-hand label report {report_path} must contain a 'frames' array")
@@ -889,7 +906,7 @@ def write_label_report(
     all_labels = [label for labels in labels_by_source.values() for label in labels]
     report = {
         "schema_version": SCHEMA_VERSION,
-        "label_type": "projected-hand-joints",
+        "label_type": PROJECTED_HAND_LABEL_TYPE,
         "camera_view": camera_view.value,
         "frame_stride": frame_stride,
         "limit_per_episode": limit_per_episode,
--- a/examples/egosuite_evaluation/tests/test_evaluation.py
+++ b/examples/egosuite_evaluation/tests/test_evaluation.py
@@ -35,6 +35,7 @@ from examples.egosuite_evaluation.evaluate import (
     select_episode_paths,
     select_stratified_labels,
     summarize_evaluation_results,
+    write_label_report,
 )
 from examples.egosuite_evaluation.geometry import (
     CameraPoseInWorld,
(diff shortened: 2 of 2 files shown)


kstonekuan (changes requested):
The validation itself is right and I would merge it as it stands. It needs a rebase, and the rebase is not mechanical, so here is what collides before you spend time on it.

#362 merged about an hour ago and rewrote the same loader: `load_projected_hand_label_report` now returns a `ProjectedHandLabelReport` and matches labels by HFlow `source_uri` instead of the basename. Merging current main into your branch auto-merges `evaluate.py` cleanly, conflicts only in the test file, and then fails five tests. Two different causes:

**Four are mechanical.** #362's new tests hand-write reports as `{"frames": [...]}` with no envelope, so your check refuses them before they reach what they are testing:

```
test_legacy_label_report_rejects_one_basename_for_multiple_sources
test_saved_label_report_matches_same_named_sources_by_canonical_provenance
test_saved_label_report_rejects_missing_or_unrelated_canonical_provenance[both cases]

  Actual message: field 'schema_version' has value None; supported value is 1
```

Adding `"schema_version": 1, "label_type": "projected-hand-joints"` to those four fixtures is the whole fix, and arguably your check is right to have caught them.

**One needs a decision from you.** You rewrote `test_saved_label_report_selects_exact_frames_for_a_canonical_episode` to build its report through `write_label_report`, which is DoD 6 and the better fixture. On main that test hand-writes a report with no `source_uri`, so it matches through #362's legacy basename path. Through `write_label_report` the report now carries a real `source_uri`, matching switches to the identity path, and the hardcoded `{"source_uri": "run-a/episode-123.mcap"}` no longer corresponds to what was recorded:

```
ValueError: label manifest has no frames for source identity 'run-a/episode-123.mcap'
```

The identity `write_label_report` records comes from `App.source_identity()` against `HFLOW_DATA_ROOT`, so for a `tmp_path` source it is an absolute path, not `run-a/...`. Reading the recorded `source_uri` back out of the written report and passing that is the robust answer, and it keeps your real-envelope fixture. Hardcoding a second identity would just move the coupling.

Nothing above is a criticism of the change. Excluding `bool` from the `schema_version` integer check is the detail I look for and most people miss: `isinstance(True, int)` is `True`, so without that line `{"schema_version": true}` would have been accepted as version 1. This repo has been bitten by that more than once.

One thing for next time, and it is the reason this is a request rather than a merge with a fixup from me: `pytest -q` reporting 11 skipped rather than 6 means it ran without `--all-extras`. `uv sync --locked --all-extras` is the documented form, and the difference is the mediapipe and vision tests.

Last thing, an invitation rather than a rule. You have eleven merged PRs here, so the `good first issue` pool is best left for people arriving after you. The higher-leverage work is the advanced backlog, and #365 was filed today by another contributor from actually running the thing against Egocentric-10K, which is the kind of issue worth more than any single starter fix. #311, #312, #320 and the bucket-backed set (#303, #304, #305) are all open and unassigned.

WilliamK112:
Thanks for the detailed rebase notes — I addressed them in 752ce42 and merged main through 65218e8.

- added the schema_version and label_type envelope to the four #362 fixture cases
- kept the write_label_report fixture for the exact-frame test
- read the source_uri written into that report and passed the recorded identity into labels_for_pipeline_episode
- preserved the strict version/type validation, including rejecting bool as an integer schema version

Validation:
- uv sync --locked --all-extras
- Python 3.11 EgoSuite workspace tests: 54 passed
- Python 3.14 EgoSuite workspace tests: 54 passed
- workspace ruff check/format and ty check: passed
- full root suite on macOS: 1467 passed, 11 skipped; the remaining 17 failures are the new current-main native-overlay tests, which intentionally reject non-Linux builds and are covered by Linux CI

The PR now merges cleanly with current main and is ready for another look.

kstonekuan (approved):
LGTM, merging. The rebase landed the way it needed to and the envelope check is properly held.

`{**envelope, "frames": [None]}` is what makes the test name true. A frame list that cannot parse means the refusal can only be coming from the envelope, so "before frames" is an assertion rather than a description. Each of the seven cases is the only failure when its own clause goes:

```
schema_version check removed   4 cases fail, no others
label_type check removed       3 cases fail, no others
```

Reading the recorded `source_uri` back out of the report instead of hardcoding `run-a/episode-123.mcap` is the better half of the change. It ties the writer's identity to the reader's lookup, and it fails now if the two drift: repointing `write_label_report` at a different label type breaks the exact-frame test.

I pushed one fixup, `de57a4d`. The bool guard was the one clause nothing held. `bool` subclasses `int`, so `True` satisfies `isinstance(_, int)` and `True != 1` is False, which means removing that clause left all 54 tests green while `{"schema_version": true}` sailed through. You called it out by name in the rebase notes, so it was worth a case of its own rather than a comment.

One thing left as is: the `isinstance(label_type, str)` half of the `label_type` check is unreachable, since only a `str` can equal the constant. Dropping it changes no test. Harmless, and it reads as intent, so I left it.

Gate on the merged result. Example project: `ruff check`, `ruff format --check`, `ty check` clean, 55 passed. Root suite: clean, 1489 passed / 6 skipped. Your 17 macOS failures were the native-overlay tests from `07dbc69`; they pass here on Linux, so your read on them was right.

Fixes #307. One ask, and it is an invitation rather than a rule: the `good first issue` label is there to give newcomers somewhere to start, and with eleven PRs you are well past needing it. Please leave that pool alone and take from the advanced list instead: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3Aadvanced>

Given your transform and video history in particular, the thing most worth your time is not an issue we wrote. Point HFlow at a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face) and tell us what breaks, what is slow, or what is awkward. #287 came out of exactly that and it has been more useful than anything we would have thought to file.

PR #366, opened 2026-09-02
feat(catalog-ui): explore bucket-backed catalogs

### LINKED ISSUE #305: Let hflow catalog ui explore bucket-backed catalogs

VARUN3WARE:
## Summary

- Allow `hflow catalog ui --catalog` to open existing S3, GCS, and Azure catalogs.
- Validate the remote catalog marker without creating or changing bucket objects.
- Sync append-only Parquet files into the existing local mirror before DuckDB opens.
- Poll for first and later completed appends without restarting the UI.
- Keep DuckDB on loopback and restrict catalog reads to the local mirror.
- Add behavioral tests and document bucket usage, mirroring, and SSH forwarding.

Closes #305.

## Why

`hflow catalog ui` previously rejected every bucket URL even though HFlow's catalog connection already supported bucket-backed catalogs through its local mirror.

This change reuses the existing catalog marker, append, and mirror semantics instead of adding direct object-store access to DuckDB or introducing a new catalog format. Local catalog behavior remains unchanged, while bucket browsing stays read-only.

## Validation

- `uv run ruff check`
  - Passed
- `uv run ruff format --check`
  - Passed; 206 files already formatted
- `uv run ty check`
  - Passed
- `uv run pytest tests/test_catalog_ui.py tests/test_cli.py tests/test_storage.py -q`
  - 92 passed, 1 skipped
- `uv run pytest -q`
  - 1402 passed, 6 skipped
- Pre-commit hooks
  - `ruff check` passed
  - `ruff format` passed

## Checklist

### DIFF
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -151,11 +151,14 @@ marker and never creates or changes objects in the bucket. It syncs append-only
 Parquet table files into the local mirror under `HFLOW_MIRROR_DIR` (or
 `$XDG_CACHE_HOME/hflow/mirrors`) before DuckDB opens, then re-syncs on its poll
 interval so the first and later completed appends become visible in the open UI
-without a restart. DuckDB reads the mirror only and continues to listen on
-loopback. A valid empty bucket catalog -- marker present, no Parquet rows yet --
-starts the UI and waits for its first completed append the same way a local
-empty catalog does. Local catalogs may still be created on startup when the
-directory does not exist yet; remote catalogs must already exist.
+without a restart. The default poll interval is 0.5 seconds; each tick lists the
+catalog table prefixes and downloads only the files the mirror still lacks,
+which is fine for an interactive laptop session. DuckDB reads the mirror only
+and continues to listen on loopback. A valid empty bucket catalog -- marker
+present, no Parquet rows yet -- starts the UI and waits for its first completed
+append the same way a local empty catalog does. Local catalogs may still be
+created on startup when the directory does not exist yet; remote catalogs must
+already exist.
 
 DuckDB UI listens on loopback. When HFlow is running on another machine, run
 the second command there and forward the port from your laptop:
--- a/docs/CATALOG.md
+++ b/docs/CATALOG.md
@@ -138,6 +138,25 @@ hflow catalog ui --catalog data/egocentric/catalog
 hflow catalog ui --catalog data/egocentric/catalog --no-browser
 ```
 
+Object-store catalogs work the same way when the optional bucket backend is
+installed (`uv sync --extra bucket`) and provider credentials are available:
+
+```bash
+hflow catalog ui --catalog s3://robot-data/production/catalog --no-browser
+hflow catalog ui --catalog gs://robot-data/production/catalog --no-browser
+```
+
+For bucket catalogs, `catalog ui` validates the existing `format_version`
+marker and never creates or changes objects in the bucket. It syncs append-only
+Parquet table files into the local mirror under `HFLOW_MIRROR_DIR` (or
+`$XDG_CACHE_HOME/hflow/mirrors`) before DuckDB opens, then re-syncs on its poll
+interval so the first and later completed appends become visible in the open UI
+without a restart. DuckDB reads the mirror only and continues to listen on
+loopback. A valid empty bucket catalog -- marker present, no Parquet rows yet --
+starts the UI and waits for its first completed append the same way a local
+empty catalog does. Local catalogs may still be created on startup when the
+directory does not exist yet; remote catalogs must already exist.
+
 DuckDB UI listens on loopback. When HFlow is running on another machine, run
 the second command there and forward the port from your laptop:
 
@@ -151,8 +170,8 @@ assets. DuckDB UI is an unrestricted local SQL explorer, so keep the port on
 loopback and use a tunnel instead of exposing it publicly.
 
 This command is separate from [`hflow serve`](./SERVE.md). `catalog ui` is
-DuckDB's browser over the local catalog. `serve` is HFlow's optional workspace
-REST API and does not ship a browser client.
+DuckDB's browser over the catalog (local or synced from a bucket). `serve` is
+HFlow's optional workspace REST API and does not ship a browser client.
 
 ### Query from Python or the CLI
 
@@ -601,9 +620,11 @@ itself. A catalog root contains the `format_version` marker.
 - The wide view binds its measurement columns to the keys present when the
   connection was opened. `hflow catalog ui` refreshes this surface after an
   initially empty catalog's first append, and `curate()` reopens per call.
-  If a later run introduces brand-new measurement keys into a long-lived UI
-  or `open_catalog_connection()` session, restart that session to add those
-  columns. New rows for existing keys appear without a restart.
+  Bucket-backed UI sessions re-sync newly appended Parquet files on each poll
+  so new rows for existing keys appear without a restart. If a later run
+  introduces brand-new measurement keys into a long-lived UI or
+  `open_catalog_connection()` session, restart that session to add those
+  columns.
 
 ## See also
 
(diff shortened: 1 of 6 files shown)


kstonekuan (changes requested):
Strong first contribution, and an advanced issue for a first PR at that. The design decision is the right one: reusing `_local_query_root`'s existing mirror sync rather than giving DuckDB object-store credentials means SQL never gets a write surface, which is DoD's loopback requirement satisfied by construction rather than by a check.

I verified the read-only claim harder than your test does. Yours compares key names; the DoD says "creates or changes", so I fingerprinted every object by sha256 and size before and after a browsing session:

```
objects before browsing: 7
objects after browsing:  7
created: []   removed: []   changed in place: []
VERDICT: unchanged
```

And the startup line prints `Catalog: file:///.../bucket/catalog`, not the mirror path, so that DoD item holds too.

One thing to fix, and it is about the tests rather than the code.

**The two mirror syncs are the mechanism for "visible without a restart", and nothing holds them.** Both can be deleted with the whole suite still green:

```
post-append mirror sync removed        -> 48 passed, nothing noticed
first-append detection stops syncing   -> 48 passed, nothing noticed
```

The cause is that `_append_remote_episode` appends through the *same* `BucketStorageRoot` the UI reads, so the writer and the reader share one mirror and the append primes it directly. The test then observes a real row-count change and passes, but it cannot distinguish "the UI synced the bucket" from "the writer already put the file in the UI's mirror". In production the ingest is a different process with its own mirror, and the sync is the only path.

Giving the writer its own mirror over the same bucket makes the difference visible. I ran your later-append flow that way:

```
                              episodes visible after a separate-process append
as submitted                  (2,)   sync worked
post-append sync removed      (1,)   NOT VISIBLE
```

So the code is right and only the fixture is too generous. Something like this is enough:

```python
writer_root = BucketStorageRoot(f"file://{remote_dir}", mirror=tmp_path / "writer-mirror")
Catalog(writer_root.child("catalog")).append_episode(...)
```

with the UI on its own `BucketStorageRoot`. Worth doing for both the first-append and later-append tests, since neither currently holds its sync.

Three smaller notes, none blocking. `FileNotFoundError` in the `except` tuple is already covered by `OSError`. Importing `obstore.exceptions` inside the handler means a bucket URL that fails for some unrelated reason without the `bucket` extra installed raises `ModuleNotFoundError` from the handler and masks the original error; importing it at the top of the branch where you already know the extra is present would avoid that. And a 0.5-second sync cadence against a real bucket is a LIST plus GETs every half second for as long as the UI is open, which is fine for a laptop session and worth a word in `docs/CATALOG.md` next to the mirror behaviour you documented.

Gate is clean otherwise: `ruff check`, `ruff format --check`, `ty check`, 1420 passed / 6 skipped with current main merged in.

VARUN3WARE:
Thanks for the careful review, especially the fingerprint check and catching the shared-mirror fixture.

Addressed in 7b4770c:
- First- and later-append tests now append through a separate `BucketStorageRoot` with its own writer mirror over the same bucket, so the UI sync path is what makes new rows visible.
- Dropped the redundant `FileNotFoundError` catch.
- Import `obstore.exceptions` before the handler so a missing extra can’t mask an unrelated failure.
- Documented the 0.5s poll LIST/GET cadence next to the mirror behavior in `docs/CATALOG.md`.

kstonekuan (approved):
LGTM, merging. Welcome, and that is a good first PR to have landed: an advanced issue, six files, and a two-part review turned around without loose ends.

The mutations that went unnoticed last round now both bite, each on its own test:

```
bucket catalogs get a Catalog() constructed  -> ..._does_not_write_bucket_objects_while_browsing
post-append mirror sync removed              -> ..._sees_a_later_remote_append_without_restart
first-append detection stops syncing         -> ..._waits_for_the_first_remote_append
```

Two of those three were "48 passed, nothing noticed" before `7b4770c`. The sync is now the mechanism the tests depend on rather than something the fixture happened to do for them.

`_writer_catalog_root` is the right shape, and its docstring is the part worth keeping:

> Production ingest is a separate process with a separate mirror. Sharing the UI's mirror would let appends prime DuckDB without exercising sync.

That sentence is why the next person will not quietly collapse the two roots back into one when the helper looks redundant.

The `obstore` import fix went further than I asked. I suggested moving it out of the handler; resolving it to `None` when the extra is absent means a bucket URL that fails for an unrelated reason still raises its own error rather than a `ModuleNotFoundError` from the rescue path, and it does not assume the extra is present just because the URL looks remote. Dropping `FileNotFoundError` from the tuple is right too, since `OSError` already covers it.

The `docs/CATALOG.md` note lands the poll cadence next to the mirror behaviour, and saying the tick "lists the catalog table prefixes and downloads only the files the mirror still lacks" is the detail an operator needs to reason about cost on a real bucket.

Gate clean with current main merged in: `ruff check`, `ruff format --check`, `ty check`, 1421 passed / 6 skipped.

Closes #305. If you want more of this shape, #303 and #304 are the other two bucket-backed gaps and neither is assigned.

PR #367, opened 2026-09-02
perf(checks): add a reproducible cold camera evidence benchmark

### LINKED ISSUE #365: Improve camera_frame_stats cold throughput on 1080p video without weakening per-frame evidence
## Current behavior

`camera_frame_stats` correctly combines blackout, freeze, exposure, and signal measurements in one FFmpeg decode, and its instrument output is cached. The remaining cold-path cost is the single full-resolution decode/filter pass itself.

On a real 30-second 1080p30 Egocentric-10K clip, a cold `camera_frame_stats` call takes about 5.3 seconds per camera. The time is essentially unchanged whether the canonical H.264 came from a native in-band H.264 landing episode or from the former JPEG landing path. This is therefore separate from the ingestion/transcode work in #287 and #291/#334.

This also does not duplicate #173/#175: that work removed repeated identical decodes. This issue is about the irreducible first decode.

## Controlled real-corpus result

Tested at HFlow commit `f6f66e85e0593dc4cc133f0eedfa0ebe16b2d14a` (current main, including #334).

- Dataset: `builddotai/Egocentric-10K`
- Revision: `3e5f87c88c54ce8343865d8e2a8c171f18385a05`
- Source: `factory051_worker001_00000.mp4`
- Source SHA-256: `f27e6b8ef4534e7e6b5320ea833ccaf58680aaf4187a0c2afaf3e2c1c1e36242`
- Source shape: HEVC, 1920x1080, 30 FPS
- Excerpt: `[10 s, 40 s)`, 900 frames
- Hardware: Ryzen 9 8945H (8C/16T), 32 GB RAM, RTX 4070 Laptop 8 GB
- FFmpeg: `n8.1.2-50-g1a748fe2cd-20260901`

Both inputs used the same source excerpt and output shape. The native route wrote `foxglove.CompressedVideo` H.264 directly. The control route used `hflow.testing.write_video_episode()` with a 1920x1080/30 FPS `VideoEpisodeSpec`, then the normal canonical transform. Every check run used a fresh `Episode.workdir`, so neither the MP4 nor instrument cache was reused.

| Route | Adapter | Transform | Landing bytes | Canonical bytes | Cold check 1 | Cold check 2 | Decoded frames |
|---|---:|---:|---:|---:|---:|---:|---:|
| native H.264 | 8.24 s | 0.32 s | 34,135,921 | 34,135,821 | 5.238 s | 5.244 s | 900 |
| JPEG-derived | 1.89 s | 8.72 s | 114,852,251 | 44,185,605 | 5.507 s | 5.479 s | 900 |

Both checks reported `black_frame_pct=0` and `freeze_total_s=0`. The native check used about 600% CPU and 176 MiB peak RSS. The JPEG-derived check used about 602% CPU and 206 MiB peak RSS.

For an exploratory hardware-decode control, I ran the exact current filter graph with CUDA decode plus `hwdownload` before the software filters:

| Decode path | Wall time | CPU | Peak RSS | NVIDIA decoder |
|---|---:|---:|---:|---:|
| current CPU path | 5.56 s | 526% | 153 MiB | 0% |
| NVDEC + hwdownload + current software filters | 6.07 s | 440% | 217 MiB | 20-22% |

The naive GPU route is slower, so this is not a request to make CUDA mandatory or to switch paths without evidence.

## Definition of done

- Add a repeatable cold-path benchmark for one canonical 1080p30 camera that reports decoded frames and wall time separately from transform time.
- Reduce the single-pass measurement time beyond run-to-run noise on the benchmark and include repeated before/after measurements in the PR.
- Preserve the current per-frame blackout, freeze, luma, and signal-quality guarantees; sampling or downscaling must not silently change their meaning.
- Preserve the existing fault-fixture results, decoded-frame accounting, intervals, and measurement values within their defined precision.
- Keep a portable CPU path. Hardware acceleration may be optional only when its availability and actual benefit are detected rather than assumed.
- If the effective measurement path changes observable semantics, version and cache/provenance identity must change with it. A stale cached instrument result must never satisfy a different path.
- Document any new operational setting and its fallback behavior.

## Non-goals

- Revisiting the WebDataset decision or the direct-H.264 adapter from #287/#291/#334
- Restoring the removed duplicate decode
- Making an NVIDIA GPU a requirement
- Trading away per-frame evidence for an undocumented sampled estimate

This is the control requested in #287 after the ingestion and check costs were separated.

Galabavamsi:
## Summary

- add a deterministic cold `camera_frame_stats` benchmark with default 1080p30, quick, and per-filter profiling modes
- time canonical transformation separately and use a fresh episode workdir for every cold check
- report FFmpeg/CPU provenance, decoded frames, and instrument-cache sizes
- document the measured bottleneck and why the shipped filter graph remains unchanged

## Why

Issue #365 asked for a reproducible benchmark and a safe speedup if one could be demonstrated. Profiling shows `signalstats=stat=tout+brng` dominates the evidence pass. Thread-count, branched-graph, and CUDA controls did not produce a repeatable semantics-preserving improvement, so this keeps product behavior unchanged and contributes the durable measurement requested by the maintainer.

Closes #365

## Validation

- `uv run ruff format --check` — 208 files already formatted
- `uv run ty check` — passed
- `uv run pytest -q` — 1410 passed, 6 skipped
- `uv run python benchmarks/camera_frame_stats_benchmark.py --quick --profile-filters` — passed
- `uv run python benchmarks/camera_frame_stats_benchmark.py --profile-filters` — all three cold 1080p30 checks decoded 900 frames; 4.897 s median
- `lychee --no-progress --include-fragments docs/BENCHMARKS.md` — 3 links passed
- pre-commit `ruff check` and `ruff format` hooks — passed

## Checklist

### DIFF
--- a/benchmarks/camera_frame_stats_benchmark.py
+++ b/benchmarks/camera_frame_stats_benchmark.py
@@ -1,0 +1,259 @@
+"""Benchmark cold ``camera_frame_stats`` throughput on canonical 1080p30 video.
+
+The default fixture is one deterministic 30-second, one-camera episode. Its
+JPEG-to-H.264 transform is timed separately, then every check repetition gets
+a fresh ``Episode`` workdir so neither the remuxed MP4 nor FFmpeg instrument
+cache can turn a cold measurement into a cache hit.
+
+Run: ``uv run python benchmarks/camera_frame_stats_benchmark.py``
+(``--quick`` for a small development fixture). Timings are wall-clock.
+"""
+
+import argparse
+import os
+import statistics
+import subprocess
+import tempfile
+import time
+from dataclasses import dataclass
+from pathlib import Path
+
+import hflow
+from hflow._video_measurements import FrameStatisticsSettings
+from hflow._video_measurements._frame_statistics import frame_statistics_filter_graph
+from hflow.checks import camera_frame_stats
+from hflow.ffmpeg import ffmpeg_path, ffmpeg_version
+from hflow.testing import SyntheticEpisodeSpec, synthesize_episode
+from hflow.transform import TransformConfig, write_canonical_episode
+
+FULL_DURATION_S = 30.0
+FULL_FRAMES_PER_SECOND = 30.0
+FULL_WIDTH = 1920
+FULL_HEIGHT = 1080
+FULL_REPETITIONS = 3
+
+QUICK_DURATION_S = 3.0
+QUICK_FRAMES_PER_SECOND = 10.0
+QUICK_WIDTH = 320
+QUICK_HEIGHT = 180
+QUICK_REPETITIONS = 2
+
+
+@dataclass(frozen=True)
+class BenchmarkSettings:
+    duration_s: float
+    frames_per_second: float
+    width: int
+    height: int
+    repetitions: int
+    profile_filters: bool = False
+
+
+@dataclass(frozen=True)
+class CheckRun:
+    seconds: float
+    decoded_frame_count: int
+    instrument_cache_path: Path
+    instrument_cache_bytes: int
+
+
+@dataclass(frozen=True)
+class FilterProfile:
+    name: str
+    seconds: float
+
+
+@dataclass(frozen=True)
+class BenchmarkResult:
+    transform_seconds: float
+    camera_topic: str
+    ffmpeg_version: str
+    logical_cpu_count: int
+    check_runs: tuple[CheckRun, ...]
+    filter_profiles: tuple[FilterProfile, ...]
+
+
+def _measure_cold_check(
+    canonical_path: Path,
+    camera_topic: str,
+    frames_per_second: float,
+    workdir: Path,
+) -> CheckRun:
+    started = time.perf_counter()
+    with hflow.Episode(canonical_path, workdir=workdir) as episode:
+        result = camera_frame_stats(
+            episode,
+            cameras=[camera_topic],
+            expected_hz={camera_topic: frames_per_second},
+        )
+    seconds = time.perf_counter() - started
+    instrument_cache_paths = list(workdir.glob("*.instrument.*.txt"))
+    if len(instrument_cache_paths) != 1:
+        raise RuntimeError(
+            f"expected one cold instrument cache in {workdir}, got {instrument_cache_paths}"
+        )
+    decoded_frame_count = result.measurements[f"{camera_topic}/decoded_frame_count"]
+    if not isinstance(decoded_frame_count, int):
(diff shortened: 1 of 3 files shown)


Galabavamsi:
@kstonekuan need your review for this pr

kstonekuan (approved):
LGTM, merging. This is the resolution I said would be legitimate on #365, and it is worth more than a speedup would have been.

Reproduced the whole thing on different hardware and a different FFmpeg build (`n8.1.2-44` against your `n8.1.2-50`, 22 logical CPUs against your 16):

```
                                        yours     mine
transform to canonical                  3.510 s   4.445 s
cold camera_frame_stats median          4.897 s   8.294 s
decode only                             0.422 s   0.695 s
decode + signalstats=stat=tout+brng     4.536 s   7.909 s
complete shipped graph                  4.711 s   8.145 s
```

Absolute numbers differ, as they should, and the finding does not: `signalstats` is essentially all of the cost above decode, 97% of the graph's margin over `decode only` on my machine against 96% on yours. Decode is under a tenth of it either way.

The stronger corroboration is accidental. When I accepted #365 I measured a synthetic 1080p30 cold check at 8.3 s with a throwaway script of my own; your benchmark reports 8.294 s on the same machine. Two independently written measurements of the same quantity agreeing to the noise is what makes this a baseline rather than a number.

The test is the part I would have asked for and did not have to. It runs the benchmark's machinery over a 160x90 one-second fixture in 1.36 s rather than importing 1080p into the suite, and this assertion is the one that matters:

```python
assert len({run.instrument_cache_path.parent for run in result.check_runs}) == 2
```

A benchmark whose "cold" runs quietly share a workdir measures a cache hit and reports it as a decode. That line is what stops this baseline rotting into a lie, and it is why the docstring's "reusing one workdir breaks this result" is a claim rather than a hope.

Three smaller things I checked. The script writes only into a `TemporaryDirectory`, so a run leaves the tree clean. `benchmarks/` is covered by `ruff` and `ty` like everything else. And the new table slots into the existing `BENCHMARKS.md` summary rather than starting a parallel document.

The negative result is written the right way round: what was tried, what it measured, and why a luma-only input is not a free win because `BRNG` deliberately reads all three planes. "Keep the measurement path unchanged" backed by controls is a finding, and the next person who wants to optimise this now has to beat a number instead of arguing with an impression.

Gate clean: `ruff check`, `ruff format --check`, `ty check`, 1410 passed / 6 skipped.

Closing #365 on this is right, and to be explicit about why: its DoD asked for a speedup, and I said when accepting it that a documented "no safe win available" would be a legitimate resolution. That was the maintainer position before you started, not a bar being moved afterwards.

PR #372, opened 2026-09-03
test(lerobot): cover the required 'index' column refusal in the curated export

### LINKED ISSUE #370: The curated export's required 'index' column refusal is untested
## Current behavior

The LeRobot curated exporter refuses a source data parquet that has no `index` column, at [`examples/lerobot/export.py:349-354`](https://github.com/Hebbian-Robotics/hflow/blob/main/examples/lerobot/export.py#L349-L354):

```python
if "index" not in cols:
    raise ValueError(
        f"source data parquet {data_local.name} has no 'index' "
        "column; selections can only be renumbered from LeRobot "
        "v3 sources"
    )
```

Nothing tests it. Delete those five lines and the whole export suite stays green.

The refusal exists for a real reason. Before #273, the renumbering fell back to `frame_index` when `index` was absent, and the fallback collided with the line above it:

```python
d["frame_index"] = local_frame
d[index_col] = len(data_frames)     # index_col == "frame_index" on the fallback
```

so `frame_index` came out global instead of per-episode. Requiring `index` removed the collision. The guard is now the only thing standing between a non-v3 source and silently wrong frame numbering, and it is the one refusal in that module with no test.

## Pattern to copy

`tests/test_lerobot_export.py` has a refusal test for every other guard: see `test_export_missing_source_episode_fails` at [`:385`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_lerobot_export.py#L385) and `test_validate_v3_rejects_non_integer_data_refs` at [`:612`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_lerobot_export.py#L612). Each builds the bad input, calls the real entry point, and asserts the refusal with `pytest.raises(..., match=...)`.

The `_fake_corpus` helper at the top of that file writes the synthetic source data parquet with an `index` column, so a fixture without one is a small variation on what is already there.

## What to build

A test that a source data parquet lacking `index` is refused, asserting the message rather than just the exception type. Several guards in that file raise `ValueError`, so the type alone does not say which one fired.

## Definition of done

1. A test drives `export.export(...)` over a source whose data parquet has no `index` column and asserts the refusal, matching on the message.
2. The test fails when the guard at `export.py:349` is deleted. Put that result in the PR body.
3. The existing tests pass untouched, and the happy-path fixture still has its `index` column.
4. Nothing in `examples/lerobot/export.py` changes: this issue adds a test, it does not change behaviour.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_lerobot_export.py
uv run pytest -q
```

## Notes

Test-only, no network, no real corpus: the module's fixtures are synthetic parquet files written in the test file itself. A good first look at how this repo proves a guard is load-bearing rather than just present.

ayam04:
Fixes #370

The guard at `examples/lerobot/export.py:349-354` was the one refusal in the export module with no test (it came from my #273 renumbering cleanup, where the `frame_index` fallback collided with the per-episode renumber). This PR adds the missing test.

**What changed**

- New `test_export_missing_index_column_fails` in `tests/test_lerobot_export.py`: builds the normal `fake_corpus`, drops the `index` column from the source data parquet (pyarrow), drives the real `export.export(...)` via a manifest, and asserts the refusal **by message** (`has no 'index' column`) rather than by exception type — several other guards in the module also raise `ValueError`.
- The happy-path fixture keeps its `index` column; nothing in `examples/lerobot/export.py` changes (test-only per the issue).

**RED/GREEN evidence**

- Guard neutered (`if False and "index" not in cols:`): the new test FAILS — the export falls through to a duckdb `BinderException` at export.py:357 instead of the refusal. (1 failed)
- Guard restored: 17/17 in `tests/test_lerobot_export.py` pass.

**Validation run**

- `uv run --no-sync pytest -q tests/test_lerobot_export.py` → 17 passed
- `uv run --no-sync ruff check tests/test_lerobot_export.py` → clean
- `uv run --no-sync ruff format --check tests/test_lerobot_export.py` → clean
- `uv run --no-sync ty check` → clean
- `uv run --no-sync pytest -q` (repo root) → 1410 passed, 6 skipped, 1 failed: `test_canonical_episode_extracts_exact_source_frame_indices` — reproduced identically on a clean origin/main worktree (same WSL ffmpeg env failure, unrelated to this change).

### DIFF
--- a/tests/test_lerobot_export.py
+++ b/tests/test_lerobot_export.py
@@ -404,6 +404,21 @@ def test_export_duplicate_episode_fails(fake_corpus: dict, tmp_path: Path) -> No
     assert not dest.exists()
 
 
+def test_export_missing_index_column_fails(fake_corpus: dict, tmp_path: Path) -> None:
+    """A source data parquet without an 'index' column is refused by message."""
+    import pyarrow.parquet as pq
+
+    src = Path(fake_corpus["cache_dir"]) / "data" / "chunk-000" / "file-000.parquet"
+    table = pq.read_table(src).drop_columns(["index"])
+    pq.write_table(table, src)
+
+    manifest = _fake_manifest(tmp_path, [{"metadata_json": _provenance_meta(0)}])
+    dest = tmp_path / "out"
+    with pytest.raises(ValueError, match=r"has no 'index' column"):
+        export.export(dest, manifest=manifest, camera_keys=CAMS)
+    assert not dest.exists()
+
+
 def test_export_sql_selection(fake_corpus: dict, tmp_path: Path) -> None:
     """SQL selection path: same outcome via a duckdb query string."""
     _make_data_local(fake_corpus)


kstonekuan (approved):
LGTM, merging.

```
guard at export.py:349 removed -> 1 failed, 16 passed
                                  FAILED test_export_missing_index_column_fails
```

Dropping the column from the source parquet rather than hand-building a broken one is the right way in: `_fake_import` runs `_make_data_local` during the stubbed materialization, so the modified file is what reaches `_fetch_data`, and the test exercises the real path instead of a shortcut into it. Matching on `has no 'index' column` matters for the reason the issue gave, since several guards in that module raise `ValueError` and the type alone would not say which fired.

`assert not dest.exists()` is a nice extra: the refusal happens during staging, so it also confirms nothing was published on the way out.

Gate clean with current main merged in: 1422 passed / 6 skipped.

Closes #370, which came out of your own #273. Good to see it come back around.

PR #373, opened 2026-09-03
Check the no-B-frame constraint in doctor and the encoder

### LINKED ISSUE #368: The no-B-frames constraint is refused on write but never checked by doctor or the encoder
## Current behavior

Canonical video requires no B-frames ([`docs/FORMAT.md:153`](https://github.com/Hebbian-Robotics/hflow/blob/main/docs/FORMAT.md#L153), item 4). Two refusals enforce it on the write side: the transform refuses pass-through video carrying B-frames, and `write_access_units_to_mp4` ([`src/hflow/video.py:592`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/video.py#L592)) refuses any B-frame payload, because a `-c:v copy` remux drops the reorder tail (#255, measured in #250).

Two places state the constraint and do not check it.

**`hflow doctor` does not classify picture coding types.** `_check_video_payload` ([`src/hflow/doctor.py:96`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/doctor.py#L96)) calls `count_h264_pictures` at `:112` for the picture count and never asks whether that picture is a B picture. `docs/FORMAT.md:161-162` says so outright:

> `hflow doctor` still does not classify picture coding types (below), so a clean report does not prove this constraint; the refusals do.

So a non-conforming file can pass `hflow doctor` and then be refused by the remux behind `Episode.video`, which is the wrong order for a validator to find out.

**The encoder does not assert what its own docstring depends on.** `encode_images_to_h264` ([`src/hflow/video.py:132`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/video.py#L132)) lists its guarantees as "enforced, raising `VideoEncodeError` otherwise", and justifies the first one with "no B-frames means decode order == presentation order". `_enforce_encode_guarantees` at `:665` checks the frame count, the keyframe cadence, and SPS/PPS on keyframes. It does not check for B-frames. The property holds only because `bframes=0` sits in the x264 parameter string at `:156`. Edit that string, or meet an ffmpeg build that ignores the flag, and HFlow's own encoder produces output that HFlow's own remux refuses, with the failure surfacing a long way from its cause.

## Why now

The instrument did not exist when either of these was written. `scan_picture_coding_types` ([`src/hflow/video.py:455`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/video.py#L455)) arrived with #345, and after #354 and #358 it is cheap and shares one NAL walk with `count_h264_pictures`. Measured on main, 640x480 bframes=3:

```
 300 frames  count_h264_pictures  1.43 ms   scan_picture_coding_types  1.68 ms
3000 frames  count_h264_pictures 14.54 ms   scan_picture_coding_types 16.93 ms
```

Both readers already walk the same NALs and decode the same two Exp-Golomb fields, so the classification is very nearly free at the point where doctor already pays for the count.

## Pattern to copy

Doctor's existing video findings and their documented codes: `video-format`, `video-invalid-slice-header`, `video-multiple-access-units` at `doctor.py:96-140`, each with a row in the finding-code table at `docs/FORMAT.md:197-199`. A new code needs a row there too.

`_enforce_encode_guarantees` already raises `VideoEncodeError` with a message naming the offending access unit; a B-frame guarantee belongs beside those. Note it receives `list[AccessUnit]`, and each unit carries `.data`, so the stream is `b"".join(unit.data for unit in access_units)`.

## What to build

A doctor finding when a canonical video message carries a B picture, and an encode guarantee that HFlow's own encoder never emits one.

**One constraint that is easy to trip over.** The two readers report an unparseable slice header differently, and #358 went out of its way to keep `count_h264_pictures`'s two messages byte-identical because `hflow doctor` displays them:

```
count_h264_pictures        "slice header has no complete first_mb_in_slice value"
                           "slice header truncates its first_mb_in_slice value"
scan_picture_coding_types  "a slice header is incomplete or truncated; picture coding types
                            cannot be classified"
```

`video-invalid-slice-header`'s text must not change. Whether you keep `count_h264_pictures` for the count and add the scan alongside, or take the count from `scan_picture_coding_types(...).picture_count` and translate its failure back into the existing text, is your call; say which you picked and why.

## Definition of done

1. `hflow doctor` reports an error finding for a canonical video message whose picture is a B picture, with a new code documented in the `docs/FORMAT.md` finding-code table.
2. `video-invalid-slice-header`'s message text is unchanged for both existing failure kinds. The current tests covering them pass untouched.
3. `_enforce_encode_guarantees` raises `VideoEncodeError` when the encoder produces a B picture, naming the access unit.
4. Both are proven by deleting the new check and watching a test go red, with the result in the PR body.
5. A conforming canonical episode still produces a clean doctor report, and `encode_images_to_h264` still succeeds. No existing fixture starts failing.
6. Doctor's per-message cost does not regress measurably. It runs this path once per video message, so do not walk the NALs twice; include a before and after measurement.
7. `docs/FORMAT.md:161-162` is updated, since it currently documents this gap as open.

## Non-goals

- Changing the refusals in `write_access_units_to_mp4` or the transform
- Changing what `bframes=0` the encoder passes to x264
- A new doctor severity level, or making this a warning rather than an error
- Reworking the shared NAL walk from #358

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
  tests/test_doctor.py tests/test_processing_regressions.py
uv run pytest -q
```

When Markdown changes, run the link check from CONTRIBUTING.md.

## Notes

Not a good first issue: it edits the canonical validation path and an error contract that `hflow doctor` displays, and getting it wrong either misses the constraint or starts failing conforming files.

Sagar-024:
Closes #368.

- Doctor now classifies picture coding types. _check_video_payload takes the count from scan_picture_coding_types(...).picture_count and reports a video-b-picture error finding when b_picture_count > 0, one NAL walk on the hot path, no double walk.
- On an unparseable slice header the doctor delegates to count_h264_pictures on the error path and lets its own ValueError emit the finding, so video-invalid-slice-header stays byte-identical for both pinned kinds and the existing guard tests pass untouched.
- Precedence: a malformed slice header wins over any B claim. The B finding is additive and never fires on an unclassifiable payload.
- _enforce_encode_guarantees scans the joined stream and raises VideoEncodeError naming the first access unit that carries a B picture. The per-unit rescan runs only on the error path.
- Mutation proof: deleting the doctor B check fails test_doctor_reports_a_b_picture_from_slice_headers, deleting the encoder B check fails both encoder tests with DID NOT RAISE VideoEncodeError.
- Timing, interleaved A/B, median of warm iterations, conforming bframes=0 streams: 300 messages 2.64 ms before and 2.96 ms after (+12.1%), 3000 messages 24.67 ms before and 26.63 ms after (+7.9%). The delta is the same count-vs-scan gap you measured on #358, now buying the doctor the B classification.
- docs/FORMAT.md closes both gap notes and adds the video-b-picture row. 1415 passed / 6 skipped, ruff, format, and ty clean.

### DIFF
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -158,18 +158,20 @@ to MP4 cannot represent the reorder tail: the trailing B frames are silently und
 the muxed file (measured in [#250](https://github.com/Hebbian-Robotics/hflow/issues/250): 303
 samples in, 301 decoded). The transform refuses pass-through video that carries B-frames, and
 the MP4 remux behind `Episode.video` refuses any B-frame payload outright, with an error naming
-the observed reorder depth and the frames at risk. `hflow doctor` still does not classify
-picture coding types (below), so a clean report does not prove this constraint; the refusals do.
+the observed reorder depth and the frames at risk. `hflow doctor` reports any B picture it
+classifies in a video message as the `video-b-picture` error finding (below), so a clean
+report now covers this constraint alongside the refusals.
 
 `hflow doctor <file.mcap> [more.mcap ...]` checks every file given and prints
 one report each in argument order; its aggregate result follows the
 [exit code rules](#exit-codes). It validates the container, summary, indexes,
 stamps, chunk purity (against the file's own group map, or a video-versus-state
 approximation when it has none), per-topic time order, per-group chunk time
-order, and the H.264 access-unit properties listed below. It does not
-currently classify H.264 picture coding types to detect
-B-frames, so a clean report is not proof of the unchecked no-B-frame
-constraint. The doctor also does not reject non-VCL NAL units before the first
+order, and the H.264 access-unit properties listed below. It classifies
+H.264 picture coding types to detect
+B-frames as the `video-b-picture` error finding. A payload whose slice headers cannot be
+parsed still cannot be classified and is reported as
+`video-invalid-slice-header`. The doctor also does not reject non-VCL NAL units before the first
 AUD, so a clean report does not prove the canonical AUD-first constraint. An
 unreadable or unparseable path is reported in place, in the same per-file shape
 (`[error] unreadable: ...`), and the run continues with the remaining files.
@@ -196,6 +198,7 @@ automation. An `error` breaks the canonical convention (or the MCAP spec); a
 | `topic-time-order` | error | A channel's `log_time` decreases between messages. |
 | `video-format` | error | A supported video message does not declare `format="h264"`. |
 | `video-invalid-slice-header` | error | The H.264 payload's picture count cannot be determined from its slice headers. |
+| `video-b-picture` | error | A video message's slice headers classify at least one picture as a B picture; canonical video requires no B-frames. |
 | `video-multiple-access-units` | error | A video message contains more than one picture or access unit. |
 | `video-not-aud-delimited` | error | No AUD is present, or VCL data precedes the first AUD. |
 | `video-keyframe-missing-parameter-sets` | error | A keyframe does not carry both SPS and PPS. |
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -167,14 +167,14 @@ one report each in argument order; its aggregate result follows the
 [exit code rules](#exit-codes). It validates the container, summary, indexes,
 stamps, chunk purity (against the file's own group map, or a video-versus-state
 approximation when it has none), per-topic time order, per-group chunk time
-order, and the H.264 access-unit properties listed below. It classifies
-H.264 picture coding types to detect
-B-frames as the `video-b-picture` error finding. A payload whose slice headers cannot be
-parsed still cannot be classified and is reported as
-`video-invalid-slice-header`. The doctor also does not reject non-VCL NAL units before the first
-AUD, so a clean report does not prove the canonical AUD-first constraint. An
-unreadable or unparseable path is reported in place, in the same per-file shape
-(`[error] unreadable: ...`), and the run continues with the remaining files.
+order, and the H.264 access-unit properties listed below. It classifies H.264
+picture coding types to detect B-frames as the `video-b-picture` error finding.
+A payload whose slice headers cannot be parsed still cannot be classified, and
+is reported as `video-invalid-slice-header` instead. The doctor does not reject
+non-VCL NAL units before the first AUD, so a clean report does not prove the
+canonical AUD-first constraint. An unreadable or unparseable path is reported
+in place, in the same per-file shape (`[error] unreadable: ...`), and the run
+continues with the remaining files.
 
 ### Doctor finding codes
 
--- a/src/hflow/doctor.py
+++ b/src/hflow/doctor.py
@@ -109,7 +109,21 @@ def _check_video_payload(
         )
         return
     try:
-        picture_count = video_module.count_h264_pictures(payload)
+        try:
+            coding_scan = video_module.scan_picture_coding_types(payload)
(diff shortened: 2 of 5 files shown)


kstonekuan (approved):
LGTM, merging. The constraint I was most worried about is the one you handled best.

I diffed every message and every doctor finding for the same payloads, main against this branch. Exactly one line changes:

```
 all-zero payload                       -> ['video-invalid-slice-header']
 terminating bit, truncated suffix      -> ['video-invalid-slice-header']
 valid first_mb, truncated slice_type   -> ['video-not-aud-delimited']
 well-formed I slice (type 2)           -> ['video-not-aud-delimited']
-well-formed B slice (type 1)           -> ['video-not-aud-delimited']
+well-formed B slice (type 1)           -> ['video-b-picture', 'video-not-aud-delimited']
 empty payload                          -> ['video-invalid-slice-header']
```

Both `count_h264_pictures` messages are byte-identical, and so is `ensure_access_unit_delimiter`'s behaviour through them. The third row is the subtle one: a header whose `slice_type` is truncated but whose `first_mb_in_slice` parses still counts and still reports no slice-header error, which is the per-field divergence #358 preserved. Delegating to `count_h264_pictures` on the error path rather than translating the scan's failure is what keeps that true, and it costs nothing on the hot path because it only runs when the scan already refused.

You also pinned that delegation, which I was about to check and did not have to. Removing it:

```
error-path delegation removed  -> 4 failed
  test_doctor_keeps_both_pinned_count_messages_when_the_scan_refuses
  test_doctor_reports_invalid_slice_header_over_b_picture_when_a_header_is_malformed
  test_nonconforming_ros2_video_is_reported
  test_nonconforming_video_is_reported
```

The first two are yours, written for exactly that property. The other two are pre-existing tests that would have caught it, which is the better kind of evidence.

Both new checks bite:

```
doctor B finding removed      -> test_doctor_reports_a_b_picture_from_slice_headers
encoder B guarantee removed   -> test_encode_guarantees_raise_on_a_b_picture_naming_the_unit,
                                 test_encode_guarantees_find_a_b_picture_in_a_later_unit
```

Scanning the joined stream first and only rescanning per unit on the error path is the right shape: the encoder pays one scan on success and the unit index costs nothing unless something is actually wrong.

On the timing, and this part is my fault rather than yours. Measured independently over real per-message access units from a conforming `bframes=0` stream:

```
              main       branch
 300 messages  3.84 ms    4.25 ms   (+10.7%)
3000 messages 38.16 ms   41.97 ms   (+10.0%)
findings on conforming video: [] on both sides
```

Your numbers and mine agree on the magnitude. DoD 6 said "does not regress measurably", which was me writing the wrong sentence: classifying a second Exp-Golomb field per NAL cannot be free, and what I actually meant was the clause right after it about not walking the NALs twice. You did that, and reported the delta instead of burying it, which is the correct response to a badly worded requirement. In absolute terms a 900-message episode pays about a millisecond against a doctor run that reads the whole MCAP.

Pushed one fixup, `dc638f7`: prose reflow only. The inserted sentences left that paragraph with lines between 30 and 96 columns where the rest wraps near 78. Also worth noting you closed only the gap this PR closes and left the non-VCL-NAL-before-first-AUD note standing, which a careless edit would have taken with it.

Gate clean: `ruff check`, `ruff format --check`, `ty check`, lychee on `FORMAT.md`, 1428 passed / 6 skipped.

Closes #368.

PR #374, opened 2026-09-03
fix: reject ingest URIs that name the data root

### LINKED ISSUE #369: Ingest URIs that name the data root itself are accepted
## Current behavior

`parse_data_root_relative_uri` ([`src/hflow/uri.py:10`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/uri.py#L10)) refuses a URI that escapes above the data root, at `:33`:

```python
normalized = normpath(candidate)
if normalized == ".." or normalized.startswith("../"):
    raise ValueError(f"{candidate!r} is not relative to the data root")
```

It does not refuse one that normalizes to the data root itself. Measured on current main:

```
'.'          -> accepted '.'
'./'         -> accepted './'
'a/..'       -> accepted 'a/..'
'a/b/../..'  -> accepted 'a/b/../..'
```

Each of those names a directory, not an episode, and the failure lands later and further away. With a local data root, `resolve_episode_reference` hands the data root directory itself to `App.process`:

```
'.'     -> /tmp/<data root>            is_dir=True
'a/..'  -> /tmp/<data root>/a/..       is_dir=False
```

With a bucket data root it is worse, because object stores do not normalize key segments the way a filesystem does. The literal `.` and `..` are sent as part of the key:

```
'.'     -> s3://bucket/root/.
'a/..'  -> s3://bucket/root/a/..
```

#314 centralized this parser exactly so the CLI, the workspace server, and the SDK agree about what a valid ingest URI is. "Names the data root rather than an episode" is a case the parser is now the right place to catch, and it already computes the normalized form it would need.

## Pattern to copy

The refusal one line above at `uri.py:33` is the model: normalize for the check, keep the candidate's own spelling in the returned value, and raise `ValueError` naming the offending URI. The three entry points already turn that into a CLI exit 2, an HTTP 400, and an SDK `ValueError` before any request, so nothing outside `uri.py` needs to change.

The boundary tests live in three places and all three should keep agreeing:

- [`tests/test_runtime_client.py:392`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_runtime_client.py#L392) parametrizes the refused shapes, and `:415` parametrizes the accepted ones.
- [`packages/hflow-server/tests/test_server_runtime.py:338`](https://github.com/Hebbian-Robotics/hflow/blob/main/packages/hflow-server/tests/test_server_runtime.py#L338) covers the server's 400.
- `tests/test_runtime_cli.py` covers the CLI's exit 2.

## What to build

Refuse a URI whose normalized form is `.`, with the same error shape as the existing refusals.

Keep `a/../b.mcap` accepted and unchanged in the value returned. #314's non-goals are explicit that a safe internal segment must not be rewritten into a different persisted identity, and there is a test asserting exactly that at `test_runtime_client.py:415`.

## Definition of done

1. `.`, `./`, `a/..`, and `a/b/../..` are refused by the parser.
2. `a/../b.mcap` is still accepted and still returned with its own spelling, not normalized to `b.mcap`.
3. The message names the URI, like the neighbouring refusals do.
4. The refusal appears at all three entry points without changing any of them, and the existing CLI exit 2 and server 400 behaviour is unchanged.
5. A test in each of the three boundary test files, added to the existing parametrize lists rather than as new functions.
6. Prove it bites: delete the new condition, watch those tests go red, put it back. Put the result in the PR body.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_runtime_cli.py tests/test_runtime_client.py \
  packages/hflow-server/tests/test_server_runtime.py
uv run pytest -q
```

## Notes

Small and self-contained: one condition in one function, plus a row in three existing parametrize lists. A good way to see how one parser at a boundary serves three entry points, which is the shape #314 was about.

fafnirkyu:
Fixes #369

## Summary

- Reject ingest URIs that normalize to the data root itself, including `.`, `./`, `a/..`, and `a/b/../..`.
- Preserve safe internal segments such as `a/../b.mcap` unchanged.
- Add regression coverage across the SDK, CLI, and server ingestion boundaries.

## Validation

- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run ty check`
- `uv run pytest -q` — 1417 passed, 6 skipped

### DIFF
--- a/packages/hflow-server/tests/test_server_runtime.py
+++ b/packages/hflow-server/tests/test_server_runtime.py
@@ -344,7 +344,15 @@ def test_ingest_rejects_absolute_and_escaping_uris_before_any_runtime(
     data_root = tmp_path / "bare-root"
     data_root.mkdir()
     client = _client_over(data_root, unbuilt_assets_dir)
-    for hostile_uri in ("/etc/passwd", "../../etc/shadow", "sub/../../escape.mcap"):
+    for hostile_uri in (
+        ".",
+        "./",
+        "a/..",
+        "a/b/../..",
+        "/etc/passwd",
+        "../../etc/shadow",
+        "sub/../../escape.mcap",
+    ):
         response = client.post(
             "/api/v1/runtime/ingest",
             json={"uris": [hostile_uri], "profile": "full", "mode": "batch"},
--- a/src/hflow/uri.py
+++ b/src/hflow/uri.py
@@ -30,7 +30,7 @@ def parse_data_root_relative_uri(uri: str) -> DataRootRelativeUri:
 
     # Preserve the candidate itself; normalize only this containment check.
     normalized = normpath(candidate)
-    if normalized == ".." or normalized.startswith("../"):
+    if normalized == "." or normalized == ".." or normalized.startswith("../"):
         raise ValueError(f"{candidate!r} is not relative to the data root")
 
     return DataRootRelativeUri(candidate)
--- a/tests/test_runtime_cli.py
+++ b/tests/test_runtime_cli.py
@@ -916,7 +916,7 @@ def test_ingest_rejects_unknown_profile(pipeline_file: Path, tmp_path: Path) -> 
     assert exit_info.value.code == 2
 
 
-@pytest.mark.parametrize("uri", ["/abs/x.mcap", "../x.mcap"])
+@pytest.mark.parametrize("uri", [".", "./", "a/..", "a/b/../..", "/abs/x.mcap", "../x.mcap"])
 def test_ingest_rejects_uris_outside_data_root(
     uri: str,
     pipeline_file: Path,
--- a/tests/test_runtime_client.py
+++ b/tests/test_runtime_client.py
@@ -376,6 +376,10 @@ def test_ingest_refuses_a_batch_count_the_run_could_not_honour(stub_server: str)
     "uri",
     [
         "   ",
+        ".",
+        "./",
+        "a/..",
+        "a/b/../..",
         "/etc/passwd",
         "../outside.mcap",
         r"\episodes-in\a.mcap",


kstonekuan (approved):
LGTM, merging. Welcome.

One condition and three parametrize rows is exactly the right size for this, and both directions are held. Removing the condition:

```
the data-root condition removed  ->  9 failed
  test_ingest_rejects_invalid_uri_before_making_http_requests        (SDK)
  test_ingest_rejects_uris_outside_data_root                         (CLI)
  test_ingest_rejects_absolute_and_escaping_uris_before_any_runtime  (server)
```

All three boundaries, which is the property #314 built the shared parser for.

The other direction matters just as much here, because the obvious "simplification" of this check is to refuse anything whose normalized form differs from the candidate. That would be wrong, and it is caught:

```
over-refusal: any normalization difference rejected  ->  test_ingest_accepts_safe_uris_unchanged fails
```

So the non-goal from #314 about not rewriting `a/../b.mcap` into a different persisted identity is protected in both directions rather than just documented. Confirmed the behaviour directly:

```
refused:   '.'  './'  'a/..'  'a/b/../..'  '..'  '../x'  '/abs/x'
accepted:  'a/../b.mcap'      -> 'a/../b.mcap'        (spelling kept, not rewritten to 'b.mcap')
           '  a/../b.mcap  '  -> 'a/../b.mcap'        (trimmed, per DoD 3 on #314)
           'CC:/notadrive.mcap' -> unchanged          (two letters is not a drive)
```

Gate clean with current main merged in: `ruff check`, `ruff format --check`, `ty check`, 1436 passed / 6 skipped.

One optional thing, not worth a push. `normalized == "." or normalized == ".."` reads a little easier as `normalized in {".", ".."}`. Ruff does not flag it and the meaning is identical, so take it or leave it.

Closes #369. If you want another, the starter list is here: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

PR #375, opened 2026-09-03
test(lerobot): use realistic valid shas in cache-namespacing test

### LINKED ISSUE #371: The importer cache namespacing test uses shas its own validator would reject
## Current behavior

`_hf_repo_info` refuses a commit sha Hugging Face returns that is not hexadecimal, at [`src/hflow/importers/lerobot.py:239-240`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L239-L240):

```python
if not re.fullmatch(r"[0-9a-f]{7,64}", resolved_revision):
    raise ValueError(f"Hugging Face returned a malformed commit sha for {repo_id}@{revision}")
```

That guard exists because the resolved sha became a path component: the importer cache is namespaced by it, so a value like `../..` would place the cache outside the intended directory.

`test_import_namespaces_source_cache_by_resolved_revision` ([`tests/test_lerobot_converter.py:393`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_lerobot_converter.py#L393)) is the test for that namespacing, and its fixtures are values the guard would reject:

```python
resolved_shas = {"branch-a": "sha-a", "branch-b": "sha-b", "tag-a": "sha-a"}
...
assert sorted(path.name for path in (tmp_path / "_lerobot_cache").iterdir()) == ["sha-a", "sha-b"]
```

`sha-a` is not `[0-9a-f]{7,64}`. The test passes because it stubs `_hf_repo_info` wholesale, so the validator never sees those values.

Nothing is broken. The two tests just tell different stories about what a resolved sha is, and the one that would catch a regression is the one whose fixtures could never occur.

## Pattern to copy

The parametrized refusal test immediately below it, at [`tests/test_lerobot_converter.py:444`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_lerobot_converter.py#L444), uses realistic hostile values (`"../../evil"`, `"/tmp/probe-328-absolute"`) for exactly this guard. The namespacing test above it should use realistic *valid* ones for the same reason.

`tests/test_lerobot_converter.py` already has a hex fixture to copy the shape from: other tests in the file use full 40-character shas.

## What to build

Replace `sha-a` and `sha-b` with distinct hexadecimal shas the validator accepts, and update the three assertions that name them. The test's meaning does not change: two revisions resolving to the same sha share one cache directory, a different sha gets its own.

Consider whether the test should stop stubbing `_hf_repo_info` wholesale and let the real validator run, so the fixtures are checked rather than assumed. If that turns out to pull in more than this issue is worth, say so in the PR and leave it; the fixture change alone is the deliverable.

## Definition of done

1. `test_import_namespaces_source_cache_by_resolved_revision` uses shas that satisfy `[0-9a-f]{7,64}`.
2. The test still asserts the same thing: same resolved sha reuses one cache directory, a different one gets its own.
3. Distinct shas stay visibly distinct in the assertions, so a reader can still see which is which. Two 40-character strings differing in the last character are worse than two obviously different ones.
4. The rest of the file is unchanged, and the whole suite passes.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

## Notes

Small on purpose. The point is not the diff, it is the habit: a fixture that production code would reject is a test that cannot fail for the right reason later.

rakesh0x:
## What was wrong

`test_import_namespaces_source_cache_by_resolved_revision` proves the cache is namespaced by the *resolved* Hugging Face sha. It stubbed `_hf_repo_info` to return the values `sha-a` and `sha-b` — neither of which would survive the production validator at `src/hflow/importers/lerobot.py:239-240` (`r"[0-9a-f]{7,64}"`). The test passed only because the stub bypassed the validator, so a future bug in the validator or in the cache-path construction that consumes the resolved sha could not manifest in this test.

A test that uses values its own code path would reject cannot fail for the right reason.

## What changed

- `tests/test_lerobot_converter.py`: replaced the non-hex `sha-a` / `sha-b` literals with two 40-character hexadecimal shas and bound them to local variables `sha_a` and `sha_b`. The fixture dict and both assertions now reference the variables, so the test no longer contradicts the production validator.
- No production code changes. `_hf_repo_info`'s validator, the cache-path construction, and the rest of the test file are untouched.
- The test's contract is unchanged: two revisions that resolve to the same sha share one cache directory, a different sha gets its own. `branch-a` and `tag-a` still resolve to the same sha on purpose — that is the two-revisions-one-cache leg of the contract.

The two shas were chosen so a reader can tell them apart at a glance (one starts with `a1b2`, the other with `f0e1`), not just at the last character.

## How the fix was validated

- `uv run ruff format` — passed (no formatting changes)
- `uv run ty check` — passed
- `uv run pytest -q tests/test_lerobot_converter.py` — **35 passed**
- `uv run pytest -q` — **1421 passed, 12 skipped** (the two pre-existing `test_checks.py` failures around `camera_motion` and Lucas-Kanade optical flow are unrelated to this change — they reproduce on plain `origin/main` without this branch's commit and concern a different module; out of scope for this PR).

## Proof the relevant test passes

```
$ uv run pytest -q tests/test_lerobot_converter.py::test_import_namespaces_source_cache_by_resolved_revision
.                                                                        [100%]
1 passed in 0.12s
```

## Out of scope

The two `test_checks.py` failures observed on the full suite are pre-existing and unrelated to this change. They will be addressed in a separate PR.

closes #371

### DIFF
--- a/tests/test_lerobot_converter.py
+++ b/tests/test_lerobot_converter.py
@@ -393,7 +393,18 @@ def test_camera_selection_validates_keys(tmp_path: Path, monkeypatch: pytest.Mon
 def test_import_namespaces_source_cache_by_resolved_revision(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
-    resolved_shas = {"branch-a": "sha-a", "branch-b": "sha-b", "tag-a": "sha-a"}
+    # Realistic valid shas the production validator at
+    # src/hflow/importers/lerobot.py would accept: 40-character hexadecimal,
+    # visibly distinct at the start so a reader can tell them apart at a
+    # glance. ``branch-a`` and ``tag-a`` resolve to the same sha on purpose:
+    # they are the two-revisions-one-cache leg of the contract.
+    sha_a = "a1b2c3d4e5f60718293a4b5c6d7e8f9001020304"
+    sha_b = "f0e1d2c3b4a5968778695a4b3c2d1e0f00112233"
+    resolved_shas = {
+        "branch-a": sha_a,
+        "branch-b": sha_b,
+        "tag-a": sha_a,
+    }
     cache_observations: list[tuple[str, Path, str]] = []
 
     monkeypatch.setattr(
@@ -431,13 +442,13 @@ def test_import_namespaces_source_cache_by_resolved_revision(
         )
 
     assert cache_observations == [
-        ("sha-a", tmp_path / "_lerobot_cache" / "sha-a", "sha-a"),
-        ("sha-b", tmp_path / "_lerobot_cache" / "sha-b", "sha-b"),
-        ("sha-a", tmp_path / "_lerobot_cache" / "sha-a", "sha-a"),
+        (sha_a, tmp_path / "_lerobot_cache" / sha_a, sha_a),
+        (sha_b, tmp_path / "_lerobot_cache" / sha_b, sha_b),
+        (sha_a, tmp_path / "_lerobot_cache" / sha_a, sha_a),
     ]
     assert sorted(path.name for path in (tmp_path / "_lerobot_cache").iterdir()) == [
-        "sha-a",
-        "sha-b",
+        sha_a,
+        sha_b,
     ]
 
 


kstonekuan (approved):
LGTM, merging.

Both shas satisfy the validator that made this worth changing, and the old ones did not:

```
sha_a      len=40  [0-9a-f]{7,64} accepts: True
sha_b      len=40  [0-9a-f]{7,64} accepts: True
'sha-a'    len= 5                 accepts: False
```

`a1b2c3...` against `f0e1d2...` is the right call on DoD 3: distinguishable at a glance, so the two-revisions-one-cache leg still reads as two revisions and one cache rather than two forty-character blurs. Naming them as constants and commenting why `branch-a` and `tag-a` share one is what keeps that legible.

On the out-of-scope note: those two `test_checks.py` failures do not reproduce here. On your branch merged with current main, `tests/test_checks.py` is 35 passed and the full suite is 1436 passed / 6 skipped with nothing failing. The likely cause is a sync without the optional extras, which is the same trap that has caught two other contributors this week:

```bash
uv sync --locked --all-extras
uv run pytest -q
```

`--all-extras` is the documented form in CONTRIBUTING, and without it a handful of vision-dependent tests behave differently. Worth confirming before opening that separate PR, since there may be nothing to fix.

Closes #371.

PR #380, opened 2026-09-03
fix(video): validate pass-through fixed GOP cadence

victorwon2001:
## Summary

- Add `video-keyframe-cadence` to `hflow doctor`, comparing structurally valid video keyframes with the message-index grid implied by stamped `gop_seconds` and the channel's measured frame rate.
- Preserve current pass-through transform behavior: irregular GOP cadence is not a `SourceNotConforming` refusal, including when the existing lossless AUD repair is needed.
- Add regression coverage proving AUD repair preserves the original H.264 payload suffix while the repaired irregular stream is diagnosed by `hflow doctor`; non-finite GOP-frame calculations are reported as cadence errors rather than raising.

Refs #376

## Why

Real-corpus evidence showed that enforcing fixed GOP during pass-through ingest would reject a recording used to validate the existing lossless AUD-repair path. The change is therefore limited to diagnosis; pass-through `provenance/v1.gop_seconds` remains the configured target, and a separate provenance revision can distinguish configured and measured cadence without changing canonical bytes in this PR.

## Validation

The focused processing/doctor suite passed with 74 tests, and the full root suite passed with 1728 tests and 6 skips. Ruff, formatting, typing, documentation links, and diff whitespace validation also passed.

```bash
[...]

PR #381, opened 2026-09-03
refactor(lerobot): type metadata at import boundary

### LINKED ISSUE #320: Parse LeRobot metadata into a typed internal archive model

victorwon2001:
## Summary

Parse LeRobot `meta/info.json` once into frozen internal dataset, episode, video-window, and numeric-feature values while keeping the downloaded JSON cache unchanged. Conversion now consumes those typed values directly, with existing paths, schemas, provenance, manifest contents, revision namespacing, and video-cache behavior preserved.

## Why

LeRobot feature specifications remained as nested raw dictionaries after the external JSON boundary, allowing metadata interpretation to be repeated later in discovery and conversion. Moving supported validation and normalization to the boundary makes invalid supported shapes unrepresentable internally without expanding dtype/rank support or changing canonical output behavior.

Fixes #320

## Validation

- `uv run ty check` — passed
- `uv run pytest -q tests/test_lerobot_converter.py` — 47 passed
- `uv run pytest -q` — 1452 passed, 6 skipped
- `git diff --check` — passed

## Checklist

- [ ] I updated documentation for changed behavior, flags, formats, or requirements. Not applicable: this keeps the public behavior and formats unchanged.

### DIFF
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -24,9 +24,9 @@ import subprocess
 import tempfile
 import urllib.request
 from collections.abc import Sequence
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from pathlib import Path
-from typing import NotRequired, TypedDict
+from typing import TypedDict
 from urllib.parse import urlsplit
 
 from mcap.writer import Writer as McapWriter
@@ -50,7 +50,17 @@ EPISODE_START_TIME_NS = 1_755_000_000_000_000_000
 HUGGING_FACE_TOKEN_ENVIRONMENT_VARIABLES = ("HF_TOKEN", "HUGGING_FACE_HUB_TOKEN")
 
 
-class _EpisodeRow(TypedDict):
+@dataclass(frozen=True)
+class _VideoWindow:
+    camera_key: str
+    chunk_index: str
+    file_index: str
+    from_timestamp: float
+    to_timestamp: float
+
+
+@dataclass(frozen=True)
+class _EpisodeRow:
     episode_index: int
     task: str
     length: int
@@ -58,14 +68,7 @@ class _EpisodeRow(TypedDict):
     data_file: str
     data_from: int
     data_to: int
-    video_windows: NotRequired[dict[str, "_VideoWindow"]]
-
-
-class _VideoWindow(TypedDict):
-    chunk_index: str
-    file_index: str
-    from_timestamp: float
-    to_timestamp: float
+    video_windows: tuple[_VideoWindow, ...] = ()
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


kstonekuan:
Heads-up for when you reopen this: #383 landed a few minutes ago and rewrote the same region. `_ensure_source_archive` now reads episode rows, video columns, and video windows from one `read_parquet([...], union_by_name=true)` relation over every `meta/episodes` shard instead of the first, and `CONVERTER_VERSION` moved to v5. Expect a real conflict rather than a clean rebase.

Nothing here is wasted. #320 asks for the metadata to be typed at the boundary, and #383 only changed where the rows come from, not what is done with them once read. The boundary is arguably cleaner to type now that there is one relation instead of a per-file loop.

On the ordering: the cap is one open pull request at a time, so #381 and #384 both closed on open. That is working as intended, but it means the queue is yours to sequence. #380 is the one currently open and it is waiting on you, so the cheapest path is to land or close that first, then reopen whichever of these two you want next. Both issues are still unassigned and unclaimed.

PR #383, opened 2026-09-04
fix(lerobot): read episode metadata from every v3 shard (fixes #293)

### LINKED ISSUE #293: LeRobot import loses episode metadata after the first v3 shard
## Current behavior

LeRobot Dataset v3 episode metadata may be split across files below `meta/episodes/`. The importer downloads each tree entry to its basename at [`src/hflow/importers/lerobot.py:331-341`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L331-L341).

That flattens paths such as:

```text
meta/episodes/chunk-000/file-000.parquet
meta/episodes/chunk-001/file-000.parquet
```

Both become `_lerobot_cache/meta/episodes/file-000.parquet`. The second entry sees the first file and skips its download. The same local file is then appended twice and read twice.

There is a second first-shard assumption at [`src/hflow/importers/lerobot.py:379-438`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L379-L438). The importer discovers video columns and reads video-window rows only from `episode_metadata_files[0]`, so episodes from later metadata files get no indexed video window.

Verified on `main` at `1f5c7c8`. The current synthetic corpus has one metadata file, so neither outcome is covered.

## Pattern to copy

Extend [`test_index_discovery_multi_camera_metadata`](https://github.com/Hebbian-Robotics/hflow/blob/main/tests/test_lerobot_converter.py#L167-L215). It already builds v3 metadata, replaces the Hugging Face network boundary, and asserts parsed episode and camera outcomes.

The data-file cache in [`src/hflow/importers/lerobot.py:656-668`](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L656-L668) keeps chunk and file identities distinct.

## What to build

Preserve enough of each `meta/episodes` tree path that files from different chunks cannot collide. Read episode rows and video windows from every downloaded metadata file.

Add a synthetic corpus with at least two metadata shards. Use the same basename in different chunk directories, put a different episode in each shard, and give each episode a distinct video window.

## Definition of done

1. Metadata files with the same basename in different chunk directories use different local paths.
2. Every metadata shard is downloaded or reused from its own cache entry.
3. Episode rows from all shards appear exactly once and remain sorted by `episode_index`.
4. Video chunk, file, and timestamp windows from every shard are attached to the matching episode.
5. The regression fails against current `main` and passes with the fix.
6. The one-shard and multi-camera behavior stays unchanged.
7. Bump `CONVERTER_VERSION` if the fixed case can currently publish different canonical bytes.
8. No new runtime dependency is added.

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

akshatpatel64:
## Summary

Multi-shard LeRobot Dataset v3 repositories now import every episode with its own video window and its own data. Before this change, two things went wrong once `meta/episodes` had more than one parquet file:

- Video columns and video windows were read from the first shard only (`lerobot.py:459` and `:482` on `main`). Every episode in a later shard converted with no window, so the whole chunk video was transcoded in place of its slice. With real frame counts that shows up as a `frame count mismatch` error; when the counts happen to agree, the wrong footage is published.
- Shards were cached under their basename (`:413`). A second chunk directory that reuses `file-000.parquet` looked already downloaded and was skipped, and the first shard was read twice. Two episodes became four published files, one of them stamped `source_episode_index=1` with episode 0's task and data, and `prepared-manifest.json` counted four.

Measured on `lerobot/droid_1.0.1` (v3.0, 95,658 episodes, seven metadata shards, revision `0eabc778`), running the first two shards through `_ensure_source_archive`: 14,152 of 28,305 episodes had no video window on `main`, 0 with this change. I measured two shards, not seven; every episode outside shard 0 is affected, so the full corpus would lose roughly 81,500 windows.

## Why

Both defects come from the same assumption, that `meta/episodes` is one file. The fix removes the assumption in the two places it lived:

- `_episode_metadata_cache_path` keeps the tree path below `meta/episodes` in the cache instead of flattening to the basename. The tree listing is remote input, so an entry that is absolute, contains `..`, or sits outside `meta/episodes/` is refused with an error that names the repository, matching the other Hugging Face boundary errors in this module.
- Episode rows, video columns, and video windows come from one `read_parquet([...], union_by_name=true)` relation over every shard. That also drops the per-file loop and the post-hoc sort.
- `CONVERTER_VERSION` moves to `lerobot-converter-v5`, following #353. Multi-shard corpora previously published episodes with the wrong window or the wrong source episode under the same output names, so those outputs should not share an identity with corrected ones. `TRANSFORM_BEHAVIOR_VERSION` does not move: the transform still writes identical bytes for identical input. This does re-mint content IDs for single-shard re-imports such as pusht; if you would rather keep v4 and treat the old multi-shard outputs as simply wrong under it, I will change that.
- One sentence in `docs/how-to/import-lerobot-v3.md`.

Tests: `test_index_discovery_reads_every_metadata_shard` is parametrized over both shard layouts (distinct basenames in one chunk, which is droid's shape, and the same basename in the next chunk). It asserts each shard downloads once to its own cache entry and is reused on a second discovery, that episode indexes come back as `[0, 1, 2, 3]`, and that every episode carries its own per-camera window. On `main` the first layout fails on the missing window and the second on the skipped download, so each defect is pinned on its own. `test_episode_metadata_cache_path_refuses_entries_outside_the_metadata_tree` covers the trust boundary.

Alternatives I did not take: running the video-window query per shard works but leaves two places that must agree on the shard list. Deduplicating `episode_rows` by `episode_index` would hide the collision instead of fixing it and would mask a genuinely malformed corpus. Reusing `hflow.storage._validated_relative_key` for the path check would work, but its message talks about storage keys. Stale flattened cache files from earlier runs are left alone; they are small and no longer read.

## Validation

```bash
uv run ruff check --fix                                   # All checks passed!
uv run ruff format                                        # 212 files already formatted
uv run ty check                                           # All checks passed!
uv run pytest -q tests/test_lerobot_converter.py          # 44 passed
uv run --python 3.14 --locked pytest -q                   # 1 failed, 1453 passed, 11 skipped
uv run --python 3.11 --locked pytest -q                   # 1 failed, 1453 passed, 11 skipped
```

The one failure is pre-existing and unrelated: `test_preview_stats_render_timestamps_in_utc_on_a_non_utc_host` asserts `"-04" not in bound_value` and fails on the fourth day of any month because the date itself contains `-04`. It fails the same way on `main` today (1 failed, 1447 passed, 11 skipped). Happy to send a one-line fix separately once this PR is closed out, given the one-open-PR rule.

## Checklist

- [x] I preserved stored-data compatibility or documented an explicit version change (`CONVERTER_VERSION` v4 to v5; canonical format and `TRANSFORM_BEHAVIOR_VERSION` unchanged).

### DIFF
--- a/docs/how-to/import-lerobot-v3.md
+++ b/docs/how-to/import-lerobot-v3.md
@@ -84,8 +84,10 @@ channels named `/observation.state` and `/action`.
 | Language tensors, depth maps, or arbitrary nested features | Not supported |
 
 The importer reads `meta/info.json`, the episode index, Parquet feature
-columns, frame rate, episode boundaries, and video-path templates. It refuses
-an unsupported or missing required feature before publishing an episode.
+columns, frame rate, episode boundaries, and video-path templates. Episode
+metadata split across several `meta/episodes` shards is read in full, so
+every episode keeps its own video window. It refuses an unsupported or
+missing required feature before publishing an episode.
 
 Every output records the source repository, resolved commit, source episode
 index, task, embodiment, importer version, and FFmpeg build. Video is encoded
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -25,7 +25,7 @@ import tempfile
 import urllib.request
 from collections.abc import Sequence
 from dataclasses import dataclass
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import NotRequired, TypedDict
 from urllib.parse import urlsplit
 
@@ -42,8 +42,14 @@ DEFAULT_REVISION = "main"
 DEFAULT_OUTPUT_DIR = Path("./data/lerobot_pusht")
 DEFAULT_CAMERA_KEY = "observation.image"
 
-CONVERTER_VERSION = "lerobot-converter-v4"
+# "v5": episode metadata is read from every meta/episodes shard, not only the
+# first, and shards are cached under their own chunk directory (#293).
+# Multi-shard corpora previously published episodes with the wrong video
+# window or the wrong source episode, so their outputs must not share an
+# identity with the corrected ones.
+CONVERTER_VERSION = "lerobot-converter-v5"
 PRESENTATION_TIMESTAMP_EPSILON_S = 0.050
+EPISODE_METADATA_TREE_PREFIX = PurePosixPath("meta/episodes")
 
 # Timestamp handling
 NANOSECONDS_PER_SECOND = 1_000_000_000
@@ -371,6 +377,32 @@ def _download_file(url: str, destination_path: Path, chunk_size: int = 1 << 20) 
             raise
 
 
+def _episode_metadata_cache_path(
+    episodes_metadata_directory: Path, tree_entry_path: str, *, repo_id: str
+) -> Path:
+    """Where one ``meta/episodes`` tree entry lands in the local cache.
+
+    The path below ``meta/episodes`` is kept rather than flattened to its
+    basename. Dataset v3 shards episode metadata as
+    ``chunk-XXX/file-YYY.parquet`` and reuses file names across chunk
+    directories, so two shards flattened to one cache file would make the
+    second look already downloaded and its episodes vanish (#293). The tree
+    listing is remote input, so an entry that would land outside the
+    metadata directory is refused rather than joined.
+    """
+    tree_path = PurePosixPath(tree_entry_path)
+    try:
+        relative_path = tree_path.relative_to(EPISODE_METADATA_TREE_PREFIX)
+    except ValueError:
+        relative_path = None
+    if relative_path is None or not relative_path.parts or ".." in relative_path.parts:
+        raise ValueError(
+            f"Hugging Face tree response for {repo_id} lists {tree_entry_path!r}, which is "
+            f"not a file below {EPISODE_METADATA_TREE_PREFIX}/"
+        )
+    return episodes_metadata_directory / relative_path
+
+
 def _ensure_source_archive(dataset_source: DatasetSource, cache_dir: Path) -> _SourceArchive:
     """Download the corpus parquets and video chunks needed for the given episodes."""
     import duckdb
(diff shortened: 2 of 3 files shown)


akshatpatel64:
CI note: both checks jobs fail on one pre-existing test, test_preview_stats_render_timestamps_in_utc_on_a_non_utc_host. It asserts "-04" not in the bound value, and the runner's date is 2026-09-04, so the date string itself trips it. It fails the same way on main today and passes on any day that is not the 4th. Everything else is green: 1458 passed on 3.11 and 3.14. I kept this PR to #293; if you would rather I fold a one-line fix for that assertion in here instead of a follow-up, say so and I will push it.

kstonekuan (approved):
LGTM, merging. Welcome, and measuring the defect on `droid_1.0.1` before fixing it is the right way to open a first PR.

The parametrization is doing real work rather than decorating the test. Each defect fails on its own shard layout:

```
mutation                                   fails
cache keyed by basename again              [chunk-001/file-000.parquet] only
episode rows from the first shard only     both layouts
video windows from the first shard only    both layouts
traversal refusal removed                  all four refusal cases
```

The first line is the point: the collision only exists when a later chunk reuses a basename, so a test built solely on droid's shape (distinct names in one chunk) would have missed it entirely. Keeping the tree path below `meta/episodes` instead of flattening is the fix for the cause rather than the symptom.

**Keep v5.** You asked whether to bump or treat the old multi-shard outputs as wrong under v4. Bump. A version constant cannot be conditional on how many shards a corpus happens to have, so the choice is between re-minting single-shard IDs that did not need it and letting a corrected output share an identity with a wrong one. The first costs a re-conversion; the second means a cache hit can serve footage from the wrong episode, which is the bug you are fixing. `TRANSFORM_BEHAVIOR_VERSION` staying put is right for the same reason: the transform's bytes did not change.

One observation, not a request. Moving column discovery onto the union relation is untested: pointing it back at the first shard leaves the whole suite green, because both shards in the fixture carry the same cameras. Real v3 shards are size-splits of one table and share a schema, so this is defensive rather than load-bearing, and I would not add a fixture asymmetry just to pin it. Worth knowing it is not covered.

Separately, `converter_version` is not asserted anywhere in the suite, so nothing catches an accidental change to a string that is part of episode identity. That is ours, not yours, and predates this.

Your date-sensitive failure report was right and it was ours: `"-04" not in bound_value` also matches the day in `2026-09-04` and the month in any April date. Fixed in #385, which credits you and @victorwon2001 for catching it independently. No need to send the one-liner.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1469 passed / 6 skipped, with the same #385 failure and nothing else.

Closes #293. If you want another, #191 is a good fit for what you have just been through: a reproducible real-camera LeRobot import through quality, curate, and export, and nobody is on it. The other thing worth your time, given you already have `droid_1.0.1` in hand, is pointing HFlow at a real corpus and reporting what breaks, what is slow, or what is awkward. #287 came out of exactly that and it is more useful to us than any issue we could write ourselves.

Issue #379, opened 2026-09-03
prepared-manifest.json records how many episodes were converted but not which ones, so a corrupt or missing episode cannot be detected from the delivery


Sagar-024:
## Current behavior

`import_lerobot_dataset` finishes by writing `prepared-manifest.json` ([src/hflow/importers/lerobot.py:676-691](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L676-L691)). Its complete contents are:

```json
{
  "schema_version": 2,
  "dataset": {"repo_id": "...", "revision": "...", "license": "..."},
  "camera_keys": ["..."],
  "episodes_converted": 3,
  "converter_version": "lerobot-converter-v4"
}
```

It records **that** episodes were converted and **how many**, but never **which**: no per-episode filename list, no size, no content address. Nothing else consumes the manifest at runtime (its only references are its own writer and a test's non-existence assertion), so this file is the delivery receipt a recipient of a prepared corpus actually has.

The content-addressing capability already exists in the codebase: `content_episode_id()` ([src/hflow/catalog.py:337-344](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/catalog.py#L337-L344)) hashes a canonical episode to a 16-hex sha256 and is load-bearing for curation dedupe ([src/hflow/app.py:2454](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/app.py#L2454), and the trade-off documented at [src/hflow/transform.py:225-232](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/transform.py#L225-L232)). The importer never calls it: `lerobot.py` contains zero references.

## Controlled result

Ran the real import path (`import_lerobot_dataset`, network and ffmpeg stubbed, 3 one-frame episodes) and then truncated one landing episode to zero bytes, as any disk accident or partial copy would:

```
--- delivery, as the buyer receives it ---
landing episodes: 3
  lerobot_episode_0001.mcap  4626 bytes
  lerobot_episode_0002.mcap  4626 bytes
  lerobot_episode_0003.mcap  4626 bytes
--- prepared-manifest.json (complete contents) ---
(above: schema_version, dataset, camera_keys, episodes_converted=3, converter_version)
episode filenames named in manifest: 0
hashes in manifest: 0
--- after one episode file is truncated to 0 bytes ---
  lerobot_episode_0003.mcap  0 bytes
[...]

PR #384, opened 2026-09-04
feat: resume interrupted LeRobot imports

### LINKED ISSUE #303: Make multi-episode LeRobot imports resumable after interruption

victorwon2001:
## Summary

Resume interrupted multi-episode LeRobot imports by reusing complete landing MCAPs only when their stored dataset revision, source episode, selected camera topics, converter version, and canonical pipeline version match the requested import. Bump `prepared-manifest.json` to schema v3 with canonical pipeline identity and per-episode output keys while preserving the existing top-level fields.

Fixes #303

## Why

A failure late in an import previously left valid earlier MCAPs but no durable way to identify them as completed on retry, so they were converted again. Invalid, summary-less, truncated, or identity-mismatched outputs now fall back to the existing atomic conversion/publish path; fresh conversion uses the same transform configuration, so canonical output behavior is unchanged.

## Validation

- `uv run ruff format --check` — passed
- `uv run ty check` — passed
- `uv run pytest tests/test_lerobot_converter.py -q` — 42 passed
- `uv run pytest -q` — 1455 passed, 6 skipped, 2 failed outside the LeRobot change paths; the bounded-concurrency test is timing-sensitive, and the UTC-preview test has a date-sensitive `-04` assertion on 2026-09-04, with the latter also reproducing against HEAD
- `lychee --no-progress --include-fragments --exclude '^https://github\.com/Hebbian-Robotics/hflow/(issues|security/advisories/new)$' --exclude-path references/mcap-spec.md --exclude-path references/foxglove-CompressedVideo.proto .` — passed with 0 errors

## Checklist

### DIFF
--- a/docs/how-to/import-lerobot-v3.md
+++ b/docs/how-to/import-lerobot-v3.md
@@ -29,7 +29,19 @@ data/lerobot_pusht/
 └── prepared-manifest.json          # source commit and import summary
 ```
 
-Re-running against the same output directory reuses downloaded source files.
+Re-running against the same output directory reuses downloaded source files. It also
+reuses a completed canonical episode when the resolved dataset commit, source episode,
+selected camera set, converter version, and canonical pipeline version all match the
+requested import. This makes an all-episodes import resumable at episode boundaries: if a
+later episode fails, retrying the same import keeps already completed episode bytes and
+converts only the remaining work.
+
+An existing landing file is reused only when it is a readable MCAP with a complete summary
+and matching identity metadata. Truncated, unreadable, or identity-mismatched files are
+converted again and atomically replaced only after the new episode succeeds. A branch or tag
+that resolves to a different commit is a different import identity, even when the revision
+name is unchanged.
+
 Run `uv run hflow doctor ./data/lerobot_pusht/landing/*.mcap` to print the
 canonical-format report.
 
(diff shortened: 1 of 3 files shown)


kstonekuan:
Before you reopen this: #390 is now an open PR against #303 and is further along, so reopening here would put two of you on the same issue. #303 still has no assignee, so nothing is formally reserved, but there is no point in you both writing it.

Worth knowing either way, because it applies to both attempts: stamping the import identity into `episode/v1` moves every episode's `content_episode_id`, since that hashes the whole canonical file. #390 hit it and needs a `CONVERTER_VERSION` bump. Your version would too.

The other half of this PR is gone: #389 merged the schema-3 per-episode entries for #379 about two hours ago, so the manifest work here is done.

#380 is still the one open under your name and it is waiting on you. Your other reopen candidate, #381 against #320, is untouched and nobody else is on it.

PR #388, opened 2026-09-04
test(lerobot): pin prepared-manifest and the converter version it carries

### LINKED ISSUE #386: prepared-manifest.json and the converter version it carries are never read back by a test
`prepared-manifest.json` is written by every LeRobot import and nothing reads it back in the test suite. The manifest test asserts only that the file exists:

```
tests/test_lerobot_converter.py:1107   assert (output_dir / "prepared-manifest.json").is_file()
```

So every field in it is unpinned. The one that matters most is `converter_version`, because it is part of episode identity rather than decoration. It is written into three places:

```
src/hflow/importers/lerobot.py:50     CONVERTER_VERSION = "lerobot-converter-v5"
src/hflow/importers/lerobot.py:739    "converter_version": CONVERTER_VERSION,   <- prepared-manifest.json
src/hflow/importers/lerobot.py:1022   "converter_version": CONVERTER_VERSION,   <- episode metadata
src/hflow/importers/lerobot.py:1028   "converter_version": CONVERTER_VERSION,   <- source-provenance/v1
```

`grep -rn converter_version tests/` returns nothing. Editing that constant, or dropping the key from any of the three writers, leaves the suite green. The constant moved from v4 to v5 in #383 for a real reason: outputs from a corrected importer must not share an identity with outputs from a broken one. A version that carries that much weight should not be changeable without a test noticing.

## What to do

Add a test that reads `prepared-manifest.json` back and asserts its contents, and a test that asserts the same version reaches the canonical episode.

`test_import_publishes_into_a_local_data_root` at `tests/test_lerobot_converter.py:1080` already runs a real import and has the manifest path in hand at `:1107`, so the manifest half can hang there or on a test in the same shape.

For reading a written JSON artifact back, copy `tests/test_sync_reuse.py:89`:

```python
marker_payload = json.loads(marker_path.read_text())
```

## Definition of done

1. `prepared-manifest.json` is parsed and its fields asserted: `schema_version`, `dataset` (`repo_id`, `revision`, `license`), `camera_keys`, `episodes_converted`, and `converter_version`.
2. `converter_version` is asserted against `prep.CONVERTER_VERSION` rather than a copy of the literal string, so the test pins that the manifest carries the module's version and does not need editing when the version legitimately moves.
3. Separately, one assertion that the version written into the canonical episode's `source-provenance/v1` metadata is the same value. That block is also unasserted today.
4. Changing `CONVERTER_VERSION`'s value alone still passes (it is not a literal-string pin); removing the key from any of the three writers fails.

## Not in scope

Do not add per-episode entries to the manifest or bump `schema_version`. That is #379 and it has a direction already. Expect this test to need one line updated when #379 lands, which is the point of having it.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

See [CONTRIBUTING.md](https://github.com/Hebbian-Robotics/hflow/blob/main/CONTRIBUTING.md).

ayam04:
Fixes #386

`prepared-manifest.json` was asserted only for existence (`is_file()`), so every field was unpinned — including `converter_version`, which is part of episode identity rather than decoration (the v4→v5 move in #383 is exactly the case that needed a net).

## What changed

**`tests/test_lerobot_converter.py`** (+82):

1. **Manifest pinned field by field** — in `test_import_returns_local_uris_and_keeps_cache_beside_landing`, the manifest is parsed and asserted as a full dict: `schema_version`, `dataset` (`repo_id`/`revision`/`license`), `camera_keys`, `episodes_converted`, and `converter_version` — the last against `prep.CONVERTER_VERSION`, not a literal copy, so it survives legitimate version moves.
2. **Canonical episode provenance pinned** — new test `test_converter_version_reaches_the_canonical_episode_provenance` runs the real `_convert_single_episode` mcap writer (network/ffmpeg stubbed, canonicalization stubbed with a copy exactly like the sibling converter tests, since fake access units are not parseable video: the staged→canonical metadata passthrough is verbatim at `transform.py:884-886`) and asserts `converter_version` in both the `episode/v1` and `source-provenance/v1` records of the landed episode.

## RED/GREEN matrix (DoD 4)

Neutering each of the three writers in turn and running the tests:

| writer neutered | result |
|---|---|
| manifest `converter_version` (`lerobot.py:739`) | `test_import_returns_local_uris_and_keeps_cache_beside_landing` FAILS |
| `episode/v1` metadata (`lerobot.py:1022`) | `test_converter_version_reaches_the_canonical_episode_provenance` FAILS |
| `source-provenance/v1` metadata (`lerobot.py:1028`) | `test_converter_version_reaches_the_canonical_episode_provenance` FAILS |
| constant value bumped alone (`v5` → `v5-testbump`) | both tests still PASS |

## Validation

- `uv run pytest tests/test_lerobot_converter.py` → 45 passed
- `uv run ruff check` / `ruff format --check` / `ty check` → clean
- `uv run pytest -q` (full suite) → 1466 passed, 6 skipped, **5 failed — all reproduced on a clean main worktree (stash-proven: none touch this change)**: `test_cli.py` module-form ×3, `test_end_to_end.py::test_canonical_episode_extracts_exact_source_frame_indices` (the known WSL ffmpeg filter_script failure), and `packages/hflow-server` non-UTC-host TZ test. Environment-only, zero overlap with this diff.

### DIFF
--- a/tests/test_lerobot_converter.py
+++ b/tests/test_lerobot_converter.py
@@ -1124,6 +1124,17 @@ def test_import_returns_local_uris_and_keeps_cache_beside_landing(
 def test_converter_version_reaches_the_canonical_episode_provenance(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
+    """What the importer writes into the episode it publishes, not what survives
+    canonicalization.
+
+    ``write_canonical_episode`` is stubbed to a copy, the idiom of the sibling
+    converter tests, because synthetic access units are not parseable video. So
+    the two records are asserted as written rather than as carried through: the
+    real transform reaches them by different routes, copying an unrecognized
+    record verbatim (``transform.py:886``) but skipping ``episode/v1`` there and
+    rewriting it from the source dict (``transform.py:889-890``). Neither route
+    is exercised here.
+    """
     corpus = _build_fake_corpus(tmp_path)
     camera_key = "observation.images.up"
     dataset_source = prep.DatasetSource(repo_id="fake/repo", revision="abc", license="apache-2.0")
--- a/tests/test_lerobot_converter.py
+++ b/tests/test_lerobot_converter.py
@@ -17,6 +17,7 @@ import pytest
 
 import hflow.importers.lerobot as prep
 from hflow.cli import main as cli_main
+from hflow.reader import open_reader
 from hflow.storage import LocalStorageRoot, StorageRoot
 
 _DERIVE = prep._derive_numeric_schema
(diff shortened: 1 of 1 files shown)


kstonekuan (approved):
LGTM, merging. Your RED/GREEN matrix reproduces exactly, all four rows:

```
manifest converter_version dropped        ..._returns_local_uris_and_keeps_cache_beside_landing
episode/v1 dropped                        ..._reaches_the_canonical_episode_provenance
source-provenance/v1 dropped              ..._reaches_the_canonical_episode_provenance
constant bumped alone                     45 passed
```

Asserting the manifest as a whole dict rather than field by field is the better choice and I would not have asked for it. It pins absence as well as presence, so a field appearing is a failure too. That matters immediately: #389 adds per-episode entries and moves `schema_version` to 3, so this test is what will make that a deliberate change rather than a silent one.

I pushed one fixup, `74afbe2`, and it is a docstring rather than a behavior change. The test does not exercise canonicalization, so it pins what the importer writes rather than what survives the transform. Both of these leave it green:

```
canonicalization stops carrying episode/v1 through        45 passed
canonicalization stops copying unknown records verbatim   45 passed
```

Stubbing there is right and it is the sibling tests' idiom, so this is a boundary worth naming rather than closing. One correction to your reasoning while it is on the record: the passthrough is not verbatim for both records. `transform.py:884-886` skips `episode/v1` and `provenance/v1` before copying, so `source-provenance/v1` rides the verbatim loop only because its name is not `provenance/v1`, and `episode/v1` is rebuilt at `:889-890` from a dict copy taken at `:787`. Both survive today, by two different routes, and neither route is covered here.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1490 passed / 6 skipped. Your five failures were environment. #385 fixed the non-UTC-host one about three hours after you opened this, and the ffmpeg `filter_script` and module-form CLI failures are WSL, not you.

Closes #386. One ask, and it is an invitation rather than a rule: `good first issue` exists to give newcomers somewhere to start, and at twenty-odd PRs you are well past needing it. I filed #386 eight minutes before this PR appeared, which is faster than any newcomer could have got to it. Please take from the advanced list instead: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3Aadvanced>

Given you wrote the v3 generalization in #230 and the export path in #273, the thing most worth your time is not an issue we wrote. Point HFlow at a real corpus (Egocentric-10K or Egocentric-100K on Hugging Face) and tell us what breaks, what is slow, or what is awkward. #287 came out of exactly that and it has been worth more than anything we would have thought to file.

PR #394, opened 2026-09-04
test(packaging): cover native overlay manifest refusals

### LINKED ISSUE #393: Every refusal in the native overlay manifest validator is untested, including the schema version in the bundle digest
`_validate_manifest` is the trust boundary for a native overlay manifest: an overlay directory is external input, and this function decides whether its `manifest.json` may be applied to a package. It has 27 refusal branches. None of them are tested.

```
$ grep -c CythonOverlayManifestError tests/test_packaging.py
0
```

The 17 tests in `tests/test_packaging.py` all exercise the apply path, and every `pytest.raises` in the file is on `CythonOverlayApplyError`:

```
tests/test_packaging.py:274   "does not match the build receipt"
tests/test_packaging.py:299   "target source is not a regular file"
tests/test_packaging.py:412   "no wheel RECORD"
tests/test_packaging.py:434   "conflicts with an existing native artifact"
tests/test_packaging.py:462   "ambiguous wheel RECORD ownership"
tests/test_packaging.py:484   "three fields"
tests/test_packaging.py:510   "cannot update a signed wheel RECORD"
```

That is good coverage of applying a valid overlay and refusing a stale one. The manifest validation that runs before any of it is uncovered.

## Verification

Against `main` at `1f725d5`, two mutations to the schema-version half:

```
_validate_manifest never refuses a version mismatch    1263 passed, nothing noticed
manifest omits schema_version entirely                 1263 passed, nothing noticed
```

The second is worth a second look. `CYTHON_OVERLAY_SCHEMA_VERSION` is part of the bundle digest payload at `src/hflow/packaging.py:1010`, so dropping it changes the digest every overlay is identified by, and the suite does not notice.

## What to do

Cover the refusals in `_validate_manifest` that are cheap to reach by building a valid manifest and changing one field. The branches worth taking first, all in `src/hflow/packaging.py`:

```
:928   unsupported native overlay schema version
:930   unsupported native overlay format
:937   artifacts must not be empty
:942   artifacts must be sorted by module_name
:945   artifact module names must be unique
:974   artifact paths must be unique
:984   bundle_digest does not match manifest components
```

## Pattern to copy

`tests/test_packaging.py:274` is the shape: build a real overlay with `build_cython_overlay`, perturb one thing, assert the refusal, then assert nothing was applied.

```python
with pytest.raises(CythonOverlayApplyError, match="does not match the build receipt"):
    apply_cython_overlay(overlay_directory, package_root)
```

A parametrized test over (field to perturb, expected message) fits these seven better than seven separate functions, since the setup is identical and only the perturbation differs.

## Definition of done

1. Each of the seven branches above is the only failure when its own `raise` is removed. Check this rather than assuming it: several of them are reachable through more than one path.
2. `schema_version` is asserted against `packaging.CYTHON_OVERLAY_SCHEMA_VERSION`, imported rather than written as `1`, so a deliberate bump moves the test and an accidental one fails it.
3. One test covers the digest: an overlay built with the constant excluded from `_calculate_bundle_digest`'s payload no longer validates. This is what makes the version load-bearing rather than decorative.
4. No test asserts on a full traceback or an internal frame; match on the message text the way the sibling tests do.

## Not in scope

The 20 other `CythonOverlayManifestError` sites, including the JSON-parsing and canonical-JSON refusals at `:638`, `:749`, and the `_require_*` helpers from `:1986` down. Those are worth covering too, and they can follow; this issue is the seven that one valid-manifest fixture reaches.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

See [CONTRIBUTING.md](https://github.com/Hebbian-Robotics/hflow/blob/main/CONTRIBUTING.md).

WilliamK112:
## Summary

- build a real two-artifact native overlay and perturb the schema version, format, empty artifact list, artifact order, and duplicate module name
- assert the specific manifest refusal and verify application leaves sources, native destinations, installed manifest, and wheel RECORD untouched
- add a load-bearing regression proving the schema-version constant participates in the bundle digest
- remove the artifact-path uniqueness refusal because it is unreachable after the unique-module-name and deterministic path/module checks

## Why the path check is removed

For any manifest reaching that branch, module names are already unique. Each source path and artifact path has also been required to equal the deterministic path derived from its module name. Those mappings are injective for validated dotted module names, so duplicate paths cannot reach the later set-membership check. A duplicate external input instead fails either module-name uniqueness or the earlier path/module-name match.

## Validation

- uv run ruff check --fix
- uv run ruff format
- uv run ty check
- uv run pytest -q --ignore=tests/test_packaging.py: 1469 passed, 11 skipped
- full local run on macOS: 1470 passed, 11 skipped; all 23 failures are the expected CPython/Linux-only native-overlay build guard (17 pre-existing packaging cases plus the 6 new cases)
- GitHub Actions complete suite on Linux, Python 3.11 and 3.14: passed
- GitHub Actions workspace example matrix: passed

AI assistance disclosure: I used OpenAI Codex to inspect, implement, and validate this change, and I reviewed the final diff and test evidence.

Closes #393

### DIFF
--- a/src/hflow/packaging.py
+++ b/src/hflow/packaging.py
@@ -943,8 +943,6 @@ def _validate_manifest(manifest: CythonOverlayManifest) -> CythonOverlayManifest
     module_names = [artifact.module_name for artifact in manifest.artifacts]
     if len(module_names) != len(set(module_names)):
         raise CythonOverlayManifestError("artifact module names must be unique")
-    artifact_paths: set[str] = set()
-    source_paths: set[str] = set()
     for artifact in manifest.artifacts:
         _require_dotted_name(artifact.module_name, "module_name")
         expected_source_path = _source_relative_path(manifest.package_name, artifact.module_name)
@@ -970,10 +968,6 @@ def _validate_manifest(manifest: CythonOverlayManifest) -> CythonOverlayManifest
         _require_nonnegative_integer(artifact.source_size_bytes, "source_size_bytes")
         _require_sha256(artifact.artifact_sha256, "artifact_sha256")
         _require_positive_integer(artifact.artifact_size_bytes, "artifact_size_bytes")
-        if artifact.source_path in source_paths or artifact.artifact_path in artifact_paths:
-            raise CythonOverlayManifestError("artifact paths must be unique")
-        source_paths.add(artifact.source_path)
-        artifact_paths.add(artifact.artifact_path)
     expected_bundle_digest = _calculate_bundle_digest(
         package_name=manifest.package_name,
         target=manifest.target,
--- a/src/hflow/packaging.py
+++ b/src/hflow/packaging.py
@@ -962,6 +962,11 @@ def _validate_manifest(manifest: CythonOverlayManifest) -> CythonOverlayManifest
             raise CythonOverlayManifestError(
                 f"artifact_path does not match module_name for {artifact.module_name}"
             )
+        # These two equalities are what make paths unique, given the
+        # module-name uniqueness checked above: both derivations are injective
+        # in the module name, so distinct names cannot converge on one path.
+        # Relaxing either equality (allowing a caller-chosen path, say) brings
+        # back the need for an explicit path-uniqueness refusal here.
         _require_relative_path(artifact.source_path, "source_path")
         _require_relative_path(artifact.artifact_path, "artifact_path")
         _require_sha256(artifact.source_sha256, "source_sha256")
--- a/tests/test_packaging.py
+++ b/tests/test_packaging.py
@@ -11,15 +11,18 @@ import subprocess
 import sys
 from dataclasses import replace
 from pathlib import Path
+from typing import cast
 
 import pytest
 
+import hflow.packaging as packaging
 from hflow.cli import main
 from hflow.packaging import (
     CYTHON_OVERLAY_MANIFEST_FILE_NAME,
     INSTALLED_CYTHON_OVERLAY_MANIFEST_FILE_NAME,
     CythonOverlayApplyError,
     CythonOverlayBuildConfig,
+    CythonOverlayManifestError,
     CythonOverlayVerificationCode,
     CythonOverlayVerificationIssue,
     apply_cython_overlay,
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging, and the removal was the right call. You did the part of the DoD that asks for judgment rather than typing: checking reachability instead of assuming it, then reporting what you found.

I tried to break the unreachability claim rather than agree with it, because `with_suffix` replaces rather than appends and `PurePosixPath` collapses empty parts, so both looked like places two module names could converge. They do not:

```
module names tried:       405
refused before deriving:  6
accepted:                 399
source_path collisions:   0
artifact_path collisions: 0
```

Both derivations are injective in the module name, and the manifest's paths are forced to equal them, so distinct unique names cannot share a path. Branch is dead.

I pushed one fixup, `bd732a4`, and it is a comment rather than a change. That reasoning lived only in your PR description, and the next person to relax either path equality (to allow a caller-chosen path, say) will be reading `_validate_manifest`, not this thread. The comment sits at the coupling that makes the guarantee true. Keeping the safety property as knowledge is better than keeping it as an untestable branch.

**Your digest test passes for the right reason, which I checked because it might not have.** It recomputes the digest over five keys with a hand-rolled canonicalization, so a mismatch could come from getting the canonicalization wrong rather than from omitting `schema_version`, and it would then keep passing if the constant were dropped from `_calculate_bundle_digest` entirely. It reproduces the real digest exactly when `schema_version` is put back:

```
real bundle_digest                  sha256:9372e337eb92...161ac29d
hand-rolled WITH schema_version     sha256:9372e337eb92...161ac29d
hand-rolled WITHOUT schema_version  sha256:0ca616339b91...af9896bc
```

So the omission is what causes the mismatch. Worth a line in the test saying that is why it is trustworthy, since the next person to touch `_calculate_bundle_digest`'s canonicalization will break this test and needs to know it is not spurious.

Asserting on every refusal that nothing was applied (sources intact, no artifacts, no installed manifest, RECORD byte-identical) is what makes these tests about the boundary rather than about the exception type. That is the sibling pattern used properly.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1514 passed / 6 skipped, and the 24 packaging tests all run here on Linux. Your macOS numbers were the build guard, as you said.

Closes #393.

**On the starter pool, and this is the second time today.** I asked on #364 five hours ago to leave `good first issue` for newcomers, and #393 is one, taken eight minutes after I filed it. I am not going to refuse the work, and the PR is good. But the pool only functions if there is something in it when a newcomer arrives, and right now it empties faster than I can fill it: #307 and #386 went the same way this morning, both to contributors with twenty-plus merged PRs.

So, concretely, rather than as an invitation I repeat and you reasonably keep ignoring: please filter the issue list by `advanced` and take from there, and treat anything labelled `good first issue` as not for you. <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3Aadvanced>

If nothing there appeals, the standing alternative is worth more to us than any of it: point HFlow at Egocentric-10K or Egocentric-100K and report what breaks, what is slow, or what is awkward. #287 came from that and has generated more real work than anything we wrote ourselves. You have been through the transform, the video path, and now the packaging boundary, so you would find things we cannot.

Issue #391, opened 2026-09-04
The dataset provenance sidecar records a schema version and coverage block that no test reads back


chiruu12:
## Current behavior

`hflow dataset create` publishes two files: the parquet manifest and a JSON provenance sidecar built by [`_sidecar_payload()`](https://github.com/Hebbian-Robotics/hflow/blob/623f9f6c789eda3cd306e933fb4c9e4be861302e/src/hflow/dataset.py#L298-L327). The sidecar is the artifact that explains a dataset without the pipeline in hand.

The only test that reads it back is `test_the_manifest_and_its_provenance_land_together` at [tests/test_dataset.py:162-167](https://github.com/Hebbian-Robotics/hflow/blob/623f9f6c789eda3cd306e933fb4c9e4be861302e/tests/test_dataset.py#L162-L167). It asserts `name`, `sql`, `row_count`, and three fields under `pipeline`. Five keys are never read:

```
src/hflow/dataset.py:311   dataset_manifest_version
src/hflow/dataset.py:313   created_at
src/hflow/dataset.py:316   total_episodes
src/hflow/dataset.py:317   coverage[]  (check_name, episodes_ran, total_episodes)
src/hflow/dataset.py:326   workspace_id
```

`DATASET_MANIFEST_VERSION` is the sharpest of these. `grep -rn DATASET_MANIFEST_VERSION src/ tests/` returns its definition at `src/hflow/dataset.py:42` and its single use at line 311, and nothing else in the repository. A schema version that no reader checks and no test pins can be bumped, or dropped, without anything noticing.

## Verification

Verified against `main` at `623f9f6`. I removed the `dataset_manifest_version` key, replaced the `coverage` list with `[]`, and forced `workspace_id` to `None`, all three at once.

```
python -m pytest tests/ -q
1236 passed
```

The whole suite stays green.

`coverage` is the one with weight beyond the version. It is the per-check record of how many episodes actually ran, which is what makes a dataset auditable rather than merely reproducible, and today an empty list reads the same as a full one.

## Pattern to copy

#386 makes this argument for `prepared-manifest.json` and `converter_version` on the importer side. The dataset sidecar is the same artifact class on the curation side, with the same hole, and the same fix shape: read the file back and assert its contents against the constant rather than a literal.

`test_the_manifest_and_its_provenance_land_together` already loads the sidecar and has an ingested fixture with a known check, so the coverage assertion has real numbers available to it.

## What to do

[...]

PR #396, opened 2026-09-04
fix(catalog): refuse a non-string interval label

### LINKED ISSUE #392: A non-string Interval label crashes append_episode, but only when another interval shares its bounds
## Current behavior

[`_normalized_intervals()`](https://github.com/Hebbian-Robotics/hflow/blob/623f9f6c789eda3cd306e933fb4c9e4be861302e/src/hflow/catalog.py#L418-L469) validates both interval bounds carefully: NumPy scalars are coerced, `bool` is refused because it subclasses `int`, out-of-BIGINT values are refused, negative and inverted spans are refused. Each error names the check and the bound.

`label` is never checked. `Interval.label` is declared `str = ""` at [src/hflow/steps.py:165](https://github.com/Hebbian-Robotics/hflow/blob/623f9f6c789eda3cd306e933fb4c9e4be861302e/src/hflow/steps.py#L165), and nothing enforces it. The value goes straight into the run fingerprint at [src/hflow/catalog.py:594](https://github.com/Hebbian-Robotics/hflow/blob/623f9f6c789eda3cd306e933fb4c9e4be861302e/src/hflow/catalog.py#L594), which sorts `(start_ns, end_ns, label)` tuples, and then into `json.dumps` at line 612.

So a check that computes labels from optional metadata, and gets `None` for one of them, hits a bare `TypeError` from inside the sort.

## What makes this worth fixing rather than documenting

The crash is data-dependent on a second interval. The bad label is only ever compared when two intervals share both bounds, because the tuple sort stops at the first differing element. The same value is stored without complaint otherwise.

All three cases below use one check returning `label=None`, through `Catalog.append_episode`, on `main` at `623f9f6`:

```
single None label                  STORED [(None,)]
None + str, same bounds            TypeError: '<' not supported between instances of 'str' and 'NoneType'
None + str, different bounds       STORED [(None,), ('freeze',)]
```

That is the shape that survives a test corpus and fails on a real one. A freeze detector that labels spans from a lookup will pass every fixture where the spans differ, then crash the moment two land on the same bounds.

Two more reachable values, same origin:

```
label=5 and label="a", same bounds    TypeError: '<' not supported between instances of 'int' and 'str'
label=object()                        TypeError: Object of type object is not JSON serializable
```

The last one comes from line 612 rather than the sort, so it fires regardless of how many intervals there are.

None of these name the check, the interval, or the field. They surface as a `TypeError` several frames inside `append_episode`, after the episode has been processed.

## Repro

```python
result = hflow.CheckResult(
    measurements={"m": 1.0},
    intervals=[
        hflow.Interval(start_ns=0, end_ns=10, label="freeze"),
        hflow.Interval(start_ns=0, end_ns=10, label=None),
    ],
)
row = CheckRunRow.from_result(
    check_name="freeze", check_version="v1", critical=False,
    status=hflow.CheckStatus.MEASURED, duration_s=1.0, error=None, result=result,
)
catalog.append_episode(
    canonical_path=canonical, stamps=stamps,
    episode_metadata={"task": "t"}, check_rows=[row],
)
```

```
File "src/hflow/catalog.py", line 837, in append_episode
    run_fingerprint = _run_fingerprint(
File "src/hflow/catalog.py", line 594, in _run_fingerprint
    "intervals": sorted(
TypeError: '<' not supported between instances of 'NoneType' and 'str'
```

## Pattern to copy

#152 fixed exactly this for a NumPy bound, and its title is the complaint: an error naming neither the check nor the field. #161 fixed the inverted span. Both landed in `_normalized_intervals`. `label` is the third field of the same dataclass and the only one that never got the treatment.

The observation path already does the string half of this: `_normalized_observations` refuses a non-string `observation_id` naming the check, and refuses empty and whitespace-padded ids.

## What to do

- Refuse a non-string `label` in `_normalized_intervals`, with an error naming the check and the bounds, matching the bound errors beside it.
- Decide whether `None` is worth coercing to `""` instead of refusing. I would refuse it: the bounds set the precedent, and silently turning a missing label into the default one hides the same bug it currently crashes on.
- Cover the same-bounds case specifically, since that is the only one today that fails.

I would like to take this one.

chiruu12:
Closes #392.

`_normalized_intervals` validated both bounds and nothing else, so a label reached the run fingerprint unchecked. The fingerprint sorts `(start_ns, end_ns, label)` tuples, so a non-string only ever met a string when two intervals shared both bounds:

```
single None label                   STORED
None + str, same bounds             TypeError
None + str, different bounds        STORED
```

A check labelling spans from a lookup passes every fixture where the spans differ, then fails on a corpus where two land together. The label is now validated first, before the bounds, since the bound errors interpolate it.

Non-string is refused rather than coerced, per your call and the precedent the bounds set with `bool`. Empty stays legal because `label: str = ""` is the dataclass default; padding is refused for the reason an `observation_id` already is, since both are stored verbatim.

Five tests beside the existing bound-rule ones. Four fail with the guard deleted: a `None` label, two intervals sharing bounds, an unserializable label, a padded label. The fifth pins that empty still stores and passes either way, so it is a regression guard rather than a reproduction and its docstring says so.

The `object()` case moved from `json.dumps` at `:612` into `_normalized_intervals`, so its message now names the check. Nothing asserted on the old text. `_run_fingerprint` is unchanged.

Full suite: 1475 passed, 11 skipped. The 17 `tests/test_packaging.py` failures are the native overlay build, which needs CPython on Linux and fails identically on unmodified `main` here. `ruff check` and `ruff format --check` clean.

### DIFF
--- a/src/hflow/catalog.py
+++ b/src/hflow/catalog.py
@@ -416,7 +416,7 @@ def _normalized_scalar_values(
 
 
 def _normalized_intervals(check_name: str, intervals: list[Interval]) -> list[Interval]:
-    """Coerce NumPy scalar interval bounds to Python ints; refuse the rest.
+    """Validate the label, coerce NumPy scalar bounds to ints, refuse the rest.
 
     A user check computing segment boundaries from ``channel.to_numpy()``
     gets ``np.int64`` bounds from plain indexing, and those reach the run
@@ -429,6 +429,29 @@ def _normalized_intervals(check_name: str, intervals: list[Interval]) -> list[In
     """
     normalized: list[Interval] = []
     for interval in intervals:
+        # The label first: the bound errors below interpolate it, and it
+        # reaches the run fingerprint, where it is sorted against the other
+        # intervals' labels. A non-string only meets one when two intervals
+        # share both bounds, so the TypeError it raises there fires on some
+        # corpora and not others, several frames from the check that caused
+        # it and naming neither the check nor the field.
+        #
+        # Empty is the dataclass default and stays legal. Padding is refused
+        # for the reason an observation id is: both are stored verbatim, so
+        # " freeze " and "freeze" would be two labels for one thing.
+        label = interval.label
+        if not isinstance(label, str):
+            raise ValueError(
+                f"check {check_name!r} set an interval label as "
+                f"{type(label).__name__} (start_ns={interval.start_ns!r}, "
+                f"end_ns={interval.end_ns!r}): interval labels are strings"
+            )
+        if label != label.strip():
+            raise ValueError(
+                f"check {check_name!r} set interval label {label!r} with leading "
+                "or trailing whitespace: interval labels are stored verbatim"
+            )
+
         bounds: dict[str, int] = {}
         for bound_name in ("start_ns", "end_ns"):
             value = getattr(interval, bound_name)
--- a/tests/test_catalog_curation.py
+++ b/tests/test_catalog_curation.py
@@ -1473,6 +1473,91 @@ def test_a_negative_interval_bound_is_refused(tmp_path: Path) -> None:
         _appended_with_interval(tmp_path, hflow.Interval(start_ns=-5, end_ns=-1, label="gap"))
 
 
+def test_a_non_string_interval_label_is_refused_naming_the_check(tmp_path: Path) -> None:
+    """A label that is not a string reaches the run fingerprint and breaks its sort.
+
+    The failure it replaces was data dependent: the fingerprint sorts
+    ``(start_ns, end_ns, label)`` tuples, so a bad label is only ever compared
+    against another when two intervals share both bounds. One interval stored
+    fine, two colliding ones raised a ``TypeError`` several frames inside
+    ``append_episode`` naming neither the check nor the field (#392).
+    """
+    with pytest.raises(ValueError, match=r"segment_check.*NoneType.*labels are strings"):
+        _appended_with_interval(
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging. Every claim in the description reproduces, and the split between the two guards is clean:

```
type guard removed      3 failures
padding guard removed   1 failure
both removed            4 failures
label coerced           3 failures
```

One thing you may not have noticed, and it argues for your empty-stays-legal call more strongly than the dataclass default does. Making empty a refusal breaks three tests, not one:

```
test_an_empty_interval_label_is_allowed
test_a_bool_interval_bound_is_refused_rather_than_stored_as_one_nanosecond
test_a_float_interval_bound_is_refused_naming_the_check_and_bound
```

The two bound tests construct intervals without a label at all, so refusing empty would make the existing bound rules unreachable for their own fixtures. Empty is not merely the documented default, it is what every caller that does not care about labels already relies on.

Validating the label before the bounds is the right order for the reason you gave, and worth noting it holds even though `!r` cannot fail: an unvalidated NumPy bound formats as `np.int64(5)` in a label error, which is ugly rather than broken, and the reverse order would put a garbage label into a bound error instead.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1519 passed / 6 skipped. Your 17 packaging failures were the native overlay build guard; #394 landed 24 of those passing on Linux this afternoon, so they should go quiet for you after a rebase.

Closes #392. #391 is still yours whenever you want it.

Issue #398, opened 2026-09-04
Development setup omits the C compiler required by the default test suite


Galabavamsi:
## Current behavior

`CONTRIBUTING.md` says a plain `uv sync --locked` creates the whole root development environment and instructs contributors to run the default suite with:

```bash
uv run pytest -q
```

Its prerequisite list includes Git, uv, Python, FFmpeg/ffprobe, conditional Docker, and conditional lychee. It does not mention a system C compiler.

The root dev group intentionally includes Cython and setuptools because the default suite exercises the real native-overlay build path, but those Python dependencies cannot provide the system compiler used by setuptools.

## Reproduction

Tested from a fresh worktree at `81a13f81f5cf97d0b3048f0a660e3a1e701d50cb` on Ubuntu Linux:

```bash
uv sync --locked
command -v cc
command -v gcc
command -v clang
uv run pytest -q tests/test_packaging.py
```

None of the compiler checks printed a path. The test result was:

```text
17 failed, 1 passed in 2.70s
```

All 17 failures reach the native Cython build and have the same root cause:

```text
FileNotFoundError: [Errno 2] No such file or directory: 'cc'
distutils.compilers.C.errors.CompileError:
    [Errno 2] No such file or directory: 'cc'
```

The full default suite in the same environment reported `17 failed, 1475 passed, 6 skipped`; every failure was in `tests/test_packaging.py`.

Searches for `C compiler`, `compiler contributing`, and `build-essential` found no existing issue or PR. #393 concerns refusal-path test coverage inside this subsystem, not the machine prerequisite needed to run those tests.

## Expected behavior

A contributor following the documented root setup should know all system prerequisites needed for the documented default suite before running it.

The smallest fix may be to document:

- that a C compiler/toolchain is required by the default native-overlay tests;
- a portable preflight such as `command -v cc`;
- OS-specific package examples where the project wants to support them (for example, build tools on Ubuntu and Xcode Command Line Tools on macOS).

If the intended contract is instead that machines without a compiler may run the default suite successfully, the packaging-test selection behavior would need to make that explicit. I have not assumed that policy here.

## Definition of done

- The development setup states whether a compiler is mandatory for the default suite.
- Linux and macOS contributors get an actionable prerequisite/preflight.
- The relationship between `uv sync --locked`, the native-build Python dependencies, and the external compiler is clear.
- The documented setup no longer reaches 17 identical missing-`cc` failures without warning.

kstonekuan:
[...]

PR #400, opened 2026-09-04
fix(lerobot): refuse depth videos before conversion

### LINKED ISSUE #399: LeRobot depth videos are accepted as RGB and lose 12-bit depth semantics
## Current behavior

The LeRobot importer discovers every feature with `dtype: video` as a selectable camera. It does not inspect the feature `info` block for `is_depth_map` (or the legacy `video.is_depth_map` spelling).

Every selected video then takes the RGB transform path:

```text
source MP4 -> libx264 -crf 23 -> foxglove.CompressedVideo
```

That is unsafe for current LeRobot depth streams. LeRobot stores depth as quantized 12-bit codes in a high-bit-depth video and persists the parameters needed to dequantize those codes to physical units. HFlow neither preserves those fields nor rejects the unsupported semantic type.

The user-facing guide currently says depth maps are unsupported, and the public API says unsupported feature layouts fail before any episode is published. A `dtype: video` depth feature bypasses that refusal.

## Controlled result

Tested at `81a13f81f5cf97d0b3048f0a660e3a1e701d50cb`.

I generated four 32 by 24 constant depth-code frames with 12-bit values `64`, `1024`, `2048`, and `4032`, encoded them losslessly as the shape LeRobot documents for video depth, then passed the MP4 through the importer transform `_transcode_mp4_to_h264`.

```text
source:    codec=hevc  pix_fmt=gray12le      frames=4
converted: codec=h264  pix_fmt=yuv420p10le

source decoded medians:    [1024, 16388, 32776, 64527]
converted decoded medians: [1025, 16400, 32767, 64510]
source unique values/frame:    [1, 1, 1, 1]
converted unique values/frame: [1, 1, 2, 1]
exact_equal: false
```

The decoded values are represented as 16-bit output values in this comparison. The relevant outcome is that a lossless 12-bit source does not survive the current CRF-23 H.264 route exactly, and the output stream has only a 10-bit pixel format.

Independently, the current metadata discovery is:

```python
feature_specification.get("dtype") == "video"
```

with no depth-semantic guard before conversion.

## Real-format evidence

Current LeRobot documentation specifies that depth maps may be stored as video streams alongside RGB, quantized to 12-bit codes and encoded as HEVC Main 12. It records `is_depth_map: true` plus `video.depth_min`, `video.depth_max`, `video.shift`, and `video.use_log` so readers can recover physical units:

- https://github.com/huggingface/lerobot/blob/main/docs/source/video_encoding_parameters.mdx
- https://github.com/huggingface/lerobot/blob/main/src/lerobot/datasets/dataset_metadata.py

There is also a pinned public official v3 depth corpus showing the active metadata contract for image-backed depth:

- `lerobot/outdoor-depth`
- revision `7c63dc16226035cdd79fffc95404e82723bda5ad`
- feature `observation.images.depth`
- `dtype: image`, `is_depth_map: true`, `depth_unit: mm`

That corpus is evidence for the semantic marker, not a video acceptance fixture.

## Immediate safe fix

Until HFlow has an agreed canonical depth representation, selecting a video feature marked as depth should fail before any episode or manifest is published. The error should name the feature and explain that the RGB H.264 path cannot preserve depth values.

Both current and legacy LeRobot marker spellings should be considered:

- `info.is_depth_map`
- `info["video.is_depth_map"]`
- `video_info["video.is_depth_map"]`

## Advanced extension to agree separately

End-to-end support needs a format decision rather than merely allowing the feature through:

1. Decode LeRobot quantized depth with its persisted parameters into a named physical unit, or deliberately preserve the codes plus their calibration metadata.
2. Choose a lossless canonical channel representation suitable for one-channel 16-bit integer and/or 32-bit floating-point depth.
3. Keep depth out of RGB luma, exposure, and blackout checks unless a depth-specific adapter is defined.
4. Preserve depth provenance through catalog and snapshot export.
5. Validate on a real public video-depth v3 corpus once one with state/action and stable pinned metadata is selected.

## Definition of done for this issue

- A selected depth-marked video cannot silently take the RGB H.264 transform path.
- Both modern and legacy marker spellings are covered.
- Refusal happens before a partial episode or prepared manifest is published.
- A normal RGB video remains accepted.
- Documentation names the refusal and points to the metadata field.
- If maintainers choose full depth support instead, exact depth values and physical-unit metadata round-trip in an outcome-focused test.

## Prior scope

#189 explicitly left depth maps out of scope unless a small extension was proposed and agreed. This issue is that extension plus the currently missing fail-loud boundary; it does not reopen the generalized RGB importer work.

Galabavamsi:
## Summary

- detect LeRobot depth video features through the current and legacy metadata markers
- refuse selected depth videos before publishing an episode or prepared manifest
- keep explicitly non-depth RGB videos on the existing import path
- document the supported boundary and why RGB H.264 cannot preserve depth values

## Why

The importer currently treats every `dtype: video` feature as RGB. Current LeRobot depth streams carry `is_depth_map` metadata and may store quantized 12-bit physical depth in HEVC Main 12.

A controlled four-frame input changed from `hevc/gray12le` to `h264/yuv420p10le` through the current transform, and its decoded values were no longer exactly equal. Allowing that path would silently discard depth precision and the metadata needed to recover physical units.

This PR takes the smallest safe scope from #399: fail loud before publication. It does not choose a canonical depth representation or claim end-to-end depth support.

Closes #399.

## Validation

- TDD red: all three depth-marker cases initially failed with `DID NOT RAISE ValueError`
- TDD green: `uv run pytest -q tests/test_lerobot_converter.py` — 50 passed
- `uv run ruff format` — 215 files unchanged
- `uv run ty check` — passed
- `uv run pytest -q --ignore=tests/test_packaging.py` — 1477 passed, 6 skipped
- `uv run pytest -q tests/test_packaging.py` — 17 failed, 1 passed because this machine has no `cc`; the same clean-main environment limitation is reported separately in #398

## Checklist

### DIFF
--- a/docs/how-to/import-lerobot-v3.md
+++ b/docs/how-to/import-lerobot-v3.md
@@ -82,17 +82,24 @@ channels named `/observation.state` and `/action`.
 
 | Feature | Support |
 |---|---|
-| Video features with `dtype: video` | Supported; select them with `--camera` |
+| RGB video features with `dtype: video` | Supported; select them with `--camera` |
 | One-dimensional fixed-width `float32` observation state | Supported |
 | One-dimensional fixed-width `float32` action | Supported |
 | Image-only features | Not supported |
-| Language tensors, depth maps, or arbitrary nested features | Not supported |
+| Video depth maps | Refused before publication; the RGB H.264 path cannot preserve their values |
+| Language tensors, image depth maps, or arbitrary nested features | Not supported |
 
 The importer reads `meta/info.json`, the episode index, Parquet feature
 columns, frame rate, episode boundaries, and video-path templates. Episode
 metadata split across several `meta/episodes` shards is read in full, so
 every episode keeps its own video window. It refuses an unsupported or
 missing required feature before publishing an episode.
+
+Depth video features are detected through LeRobot's current
+`info.is_depth_map` marker and its legacy `info["video.is_depth_map"]` and
+`video_info["video.is_depth_map"]` spellings. Selecting one fails before an
+episode or prepared manifest is published. Converting it through the RGB path
+would discard the depth quantization and physical-unit semantics.
 
 Every output records the source repository, resolved commit, source episode
 index, task, embodiment, importer version, and FFmpeg build. Video is encoded
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -2,7 +2,7 @@
 
 The converter reads repository metadata (feature schema, fps, episode
 boundaries, video paths) instead of encoding dataset-specific assumptions.
-Every selected camera is converted into its own foxglove.CompressedVideo
+Every selected RGB camera is converted into its own foxglove.CompressedVideo
 channel; numeric state and action schemas are derived from the declared
 dtype and shape, failing loud before conversion when a feature is
 unsupported.
@@ -602,6 +602,23 @@ class _NumericSchema:
     dim: int
 
 
+def _is_depth_map_feature(feature_specification: object) -> bool:
+    if not isinstance(feature_specification, dict):
+        return False
+    feature_information = feature_specification.get("info")
+    legacy_video_information = feature_specification.get("video_info")
+    return (
+        isinstance(feature_information, dict)
+        and (
+            feature_information.get("is_depth_map") is True
+            or feature_information.get("video.is_depth_map") is True
+        )
+    ) or (
+        isinstance(legacy_video_information, dict)
+        and legacy_video_information.get("video.is_depth_map") is True
+    )
+
+
 def _derive_numeric_schema(feature_name: str, feature_specification: dict) -> _NumericSchema:
     declared_dtype = feature_specification.get("dtype")
     declared_shape = feature_specification.get("shape") or []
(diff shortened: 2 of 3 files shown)


kstonekuan (approved):
Merging with one fixup. Refusing rather than guessing at a canonical depth representation is the right scope, and the issue that came with it is the most carefully sourced one we have had: separating what your `lerobot/outdoor-depth` evidence proves (the marker contract) from what it does not (video acceptance) is exactly the distinction most issues blur.

I checked the marker contract against LeRobot's source rather than the docs, and your three spellings are precisely right. Their own docstring words it the same way:

```python
"""Depth maps are flagged canonically via ``feature['info']['is_depth_map']`` or through the
legacy ``video.is_depth_map`` key, which may live in ``feature['info']`` or in a separate
``feature['video_info']`` dict."""
```

**The fixup, `29d1e33`, closes a hole in the guard.** That same function is truthy, not an identity test:

```python
    return bool(
        info.get("is_depth_map")
        or info.get("video.is_depth_map")
        or (isinstance(video_info, dict) and video_info.get("video.is_depth_map"))
    )
```

So a corpus carrying `"is_depth_map": "true"` or `1` is depth by LeRobot's own definition, and `is True` called it RGB and sent it down the H.264 path. That is the outcome the refusal exists to prevent, and it fires on exactly the non-canonical corpora most likely to be hand-edited. Restoring the identity test now fails four cases:

```
`is True` instead of truthy   4 failures (the truthy-but-not-True markers)
```

The other guards are each pinned on their own, and one thing you did deserves calling out: putting `"is_depth_map": False` in the RGB fixtures rather than omitting the key. Reading the marker as presence rather than value fails six tests because of it, so the negative case is load-bearing rather than decorative.

```
refusal removed entirely                  7 failures
canonical spelling only                   2 failures
video_info legacy dict ignored            2 failures
marker read as presence, not value        6 failures
```

One imprecision in the issue, for the record rather than for action: `depth_min`, `depth_max`, `shift`, and `use_log` are not in `dataset_metadata.py`; only `depth_unit` is. Does not affect a refusal, but it will matter if we take the advanced extension.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1526 passed / 6 skipped.

Closes #399, labelled `bug`.

On the advanced extension: your five points are the right decomposition and item 2 is the one that has to be settled first, because the canonical channel representation determines everything downstream. I am not taking that decision today. Leaving #399's closure to cover the boundary only, and if you want the format question opened as its own issue with the tradeoffs laid out the way you laid these out, that would be genuinely useful. Point 3 in particular (keeping depth out of RGB luma and exposure checks) is a live problem the moment depth exists at all, and it is a different question from the encoding.

PR #401, opened 2026-09-04
feat(snapshot): record table and copied-asset integrity in format.json (#397)

### LINKED ISSUE #397: Dataset snapshots cannot verify table or copied-media integrity after transfer
## Current behavior

Dataset snapshot format v1 identifies the format, version, media mode, media URI base, and table filenames in `format.json`. It does not record file sizes, row counts, hashes, or a copied-asset inventory, and HFlow has no snapshot verification API or CLI command.

The atomic staging/activation path proves that HFlow finished writing a snapshot. It does not let a recipient distinguish those original bytes from a snapshot damaged or changed after export.

## Controlled result

Tested at `81a13f81f5cf97d0b3048f0a660e3a1e701d50cb` (`v0.2.5-3-g81a13f8`) with a one-episode, two-measurement, one-tag snapshot exported through the public API in `copy` media mode.

The original marker contained exactly:

```text
format
format_version
media_mode
media_uri_base
tables
```

I made three independent copies and changed one thing in each:

| Mutation after export | `format.json` changed? | Current observation |
|---|---:|---|
| Replaced the copied media bytes | No | All Parquet tables still read normally; there is no signal that the asset changed |
| Deleted `tags.parquet` | No | The destination is still recognized as an overwriteable HFlow snapshot; DuckDB later raises “No files found” |
| Truncated `samples.parquet` to 64 bytes | No | The destination is still recognized as an overwriteable HFlow snapshot; DuckDB later raises “No magic bytes found at end of file” |

`hasattr(hflow, "verify_dataset_snapshot")` was false, and `hflow export snapshot --help` exposes export options but no verification path.

This means the format marker can identify a snapshot while being unable to say whether its required tables and copied assets are the bytes HFlow actually published.

## Why this matters

Snapshots are the delivery artifact intended to move independently of the catalog. A successful export followed by an interrupted copy, storage corruption, or accidental asset replacement currently looks valid until a consumer happens to read the affected file. Modified copied media may remain completely silent.

This is also consistent with adjacent dataset-format practice: Croissant strongly recommends SHA-256 checksums for every `FileObject` in a versioned dataset, and Hugging Face Datasets verification covers expected files, split metadata, and optionally checksums.

- https://docs.mlcommons.org/croissant/docs/croissant-spec.html
- https://huggingface.co/docs/datasets/about_dataset_load

## Design questions

A v2 marker plus a public verifier appears to be one possible shape, but the format contract should be decided before implementation:

1. Should every required table record `path`, `size_bytes`, `sha256`, and perhaps `row_count`?
2. In `copy` mode, should every copied asset receive the same coverage, or should the marker point to a separate deterministic asset manifest?
3. Should there also be one digest over the normalized manifest so that the delivered set itself has an identity?
4. How should v1 behave: recognizable but explicitly “unverifiable”, or rejected by a strict verifier?
5. In `references` mode, should verification cover only local tables/metadata and report remote media as unverified unless network access is explicitly requested?
6. Are unlisted extra files allowed?

## Definition of done

- Every required Parquet table in a newly exported snapshot is covered by deterministic integrity metadata.
- Every local asset in `copy` mode is covered.
- A public library and/or CLI verification path detects a missing file, truncation, and same-path byte replacement.
- An unchanged snapshot passes.
- v1 compatibility and reference-mode semantics are deliberate and tested.
- Default verification does not fetch remote media.
- The snapshot format documentation explains what is and is not guaranteed.

## Non-goals

- Cryptographic signing or publisher authenticity.
- Proving continued availability of remote reference-mode media.
- Changing the logical contents of snapshot tables.
- Making the whole catalog content-addressed.

VARUN3WARE:
## Summary
- Keep dataset snapshot `format_version` at `1` and add delivery integrity to `format.json`: each required Parquet table gets `path` / `size_bytes` / `sha256`.
- In copy mode, every local file under `assets/` gets the same receipt fields inline (no separate asset manifest); references mode leaves `assets` empty and does not fetch remote media.
- Add a `content_id` over the normalized inventory so a deleted member is detectable later without a second format change.
- Document what is and is not guaranteed; no public verifier / CLI / verify-on-export (deferred, same split as #379).

Closes #397

## Test plan
- [x] `uv run ruff check`
- [x] `uv run ty check`
- [x] `uv run pytest tests/test_dataset_snapshot.py -q` (16 passed)
- [x] `uv run pytest -q` (1529 passed, 6 skipped)

## Notes
- Follows the accepted direction on #397: marker only; verifier purely additive later.
- Inventory `content_id` width matches prepared-manifest episode `content_id` from #389 for a familiar receipt shape.

### DIFF
--- a/docs/how-to/export-dataset-snapshot.md
+++ b/docs/how-to/export-dataset-snapshot.md
@@ -62,6 +62,24 @@ each table, records the media mode, and states whether media paths are relative
 to the export directory. The JSON marker is written last and the completed
 directory is activated atomically.
 
+Under the same format version `1`, the marker also records delivery integrity
+so a later verifier (not shipped here) can tell whether the published bytes
+are still intact:
+
+| Field | Meaning |
+| --- | --- |
+| `tables.<name>.path` | Required Parquet file relative to the export root |
+| `tables.<name>.size_bytes` | Size of that file at export time |
+| `tables.<name>.sha256` | Full SHA-256 of that file's bytes |
+| `assets[]` | Same `path` / `size_bytes` / `sha256` for every regular file under `assets/` in copy mode; empty in references mode (remote media are not fetched) |
+| `content_id` | 16-hex digest of the normalized inventory (all table and asset receipts, sorted by path), so a deleted member is visible even when every remaining file still matches |
+
+This is a receipt, not a verify command: export does not re-read the
+destination after transfer, and there is no public `verify_dataset_snapshot`
+API or CLI yet. Older HFlow overwrite checks only `format` and
+`format_version`, so these added keys stay backward compatible without a
+format-version bump.
+
 The Parquet tables form one snapshot:
 
 | File | Grain and purpose |
--- a/docs/how-to/export-dataset-snapshot.md
+++ b/docs/how-to/export-dataset-snapshot.md
@@ -85,6 +85,11 @@ API or CLI yet. Older HFlow overwrite checks only `format` and
 `format_version`, and readers that only consume the string `tables` map keep
 working; the `integrity` key is purely additive under format version `1`.
 
+The receipt travels unsigned inside the `format.json` it describes, so it
+catches corruption and accidental loss, not tampering: anyone who can edit a
+table can recompute the hashes to match. Use a signature or an out-of-band
+checksum if you need to detect a deliberate change.
+
 The Parquet tables form one snapshot:
 
 | File | Grain and purpose |
--- a/docs/how-to/export-dataset-snapshot.md
+++ b/docs/how-to/export-dataset-snapshot.md
@@ -58,27 +58,32 @@ dataset-snapshot/
 ## Format contract
 
 `format.json` identifies `hflow-dataset-snapshot` format version `1`, names
-each table, records the media mode, and states whether media paths are relative
-to the export directory. The JSON marker is written last and the completed
-directory is activated atomically.
+each table (a name-to-filename map), records the media mode, and states whether
+media paths are relative to the export directory. The JSON marker is written
+last and the completed directory is activated atomically.
 
-Under the same format version `1`, the marker also records delivery integrity
-so a later verifier (not shipped here) can tell whether the published bytes
-are still intact:
+Under the same format version `1`, an additive `integrity` block records
+delivery integrity so a later verifier (not shipped here) can tell whether the
+published bytes are still intact. The original `tables` map is unchanged for
+external readers; receipts live only under `integrity`:
 
 | Field | Meaning |
 | --- | --- |
-| `tables.<name>.path` | Required Parquet file relative to the export root |
-| `tables.<name>.size_bytes` | Size of that file at export time |
-| `tables.<name>.sha256` | Full SHA-256 of that file's bytes |
-| `assets[]` | Same `path` / `size_bytes` / `sha256` for every regular file under `assets/` in copy mode; empty in references mode (remote media are not fetched) |
-| `content_id` | 16-hex digest of the normalized inventory (all table and asset receipts, sorted by path), so a deleted member is visible even when every remaining file still matches |
+| `tables.<name>` | Required Parquet filename (unchanged string map) |
+| `integrity.tables.<name>.path` | Same file relative to the export root |
+| `integrity.tables.<name>.size_bytes` | Size of that file at export time |
+| `integrity.tables.<name>.sha256` | Full SHA-256 of that file's bytes |
+| `integrity.assets[]` | Same `path` / `size_bytes` / `sha256` for every regular file under `assets/` in copy mode; empty in references mode (remote media are not fetched) |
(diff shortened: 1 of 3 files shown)


kstonekuan (changes requested):
Holding this one, and the reason is a shape change rather than anything wrong with the work. The receipt content is right: per-file `sha256`, the inventory `content_id` that makes a *deleted* member visible, and references mode leaving `assets` empty rather than fetching remote media to hash it. Gate is clean here too, 16 passed.

**`tables` changed type under an unchanged format version.** It was a name-to-filename mapping and is now a name-to-record mapping:

```python
-            "tables": {
-                "samples": _SAMPLES_TABLE_FILE_NAME,        # "samples.parquet", a string
+        tables[table_name] = _file_integrity_record(file_name, absolute_path)   # a dict
```

So `format_marker["tables"]["samples"]` returns `"samples.parquet"` before this and `{"path": ..., "size_bytes": ..., "sha256": ...}` after, with `format_version` still `1`. Main's own documentation calls that section the **Format contract** and says the marker "names each table", so the mapping is published, not incidental. Nothing in the repo reads it besides the tests you updated, but external readers are the entire audience for a snapshot: #397's premise is that this artifact moves independently of the catalog.

**This one is partly on me.** My direction said added keys are backward compatible and that adding under v1 beats a version break. That was true of *adding*, and I did not say the obvious corollary: not redefining a field that is already there. Reading it back, it invited exactly this.

**The fix, and the shape is yours to pick.** Put the integrity block under its own key and leave `tables` alone:

```json
"tables":    { "samples": "samples.parquet", ... },
"integrity": { "tables": { "samples": { "path": ..., "size_bytes": ..., "sha256": ... } },
               "assets": [ ... ],
               "content_id": "..." }
```

That keeps the whole thing genuinely additive, which makes your doc sentence about backward compatibility true rather than nearly true, and it keeps the verifier purely a reader. Nesting the per-table receipts under `integrity.tables` rather than flattening them also leaves room for a future non-table member without another reshuffle.

Two smaller things while you are in there.

`_build_snapshot_integrity_marker_fields` hashes every copied asset, so copy mode now reads every media byte a second time on export. For a snapshot with real video that is a visible slowdown and nobody has costed it. It is very likely worth paying, but say so in the PR with a rough number rather than leaving it for someone to discover.

The 16-hex truncation on `content_id` deserves a sentence. Matching #389's width is a reasonable instinct for a familiar shape, but that value identifies an episode while this one is the delivery's integrity digest, and #397 cited Croissant's full-SHA-256 recommendation. The per-file hashes are full length so the exposure is limited to the set digest, and 64 bits is ample against corruption while being weak against deliberate tampering. Either keep it and say which threat it covers, or make it full length; I lean toward full length, since nothing here needs it to be short.

The `FileNotFoundError` when a required table is missing from staging is fine and I would keep it: staging always writes all seven, so it is an invariant check rather than an error path, and failing loudly there beats publishing a marker that promises a file nobody wrote.

Nothing else needs changing, and none of the test work is wasted; the assertions move to a new path rather than being rewritten.

VARUN3WARE:
Thanks, reshaped it so this stays additive under format v1:
- Restored `tables` as the name→filename string map
- Moved receipts under `integrity.tables` / `integrity.assets` / `integrity.content_id`
- Inventory `content_id` is now full-length SHA-256 (same threat model as the per-file hashes)
- Copy mode still re-reads each copied asset once after the copy to hash it; called out in the docs
Ready for another look.

kstonekuan (approved):
The reshape is right. `tables` is a string map again, so a v1 reader that already knows the format keeps working, and everything new sits under one key it can ignore.

Validated on your branch against current main: full gate clean, 1585 passed / 6 skipped. I went after the one thing that worried me, which is that the new tests recompute their expected digests with the same helpers the implementation uses. That shape usually cannot notice a helper that stops covering something, because both sides move together. Yours does:

- Receipts stop carrying `sha256`, or stop carrying `size_bytes`: two tests red.
- Assets left out of the inventory, or the `assets/` walk skipped: red.
- Inventory digest reduced to paths only, contents ignored: `test_dataset_snapshot_marker_integrity_detects_post_export_mutations` red on its own.

That last one is the one that counts, and it works because you mutate the file on disk and let the helper recompute. A contents-blind digest then matches when the test says it must not.

Pushed one commit, `docs(snapshot): say the receipt is not a tamper defence`. Three changes:

- The `_inventory_content_id` docstring said the digest covers "deliberate set tampering". It cannot: it travels unsigned inside the `format.json` it describes, so anyone who edits a table can recompute it to match. What it does give is corruption and accidental loss, which is worth having and is what the tests demonstrate. Left the reasoning in the docstring and added the same caveat to the how-to, since someone deciding whether to trust a delivered snapshot will read the docs, not the source.
- `snapshot_module.shutil.copytree` in the new test to a plain `shutil.copytree`. The two existing uses reach through the module because they monkeypatch it; yours does not, and the idiom reads as required when it is copied without that reason.

Nice work on the second pass.

PR #402, opened 2026-09-04
test(dataset): assert the provenance sidecar payload whole

### LINKED ISSUE #391: The dataset provenance sidecar records a schema version and coverage block that no test reads back

chiruu12:
Closes #391.

The dataset provenance sidecar had five keys no test read back: `dataset_manifest_version`, `created_at`, `total_episodes`, the `coverage` block, and `workspace_id`. `test_the_manifest_and_its_provenance_land_together` already loaded the file, so this widens that test rather than adding a second one.

Following #388, the top level is asserted as a whole dict. That pins absence as well as presence. `pipeline` is popped and compared separately, since it is `json.loads(manifest.to_json())` and pinning it whole would mean restating every check on every pipeline change.

Two things the fixture needed:

- `created_at` is now passed explicitly, so the assertion pins that the sidecar records the caller's stamp rather than a second `now()` taken inside `_sidecar_payload`.
- The workspace identity is minted in the test. Ingest does not mint one, so `workspace_id` was `None` in this fixture, which is also what dropping the key would look like.

`coverage` gets real numbers: one entry for `duration`, 1 of 1. Not merely a non-empty list.

Mutations run against the assertions, each one alone:

| mutation | result |
|---|---|
| drop `dataset_manifest_version` | fails |
| set the version to a literal 2 | fails |
| drop `total_episodes` | fails |
| replace `coverage` with `[]` | fails |
| drop the `coverage` key | fails |
| write `episodes_ran` as 0 | fails |
| force `workspace_id` to None | fails |
| stamp `created_at` inside the payload | fails |

`tests/test_dataset.py` is 13 passed, 1 skipped. The full suite matches main exactly here, 31 failed and 1194 passed either way, from `hflow_server` not being installed locally.

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -8,8 +8,14 @@ import pytest
 
 import hflow
 from hflow.cli import main as cli_main
-from hflow.dataset import create_dataset, dataset_slug, default_dataset_sql
+from hflow.dataset import (
+    DATASET_MANIFEST_VERSION,
+    create_dataset,
+    dataset_slug,
+    default_dataset_sql,
+)
 from hflow.testing import SyntheticEpisodeSpec, synthesize_episode
+from hflow.workspace import Workspace
 
 PIPELINE_SOURCE = """
 import hflow
@@ -145,7 +151,16 @@ class TestArtifacts:
         _ingest(ingested_project, monkeypatch)
         app = hflow.import_pipeline_application(str(ingested_project / "pipeline.py"))
 
-        dataset = create_dataset(app, "Clean Corpus!")
+        # Minted here because ingest does not mint one, and the assertion is
+        # only worth making against a real id: with none stored the field is
+        # None, which is also what dropping it from the payload would look like.
+        workspace_id = Workspace(app.storage_root).ensure_identity().workspace_id
+
+        # Passed rather than defaulted, so the assertion below pins that the
+        # sidecar records the caller's stamp and not a second now() taken inside
+        # _sidecar_payload.
+        stamped_at = "2026-01-02T03:04:05+00:00"
+        dataset = create_dataset(app, "Clean Corpus!", created_at=stamped_at)
 
         manifest_file = Path(dataset.manifest_path)
         sidecar_file = Path(dataset.sidecar_path)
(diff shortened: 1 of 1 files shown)


kstonekuan (approved):
LGTM, merging. Your table reproduces, all eight, and two more I added for good measure (dropping `name` and `row_count`) fail too, so the whole-dict assertion covers the three keys that were already checked as well as the five that were not.

Minting the workspace identity is the part that took judgment, and your reasoning in the comment is exactly right: asserting `workspace_id is None` would have been indistinguishable from asserting the key was dropped. A test that passes for both the correct and the broken behavior is worse than no test, and you spotted that the fixture handed you one.

Passing `created_at` explicitly rather than letting it default does the same job for a different reason, and the mutation proves it: stamping a fresh `now()` inside `_sidecar_payload` fails now, so the field is pinned as the caller's value rather than merely as a plausible timestamp.

Popping `pipeline` and comparing it above is the right call. A whole-dict assertion that swallowed the entire pipeline manifest would fail on every unrelated check change, and a test that cries wolf gets edited into uselessness.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1526 passed / 6 skipped. Your 31 local failures are `hflow-server` not being installed; `uv sync --locked --all-extras` picks it up if you want them quiet.

Closes #391.

Issue #376, opened 2026-09-03
The fixed-GOP promise is stamped into provenance as "actually used" but is never measured on pass-through video


Sagar-024:
## Current behavior

FORMAT.md item 5 promises Fixed GOP (`keyint = min-keyint`, `scenecut=0`): "keyframes land exactly every `gop_frames` messages, so seek cost is uniform and predictable" ([docs/FORMAT.md:79](https://github.com/Hebbian-Robotics/hflow/blob/main/docs/FORMAT.md#L79)), and the presets section calls GOP length "effectively a training hyperparameter" ([docs/FORMAT.md:84](https://github.com/Hebbian-Robotics/hflow/blob/main/docs/FORMAT.md#L84)). The provenance record stamps `gop_seconds`, defined at [docs/FORMAT.md:129](https://github.com/Hebbian-Robotics/hflow/blob/main/docs/FORMAT.md#L129) as "The keyframe interval actually used."

On the re-encode path the promise is real: x264 gets `keyint=min-keyint=gop_frames`, `scenecut=0`, and `_enforce_encode_guarantees` refuses any keyframe that lands off the grid (`expected {keyframe_expected} (gop_frames={gop_frames})`, [src/hflow/video.py:683](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/video.py#L683)).

On the pass-through path nothing measures it:

- `_validate_passthrough_video_payload` ([src/hflow/transform.py:479](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/transform.py#L479)) checks one access unit per message, SPS/PPS on keyframes, first-message keyframe, and B-freedom. Cadence is not checked.
- `_check_video_payload` ([src/hflow/doctor.py:96](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/doctor.py#L96)) checks the same two keyframe properties (SPS/PPS at :174, first message at :180) and never the interval between them.

The provenance record is written once per episode, identically for both paths ([src/hflow/transform.py:895-896](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/transform.py#L895-L896)), so a pass-through episode is stamped with the configured interval no matter what its stream contains.

## Controlled result

[...]

PR #403, opened 2026-09-04
fix(import): report the collector's success label instead of inventing one

### LINKED ISSUE #395: The LeRobot importer ignores the source's success label and hardcodes "true" for every episode
## Current behavior

[docs/FORMAT.md:109](https://github.com/Hebbian-Robotics/hflow/blob/main/docs/FORMAT.md#L109) defines `episode/v1`'s `success` as "Collector-labeled outcome, `"true"`/`"false"`", and [:119](https://github.com/Hebbian-Robotics/hflow/blob/main/docs/FORMAT.md#L119) states the record is "copied/merged from the source recording". The canonical transform honors that: it copies `episode/v1` through from the source untouched ([src/hflow/transform.py:786-787](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/transform.py#L786-L787)).

The LeRobot importer does not. It constructs the record from scratch ([src/hflow/importers/lerobot.py:1030-1041](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L1030-L1041)) and hardcodes `"success": "true"` at [:1035](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L1035) for every episode it converts. Its episodes query ([src/hflow/importers/lerobot.py:487](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L487)) reads seven columns and never touches an outcome field.

The outcome signal it skips exists in the format and in the wild. LeRobot v3 carries `next.success` as a declared per-frame feature, and the episodes parquet aggregates it per episode as `stats/next.success/min|max`. `lerobot/pusht` — the reference dataset for the v3 docs — declares it in `meta/info.json` and carries the aggregates for all 206 episodes; in the current revision every frame of every episode reads `False`.

The label then becomes a queryable fact: `success` is one of the four episode keys promoted to first-class catalog columns ([src/hflow/catalog.py:59](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/catalog.py#L59), [:68](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/catalog.py#L68), promoted at [:863](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/catalog.py#L863)).

## Controlled result

Ran the real import path (network and ffmpeg stubbed), read the delivery back through the real `Episode` reader, appended through the real `Catalog.append_episode`, and queried with real SQL:

```
--- stage 1: what the source declares ---
  data parquet frame (episode 0): next.success = False
  episodes parquet aggregate (episode 0): stats/next.success/max = ['False']
--- stage 2: what the importer shipped ---
  lerobot_episode_0001.mcap: episode/v1 success = 'true'
--- stage 3: the buyer's query ---
  SELECT ... WHERE success = 'true'  ->  2 row(s):
    episode_id=ba19d7fb94cdba01  task='task-1'  success='true'
    episode_id=e9e1a9ad1306d39d  task='task-0'  success='true'
  (catalog holds 2 episode(s) in total)
```

A demo whose collector label reads never-succeeded is delivered as `success: "true"` and returned by the exact filter a buyer would run. On `lerobot/pusht` at scale, the same disagreement holds for all 206 episodes: hflow's column says `true` where the source says `False`, 100 percent of the time.

## Why now

Third brick of the thread #376 and #379 opened: #376 removed a stamp asserting a measurement nobody took, #389 gave the delivery a checkable receipt. This is the last record in the family that speaks without evidence, and unlike the first two it speaks about content: which demos worked. That is the column a data buyer filters on.

## What to build

Open call, not assumed:

1. **Read the label**: when the source carries an outcome feature (`next.success` or equivalent), derive the episode outcome and stamp it. Whether episode success is any-frame or last-frame is a methodology choice to name.
2. **Stamp unlabeled**: when the source carries no outcome feature, omit the key. FORMAT.md:119 already says all keys are optional; omitting is the only encoding that cannot be mistaken for a collector's judgment.
3. **Column semantics**: how the catalog column represents unlabeled episodes (NULL versus a value) is the catalog's decision to make once, deliberately.

Combinations are plausible: read when present, omit when absent, document both.

## Definition of done

1. The importer never invents a collector label: the episode record is copied from the source or omits the key.
2. When the source declares an outcome feature, the delivered record reports it, with the derivation named.
3. The catalog `success` column carries real information or NULL, pinned by tests through the real import and catalog path.
4. The transform's copy path is untouched; non-LeRobot imports are unaffected.
5. The fixture above lands as a test.

## Non-goals

- Changing the transform's episode/v1 copy path
- Judging whether pusht's upstream conversion should have populated `next.success`; that is upstream's question
- Changing catalog dedupe or the promoted-keys mechanism

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_lerobot_converter.py tests/test_catalog.py
uv run pytest -q
```

Sagar-024:
Closes #395.

The importer hardcoded `success: "true"` into every `episode/v1` record while its episodes query read seven columns and never touched an outcome field. FORMAT.md defines `success` as the collector-labeled outcome and says the record is copied/merged from the source; the v3 format carries that label as a `next.success` feature with per-episode `stats/next.success/min|max` aggregates, and `lerobot/pusht` declares it for all 206 episodes, uniformly False. So the shipped record contradicted the source it read, 100 percent of the time, on the ecosystem's reference dataset.

The fix, per the issue's open call: the importer reads `stats/next.success/max` as an eighth column when the source declares it, derives the episode outcome as MAX over the collector's frames (any success frame makes the episode a success), and stamps `success` together with `success_derivation: "max(stats/next.success)"` so the methodology travels with the data. When the source declares no outcome feature, the key is omitted entirely: FORMAT.md makes every `episode/v1` key optional, and omission is the only encoding that cannot be mistaken for a collector's judgment. Column discovery runs first, because naming a missing column would fail the read, and a corpus without the feature is normal, not malformed.

The catalog promotion already renders an omitted key as SQL NULL; that behavior is pinned now, not changed. The transform's copy path is untouched.

`CONVERTER_VERSION` moves to v7 in the same commit: the label changes episode/v1 bytes that `content_episode_id` hashes.

The #395 fixture chain lands as tests through the real import, the real `Episode` reader, real `Catalog.append_episode`, and real SQL: a False,True transition reports `true` by MAX, an all-false source reports `false`, an unlabeled source omits the key and catalogs as NULL.

Gate: ruff, format, ty clean; 1530 passed / 6 skipped.

Refs #395. Third brick of the trust thread with #376 and #389.


kstonekuan (approved):
Merging with fixups in `a6bc98b`. The change is right and follows the direction closely: `max` when the source declares the feature, omission when it does not, the derivation stamped so the methodology travels, and v7 in the same commit.

**The PR carried 34 files under `.zcode/`**, including generated MCAPs, MP4s, and catalog parquets. Removed. `.zcode/` now sits beside `.claude/` and `.agents/` in `.gitignore`, so it cannot recur for you or anyone else using that tool. That is the durable version of what your `.gitignore` edit on #389 was reaching for; worth adding it to your global gitignore too, since this repo is not the only one that would collect it. The checklist line about not adding generated media is the one to watch: `git status` before `git add -A` catches this in a second.

**One real gap, closed in the fixup.** An outcome column that exists but is empty for an episode stamped `success: "false"`, because `any([])` is False. That is the same invention the hardcoded `"true"` was, one value over: a declared feature with nothing recorded is not a judgment. Nothing caught it, so it has its own case now.

```
empty aggregate treated as a label   1 failed  <- was: 74 passed, nothing noticed
```

Everything else is properly held, and the `is not None` test rather than a truthy one is the detail that makes the False case work at all:

```
success hardcoded true again              2 failures
key stamped when the source has no label  1 failure
absent collapsed with False               1 failure
success_derivation not recorded           2 failures
outcome column never discovered           2 failures
CONVERTER_VERSION left at v6              1 failure
```

That last one is new and welcome. Three PRs in a row have needed a version bump and this is the first to pin it, so an omitted bump now fails rather than being caught in review.

**Two derivation mutations go unnoticed, and that is correct rather than a gap.** Flipping `any` to `all`, or to the last frame, changes nothing:

```
derivation flipped to MIN          75 passed, nothing noticed
derivation flipped to last frame   75 passed, nothing noticed
```

`stats/next.success/max` is already the per-episode maximum, so it arrives as a one-element list, exactly as your issue showed with `['False']`. The `any()` reduces a single value. That means the reduction is defensive against a shape that does not occur, and the honest reading of `success_derivation` is "the source's own max", not a max we computed. Worth a sentence in the code saying so, since a future reader will otherwise assume the reduction is load-bearing and try to test it. I left the code alone; it is right, just easy to misread.

Small cleanup in the same fixup: the column name and derivation string are constants now, which removed the `chr(34)` workaround for f-string quote nesting on 3.11.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1531 passed / 6 skipped.

Closes #395. The trust thread reads well as a set now: #376 removed a stamp asserting an unmeasured value, #389 gave the delivery a checkable receipt, and this one stops the record speaking for the collector. The distinction you drew, that this one is about the data rather than the pipeline, is the right note to end on.

Sagar-024:
Both fixups fair. 
Thanks for closing the gap.

PR #410, opened 2026-09-05
docs: document the LeRobot importer's episode/v1 keys in FORMAT.md

### LINKED ISSUE #409: FORMAT.md documents five episode/v1 keys; the LeRobot importer writes eleven, and five of the undocumented ones gate import resume
[docs/FORMAT.md:104-115](https://github.com/Hebbian-Robotics/hflow/blob/main/docs/FORMAT.md#L104-L115) documents `episode/v1` as five named keys plus "*(any user key)*". The LeRobot importer writes eleven, and six of them are not user keys in any meaningful sense: they are a contract between the importer and the resume path that reads them back.

What the importer writes today, at `src/hflow/importers/lerobot.py:1229-1249`:

```python
            episode_record: dict[str, str] = {
                "task": ...,                     # documented
                "operator": ...,                 # documented
                "embodiment": ...,               # documented
                "source_dataset": ...,           # NOT documented
                "source_revision": ...,          # NOT documented
                "source_episode_index": ...,     # NOT documented
                "converter_version": ...,        # NOT documented
                "camera_keys": ...,              # NOT documented
                "gop_seconds": ...,              # NOT documented
            }
            ...
                episode_record["success"] = ...            # documented
                episode_record["success_derivation"] = ... # NOT documented
```

```
$ grep -c source_episode_index docs/FORMAT.md   ->  0
$ grep -c camera_keys docs/FORMAT.md            ->  0
$ grep -c success_derivation docs/FORMAT.md     ->  0
```

## Why this is worth writing down rather than shrugging at

These are not incidental. `_episode_identity_matches` at `src/hflow/importers/lerobot.py:181-193` reads five of them back to decide whether an already-published landing file may be reused instead of reconverted (#303). A key nobody documented is now load-bearing for a correctness decision, and the person most likely to break it is someone who reads FORMAT.md, sees "any user key passes through untouched", and concludes these are free-form.

`success_derivation` is the one a data consumer actually needs, because it names the methodology behind `success`: `max(stats/next.success)` means any success frame makes the episode a success, which is a claim a buyer filtering on `success` should be able to see.

## One subtlety to get right

`gop_seconds` appears in **both** records. FORMAT.md already documents it under `provenance/v1` as "the keyframe interval actually used", and the importer now also writes it into `episode/v1` so the resume check can compare it without opening the transform's record. Same name, same value, two records, two reasons. Say so explicitly; a reader who finds it in one place will assume the other is a mistake.

## What to do

Add the six undocumented keys to the `episode/v1` table in `docs/FORMAT.md`, and note which are written by the LeRobot importer rather than by a recorder. The existing sentence "All keys are optional; the record is copied/merged from the source recording" stays true and is worth keeping, but it currently reads as though nothing but a recorder ever writes this record.

## Pattern to copy

The `provenance/v1` table immediately below it at `docs/FORMAT.md:117-126` is the right level of detail: one row per key, one clause of meaning, format stated where it is not obvious (`camera_keys` is a compact JSON array of strings, `gop_seconds` is `%g`-formatted).

## Definition of done

1. Every key `src/hflow/importers/lerobot.py:1229-1249` can write appears in the `episode/v1` table.
2. Each row says enough to reproduce the value's format, not just its meaning.
3. `gop_seconds` appearing in two records is explained rather than duplicated silently.
4. `success_derivation`'s relationship to `success` is stated: it names the derivation, and both are absent together when the source carries no outcome feature.
5. No source changes. If writing this up suggests a key is wrong or misnamed, say so on the issue rather than changing it here.

## Validation

Markdown only, so the test suite is not affected. Run the link check from [CONTRIBUTING.md](https://github.com/Hebbian-Robotics/hflow/blob/main/CONTRIBUTING.md):

```bash
```

alorentiar:
Closes #409

## What

`docs/FORMAT.md` documents `episode/v1` as five named keys plus "(any user
key)". The LeRobot importer actually writes eleven: `task`, `operator`,
`embodiment`, `success`, plus seven importer-only keys that the table does not
mention.

## Why

Six of the undocumented keys — `source_dataset`, `source_revision`,
`source_episode_index`, `converter_version`, `camera_keys`, and `gop_seconds` —
are load-bearing: `_episode_identity_matches` reads them back to decide whether
an already-published landing file can be reused instead of reconverted. A
reader who sees "any user key passes through untouched" could reasonably
conclude these are free-form, and break the resume contract without meaning to.

`success_derivation` is the key a data consumer actually needs: it names the
methodology behind `success` (`max(stats/next.success)`: any success frame
makes the episode a success). Both keys are absent together when the source
declares no outcome feature.

`gop_seconds` also appears in `provenance/v1`; for an imported episode both
records carry the same value, and the episode copy exists so the resume check
can compare the keyframe interval without opening the transform's record. The
new prose says so explicitly rather than duplicating the row silently.

## Changes

- Add the seven importer-written keys to the `episode/v1` table, with enough
  detail to reproduce each value's format (`camera_keys` is a compact JSON
  array of strings, `gop_seconds` is `%g`-formatted, `converter_version` is
  currently `lerobot-converter-v7`).
- Explain which keys the LeRobot importer writes (as opposed to a recorder),
  that six of them are a contract with the importer's resume path rather than
  free-form user keys, and how `success`/`success_derivation` relate.

Markdown only — no source or test changes.

## Validation

```bash
# 378 Total · 374 OK · 0 Errors · 4 Excluded
```

### DIFF
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -110,8 +110,32 @@ robot software that produced it. All values are strings.
 | `embodiment` | Robot/platform identifier |
 | `robot_software_version` | Software running on the robot at record time |
 | *(any user key)* | Additional semantics pass through untouched |
+| `source_dataset` | Repo id of the LeRobot dataset the episode was imported from |
+| `source_revision` | Revision of that dataset at import time (the ref the importer resolved) |
+| `source_episode_index` | Zero-based index of the episode within the source dataset, as a decimal string |
+| `converter_version` | Version stamp of the importer that produced the episode (currently `lerobot-converter-v7`) |
+| `camera_keys` | Camera keys converted to video, as a compact JSON array of strings (e.g. `["observation.image"]`) |
+| `gop_seconds` | Keyframe interval actually used, `%g`-formatted (e.g. `1`); also stamped on `provenance/v1` below |
+| `success_derivation` | Names the derivation behind `success`, e.g. `max(stats/next.success)` — any success frame makes the episode a success; written alongside `success`, and both are absent together when the source declares no outcome feature |
 
-All keys are optional; the record is copied/merged from the source recording. Recorders are encouraged to write it at collection time.
+All keys are optional; the record is copied/merged from the source recording.
+Recorders are encouraged to write it at collection time.
+
+The seven keys after *(any user key)* are written by the LeRobot importer, not
+by a recorder, and they are not free-form user keys. Six of them —
+`source_dataset`, `source_revision`, `source_episode_index`,
+`converter_version`, `camera_keys`, `gop_seconds` — are a contract between the
+importer and its resume path: when an earlier run already published a landing
+file for the same episode, the importer reuses it only while each of these
+still matches the source dataset and this conversion's settings. `gop_seconds`
+is deliberately written to both `episode/v1` and `provenance/v1` — same name,
+same value for an imported episode — so the resume check can compare the
+keyframe interval actually used without opening the transform's record.
+`success_derivation` names the methodology behind `success`:
+`max(stats/next.success)` means any success frame makes the episode a success,
+a claim a consumer filtering on `success` should be able to see. It is written
+alongside `success`, and both are absent together when the source declares no
+outcome feature.
 
 ### `provenance/v1`: what produced this file
 
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -109,28 +109,30 @@ robot software that produced it. All values are strings.
 | `success` | Collector-labeled outcome, `"true"`/`"false"` |
 | `embodiment` | Robot/platform identifier |
 | `robot_software_version` | Software running on the robot at record time |
-| *(any user key)* | Additional semantics pass through untouched |
 | `source_dataset` | Repo id of the LeRobot dataset the episode was imported from |
 | `source_revision` | Revision of that dataset at import time (the ref the importer resolved) |
 | `source_episode_index` | Zero-based index of the episode within the source dataset, as a decimal string |
-| `converter_version` | Version stamp of the importer that produced the episode (currently `lerobot-converter-v7`) |
+| `converter_version` | Version stamp of the importer that produced the episode, of the form `lerobot-converter-vN` |
 | `camera_keys` | Camera keys converted to video, as a compact JSON array of strings (e.g. `["observation.image"]`) |
 | `gop_seconds` | Keyframe interval actually used, `%g`-formatted (e.g. `1`); also stamped on `provenance/v1` below |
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging. Welcome, and this is a strong first PR.

You did the thing the issue was really asking for, which was not "list the keys" but "explain what they are for". Every format claim checks out against the source:

```
camera_keys           json.dumps(..., separators=(",", ":"))   compact JSON array   lerobot.py:138
gop_seconds           f"{IMPORT_GOP_SECONDS:g}"                %g-formatted         lerobot.py:1239
source_episode_index  str(episode_index)                       decimal string       lerobot.py:1236
```

And the six-key resume contract is exactly the six `_episode_identity_matches` compares at `lerobot.py:186-193`, which is the part a reader most needs and the part that was easiest to get subtly wrong.

The `gop_seconds` paragraph is the best thing here. Documenting why the same key appears in two records, rather than documenting it twice and leaving a reader to wonder which is authoritative, is the harder and more useful choice.

I pushed one fixup, `2e26b31`, both parts cosmetic. The new rows sat below the `*(any user key)*` row, which is a catch-all, so specific keys after it read as a kind of user key: precisely the opposite of what your prose then says. Moving them above it lets the catch-all stay last where it belongs, and the prose now names the first key rather than pointing at a position, so it survives the next key being added. I also dropped the em dashes; this repo's prose does not use them, which is a house style thing you had no way to know.

One judgment call I made rather than asked about: I changed `converter_version`'s description from "currently `lerobot-converter-v7`" to the form `lerobot-converter-vN`. The value was accurate this morning and moved twice today (v6 in #390, v7 in #403), and a version pinned in prose that nothing tests is exactly the kind of thing that quietly goes stale. Saying the shape rather than the current value keeps it true.

Gate on the merged result: link check 378 links, 374 OK, 0 errors; `ruff check`, `ruff format --check`, `ty check` clean, 1531 passed / 6 skipped.

Closes #409. If you want another, there are three more starter issues open right now: <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>

Two tips that will save you time on a second PR, and neither is a criticism of this one. Only an assignee reserves an issue, so everything unassigned is fair game and a linked PR does not block you. And we cap non-collaborators at one open PR at a time, so a bot will politely close a second and invite you back once this one lands.

chiruu12:
Flagging a line in this PR that looks unintended: it changes the `max_retries` type guard in `src/hflow/build_ai_vlm_checks.py` to `if False:`, which is on `main` now. Details and the behaviour it lets through are in #413.

alorentiar:
Thanks for flagging this, that line should not be there. I traced where it came from: my commit on the branch only touched docs/FORMAT.md, and the guard change arrived in a later commit (2e26b312b) that was pushed after mine. I restored the guard and added a regression test for it in #414.

PR #411, opened 2026-09-05
fix(build-ai): cover completeness knobs in the check version contract

### LINKED ISSUE #404: Build AI check versions exclude max_retries and request_timeout_seconds, so evaluations that can differ share one version identity
## Current behavior

`_check_version` ([src/hflow/build_ai_vlm_checks.py:652-685](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/build_ai_vlm_checks.py#L652-L685)) derives the version identity stamped into every Build AI check run. The OpenAI-compatible branch includes endpoint, model, response_format, temperature, and max_tokens. It excludes `max_retries` — validated at [:164-167](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/build_ai_vlm_checks.py#L164-L167), defaulted at [:143](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/build_ai_vlm_checks.py#L143), and live at [:715](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/build_ai_vlm_checks.py#L715), where it is passed into the OpenAI client and decides whether transient errors become retried requests or terminal failures. The hosted branch includes only the execution label and endpoint; it excludes `request_timeout_seconds` — validated at [:190-195](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/build_ai_vlm_checks.py#L190-L195) and live at [:626](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/build_ai_vlm_checks.py#L626) as the HTTP timeout.

## Controlled result

Two executions that differ only in `max_retries`, and two that differ only in `request_timeout_seconds`, run through the real `_check_version`:

```
=== OpenAI path: same version identity, different completeness ===
max_retries=0 -> build-ai-single-frame-v1-eabed1f2e257cc07
max_retries=5 -> build-ai-single-frame-v1-eabed1f2e257cc07
same version: True

=== hosted path: same version identity, different completeness ===
request_timeout_seconds=1.0  -> build-ai-single-frame-v1-ac6fa96b9b859bec
request_timeout_seconds=60.0 -> build-ai-single-frame-v1-ac6fa96b9b859bec
same version: True
```

## Why this is observable semantics, not reliability noise

The distinction that decides this issue: **sampling noise changes the value of an answer; retries and timeouts change which items produce answers at all.**

Temperature, which is already in the contract, adds sampling noise to a response that exists either way. `max_retries` and `request_timeout_seconds` change the failure mode and the completeness of a corpus: with `max_retries=0`, one transient 500 turns into a failed check run with no result; with `max_retries=5`, the same request becomes a parsed prediction. A short timeout drops slow-but-valid responses that a long timeout includes. Two corpora evaluated under these different settings have systematically different result sets — different error rates, different episode coverage — and a buyer comparing them under one version identity is told they are comparable when they are not.

That is the rule stated on [#365](https://github.com/Hebbian-Robotics/hflow/issues/365): if the effective measurement path changes observable semantics, version and cache/provenance identity must change with it. The contract's own inclusion of `temperature` and `max_tokens` already establishes that behavioral knobs are version-worthy; these two knobs meet the same standard.

## What to build

Open call, not assumed:

1. **Add the fields to the contract**: `max_retries` in the OpenAI-compatible branch, `request_timeout_seconds` in the hosted branch. The precedent is internal (temperature, max_tokens), so this is a consistency fix rather than a design change.
2. **Record the behavior as evidence instead**: leave the contract alone and stamp attempts/timeouts as measurements on every check run, so the difference is visible in the record without moving identity.

The first matches the contract's own logic; the second matches a "retries are infrastructure" reading. What cannot stand is the current middle: knobs that change outcomes, invisible to both the identity and the record.

## Definition of done

1. Every execution knob that changes which items produce results is either in the version contract or recorded per run as a measurement.
2. Two configurations differing only in the newly covered fields produce different version identities, pinned by a test through the real `_check_version`.
3. Configurations that are identical in every covered field keep their existing version, so unchanged methodology does not silently invalidate.
4. The hosted and OpenAI branches treat the rule symmetrically.
5. The demo above lands as a test.

## Non-goals

- Changing the `max_retries` or `request_timeout_seconds` defaults
- Changing how temperature participates in the contract
- Restructuring `step_version_from_contract` or the fingerprint machinery

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q tests/test_build_ai_vlm_checks.py
uv run pytest -q
```

Sagar-024:
Closes #404.

`_check_version` stamped an identity that covered endpoint, model, response_format, temperature, and max_tokens, while two execution knobs that change which items produce results stayed invisible to it: `max_retries` on the OpenAI-compatible path (a transient 500 becomes a parsed prediction under retries and a failed run without them) and `request_timeout_seconds` on the hosted path (a slow-but-valid response is included or dropped by it). The issue's controlled result showed two executions differing only in these knobs sharing one version identity.

The distinction that decided the fix: sampling noise changes the value of an answer; retries and timeouts change which items produce answers at all. `temperature`, already in the contract, only perturbs a result that exists either way. A knob that decides whether an episode yields a prediction or a failed run has a stronger claim to identity, not a weaker one.

Both fields join the contract, symmetrically across the two branches, per the issue's DoD 4: the asymmetry is how this got missed. The migration-era comment on the OpenAI branch was clarified, since its shape-stability constraint applied to that migration alone and does not bind the new fields.

Tests were written red first. Identical configurations keep one version (DoD 3), executions differing only in `max_retries` or `request_timeout_seconds` produce different versions, and a symmetry test holds both branches to the same rule. The demonstration from the issue lands as the red run: two configs sharing `build-ai-single-frame-v1-eabed1f2e257cc07` before the change, diverging after.

The named cost: adding contract fields re-mints every existing Build AI check version, so existing results become non-matching on the next run rather than being silently reused. That is the correct consequence, and DoD 3's stability test is the guard that keeps unchanged methodology from being invalidated.

Gate: ruff, format, ty clean; 1535 passed / 6 skipped.

Refs #404. Option 2 (attempts and timeouts as per-run measurements) stays unfiled per the issue direction; it is a separate enhancement when wanted.

### DIFF
--- a/src/hflow/build_ai_vlm_checks.py
+++ b/src/hflow/build_ai_vlm_checks.py
@@ -659,8 +659,13 @@ def _check_version(configuration: _RegisteredBuildAICheckConfiguration) -> StepV
     }
     match configuration.execution:
         case OpenAICompatibleExecution() as execution:
-            # Keep the existing contract shape so migrating to the explicit
-            # execution value does not invalidate otherwise identical results.
+            # The contract shape was kept stable during the migration to the
+            # explicit execution value so that migration alone did not
+            # invalidate otherwise identical results. That constraint applied
+            # to that migration only: the contract must include every knob
+            # that changes which items produce results, so max_retries is
+            # version-worthy even though it only affects request liveness
+            # (#404). Adding a field re-mints the version by design.
             version_contract.update(
                 {
                     "endpoint": execution.endpoint,
@@ -668,15 +673,19 @@ def _check_version(configuration: _RegisteredBuildAICheckConfiguration) -> StepV
                     "response_format": execution.response_format.value,
                     "temperature": execution.temperature,
                     "max_tokens": execution.max_tokens,
+                    "max_retries": execution.max_retries,
                 }
             )
         case HFlowHostedExecution() as execution:
+            # Same rule, applied symmetrically: the timeout decides whether a
+            # slow-but-valid response is included in the corpus at all.
             version_contract.update(
                 {
                     "execution": "hflow-hosted",
                     "hosted_check_endpoint": _hosted_check_endpoint(
                         execution, configuration.task_definition.task
                     ),
+                    "request_timeout_seconds": execution.request_timeout_seconds,
                 }
             )
         case unexpected_execution:
--- a/tests/test_build_ai_vlm_checks.py
+++ b/tests/test_build_ai_vlm_checks.py
@@ -310,6 +310,51 @@ def test_check_version_stable_when_every_covered_field_is_identical(tmp_path: Pa
     assert first_application.checks[0].version == second_application.checks[0].version
 
 
+# Golden versions for two fixed configurations, one per execution branch.
+#
+# The equality test above only proves _check_version is a function: it cannot
+# fail unless the same input starts producing two answers. The property DoD 3
+# actually claims is that unchanged methodology keeps its identity across
+# changes to this module, and only a recorded value can hold that. Adding a
+# field to the contract, renaming a key, or reordering nothing at all silently
+# re-mints every stored check version; here it fails instead.
+#
+# Editing these strings is the signal, not the chore. Change them only
+# together with a deliberate contract change, and say in the PR why every
+# existing Build AI result is being invalidated.
+_GOLDEN_OPENAI_CHECK_VERSION = "build-ai-single-frame-v1-2aed30388241d554"
+_GOLDEN_HOSTED_CHECK_VERSION = "build-ai-single-frame-v1-400d7f82abd83534"
+
+
+def test_check_version_is_pinned_for_a_fixed_openai_configuration(tmp_path: Path) -> None:
+    application = hflow.App("golden-openai", data_root=tmp_path, default_checks=())
+    hflow.build_ai_vlm_checks.register_hand_visibility(
+        application,
+        execution=hflow.build_ai_vlm_checks.OpenAICompatibleExecution(
+            endpoint="http://localhost:8000/v1",
+            model="model-a",
+            temperature=0.5,
+            max_tokens=512,
+            max_retries=3,
+        ),
+    )
+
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging. Both additions are held, symmetrically, which was the DoD item most likely to be half-done:

```
max_retries dropped                ..._changes_with_max_retries + ..._applies_the_rule_symmetrically
request_timeout_seconds dropped    ..._changes_with_request_timeout_seconds + ..._applies_the_rule_symmetrically
both dropped (main's contract)     3 failures
```

Rewriting the migration-era comment rather than leaving it was the right instinct. A comment saying "keep the shape stable" sitting directly above a change that deliberately breaks the shape is how the next person concludes the change was a mistake.

I pushed one fixup, `c404093`. **The DoD 3 stability test cannot fail.** It registers one execution object twice and asserts the two versions match, which is true of any pure function; it would keep passing if `_check_version` started returning a constant. The property you claimed in its docstring is stronger and better: that unchanged methodology keeps its identity across changes to this module. Only a recorded value holds that, and without one, both of these were invisible:

```
an unrelated field added to the contract   17 passed, nothing noticed
max_retries keyed under a different name   17 passed, nothing noticed
```

Either would silently re-mint every stored Build AI check version. So there is now one golden version per execution branch, with a comment saying that editing those strings is the signal rather than the chore: change them only alongside a deliberate contract change, and say in the PR why every existing result is being invalidated. Both mutations fail now, and so does dropping either field.

That is worth having beyond this PR. Three version-identity incidents landed today (#390's `episode/v1` stamps, #403's success label, and this), and in every case the version moving was correct while nothing would have noticed if it had failed to move. This is the first test in the repo that holds one still.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1546 passed / 6 skipped.

Closes #404. Option 2 staying unfiled is right; file it when someone actually wants attempt counts in the record rather than on spec.

Issue #407, opened 2026-09-04
Eight of the nine refusals in OpenAICompatibleExecution are untested, including both bool-is-an-int guards


kstonekuan:
`OpenAICompatibleExecution` is where a caller's model endpoint configuration is parsed, so its `__post_init__` is a trust boundary: it decides what reaches a remote model call. It has nine refusals. One is tested.

The existing test covers two cases, and only one of them is a `__post_init__` refusal:

```
tests/test_build_ai_vlm_checks.py:107-112
    ("localhost:8000/v1",        "model", "absolute http")           <- _require_absolute_http_url
    ("http://localhost:8000/v1", " ",     "model must not be empty") <- the one covered refusal
```

## Verification

Against `main` at `44e4f50`, each refusal's condition replaced with `if False:` in turn, running `tests/test_build_ai_vlm_checks.py`:

```
model empty                                caught
model padded                               NOTHING NOTICED
response_format wrong type                 NOTHING NOTICED
api_key_environment_variable malformed     NOTHING NOTICED
max_tokens not an integer                  NOTHING NOTICED
max_tokens <= 0                            NOTHING NOTICED
max_retries negative                       NOTHING NOTICED
temperature not finite                     NOTHING NOTICED
max_tokens bool guard alone                NOTHING NOTICED
```

Eight of nine can be deleted with the suite still green.

## The two worth caring about most

`max_tokens` and `max_retries` each carry a separate `isinstance(..., bool)` clause:

```
src/hflow/build_ai_vlm_checks.py:160   if not isinstance(self.max_tokens, int) or isinstance(self.max_tokens, bool):
src/hflow/build_ai_vlm_checks.py:164   if not isinstance(self.max_retries, int) or isinstance(self.max_retries, bool):
```

`bool` subclasses `int`, so `True` passes `isinstance(x, int)` and compares equal to `1`. Without that clause `max_tokens=True` is accepted and becomes a one-token request. Removing the `max_tokens` bool clause alone leaves the suite green, so the clause that exists precisely because this is easy to get wrong is the clause nothing holds. This repo has hit that family repeatedly (#392 for interval bounds, #364 for a schema version), which is why it is worth its own case rather than folding into a general type check.

## What to do

Extend the parametrization at `tests/test_build_ai_vlm_checks.py:107` (or add a second one beside it, since these cases need to vary fields other than `endpoint` and `model`) to cover the eight uncovered refusals, asserting the message rather than just `ValueError`.

[...]

PR #414, opened 2026-09-05
fix(checks): restore the max_retries type guard

### LINKED ISSUE #413: OpenAICompatibleExecution stopped type-checking max_retries in #410
`OpenAICompatibleExecution.__post_init__` stopped type-checking `max_retries` in #410. The change is on `main` now, at `224dc6f`.

#410 is a documentation change to `FORMAT.md`. Its title, its description and the rest of its diff are all about the LeRobot importer's `episode/v1` keys. Alongside that it carries one line in `src/hflow/build_ai_vlm_checks.py`:

```
-        if not isinstance(self.max_retries, int) or isinstance(self.max_retries, bool):
+        if False:
             raise ValueError("max_retries must be an integer")
```

`if False:` is the shape a guard takes while it is being neutered to check whether a test notices, which is the experiment #407 is built on. It looks like one of those was left in rather than backed out.

## Verification

Against `main` at `224dc6f`, constructing the dataclass directly:

```
max_retries=True   accepted, stored as True
max_retries=2.5    accepted, stored as 2.5
max_retries="5"    TypeError: '<' not supported between instances of 'str' and 'int'
max_retries=None   TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

`True` and `2.5` should raise `ValueError("max_retries must be an integer")`. `True` is the case the deleted clause named on its own: `bool` subclasses `int`, so it survives an `isinstance(x, int)` check and then compares equal to `1`.

`"5"` and `None` still fail, but from the `self.max_retries < 0` comparison on the next line rather than from the guard. The caller gets a raw `TypeError` out of a comparison instead of the message that says what was wrong.

The full suite is green with the guard disabled. That is the subject of #407: this is one of the eight refusals in this class with no test behind it, so nothing failed when it went.

## What to do

Restore the clause, and land a case that fails without it. #407 covers this guard among the other seven, so it can go there instead if you would rather keep it in one change.

## On scope

Worth saying separately from the fix: a code change to a validation path travelled in a PR whose title, description and remaining diff are all documentation. Nothing about the PR as described would lead a reviewer to look at `build_ai_vlm_checks.py`, which is how a disabled guard reached `main` on a docs merge. Keeping a PR to what its description claims is what makes review load match review risk.

alorentiar:
Restores the max_retries guard in OpenAICompatibleExecution.__post_init__, which went out disabled in #410. With the guard off, True and 2.5 were accepted as max_retries (details in #413).

The regression test covers both values and fails on current main. Note that #412 adds wider validation coverage to the same test file, so the two may conflict there; happy to rebase if needed.

Fixes #413

### DIFF
--- a/src/hflow/build_ai_vlm_checks.py
+++ b/src/hflow/build_ai_vlm_checks.py
@@ -161,7 +161,7 @@ class OpenAICompatibleExecution:
             raise ValueError("max_tokens must be an integer")
         if self.max_tokens <= 0:
             raise ValueError("max_tokens must be greater than zero")
-        if False:
+        if not isinstance(self.max_retries, int) or isinstance(self.max_retries, bool):
             raise ValueError("max_retries must be an integer")
         if self.max_retries < 0:
             raise ValueError("max_retries must not be negative")
--- a/tests/test_build_ai_vlm_checks.py
+++ b/tests/test_build_ai_vlm_checks.py
@@ -123,16 +123,28 @@ def test_openai_compatible_execution_refuses_invalid_configuration(
         )
 
 
-def test_openai_compatible_execution_refuses_non_integer_max_retries() -> None:
-    kwargs = {"endpoint": "https://example.com/v1", "model": "model"}
+@pytest.mark.parametrize("rejected_max_retries", [True, 2.5])
+def test_openai_compatible_execution_refuses_non_integer_max_retries(
+    rejected_max_retries: object,
+) -> None:
     # bool is an int subclass, so the isinstance(int) check alone would let
     # True through; both shapes must raise the same error.
-    for bad in (True, 2.5):
-        with pytest.raises(ValueError, match="max_retries must be an integer"):
-            hflow.build_ai_vlm_checks.OpenAICompatibleExecution(
-                max_retries=bad, **kwargs  # type: ignore[arg-type]
-            )
-    assert hflow.build_ai_vlm_checks.OpenAICompatibleExecution(max_retries=3, **kwargs)
+    with pytest.raises(ValueError, match="max_retries must be an integer"):
+        hflow.build_ai_vlm_checks.OpenAICompatibleExecution(
+            endpoint="https://example.com/v1",
+            model="model",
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Merging. **This regression was mine, not yours, and I want that on the record before anything else.**

The `if False:` came from my fixup commit on your #410, not from your work. I had been mutation-testing that exact guard while scoping #407, the script restores the source in a `finally` block, and the run hit a ten-minute timeout that killed the process before the restore ran. The mutated line sat in my working tree, I then ran `git add -A` while preparing the FORMAT.md fixup, and it went out inside a commit I described as cosmetic. Your PR was documentation only, and the merge commit attributing a source change to it is my doing.

Two things made it invisible, and both are uncomfortable:

- I ran the full gate on that PR and reported 1531 passing. It did pass. Nothing tests that guard, which is exactly what #407 says, so my own mutation was invisible to the suite I was using to check my own work.
- I described the change as docs-only in the review. The `+2/-1` on a source file was right there in the diff stat and I did not read it.

I have swept main for other leaked mutations and there is one and only one, this line. `rg "if False:" src/ packages/ tests/` comes back with nothing now.

On the fix itself: restoring the guard verbatim is right, and covering both `True` and `2.5` is the right pair, since they fail for different reasons and only the first is the trap. The regression test does fail on main before the change:

```
FAILED test_openai_compatible_execution_refuses_non_integer_max_retries
```

I pushed one fixup, `286a860`. The shared `**kwargs` dict inferred as `dict[str, str]`, so splatting it into `endpoint`, `model`, and `max_retries` produced seven `ty` errors on the merged result. Parametrized instead, with the positive case as its own test, and swapped `# type: ignore[arg-type]` for `# ty: ignore`, which is what this repo uses and what `ty` actually honours. Worth knowing for next time: the mypy-style code is silently ignored here.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1540 passed / 6 skipped.

Fixes #413. Two PRs merged in your first day, one of them cleaning up after me. Thank you for catching it rather than assuming the maintainer knew what he was doing.

PR #419, opened 2026-09-05
test(packaging): cover native manifest parse refusals

### LINKED ISSUE #408: The native overlay manifest parse stage refuses four things and none of them are tested
#393 covered the seven refusals in `_validate_manifest` that one valid-manifest fixture can reach, and #394 landed them. That issue explicitly deferred the rest:

> Not in scope: the 20 other `CythonOverlayManifestError` sites, including the JSON-parsing and canonical-JSON refusals at `:638`, `:749`, and the `_require_*` helpers from `:1986` down. Those are worth covering too, and they can follow.

This is that follow-up. These refusals run *before* `_validate_manifest`, on the raw bytes of an overlay directory's `manifest.json`, which is external input.

## Verification

Against `main` at `44e4f50`, each refusal neutered in turn, running `tests/test_packaging.py`:

```
native overlay manifest is not valid JSON        NOTHING NOTICED
native overlay manifest is not canonical JSON    NOTHING NOTICED
artifacts must be an array                       NOTHING NOTICED
native overlay manifest contains duplicate fields  NOTHING NOTICED
```

## What to do

Cover the parse-stage refusals, in `src/hflow/packaging.py`:

```
:638    native overlay manifest is not valid JSON
:675    artifacts must be an array
:749    native overlay manifest is not canonical JSON
:2067   native overlay manifest contains duplicate fields
```

The canonical-JSON and duplicate-fields ones are the interesting pair. Both exist because a manifest that parses is not necessarily the manifest that was signed: re-serialized JSON with different key order, or a document with a repeated key where `json.loads` silently keeps the last, would otherwise pass. A test that only feeds it garbage will not reach either.

## Pattern to copy

#394 landed the shape at `tests/test_packaging.py`: build a real overlay with `build_cython_overlay`, rewrite `manifest.json`, assert the specific `CythonOverlayManifestError`, then assert nothing was applied (sources intact, no artifacts, no installed manifest, RECORD byte-identical). The helpers `_read_manifest_payload` and `_write_manifest_payload` are already there.

Note that `_write_manifest_payload` writes with `indent=2, sort_keys=True`, which is canonical, so a canonical-JSON test needs to write the bytes directly rather than through that helper.

## Definition of done

1. Each of the four refusals is the only failure when its own `raise` is neutered.
2. The canonical-JSON case uses a payload that is valid JSON and semantically identical to the original, differing only in serialization, so it proves the refusal is about form rather than content.
3. The duplicate-fields case uses raw bytes with a repeated key, since no Python dict can express it.
4. Each case asserts nothing was applied, following #394.

## Not in scope

The `_require_*` helpers from `:1986` down (`must be an object`, `has unexpected fields`, `must be an integer`, `must be positive`, `must use POSIX separators`, `must remain within its root`, `must be a lowercase SHA-256`). Sixteen small refusals sharing one shape; they want a parametrized test of their own and would swamp this issue.

## Prerequisite

These tests build a real native overlay, so they need a system C compiler (`cc`). See the development setup in [CONTRIBUTING.md](https://github.com/Hebbian-Robotics/hflow/blob/main/CONTRIBUTING.md); #398 added that prerequisite this afternoon.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

harshitethic:
## Summary

Add focused regression coverage for the four native-overlay manifest refusals that occur before `_validate_manifest`:

- invalid JSON;
- a non-array `artifacts` field;
- valid, semantically identical but non-canonical JSON;
- duplicate JSON fields.

Each case builds a real native overlay, corrupts only `manifest.json`, calls the public apply path, and verifies that package source bytes, native artifacts, the installed manifest, and wheel `RECORD` remain untouched.

Closes #408.

## Why

These checks protect the untrusted manifest boundary. The canonical-JSON and duplicate-field cases specifically pin the exact serialized document that integrity metadata is intended to describe: a payload can decode to the same Python value while still not be the canonical/signed byte representation.

The canonical case therefore writes raw, semantically identical JSON using a different serialization. The duplicate-field case also writes raw bytes because a Python dict cannot represent repeated keys.

## Validation

GitHub Actions is green on the current head (`26a500a`), including Python 3.11/3.14 checks and both workspace-example matrices. The initial 3.14 lint feedback was a trailing-newline-only issue in the new test file and was corrected in the current head.

Repository gates exercised by CI include lint, format checking, type checking, and tests.

This supersedes #416, which was automatically closed only because the repository permits one open contributor PR at a time while #415 was under review. #415 has since merged.

AI assistance disclosure: AI assistance was used to inspect the issue and repository patterns, prepare the focused regression tests, diagnose CI feedback, and review the resulting change.


kstonekuan (approved):
LGTM, merging. All four parse-stage refusals are the only failure when their own `raise` goes:

```
not-valid-JSON refusal removed        test_apply_refuses_invalid_manifest_json_before_mutation
artifacts-must-be-an-array removed    test_apply_refuses_non_array_artifacts_before_mutation
not-canonical-JSON removed            test_apply_refuses_semantically_identical_noncanonical_manifest...
duplicate-fields removed              test_apply_refuses_duplicate_manifest_fields_before_mutation
```

The canonical-JSON test is the one I was most interested in and you built it correctly:

```python
    assert json.loads(noncanonical_bytes) == payload
    assert noncanonical_bytes != manifest_path.read_bytes()
```

Asserting both halves is what makes it a statement about *form*. Without the first assertion it could be refusing malformed content and passing for the wrong reason, which is the failure mode that makes a serialization test worthless. Same for the duplicate-field case asserting `json.loads(...)["format"] == payload["format"]`: it proves the document parses to the right value and is still refused, which is the whole point of a duplicate-key guard.

Every case asserting source bytes, artifacts, installed manifest, and RECORD are all untouched keeps these about the trust boundary rather than about exception types.

**One thing to fix later, not now.** `_write_example_distribution` and `_example_record_path` are now duplicated between this file and `tests/test_packaging.py`, and the two copies already disagree: yours returns `package_root`, the original returns a tuple. That is drift at birth rather than drift over time, and if the sample distribution ever grows a module, one copy will update and the other will quietly test a different fixture. I have filed it as #420 rather than reworking your PR, since the fix is a judgment call about where shared test fixtures should live in this repo and that is worth deciding once.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1569 passed / 6 skipped.

Closes #408.

PR #421, opened 2026-09-05
test(packaging): share native fixture helpers

### LINKED ISSUE #420: The native overlay sample distribution is built by two copies of one helper, and they already disagree
#419 added `tests/test_packaging_manifest_parsing.py` alongside the existing `tests/test_packaging.py`, and both now build the same sample distribution with their own copy of the same two helpers.

The copies already disagree:

```
tests/test_packaging.py:36                  def _write_example_distribution(
                                                temporary_directory: Path, *, include_record: bool = True
                                            ) -> tuple[Path, Path]
tests/test_packaging_manifest_parsing.py:22 def _write_example_distribution(temporary_directory: Path) -> Path
```

`_example_record_path` is duplicated verbatim in both.

This is drift at birth rather than drift over time, which is the useful thing about catching it now. Both copies write a `sample_native_package` with the same `worker.py`, the same `7.2` dist-info, and the same three RECORD rows. If that fixture ever changes (a second module, a different version, a RECORD with real hashes), one copy moves and the other keeps testing something else, and nothing fails.

## What to do

Give the sample distribution one definition and have both test modules use it.

## The decision this needs

Where shared test fixtures live in this repo, since there is more than one existing answer:

- `tests/conftest.py` for a pytest fixture, which is the idiomatic home and gives both modules it for free.
- A plain helper module, following `packages/hflow-server/tests/ui_test_fixtures.py`, which is the existing precedent for shared non-fixture builders.

Either is defensible. Pick one, say why in the PR, and prefer the one that leaves the call sites reading the way they do now: these tests want `package_root` in hand, not a fixture they have to unpack.

Note the two signatures differ for a reason, `include_record=False` is used by a test that checks the no-RECORD refusal, so the shared version has to keep that parameter. The caller that currently ignores the second tuple element should keep ignoring it, not have the return type changed to suit it.

## Definition of done

1. One definition of the sample distribution, used by both modules.
2. `include_record=False` still reaches the test that needs it.
3. No test changes behaviour: the same tests pass, and the mutation checks from #394 and #419 still fail on their own refusals.
4. `_example_record_path` likewise has one definition.

## Not in scope

Merging the two test modules into one file. Splitting parse-stage refusals from apply-path behaviour is a reasonable division and this issue is about the fixture, not the file layout.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

These tests build a real native overlay, so they need a system C compiler (`cc`); see the setup section of [CONTRIBUTING.md](https://github.com/Hebbian-Robotics/hflow/blob/main/CONTRIBUTING.md).

GruffElixir:
Closes #420

The packaging tests had two copies of the synthetic distribution and RECORD helpers. This extracts the shared fixture into tests/packaging_test_helpers.py and keeps both test modules using one implementation, including the include_record=False path used by refusal coverage.

Validation: python -m compileall -q tests; full repository checks are delegated to CI.

### DIFF
--- a/tests/packaging_test_helpers.py
+++ b/tests/packaging_test_helpers.py
@@ -1,0 +1,119 @@
+"""Shared fixtures for native overlay packaging tests."""
+
+from __future__ import annotations
+
+import base64
+import csv
+import hashlib
+import io
+import sys
+from pathlib import Path
+
+
+def _write_example_distribution(
+    temporary_directory: Path,
+    *,
+    include_record: bool = True,
+) -> tuple[Path, Path]:
+    site_packages_directory = temporary_directory / "site-packages"
+    package_root = site_packages_directory / "sample_native_package"
+    nested_package_root = package_root / "nested"
+    distribution_metadata_root = site_packages_directory / "sample_native_package-7.2.dist-info"
+    nested_package_root.mkdir(parents=True)
+    distribution_metadata_root.mkdir()
+
+    (package_root / "__init__.py").write_text(
+        "from .worker import compute\n\n__all__ = ['compute']\n",
+        encoding="utf-8",
+    )
+    (package_root / "__main__.py").write_text(
+        "from .worker import compute\n\nprint(compute(4))\n",
+        encoding="utf-8",
+    )
+    (package_root / "worker.py").write_text(
+        (
+            "from typing import NewType\n\n"
+            "Token = NewType('Token', str)\n"
+            "Token.__module__ = __name__\n\n"
+            "def compute(value: int) -> int:\n    return value * 3\n"
+        ),
+        encoding="utf-8",
+    )
+    (nested_package_root / "__init__.py").write_text(
+        "from .labels import label\n\n__all__ = ['label']\n",
+        encoding="utf-8",
+    )
+    (nested_package_root / "labels.py").write_text(
+        "def label() -> str:\n    return 'native-result'\n",
+        encoding="utf-8",
+    )
+    (package_root / "py.typed").write_text("", encoding="utf-8")
+    (package_root / "worker.pyc").write_bytes(b"legacy-bytecode")
+    (distribution_metadata_root / "METADATA").write_text(
+        "Metadata-Version: 2.4\nName: sample-native-package\nVersion: 7.2\n",
+        encoding="utf-8",
+    )
+    (distribution_metadata_root / "entry_points.txt").write_text(
+        "[console_scripts]\nsample-native = sample_native_package.worker:compute\n",
+        encoding="utf-8",
+    )
+    license_directory = distribution_metadata_root / "licenses"
+    license_directory.mkdir()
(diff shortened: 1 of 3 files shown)


GruffElixir:
@copilot review

GruffElixir:
Addressed the import feedback in c66bc60: both packaging tests now import the shared helper directly with Ruff-compliant ordering. CI run 33992966304 is green across Python 3.11 and 3.14 checks and workspace examples.

kstonekuan (approved):
LGTM, merging. Welcome.

You picked the right one of the two options and the right one of the two copies. `tests/packaging_test_helpers.py` follows the `ui_test_fixtures.py` precedent rather than inventing a third pattern, and taking `test_packaging.py`'s richer distribution as the survivor is correct: it is the one with the nested package, `py.typed`, and the `include_record` parameter that the no-RECORD refusal needs. Collapsing onto the simpler copy would have quietly reduced what the parse-stage tests build.

The property I most wanted from a test refactor is that it changes no test's meaning, and it does not. Every mutation that failed before still fails:

```
schema_version gate never refuses              test_apply_refuses_invalid_manifest_before_mutation
format gate never refuses                      test_apply_refuses_invalid_manifest_before_mutation
artifacts-must-not-be-empty removed            test_apply_refuses_invalid_manifest_before_mutation
artifact sort order not enforced               test_apply_refuses_invalid_manifest_before_mutation
schema_version dropped from the bundle digest  test_schema_version_is_bound_into_the_bundle_digest
not-valid-JSON / non-canonical / duplicate-fields / artifacts-not-an-array   all four still caught
```

1569 passed, the same count as before, so nothing was lost in the move either.

I pushed one fixup, `5f487e2`. Every name the new module exported was underscore-prefixed, so both test modules were importing private names across a boundary. `ui_test_fixtures.py` is the precedent here too: it exports `PopulatedWorkspace` and `build_populated_workspace` publicly and reserves the underscore for what stays inside. Renamed the four accordingly.

One process note, and it is genuinely not a criticism of the result. `python -m compileall` tells you the files parse, not that the tests pass, and this change moved a fixture that thirty-odd assertions depend on. Because this is your first PR here, CI is held behind a maintainer approval, so "delegated to CI" meant nobody had run it. `uv run pytest -q tests/test_packaging.py` would have taken about twenty seconds and is worth it on a refactor like this. It happens to be correct; you could not have known that.

Gate on the merged result: `ruff check`, `ruff format --check`, `ty check` clean, 1569 passed / 6 skipped.

Closes #420. If you want another, the starter list is at <https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22>. Two things worth knowing: only an assignee reserves an issue, so anything unassigned is fair game, and we cap non-collaborators at one open pull request at a time, so a bot will politely close a second and invite you back when this one lands.

PR #424, opened 2026-09-06
fix(example): include max_retries in Build AI run fingerprint

### LINKED ISSUE #422: The Build AI example's run fingerprint excludes max_retries, so two runs with different coverage share one output directory
#411 fixed this in `src/hflow/build_ai_vlm_checks.py` yesterday: `max_retries` changes which items produce answers, so it belongs in the version identity. The Build AI evaluation example has the same gap in its own run fingerprint, and it was not touched.

`max_retries` is recorded but sits outside the contract that gets hashed:

```python
examples/build_ai_evaluation/evaluate.py:509   result_contract = {
                                                   ... "temperature", "max_tokens", ...
examples/build_ai_evaluation/evaluate.py:524   }
examples/build_ai_evaluation/evaluate.py:528   "fingerprint": hflow.fingerprint_contract(result_contract),
examples/build_ai_evaluation/evaluate.py:532   "max_retries": configuration.max_retries,   <- outside result_contract
```

It reaches the real request at `:359`, so it is live rather than decorative.

## Why it matters here specifically

The fingerprint is not just a label. It gates output-directory reuse:

```python
examples/build_ai_evaluation/evaluate.py:564
        if existing_metadata.fingerprint != current_metadata.fingerprint:
            raise ValueError(
                f"{metadata_path} describes a different experiment; choose another --output"
            )
```

So a run with `max_retries=0` and a run with `max_retries=5` are currently judged the same experiment and allowed to append to one output directory. Under retries a transient 500 becomes a parsed prediction; without them it becomes a failed row. The two runs have systematically different coverage, and the guard that exists to stop incompatible experiments mixing does not see the difference.

The documentation states the current behaviour accurately, which is how it stayed invisible:

```
docs/how-to/run-build-ai-evaluation.md:315
    the endpoint, model, prompt contents, schema mode, temperature, token limit,
    selected rows, sources, and tasks
```

## What to do

Move `max_retries` into `result_contract`, and add it to that documented list.

**Do not move `worker_count`.** It is excluded for a good reason and the distinction is the interesting part of this issue: parallelism changes how fast the rows are evaluated, not which rows produce answers. Two runs differing only in worker count are genuinely the same experiment. `api_key_environment_variable` stays out for the same kind of reason: it names where a credential is read from, not what the model does.

## Pattern to copy

#411, which made exactly this argument and this change one layer in. Its tests are the shape to follow: one asserting two configurations differing only in `max_retries` produce different fingerprints, and one asserting a configuration identical in every covered field keeps its fingerprint.

## Definition of done

1. Two configurations differing only in `max_retries` produce different fingerprints, pinned by a test.
2. A configuration identical in every covered field keeps the fingerprint it has, so unchanged methodology is not invalidated. Worth a recorded value rather than comparing one config to itself; see the golden pins added to `tests/test_build_ai_vlm_checks.py` in #411 for why that distinction matters.
3. `worker_count` still does not participate, with a comment saying why.
4. `docs/how-to/run-build-ai-evaluation.md:315` lists the field.
5. Existing run directories will stop matching, which is correct. Say so in the PR.

## Validation

This lives in a workspace example, so it has its own project:

```bash
uv run --locked --project examples/build_ai_evaluation ruff check examples/build_ai_evaluation
uv run --locked --project examples/build_ai_evaluation ruff format --check examples/build_ai_evaluation
uv run --locked --project examples/build_ai_evaluation ty check --project examples/build_ai_evaluation --extra-search-path . examples/build_ai_evaluation
```

Then the root gate from [CONTRIBUTING.md](https://github.com/Hebbian-Robotics/hflow/blob/main/CONTRIBUTING.md).

harshitethic:
## Summary

Include `max_retries` in the Build AI evaluation result contract so runs with different retry policies cannot silently reuse the same output directory.

`max_retries` changes which requests recover from transient failures and therefore changes evaluation coverage. `worker_count` remains outside the fingerprint because it changes throughput only, not methodology or results.

## Changes

- move `max_retries` into the hashed result contract
- keep `worker_count` execution-only, with an explanatory comment
- add regression coverage proving retry changes alter the fingerprint while worker-count changes do not
- pin the unchanged methodology fingerprint to catch accidental identity drift
- document retry policy as part of the run fingerprint

Existing run directories created before this change will no longer match the new fingerprint, which is intentional because their retry methodology was not represented in their identity.

## Validation

- `ruff format` on the changed Python files — passed
- `ruff check` on the changed Python files — passed
- `git diff --check` — passed
- focused fingerprint behavior verified directly in the locked example environment: stable golden fingerprint, changed fingerprint for `max_retries`, unchanged fingerprint for `worker_count`
- full pytest collection is blocked on this Windows machine by the repository's POSIX `fcntl` dependency (`ModuleNotFoundError: fcntl`)

Closes #422

AI assistance disclosure: AI assistance was used to inspect the issue and code, prepare the focused implementation and tests, run validation, and review the resulting diff.

### DIFF
--- a/docs/how-to/run-build-ai-evaluation.md
+++ b/docs/how-to/run-build-ai-evaluation.md
@@ -335,7 +335,7 @@ For an unauthenticated self-hosted endpoint, add
 default because Build AI did not publish one; set it explicitly when a model
 comparison requires a fixed non-default value. The run fingerprint captures
 the endpoint, model, prompt contents, schema mode, temperature, token limit,
-selected rows, sources, and tasks so incompatible experiments cannot silently
+retry policy, selected rows, sources, and tasks so incompatible experiments cannot silently
 reuse the same output directory.
 
 ## Inspect and compare results
--- a/docs/how-to/run-build-ai-evaluation.md
+++ b/docs/how-to/run-build-ai-evaluation.md
@@ -335,8 +335,8 @@ For an unauthenticated self-hosted endpoint, add
 default because Build AI did not publish one; set it explicitly when a model
 comparison requires a fixed non-default value. The run fingerprint captures
 the endpoint, model, prompt contents, schema mode, temperature, token limit,
-retry policy, selected rows, sources, and tasks so incompatible experiments cannot silently
-reuse the same output directory.
+retry policy, selected rows, sources, and tasks so incompatible experiments
+cannot silently reuse the same output directory.
 
 ## Inspect and compare results
 
--- a/examples/build_ai_evaluation/evaluate.py
+++ b/examples/build_ai_evaluation/evaluate.py
@@ -519,6 +519,7 @@ def _run_metadata_document(configuration: EvaluationConfiguration) -> dict[str, 
         "response_format": configuration.response_format.value,
         "temperature": configuration.temperature,
         "max_tokens": configuration.max_tokens,
+        "max_retries": configuration.max_retries,
         "row_limit_per_source": configuration.row_limit_per_source,
         "prompts": prompt_metadata,
     }
@@ -528,8 +529,8 @@ def _run_metadata_document(configuration: EvaluationConfiguration) -> dict[str, 
         "fingerprint": hflow.fingerprint_contract(result_contract),
         **result_contract,
         "api_key_environment_variable": configuration.api_key_environment_variable,
+        # Execution parallelism changes throughput, not evaluation results.
         "worker_count": configuration.worker_count,
-        "max_retries": configuration.max_retries,
     }
 
 
--- a/examples/build_ai_evaluation/tests/test_evaluation.py
+++ b/examples/build_ai_evaluation/tests/test_evaluation.py
@@ -29,6 +29,7 @@ from examples.build_ai_evaluation.evaluate import (
     _argument_parser,
     _prepare_output_directory,
     _run_configuration_from_arguments,
+    _run_metadata_document,
     _sample_result,
     _sanitize_base_url,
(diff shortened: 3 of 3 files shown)


kstonekuan (approved):
The `worker_count` line earns its comment: throughput out, methodology in is the right cut, and stating it next to the field stops the next person relitigating it.

Validated on your branch against current main, since the fcntl import blocks collection on Windows: example gate clean, 36 passed, root suite 1582 passed / 6 skipped. Both new tests hold under mutation:

- Moving `max_retries` back out of the contract (the #422 bug restored) turns `test_run_fingerprint_changes_with_max_retries_but_not_worker_count` red.
- Promoting `worker_count` into the contract instead turns the same test red from the other direction.
- Dropping `temperature` turns the golden pin red.

That last one is why the pin was worth adding. It catches contract drift the behavioral test cannot see, including a change someone makes on purpose without realizing it moves every existing run's identity.

Pushed one commit, `docs(build-ai): rewrap the fingerprint paragraph`: your added clause ran the line to 91 characters against a paragraph wrapped at 78. Prose only.

Merging.

One thing, and it is an invitation rather than a rule: the `good first issue` pool is best left for people making their first contribution here, and you are five PRs in. Higher-leverage work is open if you want it. [#287](https://github.com/Hebbian-Robotics/hflow/issues/287) is the standing one: run HFlow against a real corpus such as Egocentric-10K or Egocentric-100K and report what breaks, what is slow, and what is awkward. That kind of report is worth more to us than any issue we can write from the inside, because we cannot see those failures from here.

PR #425, opened 2026-09-06
refactor: remove unreachable validation guards

### LINKED ISSUE #423: Three guards that cannot fire: two zero-size checks in the LeRobot resume path and an unreachable isinstance in the EgoSuite reader
Two guards that read as safety checks and are not. Both were noticed during review and left in place at the time; both are still dead on current `main`.

## 1. Two zero-size checks in the LeRobot resume path

```
src/hflow/importers/lerobot.py:163   if local_episode_path.stat().st_size < 1:      -> return False
src/hflow/importers/lerobot.py:209   if storage.file_size(relative_key) < 1:        -> return None
```

Removing **both** leaves the suite green:

```
lerobot: both zero-size guards removed    STILL DEAD   [75 passed in 1.46s]
```

An empty file is not a readable MCAP, so `open_reader` raises and the existing refusal already covers it. Neither check ever decides anything.

One of the two has a reason to survive: `:209` runs before `storage.fetch`, so on a bucket root it avoids downloading a zero-byte object to discover it is empty. That reason is invisible to a local-root test and belongs in a comment. `:163` runs after the fetch and has no such excuse.

## 2. An unreachable isinstance in the EgoSuite label reader

```
examples/egosuite_evaluation/evaluate.py:173
    if not isinstance(label_type, str) or label_type != PROJECTED_HAND_LABEL_TYPE:
```

Only a `str` can equal a `str` constant, so the first clause never changes the outcome. Dropping it changes no test:

```
egosuite: label_type isinstance dropped, equality kept    STILL DEAD   [55 passed in 3.27s]
```

Its sibling two lines up is genuinely load-bearing and must stay, which is what makes this worth a careful hand rather than a blanket cleanup:

```
examples/egosuite_evaluation/evaluate.py:163-167
    not isinstance(schema_version, int)
    or isinstance(schema_version, bool)      <- bool subclasses int; True == 1 without this
    or schema_version != SCHEMA_VERSION
```

## Why bother

A guard that cannot fire still costs a reader: the next person to touch `_episode_identity_matches` or `load_projected_hand_label_report` has to work out whether it protects something before changing anything near it. #419 removed exactly one such branch and #394 kept the reasoning as a comment instead; both are reasonable outcomes and this issue is about picking one deliberately per case rather than leaving all three unexamined.

## What to do

1. Remove `lerobot.py:163`.
2. Keep `lerobot.py:209` and add a comment saying it exists to avoid a download, not to catch an empty file.
3. Remove the `isinstance(label_type, str)` clause at `evaluate.py:173`.
4. Leave the `isinstance(schema_version, bool)` clause at `:165` exactly as it is.

## Definition of done

1. The three changes above, and no others.
2. No test changes, and none needed: these lines are dead, so removing them must not move a single assertion. If a test does fail, the line was not dead and the issue is wrong; say so rather than adjusting the test.
3. The bool guard at `evaluate.py:165` still fails its case when removed. That is `test_saved_label_report_rejects_unsupported_envelope_before_frames[bool-schema-version]`; run it and say you did.

## Not in scope

Any other `isinstance` or defensive branch in these files. Three specific lines, verified dead, nothing else.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

ryux1:
Closes #423.

## Summary

Removes the redundant post-fetch zero-size check from the LeRobot resume path, while documenting why the pre-fetch check remains useful for remote storage. It also simplifies the EgoSuite label-type check to direct comparison. The separate `bool` schema-version guard remains unchanged and tested.

## Why

An empty MCAP is already rejected by the existing reader exception path after fetch, and a non-string label type cannot equal the supported string constant. Removing those dead conditions reduces misleading defensive logic without changing observable behavior. The storage-level size check remains because it can avoid fetching an empty remote object.

## Validation

- `uv run ruff format --check` — 222 files already formatted
- `uv run ty check` — passed
- `uv run pytest -q tests/test_lerobot_converter.py` — 75 passed
- `uv run --locked --project examples/egosuite_evaluation ruff check examples/egosuite_evaluation` — passed
- `uv run --locked --project examples/egosuite_evaluation ruff format --check examples/egosuite_evaluation` — 6 files already formatted
- `uv run --locked --project examples/egosuite_evaluation pytest -q examples/egosuite_evaluation/tests` — 55 passed
- Required bool guard case — 1 passed
- Full root suite — 1,578 passed, 6 skipped, 1 unrelated local FFmpeg 9.0 failure (`-filter_script:v` is unavailable)

## Checklist

- [x] No test changes: #423 explicitly identifies these branches as dead; existing outcome-focused coverage passes.
- [x] No documentation update: observable behavior, flags, formats, and requirements are unchanged.
- [x] I ran `uv run ruff check`, `uv run ruff format --check`, and `uv run ty check`.
- [x] I ran the relevant pytest suites.
- [x] Stored-data compatibility is unchanged; no behavior-version bump is required.

Developed with assistance from OpenAI Codex.

### DIFF
--- a/examples/egosuite_evaluation/evaluate.py
+++ b/examples/egosuite_evaluation/evaluate.py
@@ -170,7 +170,7 @@ def load_projected_hand_label_report(
             f"{schema_version!r}; supported value is {SCHEMA_VERSION}"
         )
     label_type = report_payload.get("label_type")
-    if not isinstance(label_type, str) or label_type != PROJECTED_HAND_LABEL_TYPE:
+    if label_type != PROJECTED_HAND_LABEL_TYPE:
         raise ValueError(
             f"projected-hand label report {report_path} field 'label_type' has value "
             f"{label_type!r}; supported value is {PROJECTED_HAND_LABEL_TYPE!r}"
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -160,8 +160,6 @@ def _episode_identity_matches(
 
     reader = None
     try:
-        if local_episode_path.stat().st_size < 1:
-            return False
         reader = open_reader(local_episode_path)
         metadata_records = reader.metadata()
     except (OSError, McapError, ValueError):
@@ -206,6 +204,7 @@ def _try_reuse_completed_episode(
     if not storage.exists(relative_key):
         return None
     try:
+        # Avoid fetching an empty remote object only to reject it as invalid MCAP.
         if storage.file_size(relative_key) < 1:
             return None
         local_episode_path = storage.fetch(relative_key)


kstonekuan (approved):
Exactly the three lines, and the reasoning for keeping `:209` landed in the comment where the next reader will find it.

Validated on your branch: full gate clean, 1579 passed / 6 skipped, EgoSuite example 55 passed. Two things I checked beyond the suite, since "this line is dead" is a claim tests cannot make on their own:

- A zero-byte file and a truncated one both still make `_episode_identity_matches` return `False`, now through the reader refusal. That is the behavior the removed guard was standing in for, and it survived.
- Removing the `bool` schema-version clause you left alone turns `test_saved_label_report_rejects_unsupported_envelope_before_frames[bool-schema-version]` red on its own. Still load-bearing, as expected.

LGTM.

PR #429, opened 2026-09-06
fix(import): CRC-validate reused landing episodes before stamping receipts

### LINKED ISSUE #426: The resume path stamps a conversion receipt over reused bytes it never integrity-checked
## Current behavior

The #390 resume path reuses a landing episode when its recorded identity matches: `_episode_identity_matches` ([src/hflow/importers/lerobot.py:151-193](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L151-L193)) reads the file's metadata records and compares seven fields (source dataset, revision, episode index, converter version in two records, camera keys, gop_seconds in two records). It never reads the message payloads. `_try_reuse_completed_episode` ([src/hflow/importers/lerobot.py:197-226](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L197-L226)) then stamps a fresh conversion-grade receipt over the reused bytes: `"content_id": content_episode_id(local_episode_path)` at [:223](https://github.com/Hebbian-Robotics/hflow/blob/main/src/hflow/importers/lerobot.py#L223).

So the resume path issues a receipt over bytes it never integrity-checked. The receipt is a true hash of whatever is on disk, but nothing distinguishes those bytes from the output the original conversion produced.

## Controlled result

Built a two-episode delivery through the real import path, then flipped 400 bytes inside the first episode's chunk payload (file size unchanged, MCAP structure and all metadata records intact, exactly like on-disk bit rot or a partial external copy that survives as valid MCAP). Ran the real `_try_reuse_completed_episode` over intact and damaged bytes:

```
--- intact episode ---
identity matches: True
receipt content_id: ce5f28aabcdef987
--- damaged payload (400 bytes flipped, structure intact) ---
size unchanged: True
identity matches: True
reuse receipt issued: True
receipt content_id: 28ef502d51885a3f
content_id changed with the damage: True
verdict: the resume path stamps a fresh conversion-grade receipt on bytes it never integrity-checked
```

The resume scenario makes this quiet: run one failed mid-batch, and the manifest publishes last, so run one left no receipt at all. There is no prior content id for a comparison to catch. The damaged episode enters run two's manifest as a legitimate conversion product, and a consumer who received that delivery has no error surface.

## Why now

#423 just removed two dead guards from this same path, on the finding that a guard which cannot decide anything is noise. This issue is the sibling: the reuse path's identity check reads metadata and trusts bytes, and the receipt it stamps afterwards makes the trust invisible. The #389 receipt exists precisely so deliveries can be checked; this path stamps that receipt over unverified bytes.

## What to build

Open call, not assumed:

1. **Integrity-check at reuse**: the reuse path already fetches and reads the local file, so a CRC-validated read (`validate_crcs=True`, the same standard `hflow doctor` applies) is nearly free. A file that fails falls through to re-conversion: damaged work is not reuse, it is work to redo. No refusal is added; the caller's flow is unchanged.
2. **Record the verification**: whether the receipt gains a field or the reuse logs it, the delivery should distinguish "reused after verification" from "reused unverified".

## Definition of done

1. A reused episode's payload is CRC-validated before the receipt is stamped.
2. A payload-corrupted landing file is not reused: it falls through to re-conversion.
3. The zero-size and structural-damage cases keep their current behavior.
4. The fixture above lands as a test.

## Non-goals

- Changing the manifest schema (#389's receipt already carries the content id)
- Refusing uploads or adding CLI commands
- Changing the conversion path for fresh conversions

## Validation

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest -q
```

Sagar-024:
Closes #426.

The #390 resume path stamped conversion-grade receipts over reused bytes it never integrity-checked: `_episode_identity_matches` compared seven metadata fields and never read the message payloads. A landing file with payload damage that left the container and metadata intact (on-disk bit rot, a partial external copy) matched identity and received a fresh receipt. The resume scenario made it quiet: run one failed mid-batch and published no manifest (manifest-last), so nothing recorded the damage.

The fix: the identity check now ends with a CRC-validated full message pass, the same standard `hflow doctor` applies to any canonical file. Metadata matching alone is not integrity; the metadata records live outside the chunks, so payload damage never reaches them. A file that fails the CRC pass falls through to re-conversion: damaged work is not reuse, it is work to redo. The reuse log line now says "reusing verified" and names the content id.

Measured cost on a real canonical episode: the CRC-validated full message read is about 1.15x the content id hash pass that already ran pre-#426 (2.6 ms vs 2.3 ms at fixture scale, 51 ms vs 40 ms at 8 MB scale). The read was already happening for the receipt; the CRC flag makes the same pass verify.

tests/reuse_test_helpers.py adds `flip_chunk_payload_bytes`, a surgical corruption helper that flips bytes inside a chunk's records region while leaving the container, indexes, and metadata intact, shared so the #423-family tests can reuse it. The identity fixture now writes one chunked message (CompressionType.NONE) so the corruption is addressable in place.

Controlled result on current main: 400 payload bytes flipped, size and metadata unchanged, identity matched without the fix, and the reuse receipt carried a different content id (ce5f28aa -> 28ef502d). With the fix, the damaged file is refused and re-converted. Mutation: deleting the CRC pass turns the controlled test red.

Gate: ruff, format, ty clean; 1587 passed / 6 skipped.

Refs #426. Fourth brick of the trust thread with #376, #389, and #395.

### DIFF
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -29,6 +29,7 @@ from pathlib import Path, PurePosixPath
 from typing import NotRequired, TypedDict
 from urllib.parse import urlsplit
 
+from mcap.reader import make_reader
 from mcap.writer import Writer as McapWriter
 
 from hflow.catalog import content_episode_id
@@ -155,7 +156,13 @@ def _episode_identity_matches(
     episode_index: int,
     camera_keys: tuple[str, ...],
 ) -> bool:
-    """True when a published landing file belongs to this exact import."""
+    """True when a published landing file belongs to this exact import.
+
+    Identity over the metadata records, then a CRC-validated full message
+    pass: a reused episode must be whole, not merely labeled. Metadata-only
+    reads never touch chunk payloads, so without this pass a payload-damaged
+    file would match identity and be stamped with a fresh receipt.
+    """
     from mcap.exceptions import McapError
 
     reader = None
@@ -180,7 +187,7 @@ def _episode_identity_matches(
     if recorded_camera_keys is None:
         return False
     expected_gop = f"{IMPORT_GOP_SECONDS:g}"
-    return (
+    if not (
         episode_metadata.get("source_dataset") == dataset_source.repo_id
         and episode_metadata.get("source_revision") == dataset_source.revision
         and episode_metadata.get("source_episode_index") == str(episode_index)
@@ -189,7 +196,19 @@ def _episode_identity_matches(
         and recorded_camera_keys == camera_keys
         and episode_metadata.get("gop_seconds") == expected_gop
         and provenance_metadata.get("gop_seconds") == expected_gop
-    )
+    ):
+        return False
+    # Metadata matching is not integrity: the metadata records live outside
+    # the chunks, so payload damage never reaches them. Reuse must hold the
+    # file to the same standard hflow doctor applies to any canonical file.
+    try:
+        with local_episode_path.open("rb") as stream:
+            validated_reader = make_reader(stream, validate_crcs=True)
+            for _ in validated_reader.iter_messages(log_time_order=False):
+                pass
+    except (OSError, McapError, ValueError):
+        return False
+    return True
 
 
 def _try_reuse_completed_episode(
@@ -923,7 +942,11 @@ def import_lerobot_dataset(
             camera_keys=resolved_camera_keys,
         )
         if reused_episode is not None:
-            logger.info("reusing completed LeRobot episode %s", reused_episode["uri"])
+            logger.info(
+                "reusing verified completed LeRobot episode %s (content_id %s)",
+                reused_episode["uri"],
+                reused_episode["content_id"],
+            )
             published_episodes.append(reused_episode)
             continue
         published_episodes.append(
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -206,7 +206,17 @@ def _episode_identity_matches(
             validated_reader = make_reader(stream, validate_crcs=True)
             for _ in validated_reader.iter_messages(log_time_order=False):
                 pass
-    except (OSError, McapError, ValueError):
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
The measurement is what makes this reviewable. 1.15x on a read that was already happening turns "is this too expensive" from an argument into a number, and 51 ms against a re-conversion is not a real cost.

Validated on your branch: full gate clean, 1587 passed / 6 skipped. Four mutations, because a hand-edited corrupt file can easily be refused for a reason that has nothing to do with the feature:

- `validate_crcs=True` flipped to `False`, everything else intact: both tests red. This is the one I most wanted to see, since it proves the flag is doing the work rather than the extra read.
- The whole pass removed: red.
- The CRC failure swallowed and reused anyway: red.
- Reader opened but messages never iterated: red. Confirms iteration is load-bearing, which it is, since the CRC is checked when the chunk is decoded.

So the damaged file is refused for the stated reason and not by accident.

Pushed one commit, `feat(import): warn when a reused episode fails CRC validation`, for the half of #426's second definition-of-done item that was still open. You covered "reused after verification" with the log line. The other side was silent: a damaged episode fell through to re-conversion indistinguishably from a first import, so the import repairs the bit rot and nobody ever learns the landing tree is rotting. The warning fires only in the branch where identity already matched, so it cannot fire for a stranger's file, and it names the path. The test now asserts it, and removing the warning turns that test red on its own.

One thing I left alone: `test_reuse_accepts_an_intact_episode_after_the_crc_pass` is a strict subset of the damaged test, which already asserts both the intact receipt and the refusal after corruption. It costs nothing and reads as documentation, so it stays, but it is not adding coverage.

Putting the CRC pass inside `_episode_identity_matches` means a caller cannot tell "not mine" from "mine but damaged". That is fine while both answers mean re-convert, and the warning covers the observability gap it would otherwise leave. Worth remembering if anything ever needs to branch on the difference.

Merging.

PR #430, opened 2026-09-06
fix(example): persist Build AI missing-key policy

### LINKED ISSUE #427: run.json omits allow_missing_api_key, so a run against an unauthenticated endpoint cannot be replayed from its own record
`run.json` is the Build AI evaluation's record of how a run was configured. The how-to says it holds "request settings" ([`docs/how-to/run-build-ai-evaluation.md:359`](https://github.com/Hebbian-Robotics/hflow/blob/main/docs/how-to/run-build-ai-evaluation.md#L359)). One request setting is missing, and it is the one a replay needs.

## Current behavior

`EvaluationConfiguration` has 17 fields (`examples/build_ai_evaluation/evaluate.py:99-116`). `_run_metadata_document` at `:498` records all of them except paths, the run label, and `allow_missing_api_key`. Grep confirms the flag reaches the document nowhere:

```
examples/build_ai_evaluation/evaluate.py:108    allow_missing_api_key: bool          <- declared
examples/build_ai_evaluation/evaluate.py:883    if not api_key and not ...           <- used
examples/build_ai_evaluation/evaluate.py:1008   allow_missing_api_key=arguments...   <- parsed
```

The existing schema test enumerates every key in the document (`examples/build_ai_evaluation/tests/test_evaluation.py:604-625`); the flag is not among the 20.

That matters at `:883`:

```python
api_key = os.environ.get(configuration.api_key_environment_variable)
if not api_key and not configuration.allow_missing_api_key:
    raise ValueError(...)
```

So a run against a local unauthenticated vLLM or Ollama endpoint needs `--allow-missing-api-key`, and nothing in its `run.json` says so. Someone handed that file and asked to reproduce the run gets a `ValueError` and has to guess the flag.

## Where it goes

Outside the fingerprint, not inside. Whether the caller waived the API-key requirement does not change a single answer the model gives, so hashing it would make two identical experiments claim different identities. It belongs beside `worker_count`, which #424 just settled as the pattern for exactly this category:

```python
        "api_key_environment_variable": configuration.api_key_environment_variable,
        # Execution parallelism changes throughput, not evaluation results.
        "worker_count": configuration.worker_count,
```

`examples/build_ai_evaluation/evaluate.py:531-533`. Add the field next to it with a comment in the same spirit, saying why it is recorded but not hashed.

## Definition of done

1. `_run_metadata_document` records `allow_missing_api_key` outside `result_contract`.
2. The schema test at `:604` lists it, so the document's key set stays pinned.
3. A test proving it does not move the fingerprint: build two configurations differing only in that flag and assert the fingerprints are equal. `test_run_fingerprint_changes_with_max_retries_but_not_worker_count` at `:480` is the shape to copy.
4. The golden fingerprint pin at `:493` still passes unchanged. If it moves, the field went in the hashed contract by mistake.
5. The how-to's `run.json` row mentions it only if you think a reader needs it. Prose change optional; the record is the point.

## Not in scope

`data_directory`, `output_directory`, and `label` stay out. Paths are where a run was written, not what it did, and the label is a human name for it.

## Validation

```bash
uv run --locked --project examples/build_ai_evaluation ruff check examples/build_ai_evaluation
uv run --locked --project examples/build_ai_evaluation ruff format --check examples/build_ai_evaluation
uv run --locked --project examples/build_ai_evaluation ty check --project examples/build_ai_evaluation --extra-search-path . examples/build_ai_evaluation
```

Akimbo92i:
## Summary

Fixes the Build AI evaluation example run metadata so unauthenticated-endpoint runs can be replayed from their own `run.json` record.

## Changes

- Persist `configuration.allow_missing_api_key` in the generated run metadata document.
- Keep `allow_missing_api_key` outside the hashed `result_contract`, matching its runtime/replay setup role rather than result semantics.
- Add regression coverage for the persisted metadata field and unchanged stable fingerprint when only the missing-key policy changes.

## Testing

- `uv run --locked --project examples/build_ai_evaluation ruff check examples/build_ai_evaluation`
- `uv run --locked --project examples/build_ai_evaluation ruff format --check examples/build_ai_evaluation`
- `uv run --locked --project examples/build_ai_evaluation ty check --project examples/build_ai_evaluation --extra-search-path . examples/build_ai_evaluation`
- `uv run --locked --project examples/build_ai_evaluation pytest -q examples/build_ai_evaluation/tests`
- `git diff --check origin/main...HEAD`

## Related Issue

Fixes #427

### DIFF
--- a/examples/build_ai_evaluation/evaluate.py
+++ b/examples/build_ai_evaluation/evaluate.py
@@ -529,6 +529,8 @@ def _run_metadata_document(configuration: EvaluationConfiguration) -> dict[str, 
         "fingerprint": hflow.fingerprint_contract(result_contract),
         **result_contract,
         "api_key_environment_variable": configuration.api_key_environment_variable,
+        # Waiving API-key presence affects replay setup, not evaluation results.
+        "allow_missing_api_key": configuration.allow_missing_api_key,
         # Execution parallelism changes throughput, not evaluation results.
         "worker_count": configuration.worker_count,
     }
--- a/examples/build_ai_evaluation/tests/test_evaluation.py
+++ b/examples/build_ai_evaluation/tests/test_evaluation.py
@@ -486,9 +486,13 @@ def test_run_fingerprint_changes_with_max_retries_but_not_worker_count(tmp_path:
     parallel = _run_metadata_document(
         replace(configuration, worker_count=configuration.worker_count + 1)
     )
+    missing_key_policy_changed = _run_metadata_document(
+        replace(configuration, allow_missing_api_key=not configuration.allow_missing_api_key)
+    )
 
     assert retried["fingerprint"] != original["fingerprint"]
     assert parallel["fingerprint"] == original["fingerprint"]
+    assert missing_key_policy_changed["fingerprint"] == original["fingerprint"]
 
 
 def test_run_fingerprint_for_unchanged_methodology_is_stable(tmp_path: Path) -> None:
@@ -603,6 +607,7 @@ def test_run_metadata_document_persists_the_existing_schema(tmp_path: Path) -> N
     assert document == metadata.to_json_dict()
     assert set(document) == {
         "adapter_schema_version",
+        "allow_missing_api_key",
         "api_key_environment_variable",
         "base_url",
         "dataset_repository",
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
Right field, right half of the document, and the comment says why in the same voice as the one above it.

Validated on your branch: example gate clean, 36 passed. Two mutations, both caught:

- Moving the field into the hashed contract turns the fingerprint test and the golden pin red, which is the mistake this change was most likely to make.
- Dropping it again turns the schema test red, so #427 cannot quietly come back.

Asserting `is True` rather than truthiness in the schema test was the right instinct here. `bool` subclasses `int` and this repo has been bitten by that three times in the last week.

LGTM, merging. Thanks for the clean first contribution.

PR #433, opened 2026-09-06
refactor(lerobot): parse info.json into a typed archive model

### LINKED ISSUE #320: Parse LeRobot metadata into a typed internal archive model

chiruu12:
Closes #320.

`meta/info.json` is now parsed once, at the boundary, into immutable values. Nothing after
that point reads the external document.

## The typed model

`_parse_dataset_information(document) -> _DatasetInformation` validates and returns:

- `fps`, `data_path`, `video_path`: the existing guards, moved here unchanged
- `robot_type`: normalised to `"unknown"` once, instead of at the provenance write
- `video_feature_names`: the `dtype: video` features, the fallback for a corpus whose
  episode metadata carries no video window columns
- `numeric_features`: `dict[str, _NumericFeature]`, each carrying name, dtype and shape
- `depth_map_feature_names`: computed with the existing `_is_depth_map_feature()`

`_SourceArchive`, `_EpisodeRow` and `_VideoWindow` are frozen dataclasses. The archive holds
the parsed information, a tuple of episodes, the video keys, the cache dir and the
`DatasetSource`. `fps`, `data_path` and `video_path` stay reachable as properties so the
conversion code reads what it read before.

Shape validation stays in `_derive_numeric_schema()` and still runs only on selected
features. A corpus may declare a float32 feature HFlow never reads, and refusing the import
for it would reject datasets that convert today. `_NumericFeature` gives that function
attributes to read, so no `.get()` on a raw spec survives past the parser.

The downloaded `info.json` is still written to the cache as the original document.

## One behaviour change

The "must define a features object" refusal now fires before episode discovery instead of
after it. A corpus missing `features` used to download every `meta/episodes` shard first.
Same refusal, same message, earlier.

Everything else is the same: same selected episodes, same paths, same schemas, same
provenance, same manifest, and the fps, dtype, shape and required-feature refusals still
fail before any output.

## Tests

The archive fakes in `test_lerobot_converter.py` were dict literals. They now build through
the real parser, so a fake corpus the parser would refuse cannot reach the conversion code
under test. The `_derive_numeric_schema` cases construct `_NumericFeature` directly and keep
their exact expected messages, boolean dimension included.

## Validation

```
uv run ruff check --fix        clean
uv run ruff format             clean
uv run ty check                clean
uv run pytest -q tests/test_lerobot_converter.py    77 passed
uv run pytest -q               1554 passed, 28 failed
```

The 28 are pre-existing on this machine and unrelated: 27 in `test_packaging*` refusing to
build the native overlay off Linux, and one in `test_end_to_end` where the local ffmpeg has
no `filter_script:v`. Both fail the same way on `origin/main`.

### DIFF
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -23,10 +23,10 @@ import struct
 import subprocess
 import tempfile
 import urllib.request
-from collections.abc import Sequence
-from dataclasses import dataclass
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field, replace
 from pathlib import Path, PurePosixPath
-from typing import NotRequired, TypedDict
+from typing import TypedDict
 from urllib.parse import urlsplit
 
 from mcap.reader import make_reader
@@ -71,7 +71,20 @@ EPISODE_START_TIME_NS = 1_755_000_000_000_000_000
 HUGGING_FACE_TOKEN_ENVIRONMENT_VARIABLES = ("HF_TOKEN", "HUGGING_FACE_HUB_TOKEN")
 
 
-class _EpisodeRow(TypedDict):
+@dataclass(frozen=True)
+class _VideoWindow:
+    """One camera's slice of a chunk video for one episode."""
+
+    chunk_index: str
+    file_index: str
+    from_timestamp: float
+    to_timestamp: float
+
+
+@dataclass(frozen=True)
+class _EpisodeRow:
+    """One episode as the ``meta/episodes`` shards describe it."""
+
     episode_index: int
     task: str
     length: int
@@ -79,17 +92,11 @@ class _EpisodeRow(TypedDict):
     data_file: str
     data_from: int
     data_to: int
-    video_windows: NotRequired[dict[str, "_VideoWindow"]]
-    # MAX of the episode's collector-labeled next.success frames, present only
-    # when the source declares that outcome feature at all.
-    success_outcome: NotRequired[bool]
-
-
-class _VideoWindow(TypedDict):
-    chunk_index: str
-    file_index: str
-    from_timestamp: float
-    to_timestamp: float
+    video_windows: Mapping[str, _VideoWindow] = field(default_factory=dict)
+    # MAX of the episode's collector-labeled next.success frames. None when
+    # the source declares no such outcome feature, which carries no label
+    # either way and must not be read as a failure.
+    success_outcome: bool | None = None
 
 
 class _PublishedEpisode(TypedDict):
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
LGTM, merging.

Keeping shape validation in `_derive_numeric_schema` on selected features only is the right call: validating every feature would refuse archives that convert fine.

Issue #436, opened 2026-09-06
The camera_stability tuning knobs are undocumented, so hand-held footage reads as continuously unstable with no documented way to fix it


kstonekuan:
`camera_stability` gained `shake_threshold_dps` and `unstable_min_duration_s` (`src/hflow/checks.py:948-949`). Neither appears anywhere in `docs/`. The only description of them is the function docstring.

## Why it matters

From the commit that added them: hand-held footage clears the instrument's resolution floor almost continuously by fractions of a degree. That is real motion, but it is not what a person means by a shaky camera. Without the knobs, a user pointing `camera_stability` at ordinary hand-held recordings gets `unstable:<camera>` intervals covering nearly the whole episode, concludes the check is broken or useless, and has no way from the documentation to learn that two arguments exist to fix exactly this.

Both default to zero, so this is not a behavior change anyone will notice. It is a tuning affordance nobody can find.

## The pattern to copy

`docs/how-to/enable-built-in-checks.md` already documents parameters exactly where a user has to make a decision, and not otherwise:

- `:165-170` names the three checks that need an argument before they do anything (`topics=`, `nominal_fps=`, `topic=`).
- `:180-186` gives `dimension_scales=` a full paragraph: what it is for, and why there is deliberately no auto-scaling, because the reasoning is the part a user cannot reconstruct.

That second one is the model. The valuable content is not "this parameter exists", it is "here is the situation you are in, and here is why the obvious alternative is wrong".

`camera_stability` already has its own paragraph at `:172-177`, currently about the OpenCV extra. The knobs belong with it or just after it.

## Definition of done

1. `docs/how-to/enable-built-in-checks.md` describes both knobs: what each one does, and the hand-held-footage situation that motivates them.
2. It is clear which one affects the measurements and which does not. `shake_threshold_dps` changes which pairs count, so it moves `unstable_share` and `unstable_s` as well as the intervals; `unstable_min_duration_s` filters intervals only and leaves the measurements alone. The docstring at `:952-964` states this, and it is the kind of asymmetry a user will otherwise discover by being confused.
3. A starting point a reader can actually use. The docstring suggests a few degrees per second and a quarter second of run. Give a concrete pair rather than telling people to experiment.
4. Do not document the parameters of checks that have none of this problem. The file's discipline is that it explains a parameter when a decision hangs on it.

## Not in scope

[...]

PR #437, opened 2026-09-06
docs: document camera_stability tuning knobs

Machumerre:
### What it does
* Adds a new subsection in `docs/how-to/enable-built-in-checks.md` that explains the two optional arguments introduced in `src/hflow/checks.py:948-949`:
  * `shake_threshold_dps` – the angular velocity threshold (degrees per second) that defines a “shake”.
  * `unstable_min_duration_s` – the minimum duration (seconds) a camera must exceed the threshold to be considered unstable.
* Provides a short example of how to use these knobs when enabling the check.
* Updates the table of built-in checks to include the new parameters, mirroring the style used for `topics=`, `nominal_fps=`, and `dimension_scales=`.

### Why it matters
* Users who run `camera_stability` on hand-held footage will no longer see the entire episode marked as unstable.
* The documentation now matches the code, preventing confusion and making the check useful for real-world recordings.

## Checklist

- [ ] I added or updated outcome-focused tests for changed business logic.
- [ ] I ran `uv run ruff check --fix`, `uv run ruff format`, and `uv run ty check`.
- [ ] I ran the relevant pytest suite.
- [ ] I did not add recordings, generated media, credentials, private URLs, or runtime artifacts.

### DIFF
--- a/docs/how-to/enable-built-in-checks.md
+++ b/docs/how-to/enable-built-in-checks.md
@@ -267,3 +267,24 @@ particularly the limitations.
   surface, the measurement-key naming rules, and coverage denominators.
 - [Architecture: quality checks and curation](../ARCHITECTURE.md) -- why checks
   record evidence and gates are optional policy.
+
+
+## `camera_stability`
+
+`camera_stability` monitors the angular velocity of the camera and flags periods of excessive motion.  
+Two optional tuning knobs are available:
+
+| Argument | Default | Description |
+|----------|---------|-------------|
+| `shake_threshold_dps` | `0.0` | Angular velocity threshold (degrees per second). When the camera’s instantaneous angular velocity exceeds this value, the check considers the camera to be “shaking”. |
+| `unstable_min_duration_s` | `0.0` | Minimum duration (seconds) that the camera must stay above `shake_threshold_dps` to be reported as unstable. This prevents brief spikes from being flagged. |
+
+### Example
+
+```python
+from hflow import CameraStability
+
+check = CameraStability(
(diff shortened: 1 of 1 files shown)


kstonekuan (changes requested):
The example imports `CameraStability`, which does not exist. Use `functools.partial(camera_stability, ...)`, the idiom already at `:84-99`.

The code fence is never closed.

Both descriptions are wrong in ways that matter: `shake_threshold_dps` is not total angular velocity (a fast pan is stable), and `unstable_min_duration_s` filters intervals only, so `unstable_share` and `unstable_s` do not move when you raise it.

Missing the reason the knobs exist, which is what #436 asked for.

Also: the section landed after "See also", and the file uses straight quotes.

kstonekuan:
Still want this one? The cap means it is holding your slot.

Shortest path: replace the `CameraStability` snippet with the `functools.partial` form (that class does not exist), close the fence, move the section beside the existing `camera_stability` paragraph at `:172-177`, and say why the knobs exist: hand-held footage clears the resolution floor continuously, so the check reads as broken.

Happy to unassign if you would rather hand it back.

kstonekuan (approved):
LGTM, merging. Taking it over so the docs gap closes and your slot frees.

The section is prose beside the existing `camera_stability` paragraph now, and the example runs: `functools.partial` has no `__name__`, so `app.check` needs `name=`. Added that to the paragraph above.

`shake_threshold_dps` is a shake-versus-deliberate comparison, not angular velocity, and `unstable_min_duration_s` filters intervals only.

PR #438, opened 2026-09-06
fix(storage): refuse bucket URLs naming no object directly in fetch_uri (#435)

### LINKED ISSUE #435: fetch_uri's 'names no object' refusal is the one refusal in storage.py nothing tests, and a bucket URL with no object reports a path the caller never typed
`storage.py` has twelve refusals. I neutered each one in turn and ran `tests/test_storage.py`, `tests/test_ingest_in_process.py`, and `tests/test_cli.py`. Eleven turn a test red. One does not:

```
line  690  NOTHING NOTICED   raise ValueError(f"bucket URI {uri!r} names no object")
```

## Current behavior

`fetch_uri` splits a bucket URL into a parent prefix and an object name (`src/hflow/storage.py:679-691`):

```python
if isinstance(uri, str) and is_bucket_url(uri):
    parent_url, _, name = uri.rpartition("/")
    if not name:
        raise ValueError(f"bucket URI {uri!r} names no object")
    return BucketStorageRoot(parent_url).fetch(name)
```

Measured against the real function:

```
gs://bucket/            ValueError: bucket URI 'gs://bucket/' names no object
s3://bucket/prefix/     ValueError: bucket URI 's3://bucket/prefix/' names no object
gs://bucket//           ValueError: bucket URI 'gs://bucket//' names no object
gs://bucket             ValueError: bucket URL 'gs:/' names no bucket/container -- expected e.g. gs://bucket/prefix
s3://bucket             ValueError: bucket URL 's3:/' names no bucket/container -- expected e.g. gs://bucket/prefix
```

Two things there.

**The trailing-slash refusal is correct and untested.** `hflow doctor gs://bucket/prefix/` is a plausible typo, and the message it produces is the right one. Nothing pins it, so it can be deleted without the suite noticing.

**A bucket URL naming no object reports a string the caller never supplied.** `gs://bucket` has no trailing slash, so `rpartition("/")` takes `bucket` as the object name and leaves `gs:/` as the parent. `BucketStorageRoot("gs:/")` then refuses, and the error quotes `'gs:/'`. The user typed `gs://bucket` and is told something about `gs:/`, which is neither their input nor a thing they can fix. It should be refused by `fetch_uri` under the URI they actually passed, in the same shape as the trailing-slash case: a bucket URL that names no object.

## The pattern to copy

`TestParseStorageRoot.test_bucketless_url_is_refused` (`tests/test_storage.py:63-69`) is the shape: several inputs that must all be refused, each with `pytest.raises(ValueError, match=...)` on the distinguishing phrase. `TestFetchUri` at `:386` is the class these belong in.

## Definition of done

1. A test pins the trailing-slash refusal. Removing the `raise` at `:690` must turn it red.
2. `gs://bucket` and `s3://bucket` are refused by `fetch_uri` with the URI the caller passed in the message, not a truncated parent.
3. `match=` patterns are anchored enough to fail if the message drifts. Note that `.` in a pattern matches any character; escape it, or the test passes on messages it should not.
4. Valid bucket URIs still fetch. `TestFetchUri` has the existing cases.

## Not in scope

The message wording for `parse_storage_root` at `:439`, and anything about how `BucketStorageRoot` resolves prefixes. This is `fetch_uri` deciding what it accepts before it delegates.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

The bucket tests that need credentials skip themselves; the ones here run against local paths and the fake bucket.

Cid-oe:
## Summary
Resolves #435.

Ensures `fetch_uri` explicitly validates and refuses bucket URLs that name no object before delegating to `BucketStorageRoot`, returning an error citing the exact URI supplied by the caller instead of a truncated prefix.

## Changes
- In `src/hflow/storage.py`: checked that the bucket URL contains an object segment after the bucket/container before delegating to `BucketStorageRoot`.
- In `tests/test_storage.py`: added `test_bucket_uri_naming_no_object_is_refused` in `TestFetchUri` testing trailing slashes (`gs://bucket/`, `s3://bucket/prefix/`, `gs://bucket//`) and bare bucket names (`gs://bucket`, `s3://bucket`).
- Formatted and linted with `ruff`.

## Validation
- `uv run pytest -q tests/test_storage.py` (44 passed, 1 skipped)
- `uv run ruff check src/hflow/storage.py tests/test_storage.py`
- `uv run ruff format --check src/hflow/storage.py tests/test_storage.py`

### DIFF
--- a/src/hflow/storage.py
+++ b/src/hflow/storage.py
@@ -685,6 +685,9 @@ def fetch_uri(uri: "str | Path") -> Path:
     one-off ``app.process("gs://...")`` sources.
     """
     if isinstance(uri, str) and is_bucket_url(uri):
+        _scheme, _sep, remainder = uri.partition("://")
+        if "/" not in remainder or not remainder.partition("/")[2].strip("/"):
+            raise ValueError(f"bucket URI {uri!r} names no object")
         parent_url, _, name = uri.rpartition("/")
         if not name:
             raise ValueError(f"bucket URI {uri!r} names no object")
--- a/tests/test_storage.py
+++ b/tests/test_storage.py
@@ -12,6 +12,7 @@ A live object-store integration test runs only when
 
 import errno
 import os
+import re
 import sys
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

`re.escape(repr(uri))` is the careful version: `.` unescaped matches anything, so the pattern would have passed on messages it should not.

PR #443, opened 2026-09-06
refactor(storage): unify bucket object URI validation

### LINKED ISSUE #442: fetch_uri raises the same message from two places, which reads as a dead branch and is not
#435 landed in #438 and left `fetch_uri` refusing with the same sentence from two different lines (`src/hflow/storage.py:687-694`):

```python
if isinstance(uri, str) and is_bucket_url(uri):
    _scheme, _sep, remainder = uri.partition("://")
    if "/" not in remainder or not remainder.partition("/")[2].strip("/"):
        raise ValueError(f"bucket URI {uri!r} names no object")
    parent_url, _, name = uri.rpartition("/")
    if not name:
        raise ValueError(f"bucket URI {uri!r} names no object")
    return BucketStorageRoot(parent_url).fetch(name)
```

Both branches are live. I checked, because that is the thing you cannot assume here:

```
new pre-check removed              FAILED test_bucket_uri_naming_no_object_is_refused
original trailing-slash refusal removed   FAILED test_bucket_uri_naming_no_object_is_refused
```

The first is the only thing catching `gs://bucket`. The second is the only thing catching `s3://bucket/prefix/`. Neither is redundant.

## Why this is worth changing anyway

Two raises with a byte-identical message read as a copy-paste artifact. We deleted three genuinely unreachable guards yesterday (#423, #425) using a sweep that flags exactly this shape, and the next person running it will stop on this pair and have to re-derive that both are needed. The code is correct and its appearance is misleading, which is a cost paid by every future reader.

## What to do

One predicate covers every case. This is verified, not proposed from reading:

```python
bucket, _, object_key = uri.partition("://")[2].partition("/")
if not bucket or not object_key or object_key.endswith("/"):
    raise ValueError(f"bucket URI {uri!r} names no object")
```

Against the shipped pair over every input either branch is meant to catch, plus the ones that must pass:

```
gs://bucket                      refuse   shipped=True  single=True
s3://bucket                      refuse   shipped=True  single=True
gs://bucket/                     refuse   shipped=True  single=True
gs://bucket//                    refuse   shipped=True  single=True
s3://bucket/prefix/              refuse   shipped=True  single=True
gs://bucket///                   refuse   shipped=True  single=True
gs://bucket/object.mcap          accept   shipped=False single=False
s3://bucket/prefix/object.mcap   accept   shipped=False single=False
gs://bucket/a/b/c/object.mcap    accept   shipped=False single=False

0 disagreement(s)
```

You still need `parent_url` and `name` for the `BucketStorageRoot(parent_url).fetch(name)` call below, so keep the `rpartition`; it just stops being a guard.

Do not take the snippet on faith. Confirm the agreement yourself before you rely on it, and if you find an input where the two disagree, say so: that is a more interesting result than the refactor.

## Definition of done

1. One refusal, same message, same behavior for every input above.
2. `tests/test_storage.py::TestFetchUri::test_bucket_uri_naming_no_object_is_refused` passes unchanged. If you find yourself editing it, the behavior moved and something is wrong.
3. Removing your single guard turns that test red. Say that you checked.
4. Add `gs://bucket///` to the refused list if it is not there. It is in the table above and worth pinning.

## Not in scope

`parse_storage_root`, `BucketStorageRoot`, and the local-path branch below. This is one `if` inside one function.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

be-student:
## Summary
- replace the two live, identical bucket-object refusals with one predicate
- preserve valid nested object keys while rejecting bucket-only and trailing-slash URIs
- add the requested `gs://bucket///` regression case to the existing behavior test

Fixes #442.

## Validation
- `uv run pytest tests/test_storage.py::TestFetchUri::test_bucket_uri_naming_no_object_is_refused -q` passed
- mutation check passed: removing the unified guard makes that unchanged regression test fail
- Ruff check and format check passed for both changed files
- `uv run ty check src/hflow/storage.py` passed
- full suite: 1,565 passed, 11 skipped; 27 existing Linux-only native-overlay tests fail on macOS with the repository's explicit `native overlay builds currently require CPython on Linux` guard
- `git diff --check` passed

## AI assistance
I used an AI coding assistant to verify the branch equivalence and draft the focused refactor. I reviewed the final diff and ran the validation above locally.

### DIFF
--- a/src/hflow/storage.py
+++ b/src/hflow/storage.py
@@ -685,12 +685,10 @@ def fetch_uri(uri: "str | Path") -> Path:
     one-off ``app.process("gs://...")`` sources.
     """
     if isinstance(uri, str) and is_bucket_url(uri):
-        _scheme, _sep, remainder = uri.partition("://")
-        if "/" not in remainder or not remainder.partition("/")[2].strip("/"):
+        bucket, _, object_key = uri.partition("://")[2].partition("/")
+        if not bucket or not object_key or object_key.endswith("/"):
             raise ValueError(f"bucket URI {uri!r} names no object")
         parent_url, _, name = uri.rpartition("/")
-        if not name:
-            raise ValueError(f"bucket URI {uri!r} names no object")
         return BucketStorageRoot(parent_url).fetch(name)
     local_file = Path(uri)
     if local_file.is_dir():
--- a/src/hflow/storage.py
+++ b/src/hflow/storage.py
@@ -686,7 +686,12 @@ def fetch_uri(uri: "str | Path") -> Path:
     """
     if isinstance(uri, str) and is_bucket_url(uri):
         bucket, _, object_key = uri.partition("://")[2].partition("/")
-        if not bucket or not object_key or object_key.endswith("/"):
+        # Two different mistakes, so two different sentences. ``gs:///x`` has a
+        # key and no bucket; calling that "names no object" sends the reader
+        # looking at the part they got right.
+        if not bucket:
+            raise ValueError(f"bucket URI {uri!r} names no bucket")
+        if not object_key or object_key.endswith("/"):
             raise ValueError(f"bucket URI {uri!r} names no object")
         parent_url, _, name = uri.rpartition("/")
         return BucketStorageRoot(parent_url).fetch(name)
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Pushed a fixup: my own predicate in #442 was wrong. `gs:///x` has a key and no bucket, and the collapsed guard called that "names no object". Split back into two sentences.

PR #446, opened 2026-09-07
fix: validate horizontal field of view before camera selection

### LINKED ISSUE #445: camera_stability accepts an invalid field of view on an episode with no cameras, because that guard lives inside the per-camera loop
#441 moved two of `camera_stability`'s three numeric parameters to the top of the function, where they are refused before anything is decoded. The third was left where it was, and it is validated inside the per-camera loop. When that loop body does not run, it is not validated at all.

## Current behavior

`camera_stability` takes three numeric arguments (`src/hflow/checks.py:947-949`):

```python
horizontal_field_of_view_degrees: float = DEFAULT_HORIZONTAL_FIELD_OF_VIEW_DEGREES,
shake_threshold_dps: float = 0.0,
unstable_min_duration_s: float = 0.0,
```

The last two are checked at `:990-1000`, before `selected_cameras` is even resolved. The first is checked only when `CameraMotionSettings` is constructed, which happens per camera inside the loop.

On an episode with cameras that is fine, and the message is good:

```
fov=-5.0   ValueError: horizontal_field_of_view_degrees must be finite and in (0, 360], got -5.0
```

On an episode with no cameras the loop body never runs:

```
cameras on this episode: []
fov=-5.0   ACCEPTED, measurements={}
fov=nan    ACCEPTED, measurements={}
fov=0.0    ACCEPTED, measurements={}
shake=-5.0   refused: ValueError
shake=nan    refused: ValueError
```

Same call, same nonsense argument, and whether it is refused depends on the episode you happened to pass. State-only episodes are ordinary here: `tests/test_gates.py:20-24` builds one with `cameras=()`, and the how-to tells users to skip the joint checks on human egocentric corpora, which is the mirror of the same situation.

The practical shape of this: someone registers `functools.partial(camera_stability, horizontal_field_of_view_degrees=...)` with a typo, runs it over a corpus, and gets a clean pass on every state-only episode and a crash on the first episode with a camera. The argument was wrong the whole time.

## What to do

Validate it at the top with the other two. The guards #441 landed at `:990-1000` are the pattern, in this same function:

```python
if (
    isinstance(shake_threshold_dps, bool)
    or not np.isfinite(shake_threshold_dps)
    or shake_threshold_dps < 0
):
    raise ValueError("shake_threshold_dps must be finite and non-negative")
```

The bound is different: field of view is `(0, 360]`, so zero is invalid where zero is the *default* for the other two.

## Definition of done

1. `horizontal_field_of_view_degrees` is refused before `selected_cameras` is resolved, so a camera-less episode refuses it exactly like any other.
2. The four cases from the probe above are covered: negative, zero, non-finite, and `bool`. `bool` matters because it subclasses `int`, and `True` would otherwise pass as one degree.
3. `test_camera_stability_refuses_a_bad_tuning_value` in `tests/test_motion.py` is already parametrized over parameter and value. Extend it rather than writing a third block, and note that its current values (`-1.0`, nan, inf, `True`) do not include zero, which only this parameter rejects.
4. At least one case runs against a camera-less episode. That is the regression, and a test that only uses the video fixture would pass today.
5. Leave the `CameraMotionSettings` guard in `_video_measurements/_camera_motion.py:70-80` alone. It protects that class's other callers and is not this function's to remove.
6. Removing your guard must turn a test red. Say that you checked.

## Not in scope

The other checks in `checks.py` do not validate their numeric parameters either. That is a larger conversation. This one is worth doing now because two of three arguments in a single signature are already validated at the top, and the odd one out is silently skipped on real episodes.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

`tests/test_motion.py` needs the `motion` extra, which `--all-extras` installs.

k0shir0:
Resolves #445

## Summary
Validates `horizontal_field_of_view_degrees` at the top of `camera_stability` before `selected_cameras` is resolved, matching the existing guards for `shake_threshold_dps` and `unstable_min_duration_s`. Rejects `bool`, non-finite, `<=0` and `>360` with `ValueError: horizontal_field_of_view_degrees must be finite and in (0, 360], got <value>`. Leaves `CameraMotionSettings` guard in `_video_measurements/_camera_motion.py:73-80` intact for its other callers.

## Why
The field-of-view check lived inside the per-camera loop via `CameraMotionSettings`. On a state-only episode (`cameras=()`) the loop never runs, so the same bad value was accepted (`measurements={}`) on state-only episodes and refused on video episodes. A `functools.partial(camera_stability, horizontal_field_of_view_degrees=…)` with a typo would pass every state-only episode and crash on the first camera episode.

## Changes
- `src/hflow/checks.py`: add top-level guard `(0, 360]` before `selected_cameras`
- `tests/test_motion.py`: extend `test_camera_stability_refuses_a_bad_tuning_value` to include `horizontal_field_of_view_degrees` (12 cases) and add `test_camera_stability_refuses_a_bad_fov_on_a_camera_less_episode` covering negative, zero, non-finite, bool on a `cameras=()` episode

## Validation
```
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest tests/test_motion.py -q
```
- `ruff check` All checks passed
- `ruff format --check` 223 files already formatted
- `ty check` All checks passed
- `pytest tests/test_motion.py -q` 22 passed
- Removing the guard makes `test_camera_stability_refuses_a_bad_fov_on_a_camera_less_episode` fail with `DID NOT RAISE` (verified)

### DIFF
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -1001,6 +1001,15 @@ def camera_stability(
     ):
         raise ValueError("unstable_min_duration_s must be finite and non-negative")
 
+    if (
+        isinstance(horizontal_field_of_view_degrees, bool)
+        or not np.isfinite(horizontal_field_of_view_degrees)
+        or not 0 < horizontal_field_of_view_degrees <= 360
+    ):
+        raise ValueError(
+            f"horizontal_field_of_view_degrees must be finite and in (0, 360], got {horizontal_field_of_view_degrees}"
+        )
+
     selected_cameras = list(cameras) if cameras is not None else episode.cameras
     measurements: dict[str, MeasurementValue] = {}
     intervals: list[Interval] = []
--- a/tests/test_motion.py
+++ b/tests/test_motion.py
@@ -395,7 +395,7 @@ def test_camera_stability_refuses_a_bad_fov_on_a_camera_less_episode(
         tmp_path / "episode.mcap",
         SyntheticEpisodeSpec(duration_s=2.0, cameras=(), joint_jump_at_s=1.0),
     )
-    bad_values: list[float] = [-5.0, 0.0, float("nan"), float("inf"), True]  # type: ignore[arg-type]
+    bad_values: list[float] = [-5.0, 0.0, 361.0, float("nan"), float("inf"), True]
     for bad_value in bad_values:
         with (
             hflow.Episode(episode_path) as episode,
@@ -404,7 +404,13 @@ def test_camera_stability_refuses_a_bad_fov_on_a_camera_less_episode(
                 match=r"^horizontal_field_of_view_degrees must be finite and in \(0, 360\], got .+$",
             ),
         ):
-            camera_stability(episode, horizontal_field_of_view_degrees=bad_value)  # type: ignore[arg-type]
+            camera_stability(episode, horizontal_field_of_view_degrees=bad_value)
+
+    # The bound is inclusive at 360, and the guard must not refuse the widest
+    # legal lens. Without this the whole range check could be `< 360` and every
+    # case above would still pass.
+    with hflow.Episode(episode_path) as episode:
+        assert camera_stability(episode, horizontal_field_of_view_degrees=360.0).measurements == {}
 
 
 def test_the_check_knobs_raise_the_bar_without_changing_the_rate_measurements(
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

The camera-less test is the one that covers the bug rather than the parameter.

PR #451, opened 2026-09-07
test(server): pin the non-single-SELECT refusal on pin and preview

### LINKED ISSUE #448: The HTTP 400 for SQL that is not a single SELECT is the one curation refusal nothing tests
`_curation.py` has 13 refusals. I neutered each one in turn and ran `packages/hflow-server/tests`. Ten turn a test red. Three do not, and this is the one that matters most:

```
line  176  NOTHING NOTICED   raise HTTPException(
```

That is `_reject_non_single_select` (`packages/hflow-server/src/hflow_server/_curation.py:161-177`) mapping `NonSingleSelectQueryError` onto `400 sql must be exactly one read-only SELECT statement`. It is the boundary that decides what tenant-supplied SQL the HTTP surface will run, and deleting it breaks nothing in the suite.

## It is reachable, and it works

Measured against the real endpoints:

```
POST /api/v1/curation/preview  SELECT 1; DROP TABLE episodes      -> 400 sql must be exactly one read-only SELECT statement
POST /api/v1/curation/preview  COPY (SELECT 1) TO '/tmp/x.csv'    -> 400 sql must be exactly one read-only SELECT statement
POST /api/v1/curation/pin      DELETE FROM episodes               -> 400 sql must be exactly one read-only SELECT statement
POST /api/v1/curation/pin      CREATE TABLE mine AS SELECT 1      -> 400 sql must be exactly one read-only SELECT statement
```

So the behavior is correct today. The gap is that nothing holds it there.

The sibling 400 in the same function, the DuckDB parse-error path at `:174`, and the empty-SQL 400 at `:151`, both have tests. This one is the odd one out, and it is the half that refuses a *well-formed* statement that simply is not a read-only SELECT. That is the interesting half: SQL the parser is perfectly happy with.

## Why it is worth more than a line of coverage

`connection.execute()` runs every semicolon-separated statement and returns only the last result, which is why `hflow.reject_non_single_select` parses without executing. `src/hflow/curation.py:557-576` carries that reasoning. The SDK-level rule has tests; the HTTP mapping onto a 400 does not, so a refactor of this route could turn a refusal into a 500, or into nothing, and the suite would stay green.

Defence in depth is intact either way, and it is worth knowing that before you write anything: the server opens tenant connections with `constrained=True`, so `enable_external_access` is false and file access is refused independently:

```
POST /api/v1/curation/preview  SELECT * FROM read_csv_auto('/etc/hostname')
-> 400 Permission Error: Cannot access file "/etc/hostname" - file system operations are disabled by configuration
```

This issue is not about a hole. It is about pinning a guard that currently rests on nobody's test.

## Definition of done

1. A test asserts `400` and the exact detail string for SQL that is well-formed and not a single SELECT. Cover more than one shape: a multi-statement string, and a single statement that is not a SELECT.
2. Removing the `raise` at `:176` must turn your test red, and only yours. Say that you checked.
3. Keep it distinct from the parse-error path. A string DuckDB cannot parse produces a different 400 by design, and a test that cannot tell them apart is not testing this.
4. Anchor the assertion on the full detail string. If it drifts, the test should fail.

## The pattern to copy

`packages/hflow-server/tests/test_server_curation_pin.py` has the `writable_api` client fixture and the request shapes. `test_create_requires_nonempty_name_and_sql` is the nearest existing refusal test.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

ayam04:
Fixes #448

**The gap:** `_reject_non_single_select` (`packages/hflow-server/src/hflow_server/_curation.py:161-177`) is the boundary that decides what tenant-supplied SQL the HTTP surface runs, and the `NonSingleSelectQueryError` mapping onto the fixed 400 sentence was the one curation refusal nothing tested. The issue neutered each refusal in turn; this is the one that matters most.

**What changed:** two regression tests.

- `test_server_curation_pin.py` — four hostile-but-well-formed shapes on `POST /api/v1/curation/pin` (`SELECT 1; DROP TABLE episodes`, `COPY (SELECT 1) TO ...`, `CREATE TABLE ... AS SELECT 1`, `DELETE FROM episodes`), each asserting the fixed sentence, plus nothing registered and no parquet stranded.
- `test_server_curation_preview.py` — the multi-statement and DELETE shapes on `POST /api/v1/curation/preview` with the same sentence.

**Red/green:** neutering the `NonSingleSelectQueryError` branch turns these two tests red and only these two (the SQL then escapes into the preview wrapper / execution path and fails with a different message). Restored, both files pass: 31/31.

**Gates run:** `ruff check`, `ruff format --check`, `ty check` all clean on both files; `pytest packages/hflow-server/tests/test_server_curation_pin.py packages/hflow-server/tests/test_server_curation_preview.py` → 31 passed. No server code touched.

### DIFF
--- a/packages/hflow-server/tests/test_server_curation_pin.py
+++ b/packages/hflow-server/tests/test_server_curation_pin.py
@@ -137,6 +137,27 @@ def test_concurrent_pins_all_land_in_the_registry(
     assert len(manifest_files) == len(names)
 
 
+def test_pin_refuses_sql_that_is_not_a_single_select(
+    writable_api: TestClient, writable_workspace: PopulatedWorkspace
+) -> None:
+    # Well-formed SQL that is not exactly one read-only SELECT is refused
+    # with the fixed sentence, and nothing is registered or written.
+    hostile_sql = [
+        "SELECT 1; DROP TABLE episodes",
+        "COPY (SELECT 1) TO '/tmp/hflow-nope.csv'",
+        "CREATE TABLE should_not_exist AS SELECT 1",
+        "DELETE FROM episodes",
+    ]
+    for sql in hostile_sql:
+        response = writable_api.post("/api/v1/curation/pin", json={"sql": sql, "name": "hostile"})
+        assert response.status_code == 400
+        assert response.json()["detail"] == "sql must be exactly one read-only SELECT statement"
+    manifests_directory = writable_workspace.data_root / "manifests"
+    if manifests_directory.exists():
+        assert list(manifests_directory.glob("*.parquet")) == []
+    assert writable_api.get("/api/v1/manifests").json()["manifests"] == []
+
+
 def test_pin_with_bad_sql_is_400_and_registers_nothing(
     writable_api: TestClient, writable_workspace: PopulatedWorkspace
 ) -> None:
--- a/packages/hflow-server/tests/test_server_curation_preview.py
+++ b/packages/hflow-server/tests/test_server_curation_preview.py
@@ -111,6 +111,16 @@ def test_preview_handles_json_hostile_result_types(api: TestClient) -> None:
     assert payload["rows"] == [{"a_decimal": 1.5, "a_list": [1, 2]}]
 
 
+def test_preview_refuses_sql_that_is_not_a_single_select(api: TestClient) -> None:
+    # The parser accepts these; the gate refuses them because they are not
+    # exactly one read-only SELECT, with the same sentence pin uses.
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Covering both endpoints was right: they share the gate and did not share a test. Asserting nothing was registered or written is what makes it a refusal test rather than a status-code test.

PR #452, opened 2026-09-07
test(server): cover curation registry caps

### LINKED ISSUE #449: Neither registry cap refusal is tested, and both caps are 1000 so the obvious test would be unusably slow
Two of the three untested refusals in `_curation.py` are the registry caps:

```
line  359  NOTHING NOTICED   raise HTTPException(  # pinned manifests
line  477  NOTHING NOTICED   raise HTTPException(  # saved queries
```

Both are `409` refusals that stop a workspace's sidecar growing without bound (`packages/hflow-server/src/hflow_server/_curation.py:358-364` and `:476-482`). Neither is exercised, so either could be deleted and the suite would stay green.

## The wrinkle that makes this worth writing down

```python
_MAX_SAVED_QUERIES = 1000
_MAX_PINNED_MANIFESTS = 1000
```

`packages/hflow-server/src/hflow_server/_curation.py:92-93`. Creating 1000 pinned manifests writes 1000 Parquet files, so the obvious test is slow enough that nobody should merge it. That is very likely why these two are the untested ones.

Patch the constant down instead. Verified working:

```python
monkeypatch.setattr(_curation, "_MAX_PINNED_MANIFESTS", 1)
# first pin  -> 200
# second pin -> 409
#   "this workspace already has 1 pinned manifests (the registry cap); remove some before pinning more"
```

`test_server_curation_pin.py` already imports `_curation` directly, so the handle is there.

Note the detail string interpolates the cap, so a test that patches it to 1 must expect "1 pinned manifests" and not hardcode 1000. That is the sort of thing worth noticing before you write the assertion rather than after.

## Definition of done

1. A test for each cap: the request that fits succeeds, the one past the cap returns `409`.
2. Both assert the detail string, built from the patched cap rather than hardcoded.
3. Removing either `raise` turns the matching test red and only that one. Say that you checked both.
4. The default caps stay 1000. This is about testing them, not retuning them.
5. No test creates anything like a thousand of either thing. If your test takes longer than the rest of the file put together, patch the constant instead.

## Worth checking, not assuming

The saved-queries routes are `POST /api/v1/queries` and friends (`:457-521`), not `/saved-queries`. I guessed wrong first time and got a `405`, which is a reminder to read the router rather than the field names. Confirm the paths before writing the requests.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

harshitethic:
Fixes #449

## Summary

Add focused regression coverage for both curation sidecar registry caps without creating anything close to the production limit of 1000 entries.

- patch `_MAX_PINNED_MANIFESTS` to `1`, assert the first pin succeeds and the next returns the exact `409` detail, and verify the refused request does not strand an extra Parquet file;
- patch `_MAX_SAVED_QUERIES` to `1`, assert the first saved query succeeds and the next returns the exact `409` detail, and verify the registry remains unchanged.

The default production caps remain untouched.

## Why this shape

The refusal branches are important bounded-state protections, but testing them at the real cap would make the suite needlessly expensive, especially for pinned manifests. Monkeypatching the constants keeps the tests fast while exercising the real request and persistence paths.

## Validation

I reviewed the diff against current fork `main`; it only changes the two focused server test modules. The fork's GitHub Actions workflow may require upstream approval before jobs execute, so CI is the authoritative repository-wide validation once approved.

AI assistance disclosure: AI assistance was used to inspect the issue and repository tests, prepare the focused regression coverage, collision-check for existing work, and review the resulting diff.

### DIFF
--- a/packages/hflow-server/tests/test_server_curation_pin.py
+++ b/packages/hflow-server/tests/test_server_curation_pin.py
@@ -76,9 +76,7 @@ def test_pinned_manifest_registry_cap_refuses_the_first_entry_past_the_limit(
     monkeypatch.setattr(_curation, "_MAX_PINNED_MANIFESTS", cap)
 
     first = _pin(writable_api, "first")
-    over_cap = writable_api.post(
-        "/api/v1/curation/pin", json={"sql": OK_CUT_SQL, "name": "second"}
-    )
+    over_cap = writable_api.post("/api/v1/curation/pin", json={"sql": OK_CUT_SQL, "name": "second"})
     assert over_cap.status_code == 409
     assert over_cap.json()["detail"] == (
         f"this workspace already has {cap} pinned manifests "
--- a/packages/hflow-server/tests/test_server_curation_pin.py
+++ b/packages/hflow-server/tests/test_server_curation_pin.py
@@ -67,6 +67,29 @@ def test_second_pin_with_the_same_name_gets_a_distinct_file(
     assert listed_ids == [second_entry["id"], first_entry["id"]]
 
 
+def test_pinned_manifest_registry_cap_refuses_the_first_entry_past_the_limit(
+    writable_api: TestClient,
+    writable_workspace: PopulatedWorkspace,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    cap = 1
+    monkeypatch.setattr(_curation, "_MAX_PINNED_MANIFESTS", cap)
+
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Building the detail string from the patched cap is what stops the assertion rotting when the real cap changes.

PR #453, opened 2026-09-07
fix(curation): narrow SELECT gate to actual SELECT/WITH text (#450)

### LINKED ISSUE #450: You can pin a cut you cannot preview: preview wraps the SQL in DESCRIBE and then blames your query for the wrapper's syntax error
`POST /curation/preview` and `POST /curation/pin` both advertise the same rule, exactly one read-only SELECT, and they do not accept the same SQL.

## Measured against both endpoints

```
                       preview                                                        pin
plain select           200                                                            200
pragma database_list   400 Parser Error: syntax error at or near ")"                   200   <-- disagree
pragma show_tables     400 Parser Error: syntax error at or near ")"                   200   <-- disagree
pragma version         400 Parser Error: syntax error at or near ")"                   200   <-- disagree
describe               200                                                            200
multi statement        400 sql must be exactly one read-only SELECT statement          400
```

## Why they diverge

The shared gate accepts anything DuckDB labels `StatementType.SELECT`, and DuckDB labels the introspection pragmas that way, because it implements them as table functions:

```
PRAGMA database_list   StatementType.SELECT
PRAGMA show_tables     StatementType.SELECT
PRAGMA version         StatementType.SELECT
DESCRIBE SELECT 1      StatementType.SELECT
```

Pin runs the statement as-is, so it succeeds. Preview does not run it as-is: `_described_columns` at `packages/hflow-server/src/hflow_server/_curation.py:216` interpolates it into a wrapper.

```python
```

`PRAGMA database_list` is a legal statement and an illegal subquery, so the wrapper fails to parse.

## Two defects, one cause

**The error blames the caller for the wrapper.** The 400 body is DuckDB's parse diagnostic for the *rewritten* query, and it names a `)` the user never typed:

```
Parser Error: syntax error at or near ")"
LINE 1: DESCRIBE SELECT * FROM (PRAGMA database_list)
```

That is a message about our SQL, presented as a message about theirs, and it hands the client our internal rewrite. Someone pasting valid SQL is told it has a syntax error, with a caret pointing into text they did not write.

**The two endpoints disagree.** The UI flow is preview then pin, so a cut can be pinned that the preview pane refuses. Whatever we decide the rule is, both routes should apply the same one.

## Not a security hole, and worth knowing why before you start

The catalog is materialized in memory and tenant connections open with `constrained=True`, so `enable_external_access` is false and file access is refused regardless of statement type:

```
SELECT * FROM read_csv_auto('/etc/hostname')
-> 400 Permission Error: Cannot access file "/etc/hostname" - file system operations are disabled by configuration
```

`COPY`, `ATTACH`, `INSTALL`, `LOAD`, `SET`, `CALL`, `CREATE` and `DELETE` all carry non-SELECT statement types and are refused by the gate. What leaks through is read-only introspection of an in-memory database. This is a correctness and message-quality issue, not a containment one, and a PR that describes it as a vulnerability is overclaiming.

## What to decide

Argue for one of these in the PR rather than picking silently.

**Narrow the shared gate to statements that are actually `SELECT` text**, so `PRAGMA` and `DESCRIBE` are refused by both routes with the existing single-SELECT message. Simple and consistent, and it removes a capability that currently works on pin. Someone may be pinning `PRAGMA`-based cuts, though it is hard to imagine why.

**Stop wrapping in preview**, and get column types from the result's own description instead. Then both routes accept the same thing and the misleading parse error disappears. More work, and it needs care to keep preview from materializing a large result just to name its columns.

Either way, a parse failure of a wrapper we constructed must not be reported as the caller's syntax error.

## Definition of done

1. Preview and pin accept and refuse the same SQL. A test asserts agreement over the table above, not just one case.
2. No 400 body contains SQL the caller did not send.
3. The existing single-SELECT and parse-error refusals keep their current messages and tests. See [#448](https://github.com/Hebbian-Robotics/hflow/issues/448), which is pinning one of them.
4. Whichever direction you pick, say in the PR what a user loses.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

k0shir0:
Fixes #450

**What was broken**
`POST /curation/preview` and `POST /curation/pin` both advertise "exactly one read-only SELECT" but disagreed on PRAGMA/DESCRIBE/SHOW/SUMMARIZE. DuckDB labels those as `StatementType.SELECT` (table functions), so the shared gate accepted them. Pin runs SQL as-is → 200. Preview wraps as `DESCRIBE SELECT * FROM (<sql>)` → parser error `syntax error at or near ")"` with `LINE 1: DESCRIBE SELECT * FROM (PRAGMA database_list)` – our wrapper blamed as caller's syntax error.

**Why**
Gate checked only `StatementType.SELECT`. `PRAGMA database_list` is a legal statement but an illegal subquery, so the wrapper fails to parse. The 400 detail leaked internal rewrite. Preview→pin flow allowed pinning a cut that preview refused.

**What changed**
Narrow the shared gate in `src/hflow/curation.py:reject_non_single_select` – after type check, require stripped SQL to start with `SELECT` or `WITH` (CTE) after leading `--`/`/* */` comments (`(?i)^(SELECT|WITH)\b`). Both endpoints now reject PRAGMA/DESCRIBE/etc with the existing `"sql must be exactly one read-only SELECT statement"` and no wrapper leak. `SELECT * FROM pragma_version()` remains allowed (SELECT text).

**Why this choice**
Issue offered: narrow gate vs stop wrapping. Chose narrow gate – simple, consistent with advertised rule, minimal diff, no risk of materializing large results just to name columns. Pinning `PRAGMA`-based cuts is not a real use case; a SELECT that needs pragma data can still `SELECT * FROM pragma_version()` as SELECT text. Keeps preview's LIMIT/count/SUMMARIZE wrapping unchanged; after gate, wrapper only sees SELECT/WITH which is always wrappable, so misleading parse error disappears.

**How tested**
- Before: `PRAGMA database_list` → preview 400 (wrapper leak) vs pin 200 (divergence). After: both 400 with fixed message.
- New tests: `tests/test_catalog_curation.py` – gate rejects PRAGMA/DESCRIBE/SHOW/SUMMARIZE (incl. comment-prefixed) and accepts `--`/`/* */`+SELECT and `WITH`; `packages/hflow-server/tests/test_server_curation_preview.py` and `test_server_curation_pin.py` – preview/pin agree and no `DESCRIBE SELECT * FROM` leak.
- Suites: `pytest tests/test_catalog_curation.py` 97 passed, `pytest packages/hflow-server/tests` 234 passed, `ruff check`/`format`/`ty check` clean.

Not a security fix – `constrained=True` already blocks file access regardless of statement type.

### DIFF
--- a/packages/hflow-server/tests/test_server_curation_pin.py
+++ b/packages/hflow-server/tests/test_server_curation_pin.py
@@ -216,3 +216,26 @@ def test_manifest_download_whose_file_vanished_is_404(
     entry = _pin(writable_api, "soon gone")
     (writable_workspace.data_root / entry["manifest_path"]).unlink()
     assert writable_api.get(f"/api/v1/manifests/{entry['id']}/download").status_code == 404
+
+
+def test_pin_rejects_pragma_and_describe_consistently(
+    writable_api: TestClient, api: TestClient
+) -> None:
+    # Mirrors preview's gate: PRAGMA/DESCRIBE/SHOW are StatementType.SELECT
+    # but not SELECT text, so both endpoints must agree (issue #450).
+    for sql in (
+        "PRAGMA database_list",
+        "PRAGMA show_tables",
+        "PRAGMA version",
+        "DESCRIBE SELECT 1",
+        "SHOW TABLES",
+        "SUMMARIZE SELECT 1",
+    ):
+        pin_response = writable_api.post("/api/v1/curation/pin", json={"sql": sql, "name": "test"})
+        preview_response = api.post("/api/v1/curation/preview", json={"sql": sql})
+        assert pin_response.status_code == 400, pin_response.text
+        assert preview_response.status_code == 400, preview_response.text
+        assert pin_response.json()["detail"] == "sql must be exactly one read-only SELECT statement"
+        assert preview_response.json()["detail"] == pin_response.json()["detail"]
+        assert "DESCRIBE SELECT * FROM" not in pin_response.json()["detail"]
+        assert "syntax error at or near" not in pin_response.json()["detail"]
--- a/packages/hflow-server/tests/test_server_curation_pin.py
+++ b/packages/hflow-server/tests/test_server_curation_pin.py
@@ -239,3 +239,21 @@ def test_pin_rejects_pragma_and_describe_consistently(
         assert preview_response.json()["detail"] == pin_response.json()["detail"]
         assert "DESCRIBE SELECT * FROM" not in pin_response.json()["detail"]
         assert "syntax error at or near" not in pin_response.json()["detail"]
+
+
+def test_pin_accepts_from_first_and_parenthesized_selects(
+    writable_api: TestClient,
+) -> None:
+    # Parallel coverage with preview: the gate used to refuse every query
+    # that didn't start with SELECT or WITH, so FROM-first queries that
+    # preview accepts silently failed pin. They are legal DuckDB SELECTs
+    # that must reach both endpoints consistently.
+    for sql in (
+        "FROM episodes WHERE status = 'ok'",
+        "(SELECT episode_id FROM episodes)",
+        "VALUES (1, 'a'), (2, 'b')",
+    ):
+        response = writable_api.post(
+            "/api/v1/curation/pin", json={"sql": sql, "name": f"accepts {sql[:10]}"}
+        )
+        assert response.status_code == 200, response.text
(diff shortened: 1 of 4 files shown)


kstonekuan (changes requested):
The regex gate refuses legal read-only SELECTs. `FROM episodes WHERE status = 'ok'` is one: DuckDB labels it `StatementType.SELECT` and it does not start with the keyword.

Same for a parenthesized `(SELECT ...)` and `VALUES`.

kstonekuan (approved):
LGTM, merging.

FROM-first, parenthesized and `VALUES` reach both endpoints again.

Two filed rather than held against this: the wrapper parse refuses nothing the earlier checks do not already refuse, across 27 shapes (#481), and a binder error still returns `DESCRIBE SELECT * FROM (...)` in the 400 body (#482).

Pushed a test for `.rstrip(';')`, which had none, and one pinning that `(DESCRIBE SELECT 1)` is accepted while `DESCRIBE SELECT 1` is not.

Issue #432, opened 2026-09-06
[Feature]: LeRobot prepared-manifest receipts are written but nothing verifies a delivered import


VARUN3WARE:
### Problem

After `import_lerobot_dataset` finishes, a recipient of the delivery (landing MCAPs + `prepared-manifest.json`) cannot ask HFlow whether that delivery is still intact.

Schema-3 manifests already carry per-episode receipts (`uri`, `content_id`, `size_bytes`). The receipt docstring and `docs/how-to/import-lerobot-v3.md` say a delivery can be checked against the manifest without re-running import. There is no public library or CLI reader that performs that check:

```text
hasattr(hflow, "verify_prepared_manifest")  -> False
```

Truncation, same-path byte replacement, or a missing landing object stay silent until some later consumer fails for unrelated reasons. Hand checks exist only inside tests (`test_manifest_content_id_detects_a_truncated_episode`).

This is the LeRobot twin of the snapshot receipt/verify split (#397/#401 writer, #428 reader). Writer landed in #389; verify was deferred there. Verified against current `main` (`c6d6d3729d7d7107c72fe7de4a7caf11ee549167`).

### Proposed behavior

Add a first-party verification path that reads an existing schema-3 `prepared-manifest.json` and checks every listed episode against the landing files — without re-running conversion or touching the Hugging Face cache.

**Smallest useful outcome:** given a data root that already contains `prepared-manifest.json` and landing MCAPs, the caller learns pass/fail (and which episode failed) from the receipts HFlow already wrote (`uri`, `content_id`, `size_bytes`).

**Suggested API** (names flexible):

```python
from hflow import verify_prepared_manifest

report = verify_prepared_manifest("file:///path/to/data-root")
# report.ok is True when every episode matches
# on failure: which uri, and whether missing / size mismatch / content_id mismatch
```

**Suggested CLI** (pick one shape):

```bash
hflow verify lerobot-import file:///path/to/data-root
```

**Behavior contract:**

| Delivery state | Result |
| --- | --- |
| Unchanged after publish | Pass |
| Landing file missing | Fail, name the episode `uri` |
| Truncated / wrong `size_bytes` | Fail |
| Same-length byte replacement (`content_id` disagrees) | Fail |
| Extra files under `landing/` not listed in the manifest | Ignore (or warn — confirm with maintainers) |

[...]

Issue #428, opened 2026-09-06
Snapshot integrity receipts are written but nothing reads them: no way to verify a delivered snapshot


kstonekuan:
#401 landed per-file integrity receipts in a snapshot's `format.json` (#397). Nothing consumes them. The receipt is the half of the feature that survives a transfer; the half that checks it does not exist yet, and the how-to says so in as many words: "there is no public `verify_dataset_snapshot` API or CLI yet".

## Current state

Written at `src/hflow/snapshot.py:187`:

```json
"integrity": {
  "tables":  {"<name>": {"path": ..., "size_bytes": ..., "sha256": ...}},
  "assets":  [{"path": ..., "size_bytes": ..., "sha256": ...}],
  "content_id": "<sha256 of the normalized inventory>"
}
```

Read nowhere. The only reader of `format.json` is the overwrite guard at `:595-622`, and it looks at `format` and `format_version` only. So today the failure this feature exists to catch, a snapshot that arrived truncated or short a table, is silent: DuckDB opens what is there and returns fewer rows.

## What to build

A `verify_dataset_snapshot(snapshot_directory)` beside `export_dataset_snapshot`, exported from `hflow/__init__.py`, plus an `hflow verify snapshot` CLI command. It re-reads the delivered directory and compares it against the receipt.

The helpers are already there and are the right shape to reuse: `_sha256_hex` at `:116`, `_file_integrity_record` at `:125`, `_inventory_content_id` at `:134`.

Four distinct failures worth separating, because a caller does different things about each:

1. **A file's bytes changed.** `sha256` mismatch on a table or an asset.
2. **A file is missing.** Named in the receipt, absent on disk.
3. **A file arrived that the receipt does not name.** Extra members under `assets/`.
4. **No receipt at all.** A `format.json` from before #401. This is a valid v1 snapshot and must not be reported as corrupt. Say the receipt is absent and stop.

`size_bytes` is not a separate failure. It is a cheap pre-filter: compare it first and skip the hash when it already differs, which matters on a copy-mode snapshot holding video.

## What it is not

Not a tamper defence, and the API should not imply one. The receipt travels unsigned inside the `format.json` it describes, so anyone who can rewrite a table can rewrite the hashes to match. This catches corruption, truncation, and partial transfer. If the naming or the docstring makes it sound like a signature, that is a bug in this issue's output.

## Design questions worth answering in the PR

[...]

PR #454, opened 2026-09-07
feat(verify): check LeRobot prepared-manifest deliveries against receipts

VARUN3WARE:
## Summary
- Add shared `VerificationReport` / `VerificationFinding` / `VerificationStatus` in `src/hflow/verification.py` for the `hflow verify` family (#432; contract with #428).
- Implement `verify_lerobot_import` and CLI `hflow verify lerobot-import` to check schema-3 prepared-manifest receipts (`uri`, `content_id`, `size_bytes`) without re-converting.
- Exit codes: `0` clean, `1` damaged, `2` unreadable, `3` unverifiable; unlisted `landing/` extras ignored.
- Docs: `import-lerobot-v3.md` names the real verify path; `FORMAT.md` documents verify exit `3`.

## Test plan
- [x] `uv run ruff check` / `ruff format --check` / `ty check`
- [x] `uv run pytest tests/test_verification.py -q`
- [x] Full `uv run pytest -q` (1594 passed, 6 skipped)
- [x] Confirm CI green
- [x] @Sagar-024: please review — this lands `verification.py` first for #428 to import

### DIFF
--- a/docs/FORMAT.md
+++ b/docs/FORMAT.md
@@ -236,13 +236,23 @@ Every `hflow` command follows the same three-value convention:
 - `0` - success: the command did what it was asked.
 - `1` - ran, and found something to report: `doctor`, a non-conforming file;
   `stale --exit-code`, stale episodes found; `up`, started then failed, so
-  containers may still be running; `ingest`, episodes failed.
+  containers may still be running; `ingest`, episodes failed; `verify`, a
+  delivery whose receipts no longer match the claimed objects.
 - `2` - bad input, nothing useful happened: an invalid flag combination, an
   unreachable endpoint, a missing catalog, a `serve` launch that never started
   (a port outside 1-65535, a data root that exists and is not a directory, a
-  host with no free port), or a `doctor` run in which no file could be
-  diagnosed at all. A data root that is not there yet is not bad input: it is a
+  host with no free port), a `doctor` run in which no file could be
+  diagnosed at all, or a `verify` run whose root or receipt file cannot be
+  read. A data root that is not there yet is not bad input: it is a
   workspace nothing has ingested into, and `serve` and `up` both accept one.
+
+The `verify` family adds one further code so CI can tell an unverifiable
+delivery apart from a damaged one:
+
+- `3` - the command could read the data root, but there is no receipt to
+  verify against (for example `hflow verify lerobot-import` with no
+  `prepared-manifest.json`, or `hflow verify snapshot` with no integrity
+  receipt in `format.json`).
 
 `doctor` treats an unreadable file as at least as bad as a non-conforming one
 (exit 1), and reserves 2 for runs where nothing was diagnosed, so a batch run
--- a/docs/how-to/import-lerobot-v3.md
+++ b/docs/how-to/import-lerobot-v3.md
@@ -59,8 +59,21 @@ Durable outputs land as `landing/*.mcap` and `prepared-manifest.json` under
 that prefix. The manifest (schema version 3) carries a `episodes` receipt
 list: every delivered episode's published URI, its `content_id` (the sha256
 content address of the canonical bytes, the same value the catalog uses for
-dedupe), and its `size_bytes`, so a delivery can be checked against the
-manifest without re-running the import. Hugging Face downloads stay in the
+dedupe), and its `size_bytes`. Check a delivery without re-running the import:
+
+```bash
+uv run hflow verify lerobot-import ./data/lerobot_pusht
+# or, from Python:
+# hflow.verify_lerobot_import("./data/lerobot_pusht")
+```
+
+Verification resolves each episode as `landing/<basename>` under the data root
+you pass in (the same layout the importer writes). The receipt `uri` is
+publish-time provenance kept on findings; it is not opened as a lookup path,
+so a copied delivery is checked in place even when the original publish path
+still exists elsewhere.
+
+Hugging Face downloads stay in the
 local mirror under
 `_lerobot_cache/` (`HFLOW_MIRROR_DIR`, or `$XDG_CACHE_HOME/hflow/mirrors`) and
 are never uploaded into the bucket. The success manifest is published only
--- a/src/hflow/__init__.py
+++ b/src/hflow/__init__.py
@@ -49,7 +49,7 @@ from hflow.fingerprints import (
     step_version_from_contract,
 )
 from hflow.format import GopPreset
-from hflow.importers import import_lerobot_dataset
+from hflow.importers import import_lerobot_dataset, verify_lerobot_import
 from hflow.manifest import (
     DerivedChannelManifest,
     PipelineManifest,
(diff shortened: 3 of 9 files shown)


Sagar-024 (approved):
Approved. The part I like most is exit_code_for: 0, 1, and 3 live in one function, 2 is raised for unreadable input before a report exists, and FORMAT.md documents the split so CI can tell an unverifiable delivery from a damaged one. That split is easy to get wrong per command.

Checked against what the #432 contract needs: VerificationReport with .ok and .findings, each finding carrying .uri, .reason, and .detail, and the module importing from catalog and storage only, never from snapshot.py or lerobot.py. Extras ignored, nothing rewritten, nothing re-converted. Holds.

One coordination note for the snapshot verifier (#428): it will consume these same types, so its reasons need to live beside REASON_MISSING and friends rather than inventing a second vocabulary, and once two verifiers share this module it may be worth separating the shared types from the per-delivery verifiers. Neither blocks this PR.

VARUN3WARE:
Thanks @Sagar-024 ,  agreed on both notes for #428.

I'll keep snapshot reason strings beside the existing `REASON_*` constants in `verification.py` (same vocabulary, no second set of names). Happy to split shared types from the per-delivery verifiers in a follow-up once both verifiers are in-tree; not blocking this PR.

CI is green here; ready for maintainer merge whenever @kstonekuan has a slot.

kstonekuan (changes requested):
The contract is right; this is not about the shape.

The verifier reads the original rather than the copy, so a damaged copy verifies clean. Have a look at which path it opens.

VARUN3WARE:
@kstonekuan Fixed: episodes resolve as `landing/<basename>` under the verified root; receipt `uri` is provenance only. Empty `episodes` → ok. Added your three copy-root cases plus a bucket relative-fetch test. Ready for another look.

Sagar-024 (approved):
Re-reviewed the fix commit (25a69e3). Stale approval replaced, this still holds the #432 contract checklist:

- CLI verb: `hflow verify lerobot-import <root>` present.
- Report: `.ok` derived from status, findings in one list, no raise on first mismatch.
- Per-finding `uri` / `reason` / `detail`: present, reasons from the shared constants.
- Shared types: single `hflow.verification` module, no verifier-local report shapes.
- Exit codes: 0 clean, 1 damaged, 2 unreadable input (raises), 3 unverifiable, via `exit_code_for`.
- Extras ignored: unlisted files under `landing/` produce no finding.

The fix itself, verified in a scratch worktree (test_verification.py: 12 passed):

- Episodes now resolve as `landing/<basename>` under the root being verified; the receipt uri is provenance, never a lookup path.
- Both directions are pinned by mutation: mutating only the original keeps the copy green, damaging only the copy goes red.
- The deleted-original case and the bucket-root case are covered, plus the empty-claim receipt flipping to clean (the right call: nothing claimed, nothing damaged).
- Publish-time absolute uris still parse after a copy, so no schema bump is needed.

One note, non-blocking: `.zcode` shows a stray pr454.diff in my scratch from the review run; it is mine, not part of the PR.

Re-approving on the fix commit.

Sagar-024:
@VARUN3WARE Kingston unblocked the order, so I am landing #428 first with your type shape verbatim: VerificationStatus, VerificationFinding, VerificationReport with the ok property, your three REASON constants and exit_code_for, plus one REASON_NO_RECEIPT beside yours, hyphenated to match your convention. When you rebase #454, drop your copy of verification.py and import from the landed one. Your fix stays in the import-verifier half (the landing resolution and the empty-claim case), so nothing is lost. Ping me after the rebase and I will review again.

Sagar-024:
@kstonekuan carried into #428: the moved-root test exports a snapshot to one root, copies it to a second, deletes the first entirely, then verifies the copy clean and then damaged, plus a variant where the original stays alive and clean while the copy is damaged, so damage is never masked by the export root. PR #457 is up with it, and the shared types land in VARUN's exact shape so #454 re-lands as an importer.

VARUN3WARE:
@Sagar-024 Thanks for the re-review. Ack on order: once #457 lands I’ll rebase #454, import the shared types from your `verification.py`, and keep only the lerobot-import resolution / empty-claim / CLI side. Will ping you after the rebase.

kstonekuan:
Fix confirmed with the same probe that caught it. Before, the third line read `ok`; now:

```
original verify           -> ok
copy, original present    -> ok       findings=0
copy damaged, original ok  -> damaged  findings=['content-id-mismatch']
copy alone, original gone  -> damaged  findings=['content-id-mismatch']
```

The third line is the one that mattered: damage in the copy is found even with an intact original sitting next to it, so the verifier is reading what it was handed. The fourth is my probe still holding the damaged copy from the step before, so `damaged` is the right answer there too.

All four cases from the review are in, and `test_verify_lerobot_import_resolves_bucket_deliveries_under_the_verified_root` goes past what I asked for. Reasoning the bucket case rather than testing it would have been acceptable and testing it is better.

You also changed `episodes: []` from unverifiable to ok, which I agree with. A manifest in a format we understand that claims nothing has been verified against everything it claims. Exit 3 should mean "go and look", and there is nothing to look at.

Not approving yet only because main moved underneath you: #457 merged and landed `src/hflow/verification.py`, so this now has an add/add conflict there plus one in `__init__.py`. That is the rebase you and @Sagar-024 planned for, in the order you agreed.

When you do it: take the merged module as the base and add to it rather than reconciling two copies. `VerificationStatus`, `VerificationFinding`, `VerificationReport`, `exit_code_for` and the three shared reason strings are already there, and `REASON_NO_RECEIPT` arrived with the snapshot verifier. Your `verify_lerobot_import` and `_landing_relative_key_from_receipt_uri` are the new parts. If you find yourself editing the shared types to fit, say what does not fit instead, because that is a contract question rather than a merge.

The basename mapping is worth one line in the PR body when you push: it means the verifier and the importer share the landing layout without the manifest recording it. That is fine, and it is the sort of coupling that should be stated rather than discovered.

Ping me when it is clean and I will merge.

VARUN3WARE:
@kstonekuan  Rebased onto main after #457. Dropped our duplicate `verification.py` types and import yours; LeRobot half lives in `importers/lerobot_verify.py` (root-relative `landing/<basename>`, empty claim → ok, copy-root tests). Ready for another look when you have a minute.

kstonekuan (approved):
Clean rebase. The verifier moved to `src/hflow/importers/lerobot_verify.py`, the shared types are imported rather than reconciled, and the only edit to `verification.py` is a docstring now naming both concrete verifiers. That is the right shape: one contract, two implementations, neither owning the other.

Validated on the rebased branch: full gate clean, 1632 passed / 6 skipped. The path fix stays fixed, which I checked with the probe from the first round rather than trusting the rebase.

Merging. Between this and #457 the verify family is real: same report, same reason strings, same exit codes, two receipt formats.

Worth writing down somewhere eventually, though not in this PR: the basename mapping means the verifier and the importer share the landing layout without the manifest recording it. You noted it, and the next person to change `_landing_relative_key` needs to know a verifier depends on it. If a schema-4 receipt ever happens, carrying the relative key is the thing that removes the coupling.

Good work on both rounds, and on settling the contract with @Sagar-024 in writing before either of you wrote code. That is why these merged in two days instead of arguing in review.

PR #459, opened 2026-09-07
fix(checks): hoist camera threshold validations to fail early on came…

### LINKED ISSUE #447: camera_frame_stats and camera_signal_quality skip their threshold guards on a camera-less episode, and name internal fields when they do fire
> **Held for @Chama2001.** I have twice pointed them at an issue and had it taken before they replied, so this one is spoken for. GitHub will not let me assign someone who has not commented on the issue yet, and it fails silently rather than saying so. Chama2001: comment here and I will assign it. If a week passes with no comment, it is open to anyone.

#445 was one instance of a pattern. Here are five more, across two checks, verified on current main.

## Both halves, measured

`camera_frame_stats` (`src/hflow/checks.py:567-576`) and `camera_signal_quality` (`:1315-1321`) hand their thresholds to `FrameStatisticsSettings`, whose `__post_init__` (`src/hflow/_video_measurements/_frame_statistics.py:79`) does the refusing. That dataclass is constructed inside the per-camera loop.

```
camera_frame_stats.black_pixel_threshold=-5
    with camera : ValueError: black_pixel_luma_threshold must be between 0 and 255
    no camera   : ACCEPTED

camera_frame_stats.black_frame_amount_pct=900
    with camera : ValueError: black_frame_minimum_pixel_share_percent must be between 0 and 100
    no camera   : ACCEPTED

camera_frame_stats.freeze_min_duration_s=nan
    with camera : ValueError: freeze_minimum_duration_seconds must be finite and positive
    no camera   : ACCEPTED

camera_signal_quality.black_pixel_threshold=999
    with camera : ValueError: black_pixel_luma_threshold must be between 0 and 255
    no camera   : ACCEPTED

camera_signal_quality.freeze_noise_db=nan
    with camera : ValueError: freeze_noise_tolerance_decibels must be finite
    no camera   : ACCEPTED
```

Two separate defects in that table.

**The guard is skipped when the loop body does not run.** A state-only episode accepts `black_pixel_threshold=-5` and returns a clean result. Register the check over a corpus with a typo in a threshold and every camera-less episode passes while the first episode with a camera crashes. `tests/test_gates.py:20-24` builds a camera-less episode, and the how-to tells users to skip joint checks on human egocentric corpora, so mixed corpora are ordinary here.

**Every message names a field the caller cannot have typed.** You pass `black_pixel_threshold` and are told about `black_pixel_luma_threshold`. You pass `freeze_noise_db` and are told about `freeze_noise_tolerance_decibels`. Four of the five above rename the parameter in the error. This is the same problem #434 fixed for `shake_threshold_dps`, which used to report `minimum_shake_degrees_per_second`.

## The pattern to copy

Both fixes already exist in this file. #441 and #446 put `camera_stability`'s three numeric arguments at the top of the function, before `selected_cameras` is resolved, each refused under its own name. Do the same thing here.

Keep the `FrameStatisticsSettings` guards where they are. They protect that class's other callers, and they are the reason the bounds are known. You are adding an earlier refusal in the caller's vocabulary, not moving one.

## Definition of done

1. Both functions refuse a bad threshold before `selected_cameras` is resolved.
2. Each message names the parameter as the check's signature spells it.
3. Every case in the table above is covered, and at least one runs against a camera-less episode. A test that only uses a video episode passes today and proves nothing.
4. `bool` is rejected for the int-typed thresholds. `bool` subclasses `int`, so `True` otherwise passes as 1.
5. `expected_hz` and the other non-numeric arguments are out of scope.
6. Removing any guard you add must turn a test red. Say which ones you checked.

## Worth deciding rather than assuming

The five rows are not the whole surface: `bright_luma_threshold`, and `camera_fps_conformance`'s `max_plausible_fps` and `downsample_tolerance_fps`, are the same shape and I did not verify them. Check them yourself and either include them or say why not. Do not take my list as the boundary of the problem.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Chama2001:
Closes #447

## Summary
Moved threshold guards before the camera loop in `camera_frame_stats` and `camera_signal_quality`.

## Why
As discussed in #447, previously these guards were inside the `for topic in selected_cameras:` loop, meaning they were skipped entirely on camera-less episodes. This change hoists them to the top of the functions so they fail early and correctly.

## Validation
Applied the exact pattern requested in the issue description.

### DIFF
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -16,7 +16,6 @@ rather than a shared ``message_count``.
 """
 
 import hashlib
-import math
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass
 
@@ -605,26 +604,35 @@ def camera_frame_stats(
     lookup per key on top of the ffmpeg decode each topic already pays (#182).
     """
 
-    import math
-
+    # Ahead of selected_cameras so a bad threshold is refused on a camera-less
+    # episode too, which is where #447 found them silently skipped. bool is
+    # refused explicitly on every one: it subclasses int, so True reads as 1
+    # and sails through both the range comparisons and np.isfinite.
     if isinstance(black_pixel_threshold, bool):
-        raise TypeError("black_pixel_threshold must be an integer")
+        raise ValueError("black_pixel_threshold must be an int, got bool")
     if not (0 <= black_pixel_threshold <= 255):
         raise ValueError("black_pixel_threshold must be between 0 and 255")
 
-    if type(black_frame_amount_pct) is bool:
-        raise TypeError("black_frame_amount_pct must be an integer")
+    if isinstance(black_frame_amount_pct, bool):
+        raise ValueError("black_frame_amount_pct must be an int, got bool")
     if not (0 <= black_frame_amount_pct <= 100):
         raise ValueError("black_frame_amount_pct must be between 0 and 100")
 
-    if not math.isfinite(freeze_min_duration_s) or freeze_min_duration_s <= 0:
+    if isinstance(freeze_min_duration_s, bool):
+        raise ValueError("freeze_min_duration_s must be a float, got bool")
+    if not np.isfinite(freeze_min_duration_s) or freeze_min_duration_s <= 0:
         raise ValueError("freeze_min_duration_s must be finite and positive")
 
-    if not math.isfinite(freeze_noise_db):
+    if isinstance(freeze_noise_db, bool):
+        raise ValueError("freeze_noise_db must be a float, got bool")
+    if not np.isfinite(freeze_noise_db):
         raise ValueError("freeze_noise_db must be finite")
 
+    if isinstance(bright_luma_threshold, bool):
+        raise ValueError("bright_luma_threshold must be an int, got bool")
     if not (0 <= bright_luma_threshold <= 255):
         raise ValueError("bright_luma_threshold must be between 0 and 255")
+
     selected_cameras = list(cameras) if cameras is not None else episode.cameras
     intermediates_by_topic = {
         topic: _camera_intermediates(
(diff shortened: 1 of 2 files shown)


kstonekuan (changes requested):
Four things.

Nothing tests any of it. None of the five messages appears anywhere in the repo, so removing any line you added leaves the suite green. #447 asks for every case in its table, at least one on a camera-less episode.

Three of the five parameters still accept `True`. `freeze_min_duration_s`, `freeze_noise_db` and `bright_luma_threshold` pass as 1.

`type(x) is bool` and `TypeError`: everywhere else here is `isinstance` and `ValueError`. See `_field_guards.require_int` and #441/#446.

`import math` inside two function bodies: use the `np.isfinite` already imported.

Also: #447 asked you to rule on `camera_fps_conformance` either way. Say which.

kstonekuan (changes requested):
Getting close. `camera_signal_quality` is done; `camera_frame_stats` still has no tests.

Pushed the three mechanical items rather than sending you back: the missing bool clauses, `TypeError` to `ValueError`, `isinstance`, and `np.isfinite`. All eight parameters now refuse `True`.

Also rewrote your test. It used an episode with cameras and a bare `pytest.raises(ValueError)`, which a camera-processing failure also satisfies, so it passed whether or not the guard was there. That is the bug #447 is about, inside its own test.

What is left: the same treatment for `camera_frame_stats`, ten rows.

On `camera_fps_conformance`: leave it out, I will file it.

kstonekuan (approved):
Taking this over after five days, so the #447 gap closes and your slot frees.

Your diagnosis was right and it is what the fix does: the guards were inside the loop over `selected_cameras`, so a camera-less episode ran none of them.

Force-pushed onto current main, because the branch no longer rebased: `_field_guards` and `_resolve_selected_cameras` both landed underneath it. The guards now compose from the helpers rather than hand-rolling, which also takes checks.py's share of #498.

`bright_luma_threshold` is a float, so it does not map onto `require_int_in_range`; the helper owns finiteness and the bound stays local.

Tests are camera-less with anchored patterns, which matters here: on an episode with cameras a camera-processing failure also raises `ValueError`, so an unanchored assertion passes whether or not the guard exists.

Mutations: removing either function's guards fails 8 and 6 cases; moving one below the loop fails 2.

PR #461, opened 2026-09-07
fix(lerobot): stop flooring a fractional source fps

### LINKED ISSUE #458: A fractional fps in meta/info.json is floored before conversion, stretching the episode's time axis
`_parse_dataset_information` accepts any finite positive fps, int or float, and stores it
as one. #313 made that explicit when it added the non-finite refusal. The single call
site then floors it:

```python
frames_per_second=int(source_archive.fps),
```

so a corpus declaring `"fps": 29.97` is converted as if it were 29. Every message log
time is computed as `frame_index * 1e9 / frames_per_second`, so the whole time axis
stretches:

```
frame     1: source    0.0334s   written    0.0345s   drift +0.0011s
frame    30: source    1.0010s   written    1.0345s   drift +0.0335s
frame   900: source   30.0300s   written   31.0345s   drift +1.0045s
frame  5400: source  180.1802s   written  186.2069s   drift +6.0267s
```

A second off after thirty seconds, six off across a three minute clip. The clips in #287
are about 180 s, so this is the size of corpus where it lands.

It also moves the keyframe interval. `_transcode_mp4_to_h264` computes
`round(gop_seconds * frames_per_second)`, which is 30 at 29.97 and 29 at the floored
value, so the file does not carry the GOP its own provenance record claims.

29.97 is not exotic. It is what NTSC-derived capture writes, and nothing upstream of the
converter rejects it, so the import succeeds and the damage is silent.

Every fps in the test suite is the integer 30, which is why this has not shown up.

Expected: the source fps is used as declared. Both consumers already take a float
(`_transcode_mp4_to_h264` annotates `frames_per_second: float`, and the log time is a
division), so only the call site and one annotation force the truncation.

I have a fix and a test.

chiruu12:
Fixes #458

`meta/info.json` is allowed a fractional fps and `_parse_dataset_information` stores it as
one, but the single call site floored it with `int(source_archive.fps)`. A corpus
declaring 29.97 was converted as if it were 29.

The keyframe interval is the primary half. `_transcode_mp4_to_h264` computes
`round(gop_seconds * frames_per_second)`, which is 30 at the declared rate and 29 at the
floored one, so the file carried a 29 frame GOP while `provenance/v1` stamped
`gop_seconds` as `1`. That field is recorded as what was actually used and nothing checks
it, which is what #376 is open about. This was one of the ways it could already be false.

The time axis is the visible half. Log times are `frame_index * 1e9 / frames_per_second`,
so the whole episode stretches:

```
frame     1: source    0.0334s   written    0.0345s   drift +0.0011s
frame    30: source    1.0010s   written    1.0345s   drift +0.0335s
frame   900: source   30.0300s   written   31.0345s   drift +1.0045s
frame  5400: source  180.1802s   written  186.2069s   drift +6.0267s
```

Both consumers already took a float, so the fix is the `int()` at the call site and the
annotation that forced it.

`CONVERTER_VERSION` moves to v8, and the reason is resume rather than bytes.
`_episode_identity_matches` compares dataset, revision, episode index, camera keys,
converter version and gop_seconds. It never looks at fps, so an episode delivered before
this fix with the stretched time axis still matches on every field and the resume path
would hand it back as completed work. The bump is what makes those episodes fail identity
and be re-converted. This differs from the bumps in #390 and #403: those moved every
episode's content id, this one changes bytes only for fractional-fps sources, and an
integer-fps corpus produces identical output either way.

Tests, in the order they were asked for:

- the declared fps reaches conversion untruncated
- the log times come from the declared rate, with the floored value asserted absent
- the rate the transcoder receives yields a keyframe interval of 30 rather than 29
- an episode stamped with the previous converter version is refused by reuse

All four fail on the commit before the fix. The corpus helper now takes an fps so the
end-to-end import runs at 29.97 rather than the integer 30 every existing fixture uses,
which is why this survived: nine fixtures, all at 30, and `int(30.0)` is 30.

`tests/test_lerobot_converter.py tests/test_lerobot_metadata_refusals.py
tests/test_lerobot_export.py`: 107 passed. `ruff check`, `ruff format --check` and
`ty check` are clean.

### DIFF
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -1053,7 +1053,7 @@ def import_lerobot_dataset(
                 episode_index=selected_episode_index,
                 camera_keys=resolved_camera_keys,
                 numeric_schemas=numeric_schemas,
-                frames_per_second=int(source_archive.fps),
+                frames_per_second=source_archive.fps,
             )
         )
         episodes_converted += 1
@@ -1089,7 +1089,7 @@ def _convert_single_episode(
     episode_index: int,
     camera_keys: tuple[str, ...],
     numeric_schemas: dict[str, _NumericSchema],
-    frames_per_second: int,
+    frames_per_second: int | float,
 ) -> _PublishedEpisode:
     """Convert a single episode to canonical MCAP and publish it.
 
@@ -1312,6 +1312,9 @@ def _convert_single_episode(
             )
 
             for frame_index in range(frame_count):
+                # Divided at the source rate, not a truncated one. meta/info.json
+                # is allowed a fractional fps, and 29.97 floored to 29 stretches
+                # the time axis by about a second every thirty.
                 log_time_ns = EPISODE_START_TIME_NS + round(
                     frame_index * NANOSECONDS_PER_SECOND / frames_per_second
                 )
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -55,7 +55,11 @@ DEFAULT_CAMERA_KEY = "observation.image"
 # can prove a landing file belongs to this exact selection (#303). Those
 # fields change the canonical bytes that content_episode_id hashes, so v5
 # and v6 outputs must not share a converter identity.
-CONVERTER_VERSION = "lerobot-converter-v7"
+# "v8": a fractional source fps is no longer floored, so every message log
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Both halves are held separately. Dropping the truncation fixes new conversions; the version bump is what stops resume handing back episodes already written with a stretched time axis.

PR #462, opened 2026-09-07
fix(ingest): CRC-validate the primary ingest read, not just the resume path

### LINKED ISSUE #431: Ingest transcodes a payload-damaged source without noticing, so the canonical episode gets a valid receipt for corrupt bytes
#429 fixed this on the LeRobot resume path. The same hole is open one stage earlier, on the primary path every ingest takes.

## The two facts

**A non-validating MCAP reader does not see payload damage.** This is measured, not assumed. On #429 I mutated the new guard from `validate_crcs=True` to `validate_crcs=False`, changing nothing else, and the payload-corruption test went green again: the damaged file read cleanly, all messages iterated, no error. Chunk CRCs are only checked when the reader is told to check them.

**Ingest reads through exactly such a reader.** `transform.py:576` does `reader = open_reader(source_path)`. `open_reader` (`reader.py:296`) returns a `PythonMcapEpisodeReader`, whose constructor calls `make_reader(self._stream)` at `reader.py:147` with no `validate_crcs`. The only two validating readers in the tree are `doctor.py:207` and the one #429 just added at `lerobot.py:206`.

So a source recording with bit rot inside a chunk, structurally valid and correct in every metadata record, transcodes without complaint. The canonical episode that comes out gets a `content_id` that is a true hash of bytes derived from corrupt input. That is #426's finding exactly: a receipt issued over bytes nobody checked.

## Why the existing tests do not cover it

Every corrupt-source test in the suite writes `b"not an mcap file"` (`tests/test_ingest_in_process.py:136`, `:188`, `:232`, `:254`, `tests/test_stage_planning.py:393`). That is caught by magic-byte validation in the reader constructor and classified `source-unreadable` (`ingest_ledger.py:65`). Structural damage is covered. Payload damage inside a valid container is not tested anywhere on this path.

`tests/reuse_test_helpers.py` landed with #429 and has `flip_chunk_payload_bytes`, which produces precisely this file: container, indexes, and metadata intact, chunk CRC no longer matching. The fixture for this issue already exists.

## Why it may be worse here than on the resume path

Resume reuses bytes it already produced, so the damage is HFlow's own output rotting on disk. Ingest is where someone else's recording enters the system. The corrupt bytes are transcoded into a canonical episode, and from that point every downstream receipt, content id, and check result is a faithful record of corrupt data. Nothing later re-reads the source to notice.

## What to decide

The fix is not just "add the flag", because ingest is the path where cost and policy both bite:

- **Cost.** The transcode already reads every message, so validation is close to free in the same sense #429 measured (1.15x on a read that was already happening). Confirm that on this path rather than inheriting the number. Ingest handles much larger files.
- **Outcome.** A payload-damaged source should almost certainly fail as `source-unreadable`, matching how the structurally invalid case is already classified. Argue it, and check whether the ledger's vocabulary should distinguish "not MCAP" from "MCAP with damaged payload". An operator does different things about those two.
- **Blast radius.** `open_reader` is used by `app.py:1390`, `episode.py:311`, `transform.py:576`, and `runtime/_templates.py:717`. Turning validation on inside `open_reader` itself would change all of them at once, including hot read paths that have nothing to do with trusting new input. Decide whether this is a new validating entry point or a flag on the existing one, and say why.

## Definition of done

1. A source MCAP with a damaged chunk payload does not produce a canonical episode with a fresh receipt.
2. The failure is classified in the ingest ledger, with a test asserting which outcome.
3. The structurally invalid case keeps its current classification and message.
4. The cost of validation on this path is measured and stated in the PR.
5. Read paths that are not ingesting new input are not slowed down by the change, or the PR explains why the cost is acceptable.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Sravan1011:
Closes #431.

## Summary

#429 fixed the LeRobot resume path. The same hole was open one stage earlier, on the primary path every ingest takes: `transform.py`'s read of the source file (`open_reader`) never asked for CRC validation, so a structurally valid MCAP with a damaged chunk payload transcoded without complaint and got a fresh, true receipt over corrupt bytes — #426's finding, one stage upstream.

## What changed

- `open_reader` / `PythonMcapEpisodeReader` (`reader.py`) gain `validate_crcs: bool = False`, threaded straight into the existing `make_reader(..., validate_crcs=...)` call — the same shape `doctor.py` and #429's fix already use.
- Default stays `False`, so `app.py:1390`, `episode.py:311`, and `runtime/_templates.py:717` are unaffected: those read canonical files HFlow already produced and already identifies by content hash. Only `transform.py`'s ingest read opts in with `validate_crcs=True`, since that's the one call reading someone else's file for the first time.
- `ingest_ledger.classify_ingest_failure` gets a new branch for `mcap.stream_reader.CRCValidationError`. This is not cosmetic: `CRCValidationError` subclasses `ValueError`, **not** `McapError`, so without this branch a payload-damaged source would silently fall through to `IngestFailureKind.INFRASTRUCTURE` — blaming the platform for a bad recording. It classifies to the same `SOURCE_UNREADABLE` kind as the not-MCAP case; `error_type` (`InvalidMagic` vs `CRCValidationError`) already stays stored verbatim beside the classification, so the two stay distinguishable in the ledger without a new `IngestFailureKind` member.

## Cost

The transcode already decodes every chunk over one full read (`reader.iter_batches()`), so `validate_crcs=True` checks a CRC on a pass that was already happening rather than adding one. Measured directly on ingest-scale sources (not inherited from #429's smaller fixture number):

| Source size | `validate_crcs=False` | `validate_crcs=True` | Ratio |
|---|---|---|---|
| 30 MB (2 cameras, 120s, 640x480 @ 15Hz) | 269.7 ms | 300.1 ms | 1.113x |
| 75 MB (2 cameras, 300s, 640x480 @ 15Hz) | 783.3 ms | 810.7 ms | 1.035x |

Best-of-5 wall time, filesystem cache warmed identically before each series.

## Tests

- `tests/test_ingest_ledger.py::test_classify_crc_validation_error_as_source_unreadable` — pins the classifier branch directly.
- `tests/test_ingest_ledger.py::test_ingest_refuses_a_source_with_a_damaged_chunk_payload` — a source with a damaged chunk payload (via the shared `flip_chunk_payload_bytes` fixture from #429) raises through `write_canonical_episode` and produces no output file.
- `tests/test_ingest_in_process.py::test_a_payload_damaged_source_is_classified_the_same_as_unreadable` — end-to-end through `hflow ingest`: ledger row lands with `failure_kind = source-unreadable`, `error_type = CRCValidationError`.
- The existing not-an-MCAP tests (`InvalidMagic` → `source-unreadable`) are unchanged and still pass, confirming the structurally-invalid case keeps its current classification.

## Validation

```
uv sync --locked --all-extras
uv run pytest -q                  # 1623 passed, 6 skipped
```

### DIFF
--- a/src/hflow/ingest_ledger.py
+++ b/src/hflow/ingest_ledger.py
@@ -100,12 +100,24 @@ def classify_ingest_failure(error: BaseException) -> IngestFailureKind:
         return IngestFailureKind.SOURCE_UNSUPPORTED
     try:
         from mcap.exceptions import McapError
+        from mcap.stream_reader import CRCValidationError
     except ImportError:  # pragma: no cover - mcap is a hard dependency
         return IngestFailureKind.INFRASTRUCTURE
     if isinstance(error, McapError):
         # The single base of InvalidMagic, EndOfFile, RecordLengthLimitExceeded
         # and friends: the file is not a readable MCAP, which is a fact about
         # the recording rather than about this machine.
+        return IngestFailureKind.SOURCE_UNREADABLE
+    if isinstance(error, CRCValidationError):
+        # Raised by a CRC-validated read (transform.py's ingest read passes
+        # validate_crcs=True) on a structurally valid MCAP whose chunk payload
+        # does not match its recorded checksum. mcap does not root this on
+        # McapError -- it subclasses ValueError instead -- so it needs its own
+        # branch or it silently falls to INFRASTRUCTURE below, blaming the
+        # platform for a damaged recording. Same failure kind as McapError:
+        # the file is a fact about the recording, not the machine. error_type
+        # still distinguishes "not MCAP" (InvalidMagic) from "MCAP with a
+        # damaged payload" (CRCValidationError) in the stored row.
         return IngestFailureKind.SOURCE_UNREADABLE
     return IngestFailureKind.INFRASTRUCTURE
 
--- a/src/hflow/reader.py
+++ b/src/hflow/reader.py
@@ -140,11 +140,11 @@ class EpisodeReader(Protocol):
 class PythonMcapEpisodeReader:
     """Pure-Python :class:`EpisodeReader` backend over the stock ``mcap`` package."""
 
-    def __init__(self, path: Path | str) -> None:
+    def __init__(self, path: Path | str, *, validate_crcs: bool = False) -> None:
         self.path = Path(path)
         self._stream: IO[bytes] = self.path.open("rb")
         try:
-            self._reader: McapReader = make_reader(self._stream)
+            self._reader: McapReader = make_reader(self._stream, validate_crcs=validate_crcs)
         except BaseException:
             # make_reader validates the magic bytes; don't leak the handle
             # when it rejects a non-MCAP, empty, or truncated file.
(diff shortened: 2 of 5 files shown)


kstonekuan (approved):
LGTM, merging.

Checked the MRO rather than taking it: `CRCValidationError` subclasses `ValueError`, not `McapError`, so the classification branch is needed rather than defensive.

PR #464, opened 2026-09-08
refactor: move require_int/require_float to hflow._field_guards

### LINKED ISSUE #463: testing.py carries a byte-identical copy of _field_guards.require_int/require_float
`src/hflow/testing.py:60-67` defines two helpers that are byte-for-byte identical to `src/hflow/_video_measurements/_field_guards.py:17-30`, message strings included.

```python
# testing.py
def _require_int(value: object, name: str) -> None:
    if isinstance(value, bool) or not isinstance(value, int):
        raise ValueError(f"{name} must be an int, got {type(value).__name__}")

# _field_guards.py
def require_int(value: object, name: str) -> None:
    """Refuse anything but a plain ``int``, ``bool`` included."""
    if isinstance(value, bool) or not isinstance(value, int):
        raise ValueError(f"{name} must be an int, got {type(value).__name__}")
```

Same for `_require_float` / `require_float`. Only the docstring differs.

## Why it is worth one small change

`_field_guards` is imported by exactly two modules today (`_frame_statistics.py:17`, `_camera_motion.py:10`), both inside `_video_measurements`. Its docstring says it is "for the video-measurement settings dataclasses", so the next person who needs this check does what `testing.py` did and writes it again. That is how a shared helper stops being shared.

`testing.py` uses its copy six times, so this is a real import, not a dead definition.

## What to do

1. Move the two functions to a home that is honest about being general. `_field_guards.py` sits under `_video_measurements` and nothing about `require_int` is video-specific, so a top-level private module (`src/hflow/_field_guards.py`) fits better than importing across package boundaries.
2. Re-point `_frame_statistics.py`, `_camera_motion.py`, and `testing.py` at it. Delete `testing.py`'s copy.
3. Keep both public names as they are. This is a move, not a rename.

## Explicitly not in scope

**Do not sweep the other 50 call sites.** There are about fifty places in `src/` that write `isinstance(x, bool)` inline, and most of them should stay exactly as they are. They combine the type check with a finiteness and range check in one condition, and their message names the parameter the caller typed:

```python
if (
    isinstance(horizontal_field_of_view_degrees, bool)
    or not np.isfinite(horizontal_field_of_view_degrees)
    or not 0 < horizontal_field_of_view_degrees <= 360
):
    raise ValueError("horizontal_field_of_view_degrees must be finite and in (0, 360], got ...")
```

`src/hflow/checks.py:1004-1010`. That shape is deliberate, and #434, #445 and #447 exist precisely because the alternative reported an internal field name the caller never typed. Replacing those with a shared helper would undo work merged this week. If a call site does nothing but the type check, folding it in is fine; if it carries a range or a caller-facing message, leave it.

`packaging.py:1998-2056` has its own `_require_integer` family that returns the parsed value rather than returning `None`. Different contract, different module, out of scope.

## Definition of done

1. One definition of each helper, imported by all three current users.
2. No behavior change anywhere: same exceptions, same message strings.
3. The full suite passes with no test edits. If a test needs changing, the move changed behavior and something is wrong.
4. Deleting the moved definition breaks the import rather than silently falling back to a copy.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

7487:
## Summary

`require_int` / `require_float` now have one definition, `src/hflow/_field_guards.py`, imported by `_frame_statistics.py`, `_camera_motion.py` and `testing.py`. The byte-identical private copy in `testing.py` is gone. Same exception types, same message strings, same public names. No behavior change; the other ~50 inline `isinstance` sites are untouched, per the issue.

Fixes #463

## Why

The move is what the issue asks for. One thing did not go as the issue's definition of done predicted, and it is worth being explicit about:

**One test edit was required.** `tests/test_module_boundaries.py` enforces two import contracts that any single-definition layout has to cross:

- `_video_measurements/*` may import nothing from `hflow.*` (empty allowlist), and
- `hflow.testing` may import only `hflow.ffmpeg` and `hflow.format`.

With the layout from the issue and no test change, the guard fails with exactly the three imports this PR adds:

```
hflow/_video_measurements/_camera_motion.py:10: from hflow._field_guards import require_float -- the video-measurements package must not import hflow or any hflow.* module because it is designed to be extracted into a standalone package
hflow/_video_measurements/_frame_statistics.py:17: from hflow._field_guards import require_float, require_int -- ...
hflow/testing.py:49: from hflow._field_guards import require_float, require_int -- hflow.testing may import only hflow.ffmpeg and hflow.format; ...
```

The AST guard skips relative imports, so `from .._field_guards import ...` inside `_video_measurements` would have passed it silently. I did not do that: it is the same dependency, just hidden from the test. Instead both boundaries allowlist `hflow._field_guards` by name, with the rule text updated to say so. `_field_guards` has no imports of its own, so extracting `_video_measurements` later means vendoring 14 lines. I mutation-checked that a `from hflow.storage import ...` inside `_camera_motion.py` is still red under the new rule.

If you would rather keep the `_video_measurements` allowlist empty, the alternative is to leave `_field_guards.py` where it was and have `testing.py` import it from there (only the `hflow.testing` allowlist changes). Happy to flip; it is a one-line direction change.

Also removed the `testing.py` module-docstring bullet that explained why the copy existed, since it no longer does.

## Validation

```
uv sync --locked
uv run ruff check                 # All checks passed!
uv run ruff format --check        # 227 files already formatted
uv run ty check                   # All checks passed!
uv run pytest -q                  # 1609 passed, 9 skipped, 27 failed (see note)
```

Ran on macOS (CI is Linux-only). The 27 failures are all `tests/test_packaging*.py` raising `CythonOverlayBuildError: native overlay builds currently require CPython on Linux; found cpython on darwin`; the identical 27 fail on pristine `upstream/main` (7df224a) in the same environment, so they are platform-only and CI covers them.

Additional checks:

- `uv run pytest -q tests/test_module_boundaries.py` with the source change and the *unmodified* test: 1 failed with the three violations quoted above, then 2 passed with the allowlist.
- Mutation: appended `from hflow.storage import _load_obstore` to `_camera_motion.py`, guard went red naming that line; restored, green.
- `tests/test_testing.py`, `tests/test_video_measurement_settings_types.py`, `tests/test_video_measurement_settings_ranges.py`: 103 passed, unmodified. These already pin every `must be an int[, got ...]` / `must be an int or float, got ...` message through the public constructors, which is why no new test is added for a pure move.
- `git diff --check` clean.

## Checklist

- [ ] I added or updated outcome-focused tests for changed business logic. -- no business logic changed; existing outcome tests cover both consumers and pass unmodified. The only test edit is the boundary allowlist described above.
- [ ] I updated documentation for changed behavior, flags, formats, or requirements. -- no user-facing behavior changed; docs do not reference `_field_guards`.

### DIFF
--- a/src/hflow/_field_guards.py
+++ b/src/hflow/_field_guards.py
@@ -1,13 +1,13 @@
-"""Shared numeric-field validation for the video-measurement settings dataclasses.
+"""Shared numeric-field type guards for caller-constructed settings.
 
 Range checks alone (``0 <= x <= 100``) do not reject the wrong *type*: ``bool``
 subclasses ``int``, so ``True``/``False`` satisfy every numeric comparison and
-range test the settings below already run, and a ``str``/``None`` would raise
-a bare ``TypeError`` from the comparison itself instead of a clear message
+range test a settings dataclass runs, and a ``str``/``None`` would raise a
+bare ``TypeError`` from the comparison itself instead of a clear message
 naming the field. This closes that hole the same way ``catalog.py`` already
 does for interval bounds and measurement values.
 
-These settings are caller-constructed configuration, not measurement-pipeline
+These guards are for caller-constructed configuration, not measurement-pipeline
 output, so unlike ``catalog.py``'s NumPy-scalar coercion, no NumPy handling is
 added here: a caller building a ``np.float64`` threshold can call ``.item()``
 itself, the same way any other non-native-Python value would need to.
--- a/src/hflow/_video_measurements/_camera_motion.py
+++ b/src/hflow/_video_measurements/_camera_motion.py
@@ -7,7 +7,8 @@ from types import ModuleType
 
 import numpy as np
 
-from ._field_guards import require_float
+from hflow._field_guards import require_float
+
 from ._raw_frames import LUMA_FRAME_FILTER_GRAPH, luma_frames
 from ._toolchain import VideoMeasurementToolchain
 
--- a/src/hflow/_video_measurements/_frame_statistics.py
+++ b/src/hflow/_video_measurements/_frame_statistics.py
@@ -14,7 +14,8 @@ from io import StringIO
 from pathlib import Path
 from typing import Protocol
 
-from ._field_guards import require_float, require_int
+from hflow._field_guards import require_float, require_int
+
 from ._toolchain import VideoMeasurementToolchain
 
 FRAME_STATISTICS_DEFINITION_VERSION = "video-frame-statistics/v1"
--- a/src/hflow/testing.py
+++ b/src/hflow/testing.py
@@ -31,9 +31,6 @@ Implementation notes:
   ``MSG: pkg/msg/Type`` headers) so any ROS 2-aware reader decodes them.
 - Everything is deterministic: fixed ``start_time_ns``, seeded noise, no
   wall-clock reads.
-- ``_require_int`` / ``_require_float`` are copied here rather than imported
-  from ``hflow._video_measurements`` so that incubating package can stay
-  dependency-free and extractable (see ``src/hflow/_video_measurements/__init__.py``).
 """
 
 import errno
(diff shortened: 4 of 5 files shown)


kstonekuan (approved):
Gate clean, 1709 passed / 6 skipped.

My issue was wrong and you caught it. `testing.py` carried a comment saying the copy was deliberate, so the incubating package could stay dependency-free and extractable, and I filed #463 calling it a duplicate without reading it. The trade was eight duplicated lines for a sealed boundary.

Some boundary has to give to fix it at all. `testing.py` importing from `_video_measurements` would couple the fixture to the incubating package, which is the worse direction. Moving the guards to a top-level leaf module and widening the allowlist by one entry is the cheaper side.

The widened rule still bites:

```
video-measurements imports a domain module   FAILED test_documented_hflow_import_boundaries
video-measurements imports hflow.format      FAILED test_documented_hflow_import_boundaries
```

Both refused. The allowlist names one module, not a prefix.

One cost remains: extracting `_video_measurements` now means taking `hflow/_field_guards.py` along. That's one twelve-line file with no imports, and your rule text says so.

Merging.

PR #465, opened 2026-09-08
fix(doctor): report unsupported video encodings and missing schemas separately from read-failed

### LINKED ISSUE #460: doctor labels a healthy video corrupt when a decoder is unavailable
Current behavior
The doctor video check reads messages inside one broad try. When the machine cannot decode a stream, meaning there is no decoder available for that video, the read raises inside the loop and is reported as:

read-failed: reading messages failed (corrupt chunk or bad CRC?)

The file is intact. Every byte validates. The machine simply cannot decode it. But the diagnostic asserts "corrupt chunk or bad CRC," which points the operator at data corruption when the real cause is a missing capability.

Controlled repro
- Take a known-good video that passes doctor on a machine with the codec.
- Make that codec unavailable on the machine.
- Run hflow doctor. The intact file reports read-failed, corrupt chunk or bad CRC, even though CRCs validate and nothing is damaged.

Why it matters
This is a dishonest diagnostic. The finding asserts corruption when the true cause is environment capability. Same honesty line as the provenance stamps: a report should say what is actually true. An operator told "corrupt chunk" will distrust a healthy file. The correct action is to install the decoder, not quarantine the data.

Proposed fix
Separate decoder resolution from the read. Catch the decoder-unavailable case at codec resolution and report it as its own finding, for example video-decoder-unavailable, naming the stream. Reserve read-failed for genuine read and CRC errors. Small diff, one test.

Definition of done
- A missing decoder reports video-decoder-unavailable, not read-failed
- A genuine corrupt or CRC-failed read still reports read-failed
- A decodable healthy video reports clean
- Mutation: make decoder resolution always fail, confirm the new finding fires and read-failed does not
- Existing doctor tests unaffected

Scope note
This only splits the decoder-unavailable case out of the broad read handler. It does not change CRC validation, the severity model, or any other doctor check. Claiming this, will build once accepted.

Sagar-024:
# fix(doctor): report unsupported video encodings and missing schemas separately from read-failed

Closes #460.

## What changed

Doctor's video check previously ran inside one broad `try` block. That meant two failures that weren't actually corruption problems ended up being reported as `read-failed`, along with the misleading "corrupt chunk or bad CRC?" assertion.

### Unsupported video encodings

`_resolve_video_decoder` raises when neither supported decoder factory (ROS2 or protobuf) can handle a channel's `message_encoding`.

That isn't evidence of corrupted bytes or a machine codec problem. It simply means the file uses an encoding that isn't supported by our current decoder set.

This now reports as `video-encoding-unsupported`, including the topic name and a detail message making it clear that the file bytes themselves are not implicated. The resolver now raises the typed `VideoEncodingUnsupported`, which remains a `ValueError` subclass.

### Missing schemas

The existing in-loop `schema is None` check was effectively dead code.

There are two reasons:

* A channel with a dangling schema gets an empty schema name, so it never gets classified as a video channel.
* The reader raises `KeyError` during iteration before the loop body can reach that check.

The missing-schema case is now handled in a summary pre-pass instead. Any channel that references a schema ID not present in the file reports `channel-schema-missing`.

`schema_id == 0` remains valid and means the channel is schemaless.

Channels with missing schemas are excluded from the read pass using the reader's `topics` filter, which keeps `read-failed` reserved for actual read and CRC failures.

## Severity

Both findings remain **ERROR**.

Doctor is certifying episode conformance. If a video topic uses an encoding our decoder set cannot handle, the episode can't be fully certified because downstream readers rely on that same decoder support.

Making this a WARNING would allow Doctor to certify a file whose video was never actually checked.

The distinction is reflected in the finding codes and messages, which now describe what actually happened. The severity represents "cannot certify," consistent with findings such as `no-statistics` and `no-chunk-indexes`.

## Behavior notes

* **Healthy files:** No behavioral change. When all channels are valid, the read pass remains unfiltered and follows the exact same path as before.
* **Excluded topics:** Chunks containing only excluded topics are no longer CRC-validated during the read pass. Previously, iteration would stop at the first exception anyway, so CRC coverage for later chunks in already-broken files was already incomplete. Healthy files still receive the full pass.
* **Zero-message channels:** Decoder resolution now happens during the summary pre-pass, so unsupported video channels with zero messages are reported instead of remaining silent.

## Tests

`23 passed` in `tests/test_doctor.py`, including two new real-path tests with no stubs:

* A JSON-encoded video channel reports `video-encoding-unsupported` and never `read-failed`.
* A channel with a fabricated schema ID (`99`) alongside a healthy sibling reports `channel-schema-missing` and never `read-failed`, while the sibling's video checks still run.

Mutation checks also confirmed the new coverage: removing either the pre-pass or the encoding catch makes exactly the two new tests fail. Restoring them brings everything back to green.

`ruff check`, `ruff format`, and `ty` are all clean.

### DIFF
--- a/src/hflow/doctor.py
+++ b/src/hflow/doctor.py
@@ -181,6 +181,17 @@ def _check_video_payload(
         )
 
 
+class VideoEncodingUnsupported(ValueError):
+    """No decoder factory in the supported set handles this message encoding.
+
+    A subclass of ``ValueError`` because every existing caller treats
+    resolution failure as a ``ValueError``; the distinct type lets the video
+    check report the true cause, an encoding outside the supported set, a
+    fact about the file paired with the reader, instead of wrapping it in a
+    corruption assertion. #460.
+    """
+
+
 def _resolve_video_decoder(topic: str, channel: Channel, schema: Schema) -> Callable[[bytes], Any]:
     """Resolve either supported encoded-video payload representation."""
     from mcap_protobuf.decoder import DecoderFactory as ProtobufDecoderFactory
@@ -190,7 +201,7 @@ def _resolve_video_decoder(topic: str, channel: Channel, schema: Schema) -> Call
         decoder = factory.decoder_for(channel.message_encoding, schema)
         if decoder is not None:
             return decoder
-    raise ValueError(
+    raise VideoEncodingUnsupported(
         f"video topic {topic!r} has message encoding {channel.message_encoding!r} "
         "that no available decoder handles"
     )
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Both findings hold under mutation, and `read-failed` stays reserved for what it names.

PR #468, opened 2026-09-08
refactor: centralize unusable file path checks

### LINKED ISSUE #467: Four sites hand-roll the same is-a-directory / not-found errno pair, two with byte-identical comments
Four places refuse an unusable input path with the same two-step check and the same three-argument `FileNotFoundError`:

| Site | Lines |
| --- | --- |
| `src/hflow/storage.py` | `:373` / `:375` |
| `src/hflow/storage.py` | `:701` / `:703` |
| `src/hflow/runtime/_deploy.py` | `:180` / `:182` |
| `src/hflow/runtime/_bundle.py` | `:1037` / `:1041` |

Each is:

```python
if <path>.is_dir():
    raise FileNotFoundError(errno.EISDIR, os.strerror(errno.EISDIR), str(<path>))
if not <path>.is_file():
    raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), str(<path>))
```

The duplication is not only in the code. `_deploy.py:176-179` and `_bundle.py:1033-1036` carry byte-identical four-line comments, down to "thirteen more name it elsewhere under `src/hflow/`" and the `#100` reference. When two comments are copies of each other, the code under them usually is too.

## Why this one is worth doing

The three-argument form is a deliberate convention here, not an accident: `str(path)` as the third argument makes `errno` supply the text, so a CLI prints why the path failed rather than just naming it. `storage.py:368-372` explains the choice of `FileNotFoundError` over `IsADirectoryError`, and `#100` and `#144` decided it. That reasoning currently lives in four places and has already drifted: `_bundle.py:1039-1040` has an extra comment the others lack.

One helper means the convention is stated once and the next path check inherits it instead of reinventing it.

## What to do

Add one private helper somewhere both `storage.py` and `runtime/` can import, and call it from all four sites:

```python
def _refuse_unusable_file_path(path: Path) -> None:
    """Raise the errno-carrying FileNotFoundError for a directory or missing file."""
```

Keep the reasoning comment on the helper and delete the four copies. Placement is your call: `storage.py` already owns path handling and both `runtime/` modules import from it, so exporting a private helper from there is one option; a small shared module is another. Say which you picked and why.

## Definition of done

1. One definition, four call sites, no behaviour change: same exception class, same errno, same third argument.
2. The existing tests pass untouched. `test_fetch_directory_says_is_a_directory`, `test_fetch_missing_has_oserror_filename`, `test_fetch_local_directory_says_is_a_dir` and `test_fetch_local_missing_has_oserror_filename` already pin the storage sites, so they are your regression net.
3. Removing the helper's body breaks those tests. Confirm it, and say which of the four sites has no test today, because at least one does not.
4. The reasoning comment survives in one place rather than being deleted along with the copies.

## Not in scope

`packaging.py:1841-1980` has its own `_require_regular_path_below_root` family doing symlink and containment checks. Different question, leave it.

Do not widen this to every `raise FileNotFoundError` in the tree. Only the four sites above run the directory check followed by the existence check as a pair.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Kaileshwar16:
## Summary

Centralizes the repeated unusable-file-path validation used across storage and runtime code.

The new private `_refuse_unusable_file_path()` helper lives in `storage.py`, since that module already owns local path handling and both runtime modules already depend on it.

## Changes

* Added `_refuse_unusable_file_path(path: Path)` in `src/hflow/storage.py`
* Replaced the duplicated directory / missing-file checks in:

  * `LocalStorageRoot.fetch()`
  * `fetch_uri()`
  * `render_deploy_bundle()`
  * `render_bundle()`
* Preserved the existing behavior:

  * directories raise `FileNotFoundError` with `errno.EISDIR`
  * missing files raise `FileNotFoundError` with `errno.ENOENT`
  * the path is still passed as the third argument so `filename` and OS error text remain unchanged
* Kept the reasoning for using `FileNotFoundError` instead of `IsADirectoryError` in one place on the helper

## Validation

Ran:

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Result:

```text
1638 passed, 7 skipped
```

Also confirmed the four existing storage regression tests pass with the helper in place.

As requested in the issue, I temporarily replaced the helper body with `pass` and confirmed all four regression tests fail because `FileNotFoundError` is no longer raised.

I also checked the current test suite for the four call sites. On the current `main`, both runtime sites now have missing-file and directory-path coverage, so the issue note that at least one site is untested appears to be outdated.

Closes #467

### DIFF
--- a/src/hflow/runtime/_bundle.py
+++ b/src/hflow/runtime/_bundle.py
@@ -61,7 +61,6 @@ extracts them to temp files):
    mass failure visible.
 """
 
-import errno
 import hashlib
 import json
 import logging
@@ -1065,15 +1064,7 @@ def render_bundle(config: RuntimeConfig, bundle_dir: Path | str) -> BundlePaths:
     )
     if config.requirements_file is not None:
         requirements_source = Path(config.requirements_file)
-        if requirements_source.is_dir():
-            # FileNotFoundError, not IsADirectoryError, for the reason above.
-            raise FileNotFoundError(
-                errno.EISDIR, os.strerror(errno.EISDIR), str(requirements_source)
-            )
-        if not requirements_source.is_file():
-            raise FileNotFoundError(
-                errno.ENOENT, os.strerror(errno.ENOENT), str(requirements_source)
-            )
+        _refuse_unusable_file_path(requirements_source)
         shutil.copyfile(requirements_source, user_dir / "requirements.txt")
 
     hflow_source = (
--- a/src/hflow/runtime/_bundle.py
+++ b/src/hflow/runtime/_bundle.py
@@ -91,6 +91,7 @@ from hflow.storage import (
     BucketStorageRoot,
     LocalStorageRoot,
     StorageRoot,
+    _refuse_unusable_file_path,
     is_bucket_url,
     parse_storage_root,
 )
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
LGTM, merging.

Both runtime sites do have coverage now, so ignore point 3 of the issue.

The issue undercounted: six sites, not four. `render_bundle` and `render_deploy_bundle` each run the same pair on `config.requirements_file`, and both carried a comment pointing at the explanation this PR moved. Pushed a commit routing them through the helper.

PR #476, opened 2026-09-09
perf: pass channel topics down to the MCAP reader

### LINKED ISSUE #470: Episode.channel() reads unrelated MCAP topics before channel filtering
`Episode.channel()` currently selects a channel by `channel_id`, but that information is not propagated to the underlying MCAP reader as a topic filter.

Because of this, the MCAP reader can yield messages from unrelated topics, and HFlow only removes them afterwards in Python.

For episodes containing large camera/image streams, reading a small state/action channel may therefore cause unnecessary MCAP reads and decoding.

## Where the issue comes from

The read path is roughly:

```text
Episode.channel()
    ↓
PythonMcapEpisodeReader.iter_batches(channel_ids=[...])
    ↓
MCAP.iter_messages(topics=None)
    ↓
messages from all topics are yielded
    ↓
HFlow filters by channel.id
```

### 1. `Episode.channel()` only passes the channel ID

In `src/hflow/episode.py`, `Episode.channel()` resolves the requested channel and calls the reader using its channel ID:

```python
self._reader.iter_batches(
    channel_ids=[info.channel_id],
)
```

At this point HFlow already knows the channel's topic through `info`, but the topic is not passed down.

### 2. `PythonMcapEpisodeReader.iter_batches()` only forwards `topics` to MCAP

In `src/hflow/reader.py`, the underlying MCAP reader is called roughly as:

```python
self._reader.iter_messages(
    topics=list(topics) if topics is not None else None,
    ...
)
```

Since `Episode.channel()` did not provide `topics`, this becomes:

```python
iter_messages(topics=None)
```

### 3. `channel_ids` are filtered only after MCAP yields the message

The channel-ID check happens afterwards in HFlow:

```python
if wanted_channel_ids is not None and channel.id not in wanted_channel_ids:
    continue
```

So the channel filtering is correct from a result perspective, but it happens too late to help the underlying MCAP reader avoid unrelated data.

## Reproduction

I reproduced this using an MCAP containing:

* `/target`: 1 message
* `/camera`: 8 unrelated messages

The messages were made large enough, with a small MCAP chunk size, to keep the streams in separate chunks.

I then instrumented the underlying MCAP reader to record:

1. what `topics` HFlow passes to `iter_messages()`
2. which messages MCAP yields before HFlow applies its `channel_id` filter

Reproduction script:

```python
from collections import Counter
from pathlib import Path

from mcap.writer import CompressionType, Writer
from hflow.episode import Episode

path = Path("/tmp/hflow-channel-filter-probe.mcap")

with path.open("wb") as f:
    writer = Writer(
        f,
        chunk_size=64 * 1024,
        compression=CompressionType.NONE,
    )
    writer.start()

    target = writer.register_channel(
        topic="/target",
        message_encoding="json",
        schema_id=0,
    )

    camera = writer.register_channel(
        topic="/camera",
        message_encoding="json",
        schema_id=0,
    )

    writer.add_message(
        channel_id=target,
        log_time=1,
        publish_time=1,
        data=b"x" * 100_000,
    )

    for i in range(8):
        writer.add_message(
            channel_id=camera,
            log_time=10 + i,
            publish_time=10 + i,
            data=b"y" * 100_000,
        )

    writer.finish()


ep = Episode(path)

hflow_reader = ep._reader
mcap_reader = hflow_reader._reader

original_iter_messages = mcap_reader.iter_messages

calls = []
seen_topics = []


def traced_iter_messages(*args, **kwargs):
    topics = kwargs.get(
        "topics",
        args[0] if args else None,
    )

    calls.append(topics)

    for schema, channel, message in original_iter_messages(*args, **kwargs):
        seen_topics.append(channel.topic)
        yield schema, channel, message


mcap_reader.iter_messages = traced_iter_messages

result = ep.channel("/target")

print("requested channel: /target")
print("messages returned by HFlow:", len(result))
print("topics passed to MCAP:", calls)
print("messages yielded by MCAP before HFlow filter:")
print(Counter(seen_topics))

ep.close()
path.unlink()
```

Observed output:

```text
requested channel: /target
messages returned by HFlow: 1
topics passed to MCAP: [None]
messages yielded by MCAP before HFlow filter:
Counter({'/camera': 8, '/target': 1})
```

So although only `/target` was requested, the underlying MCAP reader yielded all 8 `/camera` messages as well.

HFlow eventually returns the correct result, but the unrelated messages have already crossed the underlying reader boundary.

## Why this matters

This is potentially significant for multimodal robotics episodes, where channel sizes can be very different.

For example:

```text
/joint_states      50 MB
/actions           20 MB
/front_camera       6 GB
/wrist_camera       4 GB
/depth_camera       5 GB
```

Code such as:

```python
joints = episode.channel("/joint_states")
```

should ideally constrain the underlying MCAP read to the relevant topic instead of allowing unrelated camera streams to be yielded and then discarded.

The overhead may become more noticeable when several individual channels are accessed independently.

## Expected behavior

When HFlow already knows the topic associated with the requested channel, it should propagate that information to the MCAP reader so unrelated topics can be filtered as early as possible.

Conceptually, the read should become:

```text
Episode.channel("/target")
    ↓
iter_batches(
    topics=["/target"],
    channel_ids=[target_id],
)
    ↓
MCAP.iter_messages(topics=["/target"])
    ↓
only messages from the relevant topic reach HFlow
    ↓
channel_id filtering preserves exact channel semantics
```

### Option 1: Optimize `Episode.channel()`

Since `Episode.channel()` already has the channel metadata, it could pass both the topic and channel ID:

```python
self._reader.iter_batches(
    topics=[info.topic],
    channel_ids=[info.channel_id],
)
```

This is probably the smallest change.

The channel-ID filter should remain because multiple MCAP channels may share the same topic.

### Option 2: Optimize `PythonMcapEpisodeReader.iter_batches()`

A more general solution could be to derive the corresponding topic(s) whenever `channel_ids` are supplied.

For example, given:

```python
iter_batches(channel_ids=[1, 4])
```

the reader could resolve those channel IDs to their topics and pass those topics to:

```python
self._reader.iter_messages(topics=[...])
```

while still applying the existing channel-ID filter afterwards.

This would benefit any caller using `channel_ids`, rather than only `Episode.channel()`.

## Important correctness case

Topic filtering alone should not replace channel-ID filtering.

MCAP can have multiple channels associated with the same topic, so:

```python
topics=["/target"]
```

may identify a broader set than:

```python
channel_ids=[specific_channel_id]
```

The optimization should therefore use topic filtering to reduce the underlying read and retain the channel-ID check for exact selection.

## Suggested regression test

A regression test could create an MCAP containing two topics in separate chunks:

```text
/target
/camera
```

Then request only `/target` through `Episode.channel()` and instrument/mock the underlying MCAP reader.

The test should verify that MCAP receives:

```python
topics=["/target"]
```

rather than:

```python
topics=None
```

A second test would be useful for two channels sharing the same topic to ensure the existing channel-ID semantics are preserved.

## Expected outcome

After the fix:

```python
episode.channel("/target")
```

should still return exactly the same channel data, but the underlying MCAP call should be constrained to the relevant topic instead of reading unrelated topics and discarding them afterwards.

mulkakhileshmj:
Closes #470.

## Problem

`Episode.channel()` selects a channel by `channel_id`, but that never became a topic filter for the underlying MCAP reader, so `iter_messages(topics=None)` yielded every stream in the file and HFlow discarded the unrelated ones in Python. Reading a small state or action channel could decompress gigabytes of camera data for nothing.

## Fix

I took Option 2 from the issue: `PythonMcapEpisodeReader.iter_batches()` now derives the topics for the requested `channel_ids` from the file summary and passes them to `iter_messages()`, so the seeking reader can skip unrelated chunks. This covers every `channel_ids` caller, including `Episode.iter_decoded_batches()`, not just `Episode.channel()`.

Correctness cases from the issue are preserved:

- The channel-id filter is unchanged. Several channels can share one topic, and topic filtering only narrows the read, it never selects.
- Caller-supplied `topics` are never overridden.
- A file with no summary section, or all-unknown channel ids, falls back to the unconstrained read, preserving current behavior.

## Coverage

Two tests added to `tests/test_multichannel.py`, tracing at the mcap library boundary:

- Two topics in separate chunks: requesting `/target` passes `topics=["/target"]` down and only `/target` messages cross the boundary, with identical returned data.
- Two channels sharing one topic (the file's existing fixture): the shared topic is passed down, both channels are yielded by mcap, and exactly the requested channel's payloads come back, pinning that channel-id selection still wins.

## Validation

```
uv sync --locked
uv run pytest -q           # 1713 passed, 6 skipped
```

Mutation check: with `src/hflow/reader.py` reverted to main, both new tests fail; restored, they pass. The touched test file also passes under Python 3.11.

### DIFF
--- a/src/hflow/reader.py
+++ b/src/hflow/reader.py
@@ -246,8 +246,12 @@ class PythonMcapEpisodeReader:
             try:
                 known_channels = self.channels()
             except ValueError:
-                # No summary section to derive topics from (unindexed or
-                # truncated file); fall back to the unconstrained read.
+                # No summary section to derive topics from: a legitimately
+                # unindexed file, still readable by linear scan, so fall back
+                # to the unconstrained read. A truncated file is not this
+                # case -- channels() raises mcap's RecordLengthLimitExceeded,
+                # and the same error surfaces from iter_messages below with
+                # or without this catch, so damage stays loud.
                 known_channels = {}
             derived_topics = sorted(
                 {
--- a/src/hflow/reader.py
+++ b/src/hflow/reader.py
@@ -236,6 +236,28 @@ class PythonMcapEpisodeReader:
         batch_max_bytes: int = DEFAULT_BATCH_MAX_BYTES,
     ) -> Iterator[MessageBatch]:
         wanted_channel_ids = frozenset(channel_ids) if channel_ids is not None else None
+        if topics is None and wanted_channel_ids is not None:
+            # Constrain the underlying read to the topics the requested
+            # channels live on, so the MCAP reader can skip unrelated streams
+            # (e.g. multi-gigabyte camera topics) instead of yielding messages
+            # that would be discarded below. Topic filtering alone is not
+            # exact -- several channels may share one topic -- so the
+            # channel-id filter in the loop still applies.
+            try:
+                known_channels = self.channels()
+            except ValueError:
+                # No summary section to derive topics from (unindexed or
+                # truncated file); fall back to the unconstrained read.
+                known_channels = {}
+            derived_topics = sorted(
+                {
+                    known_channels[channel_id].topic
+                    for channel_id in wanted_channel_ids
+                    if channel_id in known_channels
+                }
+            )
+            if derived_topics:
+                topics = derived_topics
         topics_by_channel_id: dict[int, str] = {}
         pending_log_times: dict[int, list[int]] = {}
         pending_publish_times: dict[int, list[int]] = {}
--- a/tests/test_multichannel.py
+++ b/tests/test_multichannel.py
@@ -19,7 +19,7 @@ from mcap.writer import Writer as StockWriter
 from hflow.doctor import diagnose
 from hflow.episode import Episode
 from hflow.format import METADATA_RECORD_EPISODE
-from hflow.reader import PythonMcapEpisodeReader, open_reader
+from hflow.reader import EpisodeReader, PythonMcapEpisodeReader, open_reader
 from hflow.transform import write_canonical_episode
 
 SHARED_TOPIC = "/status"
@@ -124,12 +124,45 @@ def test_episode_addresses_channels_by_id(dual_channel_source: Path) -> None:
         ]
 
 
+def _write_two_topic_mcap(
+    path: Path,
+    target_payloads: list[bytes],
+    camera_payloads: list[bytes],
+    **writer_kwargs: Any,
+) -> tuple[int, int]:
+    """Write a small ``/target`` channel and a large ``/camera`` channel.
+
+    The default small chunk size keeps the two streams in separate chunks,
+    the layout where an early topic filter can skip unrelated data. Returns
+    the two channel ids in registration order.
+    """
+    writer_kwargs.setdefault("chunk_size", 1024)
+    with path.open("wb") as stream:
+        writer = StockWriter(stream, compression=CompressionType.NONE, **writer_kwargs)
+        writer.start(profile="", library="test")
+        target_channel_id = writer.register_channel(
+            topic="/target", message_encoding="json", schema_id=0
+        )
+        camera_channel_id = writer.register_channel(
+            topic="/camera", message_encoding="json", schema_id=0
+        )
+        for index, payload in enumerate(target_payloads):
+            writer.add_message(
+                target_channel_id, log_time=1 + index, data=payload, publish_time=1 + index
+            )
+        for index, payload in enumerate(camera_payloads):
+            writer.add_message(
+                camera_channel_id, log_time=10 + index, data=payload, publish_time=10 + index
+            )
+        writer.finish()
+    return target_channel_id, camera_channel_id
+
+
 def _trace_mcap_iter_messages(
-    episode: Episode, monkeypatch: pytest.MonkeyPatch
+    hflow_reader: EpisodeReader, monkeypatch: pytest.MonkeyPatch
 ) -> tuple[list[Any], list[str]]:
     """Record the ``topics`` argument HFlow passes to the stock mcap reader
     and the topic of every message that crosses that library boundary."""
-    hflow_reader = episode._reader
     assert isinstance(hflow_reader, PythonMcapEpisodeReader)
     mcap_reader = hflow_reader._reader
     original_iter_messages = mcap_reader.iter_messages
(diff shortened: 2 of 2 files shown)


kstonekuan (changes requested):
The layer is right and #475 does not cover it.

But rebased on merged main your tests no longer reach the derivation: deleting the whole block leaves all 8 passing, because they go through `Episode.channel()`, which now passes topics explicitly. Needs a case reaching `iter_batches` with `channel_ids` set and `topics` left `None`.

Two smaller ones: the `except ValueError` fallback is untested, and check whether `channels()` can raise something wider on a truncated file.

mulkakhileshmj:
Thanks for the review, and for catching the sequencing with #475. You are right: once it passed the topic explicitly on the `channel()` path, my tests stopped reaching the derivation. Fixed as suggested, plus the two smaller points, measured rather than assumed.

**The missing pin.** A new test drives `Episode.iter_decoded_batches(channel_ids=...)`, the remaining caller that reaches the reader with `topics` left `None`. It asserts the derived topic is what the mcap reader receives and that only the requested stream crosses the library boundary. Mutation checked: with the derivation block deleted, this test fails; restored, it passes. The two `Episode.channel()` tests stay unchanged, agreed they now pin #475's forwarding.

**The no-summary fallback is reachable, and now exercised.** A file written with `use_chunking=False, use_statistics=False, use_summary_offsets=False, repeat_channels=False, repeat_schemas=False` has no summary section at all: `get_summary()` returns `None`, `channels()` raises the ValueError, and a channel id read falls back to the unconstrained scan while still returning only the requested channel. The new test pins exactly that.

**Truncated files, measured before deciding the catch width.** On a file cut mid-data or inside the summary region, `channels()` raises mcap's `RecordLengthLimitExceeded`, not ValueError, and `iter_messages` on the same file raises the identical error whether or not the derivation catches it. Widening the catch would only move the same failure two lines down, so it stays narrow and the code comment now records that reasoning.

One side finding from the same probe, reported rather than acted on: a chunked file whose summary carries chunk indexes but no repeated channel records (`repeat_channels=False` with chunking on) makes the stock mcap `SeekingReader` itself raise `KeyError` from `summary.channels` inside `iter_messages`. That happens with or without this change, so it is mcap behavior on a degenerate writer configuration, not something this PR touches. Happy to file it separately if it is worth tracking.

Gate rerun on the updated branch: ruff, format, and ty clean; full suite 1717 passed, 6 skipped; the multichannel file also passes on Python 3.11. The three fixture writers now share one helper, so the file builds its two-topic layouts in one place.

kstonekuan (approved):
LGTM, merging.

Pushed one commit: dropping the `topics is None` half of the condition left the whole suite green, because `Episode.channel()` passes both and the derived topic equals the explicit one. Asking for `/target` while naming a channel on `/camera` separates them.

Please do file the `SeekingReader` `KeyError` you hit.

Issue #474, opened 2026-09-09
check lanes re-certify a decayed canonical episode: post-sync reads never validate chunk CRCs


Sagar-024:
### Version or commit

ff1e218 (current main)

### Environment

Ubuntu 24.04 (WSL2), Python 3.12, x86_64

### Summary

Every post-sync read of a canonical episode runs with chunk CRC validation off. `Episode._reader` opens the file through `open_reader(path)` (src/hflow/episode.py:311), whose default is `validate_crcs=False` (src/hflow/reader.py:296), and every check consumes that reader. So the META lane, the relabel lane, and the online re-check flow re-read a canonical episode that has decayed on disk since sync and re-certify it: checks stamp `measured` findings over bytes that fail the file's own integrity stamp. The reader docstring justifies the default with "already identified by content hash", but that hash proves the source at sync time, not the state of the file at re-read time. #426 and #462 established the opposite rule for sources: file existence is not proof the bytes are good.

### Measured repro

On current main. Flip the stored `uncompressed_crc` of the first chunk of a published canonical episode (real-world header rot: the payload decompresses fine, the bytes no longer match the file's own stamp), then run the production re-check lane:

```python
import io, sys
from pathlib import Path
sys.path.insert(0, "tests")
import hflow
from hflow.testing import SyntheticEpisodeSpec, synthesize_episode
from mcap.reader import make_reader
from mcap.stream_reader import CRCValidationError
from hflow.steps import Stage

tmp = Path("/tmp/check-crc-repro")
tmp.mkdir(parents=True)

source = synthesize_episode(tmp / "source.mcap", SyntheticEpisodeSpec(duration_s=4.0))
app = hflow.App("emi", data_root=tmp / "data")
report = app.process(source, verbose=False)
clean = [(c.check.name, c.status.value) for c in report.checks]
print("healthy canonical:", clean)

canonical = sorted(tmp.rglob("*.canonical.mcap"))[0]
data = bytearray(canonical.read_bytes())
summary = make_reader(io.BytesIO(bytes(data))).get_summary()
crc_offset = summary.chunk_indexes[0].chunk_start_offset + 33  # chunk: opcode(1) len(8) start(8) end(8) usize(8) crc(4)
data[crc_offset] ^= 0x01
canonical.write_bytes(bytes(data))

# A CRC-validating read refuses the file:
with canonical.open("rb") as stream:
    try:
[...]

Issue #472, opened 2026-09-09
verify snapshot accepts directories the exporter would refuse: no format identity check at the verifier boundary


Sagar-024:
### Version or commit

ff1e218 (current main)

### Environment

Ubuntu-derived (WSL2), Python 3.12, x86_64

### Summary

`export_dataset_snapshot` pins the marker's format identity before doing anything: it refuses when `format` is not `hflow-dataset-snapshot` or `format_version` is not `1` (src/hflow/snapshot.py:618-625, landed in #401). `verify_dataset_snapshot`, the receipt's promised consumer, never performs the same check. It reads any parseable JSON object that carries an `integrity` dict and verifies whatever paths that object names, so a directory produced by a different tool is accepted as a delivered HFlow snapshot and can be certified clean.

### Measured repro

A directory written by a foreign tool, no `format` key, no `format_version` key:

```python
import hashlib, json, tempfile
from pathlib import Path
from hflow.snapshot import verify_dataset_snapshot

with tempfile.TemporaryDirectory() as temporary_directory:
    foreign = Path(temporary_directory) / "some-other-tools-output"
    foreign.mkdir()
    payload = b"not-a-hflow-snapshot-at-all"
    (foreign / "data.parquet").write_bytes(payload)
    (foreign / "format.json").write_text(json.dumps({
        "producer": "not-hflow",
        "integrity": {"tables": {"data": {
            "path": "data.parquet",
            "size_bytes": len(payload),
            "sha256": hashlib.sha256(payload).hexdigest(),
        }}},
    }))

    report = verify_dataset_snapshot(foreign)
    print(report.status, report.ok)
    # VerificationStatus.OK True -- a non-HFlow directory is certified
```

The same verifier verifies a hypothetical `format_version: 2` marker as if its receipts carried v1 semantics, because the version field is never consulted. The exporter refuses both inputs; the asymmetry between :618 and :806 is the bug.

### Why it matters

`hflow verify snapshot` exists so a recipient can trust a delivered directory. Exit 0 on a directory the exporter would have refused is a false pass of the same species the LeRobot verifier avoids: `verify_lerobot_import` raises on unsupported `schema_version` (importers/lerobot_verify.py:67-71), so the two verifiers disagree on the same contract point. The docs tell recipients to run this command as the trust check; its clean answer currently means "the receipt matched", not "this is an HFlow snapshot and the receipt matched".

### Proposed fix

[...]

PR #483, opened 2026-09-10
fix(snapshot): enforce the receipt inventory content_id at verify time

### LINKED ISSUE #473: verify snapshot never checks integrity.content_id, so the documented deleted-member guarantee is unenforced
### Version or commit

ff1e218 (current main)

### Environment

Ubuntu-derived (WSL2), Python 3.12, x86_64

### Summary

The integrity receipt added in #401 documents `integrity.content_id` as "Full SHA-256 of the normalized inventory (all table and asset receipts, sorted by path), so a deleted member is visible even when every remaining file still matches" (docs/how-to/export-dataset-snapshot.md:77). `verify_dataset_snapshot`, the consumer that documentation promises, never reads the field: it iterates `integrity.tables` and `integrity.assets` (src/hflow/snapshot.py:880-884) and stops. The deleted-member guarantee is written in the docs and enforced nowhere.

### Measured repro

Delete one table entry from the receipt and delete that table's file. Every remaining file still matches, and the deleted member is invisible:

```python
import json, tempfile
from pathlib import Path
import sys
sys.path.insert(0, "tests")
from test_dataset_snapshot import _append_snapshot_episode
import hflow
from hflow.catalog import Catalog

    tmp = Path(temporary_directory)
    catalog = Catalog(tmp / "catalog")
    episode_id, _ = _append_snapshot_episode(catalog, tmp, name="t", score=1.0, with_media=False)
    manifest = tmp / "manifest.parquet"
    hflow.curate(catalog.location,
                 f"SELECT episode_id FROM episodes WHERE episode_id = '{episode_id}'",
                 output=manifest)
    out = tmp / "snap"
    hflow.export_dataset_snapshot(catalog.location, out, manifest=manifest, media_mode="references")

    marker_path = out / "format.json"
    marker = json.loads(marker_path.read_text())
    entry = marker["integrity"]["tables"].pop("measurements")
    marker_path.write_text(json.dumps(marker, indent=2, sort_keys=True) + "\n")
    (out / entry["path"]).unlink()

    report = verify_dataset_snapshot(out)
    print(report.status, report.ok, report.findings)
    # VerificationStatus.OK True [] -- docs say this deletion must be visible
```

No hash needs to be recomputed by an attacker; removing an entry and its file produces a clean exit 0. The "not a tamper defence" caveat in the docs covers recipients who recompute the receipt to match edited bytes; it does not cover this case, which is exactly the one `content_id` was added to catch.

### Why it matters

The receipt's own table documents this field's purpose, and #428's verifier was reviewed against that table. Everything else the receipt records is checked; the one field whose stated job is catching receipt-level deletion is not. A verifier that answers "clean" on a delivery whose member set silently shrank gives the deleted member exactly the false pass the field exists to prevent.

### Proposed fix

Recompute the inventory content id from the receipt entries under the same normalization #401 publishes (sorted by path, tables then assets), compare against `integrity.content_id`, and report a finding on `format.json` when they differ. Pre-#401 snapshots keep the `no-receipt` path. One test pins the deleted-member case; the normalization is already a helper on the export side (`_inventory_content_id`, src/hflow/snapshot.py:134), so the check is small.

Related: #469 and the format-identity asymmetry are separate boundary holes in the same verifier; this issue covers only the unenforced `content_id`.

Sagar-024:
# fix(snapshot): enforce the receipt inventory content_id at verify time

Closes #473.

## What changed

`verify_dataset_snapshot` now enforces the `integrity.content_id` guarantee documented at docs/how-to/export-dataset-snapshot.md:77. After the receipt is parsed and before the per-file loop, the verifier recomputes the inventory hash from the receipt's own table and asset entries (the identical normalization the exporter uses at snapshot.py:194, via the shared `_inventory_content_id` helper) and compares it against the stored value.

## The exit-2 decision

A `content_id` that disagrees with the recomputed inventory means the marker is internally inconsistent, which points at tampering or a truncated write, not at damaged bytes. Damaged bytes are what findings and exit 1 are for; an inconsistent receipt is unreadable input. The gate therefore raises through the same `ValueError` the unparsable-marker path uses (snapshot.py:862), so the CLI maps it to exit 2 with no new exit code. A DAMAGED finding was considered and rejected: findings are evidence about delivered files, and the failure here is in the receipt itself, which per-file findings cannot express without inventing a new reason code for a non-file.

The same handling covers a missing or malformed `content_id` on a marker that carries an `integrity` block: one gate, one behavior, one exit code. Pre-#401 markers with no `integrity` key keep the existing `no-receipt` UNVERIFIABLE path untouched.

## Test shape, and why

- Pinning test: a receipt entry is removed from the marker and its file is deleted, then verify must raise (and the CLI must map to exit 2). This is the case only the stored `content_id` catches: the surviving entries agree with each other, so every per-file check passes; the comparison against the hash of the original set is the only witness.
- Negative control (as asked on the issue): the file is deleted but the receipt entry is kept. This exercises the ordinary `missing` path and passes without the fix; under the mutation run it stayed green while the pinning test went red, proving the two tests pin different behaviors.

## Related issues, separate PRs

- #472 (format-identity gate at the verifier boundary): separate one-line gate, separate PR.
- #474 (post-sync reads never validate chunk CRCs): the reader/CRC lane, not touched here.

## Gates

15 passed in tests/test_snapshot_verify.py (13 existing + pinning test + negative control), 31 passed across the verify and dataset-snapshot suites, ruff check / ruff format / ty clean, zero non-ASCII and zero em dashes in changed files. Mutation run: inventory comparison neutered, pinning test red, negative control green, restored, all green.

### DIFF
--- a/src/hflow/snapshot.py
+++ b/src/hflow/snapshot.py
@@ -828,7 +828,11 @@ def verify_dataset_snapshot(
 
     Unreadable input (a missing directory, a missing or unparsable
     ``format.json``) raises instead: that is not a finding about a delivered
-    snapshot, it is the wrong input entirely.
+    snapshot, it is the wrong input entirely. The same applies to a receipt
+    that is internally inconsistent with itself: a missing or malformed
+    ``content_id``, or a recomputed inventory hash that disagrees with the
+    stored one, means the receipt no longer describes the delivered set
+    (tampering or a truncated write), which is exit 2, not damaged bytes.
 
     Extra files under ``assets/`` that the receipt does not name are ignored:
     the receipt covers what was exported, not everything a recipient may add.
@@ -882,6 +886,29 @@ def verify_dataset_snapshot(
         *integrity.get("tables", {}).values(),
         *integrity.get("assets", []),
     ]
+
+    # The deleted-member gate (#473): when a receipt entry and its file are
+    # both gone, the surviving entries are self-consistent and every per-file
+    # check passes; only the stored inventory hash, computed over the original
+    # set, differs from the hash of what remains. Recompute it exactly as the
+    # exporter did (:194) and refuse a receipt that no longer describes the
+    # delivered set. Internal inconsistency is unreadable input, not damage,
+    # so it raises to exit 2 like an unparsable marker rather than reporting
+    # findings.
+    stored_content_id = integrity.get("content_id")
+    if not isinstance(stored_content_id, str) or not stored_content_id:
+        raise ValueError(
+            "format.json integrity receipt carries no usable content_id; "
+            "the delivered member set cannot be checked against the receipt"
+        )
+    recomputed_content_id = _inventory_content_id(receipt_entries)
+    if recomputed_content_id != stored_content_id:
+        raise ValueError(
+            "format.json integrity receipt is internally inconsistent: recomputed "
+            f"inventory content_id {recomputed_content_id!r} != stored "
+            f"{stored_content_id!r}; the receipt no longer describes the delivered set"
+        )
+
     for entry in receipt_entries:
         relative_path = str(entry["path"])
         delivered_path = resolved_directory / relative_path
--- a/tests/test_snapshot_verify.py
+++ b/tests/test_snapshot_verify.py
@@ -89,6 +89,49 @@ def test_missing_file_reports_missing_alone(tmp_path: Path) -> None:
     assert marker["integrity"]["tables"]["samples"]["path"] in report.findings[0].uri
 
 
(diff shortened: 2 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Pushed one commit: the missing-`content_id` branch had no case, so four shapes through the API and the CLI exit code.

`integrity` and `content_id` arrived in the same commit (#401), so no snapshot we wrote carries one without the other. That is what makes it unreadable input rather than damage.

Sagar-024:
Thanks @kstonekuan  .

#472 is next then i will see what i can do on   (Egocentric-10K or Egocentric-100K on Hugging Face)

PR #485, opened 2026-09-10
perf: avoid repeated keyframe payload scans

### LINKED ISSUE #480: keyframe_interval() repeatedly rescans camera payloads, causing O(C²) keyframe parsing
`keyframe_interval()` appears to scan the same camera payloads multiple times.

The expensive part is `_keyframe_indices()`, which walks every raw payload in a camera channel and checks whether it starts with a keyframe.

Inside `keyframe_interval()`, that scan is already done once for the current camera:

```python
inter = _KeyframeIntervalPerCamera(
    ...
    keyframe_indices=_keyframe_indices(channel),
)
```

But while still inside the per-camera loop, it also calls:

```python
_keyframe_interval_keys(
    episode,
    cameras=selected_cameras,
)
```

`_keyframe_interval_keys()` loops over all selected cameras and calls `_keyframe_indices()` for each of them again.

So with `C` cameras, the number of full camera scans is roughly:

```text
C + C²
```

instead of ideally:

```text
C
```

### Reproduction

I instrumented `_payload_starts_a_keyframe()` to count how many payloads are inspected.

The probe used:

* 3 camera channels
* 5 payloads per camera

If each camera were scanned once, that would be:

```text
3 × 5 = 15 payload checks
```

Observed output:

```text
Payload checks per camera:
  cam0: 20
  cam1: 20
  cam2: 20

Total payload checks: 60
Ideal if each camera were scanned once: 15
```

So the 3-camera case performs 12 full camera scans:

```text
3 + 3² = 12
```

and therefore:

```text
12 × 5 = 60 payload checks
```

### Reproduction script

```python
from collections import Counter
from types import SimpleNamespace

import numpy as np
import hflow.checks as checks


camera_names = ["/cam1", "/cam2", "/cam3"]

channels = {}

for i, topic in enumerate(camera_names):
    payloads = [
        f"cam{i}:frame{j}".encode()
        for j in range(5)
    ]

    channels[topic] = SimpleNamespace(
        topic=topic,
        timestamps=np.arange(5, dtype=np.int64) * 1_000_000_000,
        raw=payloads,
    )


class FakeEpisode:
    @property
    def cameras(self):
        return camera_names

    def channel(self, topic):
        return channels[topic]


episode = FakeEpisode()

original = checks._payload_starts_a_keyframe

payload_checks = Counter()


def traced_payload_check(payload):
    camera = payload.split(b":", 1)[0].decode()
    payload_checks[camera] += 1

    # Only counting scans for this probe.
    return False


checks._payload_starts_a_keyframe = traced_payload_check

try:
    checks.keyframe_interval(episode)
finally:
    checks._payload_starts_a_keyframe = original


print("Payload checks per camera:")
for camera, count in sorted(payload_checks.items()):
    print(f"  {camera}: {count}")

print()
print("Total payload checks:", sum(payload_checks.values()))
print("Ideal if each camera were scanned once:", 3 * 5)
```

### Possible fix

The keyframe information for each selected camera could be computed once and reused both for the per-camera intermediate and for determining the measurement keys.

For example, build the intermediates first:

```python
intermediates = {}

for topic in selected_cameras:
    channel = episode.channel(topic)

    intermediates[topic] = _KeyframeIntervalPerCamera(
        frame_count=channel.timestamps.size,
        keyframe_indices=_keyframe_indices(channel),
    )
```

Then derive the required keys from those cached intermediates instead of calling `_keyframe_indices()` again through `_keyframe_interval_keys()`.

Even just moving the `_keyframe_interval_keys()` call outside the outer camera loop would reduce the repeated scans from roughly:

```text
C² + C
```

to:

```text
2C
```

but reusing the already-computed keyframe indices should allow it to stay at roughly one scan per camera.

The output of `keyframe_interval()` should remain unchanged; this would only avoid repeatedly parsing the same camera payloads.

Kaileshwar16:
Fixes #480.

`keyframe_interval()` was rescanning camera payloads multiple times while determining both per-camera measurements and the overall measurement key set.

With `C` selected cameras, the old path performed roughly:

```text
C + C²
```

full keyframe scans.

This change computes the per-camera keyframe intermediates once and reuses them, reducing the expensive payload parsing to roughly:

```text
C
```

while keeping `_keyframe_interval_keys()` as the single source of truth for the measurement key set.

## What changed

* Added a helper to compute `_KeyframeIntervalPerCamera` once per camera.
* `keyframe_interval()` now:

  * loads each unique selected camera once
  * stores only the timestamps and small keyframe intermediate
  * reuses those intermediates when determining measurement keys
* `_keyframe_interval_keys()` can consume precomputed intermediates from `keyframe_interval()`.
* Standalone callers of `_keyframe_interval_keys()` keep the existing behavior and compute the information themselves.
* Preserved the empty-camera behavior and App pre-decode supersession behavior.

## Performance regression test

Added a test that counts calls to `_payload_starts_a_keyframe()` and `Episode.channel()`.

For 3 cameras with 5 payloads each:

```text
Before:
60 payload checks

After:
15 payload checks
```

So each payload is inspected once instead of repeatedly rescanning all selected cameras.

The test also covers:

* all cameras
* no cameras
* empty camera
* one camera
* non-empty + empty cameras
* repeated camera selection

## Behavior

The emitted measurements remain unchanged, including the empty-camera case where only `scanned_frame_count` is emitted.

The App supersession path is also covered to ensure `_keyframe_interval_keys()` still produces the same key set.

## Validation

```text
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Result:

```text
1720 passed, 7 skipped
```

### DIFF
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -1837,10 +1837,16 @@ def _keyframe_indices(channel: ChannelData) -> tuple[int, ...]:
 
 
 def _keyframe_interval_intermediates(channel: ChannelData) -> _KeyframeIntervalPerCamera:
-    frame_count = channel.timestamps.size
+    """One camera's keyframe scan, computed once and shared by the body and
+    the key-set fact (#480).
+
+    No empty-channel shortcut: ``_keyframe_indices`` on a channel with no
+    messages already returns ``()`` without inspecting a payload, so guarding
+    it would only add a branch that reads as if it saved something.
+    """
     return _KeyframeIntervalPerCamera(
-        frame_count=frame_count,
-        keyframe_indices=_keyframe_indices(channel) if frame_count else (),
+        frame_count=channel.timestamps.size,
+        keyframe_indices=_keyframe_indices(channel),
     )
 
 
@@ -1940,11 +1946,15 @@ def keyframe_interval(episode: Episode, *, cameras: Sequence[str] | None = None)
     selected_cameras = list(cameras) if cameras is not None else episode.cameras
     timestamps_by_camera: dict[str, np.ndarray] = {}
     intermediates_by_camera: dict[str, _KeyframeIntervalPerCamera] = {}
+    # dict.fromkeys, not set: a repeated camera is loaded once but the
+    # measurement loop below still walks selected_cameras in the order the
+    # caller gave. Only the timestamps and the small keyframe struct are kept,
+    # so the channel's raw payloads are released as the loop moves on rather
+    # than all C being held at once.
     for topic in dict.fromkeys(selected_cameras):
         channel = episode.channel(topic)
         timestamps_by_camera[topic] = channel.timestamps
         intermediates_by_camera[topic] = _keyframe_interval_intermediates(channel)
-        del channel
     keys = sorted(
         _keyframe_interval_keys(
             episode,
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -16,7 +16,7 @@ rather than a shared ``message_count``.
 """
 
 import hashlib
-from collections.abc import Callable, Sequence
+from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
LGTM, merging.

Dropped the empty-channel shortcut: `_keyframe_indices` already returns `()` without inspecting a payload, so the guard saved nothing and deleting it left all 80 tests green. Also wrote down why `dict.fromkeys` and not `set`.

Your `batch_max_messages=2` case is what catches collapsing the batch end bound.

Issue #469, opened 2026-09-08
[Bug]: Snapshot verify can read outside the handed directory via receipt paths


VARUN3WARE:
### Version or commit

7df224adf0ebcd18cd18e1323a1abd0207c1131c

### Environment

Ubuntu 24.04, Python 3.11, x86_64

### Minimal reproduction

```python
from pathlib import Path
import json, tempfile, hashlib
from hflow.snapshot import verify_dataset_snapshot

def sha256_hex(path: Path) -> str:
    digest = hashlib.sha256()
    digest.update(path.read_bytes())
    return digest.hexdigest()

with tempfile.TemporaryDirectory() as temporary_directory:
    root = Path(temporary_directory)
    snap = root / "snap"
    outside = root / "outside"
    snap.mkdir()
    outside.mkdir()
    secret = outside / "secret.bin"
    secret.write_bytes(b"secret-bytes")

    (snap / "format.json").write_text(
        json.dumps(
            {
                "format": "hflow-dataset-snapshot",
                "format_version": 1,
                "media_mode": "references",
                "media_uri_base": None,
                "tables": ["samples.parquet"],
                "integrity": {
                    "tables": {
                        "samples": {
                            "path": "../outside/secret.bin",
                            "size_bytes": secret.stat().st_size,
                            "sha256": sha256_hex(secret),
                        }
                    },
                    "assets": [],
                    "content_id": "00" * 32,
                },
            }
        )
    )

    report = verify_dataset_snapshot(snap)
    print(report.status, report.ok, report.findings)

### Expected behavior

Verify is scoped to the snapshot directory handed in. Docs say recorded paths are “joined only onto the handed directory” (`docs/how-to/export-dataset-snapshot.md`).
A receipt `path` with `..`, an absolute filesystem path, or a drive/UNC form should be refused as unreadable input (exit `2` / `ValueError`) — same spirit as `_validated_relative_key` in `src/hflow/storage.py` — not accepted by hashing bytes outside the delivery.
A normal relative path under that root should still pass when size and sha256 match.

### Actual behavior

Implementation is a bare join in `verify_dataset_snapshot`:
```python
delivered_path = resolved_directory / relative_path
```
(src/hflow/snapshot.py, integrity tables/assets loop)

On the reproduction above:
```
status=ok  ok=True  findings=[]
```
[...]

PR #486, opened 2026-09-10
fix(snapshot): refuse markers whose format identity the exporter would refuse

### LINKED ISSUE #472: verify snapshot accepts directories the exporter would refuse: no format identity check at the verifier boundary

Sagar-024:
# fix(snapshot): refuse markers whose format identity the exporter would refuse

Closes #472.

## What changed

`verify_dataset_snapshot` now pins the marker's format identity at the boundary, immediately after the marker is parsed and before any receipt logic runs. The comparison is deliberately identical to the exporter's replace guard (snapshot.py:618-625) and the writer (:772-773): `format` must equal `hflow-dataset-snapshot` and `format_version` must equal `"1"`, so there is exactly one answer to "is this an HFlow snapshot".

A marker without that identity, or with a future version, raises through the same `ValueError` the unparsable-marker path uses, so the CLI maps it to exit 2 with no new exit code. A foreign directory that happens to carry an integrity-shaped key can no longer be certified; on this base the repro from the issue returned `ok: True` for exactly that input.

## The maintainer's two notes, applied

- **Type strictness**: the issue repro wrote `"format_version": 1` as a JSON int, and the writer records the string `"1"`. The comparison inherits the export guard's strictness, so the int is refused, and per the request the error message says exactly why: it names the found values, the expected values, and states that the version's type is checked because the writer records it as a string.
- **Precedent**: version `1` is the only version there has ever been; refusing anything else at the same boundary sets the version-2 precedent (a future verifier decides what a newer marker can partly read, and that choice gets made then, deliberately).

## Pre-#401 probe result

The very first exporter (402688e, #132) already wrote both fields unconditionally (writer lines :645-646 of that revision), and every revision since has kept them. So no marker the exporter has ever produced lacks the identity, no valid historical input changes answer, and the pre-#401 no-integrity marker (identity present, `integrity` absent) still returns the `no-receipt` UNVERIFIABLE path, not a raise.

## Boundary family and sequencing

These three sit on one function's input boundary, and the maintainer asked for one change rather than three where possible:

- #483 (merged): the inventory `content_id` gate, built to the maintainer's spec on the issue; this branch is rebased onto it and both gates pass the same snapshots.
- This PR: the format-identity gate. Independent one-gate diff; if the maintainer prefers a single combined boundary PR, this commit rebases cleanly onto #483's branch since the gates sit at different points of the same function.
- #469 (VARUN3WARE): receipt-path containment, the last sibling, remains with him.

## Gates

15 passed in tests/test_snapshot_verify.py (13 existing + the foreign-marker test + the version test covering both the future-version and int-1 refusals), 31 passed across the verify and dataset-snapshot suites, ruff check / ruff format / ty clean. Mutation run: with the identity gate disabled, exactly the two new raise tests go red while the clean, pre-#401, and moved-root tests stay green; restored, all green. Zero non-ASCII and zero em dashes in changed files.

### DIFF
--- a/src/hflow/snapshot.py
+++ b/src/hflow/snapshot.py
@@ -126,6 +126,25 @@ def _sha256_hex(path: Path) -> str:
     return digest.hexdigest()
 
 
+def _marker_identifies_dataset_snapshot(format_marker: dict) -> bool:
+    """Whether ``format_marker`` claims to be a snapshot this version handles.
+
+    Two callers ask this and must never disagree: the exporter deciding
+    whether a destination is one of ours to replace, and the verifier
+    deciding whether a directory is one of ours to certify (#472). If the
+    verifier were the looser of the two, exit 0 would mean "some directory
+    with an integrity-shaped key matched".
+
+    Deliberately strict about the version's type. The writer records the
+    string ``"1"``, so a JSON number ``1`` is a different value and is
+    refused. Callers phrase their own message; only the predicate is shared.
+    """
+    return (
+        format_marker.get("format") == DATASET_SNAPSHOT_FORMAT_NAME
+        and format_marker.get("format_version") == DATASET_SNAPSHOT_FORMAT_VERSION
+    )
+
+
 def _file_integrity_record(relative_path: str, absolute_path: Path) -> dict[str, str | int]:
     """Receipt for one delivered snapshot file (table or copied asset)."""
     return {
@@ -614,10 +633,7 @@ def _parse_dataset_snapshot_destination(
             f"snapshot export destination {output_directory} cannot be replaced because "
             f"{_FORMAT_MARKER_FILE_NAME} is not a JSON object"
         )
-    if (
-        format_marker.get("format") != DATASET_SNAPSHOT_FORMAT_NAME
-        or format_marker.get("format_version") != DATASET_SNAPSHOT_FORMAT_VERSION
-    ):
+    if not _marker_identifies_dataset_snapshot(format_marker):
         raise ValueError(
             f"snapshot export destination {output_directory} cannot be replaced because "
             f"{_FORMAT_MARKER_FILE_NAME} does not identify supported "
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
LGTM, merging.

Pushed two commits. The gate repeated the exporter's replace guard verbatim, so both now share `_marker_identifies_dataset_snapshot` with their own messages. Then the tests that makes safe: dropping the format-name comparison and disabling the exporter guard both passed before, the second because that guard had no test anywhere.

PR #490, opened 2026-09-10
fix(verification): type finding reasons as an enum

### LINKED ISSUE #488: VerificationFinding.reason is a bare str with four closed values, ten lines below a StrEnum doing the same job
`VerificationFinding.reason` is typed `str` (`src/hflow/verification.py:43`) but its value set is closed: four module-level constants right above it in the same file.

```python
REASON_MISSING = "missing"                        # :23
REASON_SIZE_MISMATCH = "size-mismatch"            # :24
REASON_CONTENT_ID_MISMATCH = "content-id-mismatch"  # :25
REASON_NO_RECEIPT = "no-receipt"                  # :27
```

All 7 construction sites in the tree pass one of those constants, so the set really is closed and nothing else has ever been used:

```
src/hflow/snapshot.py               REASON_MISSING, REASON_SIZE_MISMATCH,
                                    REASON_CONTENT_ID_MISMATCH, REASON_NO_RECEIPT
src/hflow/importers/lerobot_verify.py  REASON_MISSING x2, REASON_SIZE_MISMATCH,
                                    REASON_CONTENT_ID_MISMATCH
```

Nothing validates the field. `VerificationFinding(uri="x", reason="mising", detail="")` constructs, serializes, prints, and reaches a user.

Ten lines up, `VerificationStatus` (`:30`) is a `StrEnum` for exactly this reason. There are 20 `StrEnum`s in `src/hflow`. This field is the odd one out in a codebase that already made the decision.

## Why it is worth the change

`reason` is the machine-readable half of a finding. `exit_code_for` keys off `status`, but a caller filtering findings, or a future route serving them, has nothing but a string to match on, and no list of what strings exist except by reading two modules.

## Why it is small

`StrEnum` members equal and format as their values, so this is behaviour-preserving:

- `f"[{finding.reason}]"` at `cli.py:1533` still prints `[missing]`
- `finding.reason == "missing"` still True
- `[f.reason for f in report.findings] == ["missing"]` in the existing tests still passes
- JSON serialization through `str` still yields `"missing"`

## What to do

Turn the four constants into one `StrEnum` and type the field with it:

```python
class VerificationReason(StrEnum):
    MISSING = "missing"
    SIZE_MISMATCH = "size-mismatch"
    CONTENT_ID_MISMATCH = "content-id-mismatch"
    NO_RECEIPT = "no-receipt"
```

Keep the four `REASON_*` names as aliases pointing at the members, or update the 7 call sites, whichever you prefer. Say which and why. If you keep the aliases, they are the public spelling and the enum is the type; do not leave two ways to say the same thing without a reason.

Preserve the comments currently attached to the individual constants (`:26` explains `no-receipt` and cites #428). Those belong on the members.

## Definition of done

1. `VerificationFinding.reason` is typed as the enum, and `ty check` passes.
2. No string literal reason is constructible through the public API without going through the enum.
3. Every existing verify test passes untouched, including the ones comparing `reason` to a plain string. If one needs editing, the change is not behaviour-preserving and is worth a comment saying why.
4. `hflow-server`'s suite passes too: `cd packages/hflow-server && uv run --project . pytest -q`.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

stgomoyaa:
## Summary

- Add `VerificationReason` as the closed `StrEnum` for verification findings.
- Preserve the existing `REASON_*` names as aliases to the enum members.
- Export `VerificationReason` from the package root and cover the enum aliases with a focused test.

## Why

`VerificationFinding.reason` had a plain `str` type even though every valid value was already a fixed set. The enum makes that contract explicit without changing the string values or existing constant imports.

## Validation

- `uv run ruff check --fix` passed.
- `uv run ruff format` passed.
- `uv run ty check` passed.
- `uv run pytest tests/test_verification.py -q` passed: 13 tests.
- `uv run pytest -q` completed with 1741 passed, 6 skipped, and 3 pre-existing failures. The failures are in unrelated ffmpeg tests because this host has ffmpeg 4.4.2, which does not support `-fps_mode`; the same failures occur on the unchanged base branch.

## Checklist

- [x] I updated documentation for changed behavior, flags, formats, or requirements. No user-facing documentation changed because this is a type/API contract clarification.
- [x] I preserved stored-data compatibility; no behavior version change is needed.

Closes #488

### DIFF
--- a/src/hflow/__init__.py
+++ b/src/hflow/__init__.py
@@ -122,6 +122,7 @@ from hflow.storage import (
 from hflow.transform import EpisodeStamps, TransformConfig, write_canonical_episode
 from hflow.verification import (
     VerificationFinding,
+    VerificationReason,
     VerificationReport,
     VerificationStatus,
 )
@@ -206,6 +207,7 @@ __all__ = [
     "TopicInfo",
     "TransformConfig",
     "VerificationFinding",
+    "VerificationReason",
     "VerificationReport",
     "VerificationStatus",
     "VideoImportConfig",
--- a/src/hflow/verification.py
+++ b/src/hflow/verification.py
@@ -9,7 +9,8 @@ dataset snapshot verifier (:func:`hflow.snapshot.verify_dataset_snapshot`,
 #428) both return this shape, so one CLI and one exit-code mapping cover
 every delivered artifact.
 
-Reasons are fixed strings so callers can branch on them.
+Reasons use a closed string enum so callers can branch on a documented set of
+values.
 ``exit_code_for`` maps a report to the verify-family exit codes: 0 clean,
 1 damaged, 3 unverifiable. Exit 2 (unreadable input) is raised as an
 exception before a report exists.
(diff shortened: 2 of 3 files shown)


kstonekuan (approved):
LGTM, merging unchanged. Welcome.

Checked the guarantee rather than the annotation: `reason="mising"` and `reason="missing"` are both refused by `ty`, and CI runs it.

Worth knowing the net here is the type checker, not pytest: reverting the annotation to `str` leaves your test green, because `StrEnum` members are strings at runtime. That is the right shape, not a gap.

Filed #491 from your ffmpeg report.

PR #492, opened 2026-09-10
fix: validate camera FPS thresholds

MehulNegi:
## Summary

`camera_fps_conformance` now validates both `max_plausible_fps` and `downsample_tolerance_fps` before processing an episode. Invalid boolean, non-finite, and out-of-range values now raise a clear `ValueError`.

## Why

Previously, both threshold parameters accepted invalid values such as `True`, negative values, `NaN`, and infinity. This could lead to incorrect FPS classification or silently bypass validation, especially for episodes without cameras.

This change adds parameter validation at the start of `camera_fps_conformance` and adds regression coverage using a camera-less episode. No valid-argument behavior is changed.

## Validation

* `uv run pytest tests/test_checks.py -k "fps_conformance or invalid_thresholds" -q` — 2 passed
* `uv run ruff check` — passed
* `uv run ruff format --check` — passed
* `uv run ty check` — passed
* `uv run pytest tests/test_checks.py -k invalid_thresholds -q` — passed
* `uv run pytest -q` — 1691 passed, 28 failed, 11 skipped

  * The 28 failures are unrelated existing environment/platform failures in packaging tests and the ffmpeg end-to-end test.

## Checklist

### DIFF
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -1626,8 +1626,8 @@ def camera_fps_conformance(
     *,
     nominal_fps: dict[str, int] | None = None,
     cameras: Sequence[str] | None = None,
-    max_plausible_fps: int = 240,
-    downsample_tolerance_fps: int = 1,
+    max_plausible_fps: float = 240,
+    downsample_tolerance_fps: float = 1,
 ) -> CheckResult:
     """Classify each camera's timestamp-derived rate against the rate declared.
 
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Pushed one commit pinning two decisions the PR left silent. Zero `downsample_tolerance_fps` is allowed and a zero plausibility ceiling is not, so the two parameters take different bars; tightening it went unnoticed by all 81 tests. And the `int` to `float` widening is right because 29.97 and 23.976 are real rates.

Your 28 failures are environment, not branch. Zero here.

PR #493, opened 2026-09-10
refactor(snapshot): type receipt entries while preserving the content_id invariant

### LINKED ISSUE #489: Integrity receipt entries are dict[str, str | int], so four use sites coerce and one of them forgets
Integrity receipt entries travel as `dict[str, str | int]` (`src/hflow/snapshot.py:129`, `:138`, `:170`, `:180`, `:885`). Every field is a union, so every read coerces, and the coercions disagree with each other:

```python
src/hflow/snapshot.py:153   key=lambda entry: str(entry["path"])       # coerced
src/hflow/snapshot.py:913   relative_path = str(entry["path"])         # coerced again
src/hflow/snapshot.py:928   receipt_size = int(entry["size_bytes"])    # coerced
src/hflow/snapshot.py:942   if delivered_sha256 != entry["sha256"]:    # NOT coerced
```

`:942` compares a `str` against `str | int`. That comparison is well-typed and always False for the `int` branch, so a receipt whose `sha256` arrived as a number would silently report a content mismatch rather than refusing a malformed receipt. Nothing constructs that today, because `_file_integrity_record` writes a hex string, but the type says it is possible and the code at `:942` does not act like it.

Three fields, three known types, one union covering all of them. The receipt is parsed from external JSON at `:861` and then re-interrogated at every use.

## What to do

One frozen dataclass, constructed at the boundary where the marker is read, and used by type everywhere after:

```python
@dataclass(frozen=True)
class FileIntegrityRecord:
    path: str
    size_bytes: int
    sha256: str
```

Parse it once, next to the `content_id` gate that #483 just added, and refuse a receipt entry whose fields are the wrong type there rather than coercing at four separate use sites.

## The constraint that makes this non-trivial

`_inventory_content_id` (`:138`) hashes `json.dumps(normalized, sort_keys=True, separators=(",", ":"))` over these dicts, and #483 made every verify compare that digest against the stored `content_id`. **The serialized form must not change by one byte**, or every snapshot ever exported fails verification.

So the dataclass needs an explicit conversion back to the same dict shape for hashing, and the digest must be computed over that, not over the dataclass. This is the whole risk of the change and the reason it is labelled advanced.

## Definition of done

1. `FileIntegrityRecord` is constructed once per entry, at the marker-reading boundary, and the per-file loop takes records rather than dicts.
2. A receipt entry with a non-string `path` or `sha256`, or a non-int `size_bytes`, is refused with a message naming the field, not coerced.
3. `content_id` digests are byte-identical. Prove it: export a snapshot on current main, keep the `format.json`, then verify that same directory with the change applied. It must report OK. A test asserting a known digest over a fixed entry set is the durable version.
4. The exporter writes the same `format.json` as before. Compare a full marker before and after.
5. Mutation: change one field's type in the parse step and confirm a test goes red.

## If the digest cannot be preserved

Say so on the issue and stop. Keeping the receipt format stable is worth more than the type cleanup, and a serialization change here would need its own decision about format versioning.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Sagar-024:
# refactor(snapshot): type receipt entries while preserving the content_id invariant

Closes #489.

## What changed

Receipt entries no longer travel as `dict[str, str | int]` through the verifier. A frozen `FileIntegrityRecord` (path: str, size_bytes: int, sha256: str) is parsed once per entry at the marker-reading boundary, next to the #473 content_id gate, and everything after the boundary is typed: the inventory gate, the per-file loop, and the sha256 comparison at what was :942.

The boundary refuses without coercion, naming the exact field: a receipt whose `sha256` arrived as a JSON number used to reach a comparison that can never succeed and was reported as damaged bytes (exit 1, "your data is damaged"); the truth is that the receipt is malformed, which is unreadable input (exit 2, "your receipt is written wrong"). The same refusal covers a non-string path, a non-int size_bytes, and bool masquerading as int.

## The invariant, proven

The hard constraint: `_inventory_content_id` serializes entries with `json.dumps(sorted-by-path, sort_keys=True, separators=(",", ":"))`, and #483 made every verify compare that digest against the stored `content_id`. Moving one byte breaks every snapshot ever exported.

Proof from this branch: the exact serialized string over an 8-entry inventory (7 tables, 1 asset, 1000 characters) and its digest are byte-identical before and after the refactor. Digest before and after: `c117d138cfdd8cd4c613906b8813aa92a89a96192c89ee6522e7a30fc9b69b9a`. The bridge is `to_dict_for_hashing()`, which rebuilds exactly the dict shape the exporter has always serialized, and the digest is computed over that dict form, never over the dataclass.

Durable pins in the suite: an invariant test that computes the digest the old raw-dict way and the new records way and asserts both equal a computed golden value, and a refusal test feeding a numeric sha256 and asserting the boundary raises naming the field.

## Exporter unchanged, proven

The exporter now builds records and serializes them through `to_dict_for_hashing` when writing the marker, so the emitted `format.json` carries the identical dict shapes. The dataset-snapshot suite, which pins the marker's structure and recomputes its `content_id`, passes unchanged apart from parsing records where it recomputes.

## Mutation

With the parser changed to accept an int `sha256`, exactly the refusal test fails and the other 42 stay green. Restored, all green.

## Validation

43 passed across tests/test_snapshot_verify.py and tests/test_dataset_snapshot.py, ruff check, ruff format, ty clean on the changed files, zero non-ASCII and zero em dashes. Boundary-family note: #486 (format identity) and #483 (inventory gate) are merged; #469 (path containment) stays with VARUN3WARE; #474 (post-sync CRC reads) is claimed and next.

### DIFF
--- a/src/hflow/snapshot.py
+++ b/src/hflow/snapshot.py
@@ -154,6 +154,16 @@ class FileIntegrityRecord:
     them. ``to_dict_for_hashing`` rebuilds exactly the dict shape the
     exporter has always serialized, so the ``content_id`` digest stays
     byte-identical with every snapshot ever exported (#489).
+
+    Typing the entries also fixed what the digest covers. Hashing raw dicts
+    made it depend on every key an entry happened to carry; it now depends on
+    these three fields, which are the ones that define a delivery. So an
+    entry with an extra key hashes the same, where it used to hash
+    differently. That is deliberate: a later format revision can add metadata
+    without invalidating the digest of every snapshot already exported. It
+    does mean a marker edited to add a field is not caught here, which costs
+    nothing, because this receipt travels unsigned inside the file it
+    describes and was never a tamper defence.
     """
 
     path: str
--- a/src/hflow/snapshot.py
+++ b/src/hflow/snapshot.py
@@ -145,16 +145,69 @@ def _marker_identifies_dataset_snapshot(format_marker: dict) -> bool:
     )
 
 
-def _file_integrity_record(relative_path: str, absolute_path: Path) -> dict[str, str | int]:
+@dataclass(frozen=True)
+class FileIntegrityRecord:
+    """One receipt entry with known field types.
+
+    Receipt entries arrive from external JSON where every field could be
+    anything; this type is what they become once the boundary has checked
+    them. ``to_dict_for_hashing`` rebuilds exactly the dict shape the
+    exporter has always serialized, so the ``content_id`` digest stays
+    byte-identical with every snapshot ever exported (#489).
+    """
+
+    path: str
+    size_bytes: int
+    sha256: str
+
+    def to_dict_for_hashing(self) -> dict[str, str | int]:
+        return {
+            "path": self.path,
+            "size_bytes": self.size_bytes,
+            "sha256": self.sha256,
+        }
+
+
+def _parse_file_integrity_record(entry: object) -> FileIntegrityRecord:
+    """Check one raw marker entry at the boundary and type it.
+
+    Strict, no coercion: a receipt whose ``sha256`` arrived as a JSON number
+    used to fall through to a per-file comparison that can never succeed and
+    was reported as damaged bytes; the truth is that the receipt itself is
+    malformed, which is unreadable input (#489).
+    """
+    if not isinstance(entry, dict):
+        raise ValueError(
+            f"integrity receipt entry must be a JSON object, got {type(entry).__name__}"
+        )
+    for field_name, expected_type in (
+        ("path", str),
+        ("sha256", str),
+        ("size_bytes", int),
+    ):
+        value = entry.get(field_name)
+        if isinstance(value, bool) or not isinstance(value, expected_type):
+            raise ValueError(
+                f"integrity receipt entry field {field_name!r} must be "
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
LGTM, merging.

Verified across versions rather than within one: exported a copy-mode snapshot on `main`, then ran this branch's verifier against it. Digest identical.

Pushed one commit. The digest now covers `path`, `size_bytes`, `sha256` and nothing else, so an entry with an extra key hashes the same where it used to differ. That is the right trade (a later revision can add metadata without invalidating existing snapshots) but it was silent in the code and the tests.

Sagar-024:
Fair point on the cross-version check. I proved the two paths agreed on the branch, which only proves internal consistency, not stability. Exporting on main to verify against the branch is exactly the blind spot I missed.

The extra-key delta is a sharp catch too. I was so focused on byte-identity for today's shape that I didn't map out the input space for future metadata. I will carry that four-point checklist (added, removed, reordered, retyped) into anything involving serialization from now on.

I am starting the lane-entry validation for #474 now.
Hope Demo Day week leaves you enough time to sleep. Catch you in the next review. @kstonekuan

PR #494, opened 2026-09-10
Centralize numeric range validation

### LINKED ISSUE #487: _field_guards covers the type half only, so 49 sites hand-roll the bool guard and 7 issues have been the same bug
`src/hflow/_field_guards.py` exists to be the one statement of "reject a bool where a number is wanted". It is imported by 4 modules. The same check is hand-rolled 49 times across 17 others.

## Measured on f484656

```
isinstance(x, bool) guard sites outside _field_guards.py:   49  in 17 files
modules that import _field_guards:                           4
distinct spellings of the "must be an int" message:         18
```

Worst offenders:

| file | sites |
| --- | --- |
| `build_ai_vlm_checks.py` | 13 |
| `catalog.py` | 6 |
| `batching.py` | 4 |
| `transform.py`, `importers/video.py`, `ffmpeg/_contact_sheet.py`, `checks.py` | 3 each |

And the message text has drifted apart:

```
"batch_count must be an int, got {}"          batching.py
"max_tokens must be an integer"               build_ai_vlm_checks.py
"api_port must be an int, not {}: {}"         runtime/_bundle.py
"episodes[{}].size_bytes must be an integer"  importers/lerobot_verify.py
"{} must be an int, got {}"                   _field_guards.py  (the one that was supposed to win)
```

## Why adoption stalled, which is the actual bug

`_field_guards` only covers the **type** half. Almost every real field also needs finite and range checks, and there is no helper for those, so the call site writes the whole thing inline. Once it is writing the range check inline, calling `require_int` for the type half buys nothing, so it inlines that too.

There are 34 `isfinite` guard sites and not one of them can use `_field_guards` today. `build_ai_vlm_checks.py:265-268` is the shape:

```python
if (
    isinstance(self.request_timeout_seconds, bool)
    or not isinstance(self.request_timeout_seconds, int | float)
    or not math.isfinite(self.request_timeout_seconds)
    or self.request_timeout_seconds <= 0
):
```

The first two lines are `require_float` spelled longhand. The last two are what is missing.

`build_ai_vlm_checks.py:218` and `:222` are `require_int` spelled longhand, and `:222` and `:253` are byte-identical to each other in two different dataclasses in the same file.

## What this has cost

Seven issues in this family, five already fixed one site at a time: #392, #407, #434, #445, #447 (open), plus #459 and #484 this week. Each one is the same defect in a new place, because there is no single guard to fix.

## What to do

Give `_field_guards` the range half, so a call site can express the whole invariant by composing:

```python
def require_positive_float(value: object, name: str) -> None:
    """Refuse anything but a finite, strictly positive int or float."""

def require_non_negative_int(value: object, name: str) -> None:
def require_int_in_range(value: object, name: str, *, minimum: int, maximum: int) -> None:
```

Pick the set from what the 34 `isfinite` sites actually need; do not invent guards nobody calls. Then adopt them in `build_ai_vlm_checks.py` (13 sites) and `batching.py` (4 sites) and stop there. The remaining 15 files are follow-ups, one per file, and easier once the guard set has settled.

Every message ends up in `_field_guards`'s spelling. That is the point of the change, so a few test strings will need updating and that is expected.

## Definition of done

1. The new guards live only in `_field_guards.py`, with the reasoning comment that file already carries.
2. `build_ai_vlm_checks.py` and `batching.py` have no `isinstance(x, bool)` left; `rg -c "isinstance\([a-z_.]+, bool\)"` reports zero for both.
3. Every refusal those files made before, they still make. `True`, `False`, `NaN`, infinity, out-of-range and wrong-type all still raise `ValueError`.
4. Mutation: neuter each new guard in turn and confirm a test goes red. If a guard has no test, it is not done: that is exactly how this family keeps recurring.
5. No behaviour change for valid input.

## Not in scope

`catalog.py`'s 6 sites. It coerces NumPy scalars, which `_field_guards` deliberately does not (see its module docstring), so it is a different question. Leave it.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Kaileshwar16:
Adds shared numeric guards for finite-value and bounds validation, then uses them in `build_ai_vlm_checks.py` and `batching.py`.

This removes duplicated bool checks and standardizes numeric validation errors while preserving existing valid-input behavior, including batching's support for fractional values.

`catalog.py` remains unchanged.

### Validation

* `uv sync --locked --all-extras`
* `uv run ruff check`
* `uv run ruff format --check`
* `uv run ty check`
* `uv run pytest -q` — **1,780 passed, 7 skipped**
* Verified that disabling each of the seven new guards causes a caller test to fail.
* Confirmed there are no remaining inline bool guards in either target file.

Closes #487

### DIFF
--- a/examples/build_ai_evaluation/tests/test_evaluation.py
+++ b/examples/build_ai_evaluation/tests/test_evaluation.py
@@ -140,20 +140,11 @@ def test_hand_count_parser_accepts_published_and_compatible_shapes(
     assert parse_hand_count_response(response_text) == expected_hand_count
 
 
-@pytest.mark.parametrize(
-    ("response_text", "message"),
-    [
-        ("3", r"hand count must be in \[0, 2\], got 3"),
-        ('{"hand_count": true}', "hand count must be an int, got bool"),
-        ("two", "hand count must be an int, got str"),
-        ("", "hand count must be an int, got str"),
-    ],
-)
+@pytest.mark.parametrize("response_text", ["3", '{"hand_count": true}', "two", ""])
 def test_hand_count_parser_rejects_values_outside_the_evaluation_contract(
     response_text: str,
-    message: str,
 ) -> None:
-    with pytest.raises(ValueError, match=message):
+    with pytest.raises(ValueError, match="0, 1, or 2"):
         parse_hand_count_response(response_text)
 
 
@@ -246,7 +237,7 @@ def test_unparsed_model_judgment_is_an_explicit_recoverable_outcome() -> None:
         timestamp_ns=123,
     )
     assert check_result.measurements["build_ai/hand_count/parse_error"] == (
-        "hand count must be an int, got str"
+        "hand count must be 0, 1, or 2"
     )
     assert check_result.observations[0].values["valid"] is False
     assert check_result.tags == ["build_ai/hand_count/unparsed"]
@@ -426,7 +417,7 @@ def test_sample_result_outcome_variants_are_exclusive() -> None:
     assert not hasattr(success, "error")
     invalid = results[1].outcome
     assert isinstance(invalid, InvalidResponseSampleOutcome)
-    assert invalid.parse_error == "hand count must be an int, got str"
+    assert invalid.parse_error == "hand count must be 0, 1, or 2"
     assert invalid.raw_response == "not a count"
     assert not hasattr(invalid, "predicted_value")
     error = results[2].outcome
--- a/examples/build_ai_evaluation/tests/test_evaluation.py
+++ b/examples/build_ai_evaluation/tests/test_evaluation.py
@@ -140,11 +140,20 @@ def test_hand_count_parser_accepts_published_and_compatible_shapes(
     assert parse_hand_count_response(response_text) == expected_hand_count
 
 
-@pytest.mark.parametrize("response_text", ["3", '{"hand_count": true}', "two", ""])
+@pytest.mark.parametrize(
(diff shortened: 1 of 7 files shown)


kstonekuan (changes requested):
The guard set and composition are right, and `require_non_negative_real` keeps batching's `numbers.Real` contract rather than narrowing it.

Three conversions are not. Every place you needed a `cast` or a `try/except` is a place the guard does not belong: `:366`, `:520`, `:789` are the only bool sites in that file outside a `__post_init__`, and all three parse an external model response. `_field_guards` says it is for caller-constructed configuration.

The cost shows in your own test edit: `parse_hand_count_response("two")` went from "hand count must be 0, 1, or 2" to "must be an int, got str". The second is a fact about Python types; the first is the contract a model failed.

Revert those three, keep the ten. Separately, have `require_int` return the value instead of `None` and the remaining seven casts go away.

Kaileshwar16:
Got it, thanks fir review @kstonekuan , I reverted the three response-handling sites and updated the guards to return the narrowed value so the casts are gone. Pushed the changes

kstonekuan (approved):
LGTM, merging.

Casts gone entirely, and the three response sites are byte-identical to main.

Pushed one commit: `skip_black_frames` was the only guard left unheld, and `type(x) is not bool` went back to `isinstance` (bool cannot be subclassed, so they are equivalent).

PR #495, opened 2026-09-10
fix(lerobot): validate the path templates at the info.json boundary

### LINKED ISSUE #471: LeRobot: the data_path and video_path templates escape the info.json validation boundary
`_parse_dataset_information` is documented as the one place the external document is inspected:

> Every supported field is validated here, once, so that nothing after this boundary inspects the external document again.

`fps` and `features` hold to that. The two path templates do not. They are checked for being non-empty strings and nothing else:

```python
data_path_template = dataset_information.get("data_path")
if not isinstance(data_path_template, str) or not data_path_template.strip():
    raise ValueError("LeRobot meta/info.json must define a non-empty data_path template")
```

They are then handed to `str.format` about five hundred lines later, at `_convert_episode` line 1118 and line 1209, with `chunk_index`, `file_index`, `video_key` and `camera_key`. A template that names anything else, or that is not a well-formed format string, passes the boundary and raises out of the conversion instead.

## Measured

Each of these is accepted by `_parse_dataset_information` and fails at the `format` call:

| `data_path` | raised |
|---|---|
| `data/{episode_index:03d}/f.parquet` | `KeyError: 'episode_index'` |
| `data/{chunk_index:03d/f.parquet` | `ValueError: unmatched '{' in format spec` |
| `data/{chunk_index:qq}/f.parquet` | `ValueError: Invalid format specifier 'qq' for object of type 'int'` |
| `data/{0}/f.parquet` | `IndexError: Replacement index 0 out of range for positional args tuple` |
| `data/}chunk/f.parquet` | `ValueError: Single '}' encountered in format string` |

None of them names the file or the field.

## What it costs

The boundary is at line 691. Between it and the first `format`, the import lists `meta/episodes` (699), downloads every episode metadata parquet (706), and reads them through DuckDB. So the refusal lands after the metadata download rather than before it, and lands as a bare `KeyError` rather than as the "LeRobot meta/info.json ..." message every other invalid field produces.

The unknown-placeholder row is the one likely to be met by accident: a template that indexes episodes rather than chunk and file is a plausible thing to find in a repository that is not the v3 layout the importer expects.

## Suggested fix

Format both templates once at the boundary with the four field names the converter supplies, and refuse the ones that raise, in the same shape as the existing messages:

```
LeRobot meta/info.json has an invalid data_path template 'data/{episode_index:03d}/f.parquet': unknown field 'episode_index'
```

That keeps the documented invariant true and moves five late failures to the place the docstring says they belong. The refusal tests added in #415 pin the non-empty case for both templates, so this would extend those rather than add a new file.

Happy to send a PR if you want it.

chiruu12:
Closes #471.

Both templates are now formatted once inside `_parse_dataset_information`, and a template that raises is refused there with the same message shape as the other invalid fields:

```
LeRobot meta/info.json has an invalid data_path template 'data/{episode_index:03d}/f.parquet': unknown field 'episode_index'
```

Each template is checked against the fields its own call site supplies, as you asked: `data_path` against `chunk_index` and `file_index`, `video_path` against those plus `video_key` and `camera_key`. The sample values match the types the call sites pass, ints for the two indexes and strings for the keys, because a format spec is only valid against a type. `{chunk_index:03d}` has to see an int to be checked at all.

Three failure modes reach the message: `KeyError` becomes `unknown field 'x'`, `IndexError` becomes a line saying positional fields are not supported, and a `ValueError` from `str.format` is passed through as its own text.

The refusal now lands before `meta/episodes` is listed and before any metadata parquet is downloaded. The tests assert that directly, not just the message: each case records what `_hf_tree` was asked for and requires the list to be empty. Asserting the message alone would pass either way.

`test_the_two_templates_are_checked_against_different_field_sets` is the one covering your point about the union. It requires `camera_key` to be legal in `video_path` and refused in `data_path`. I checked it bites by validating both against the union: that test fails and nothing else does.

Removing both validation calls fails 8 of the new tests.

These extend the #415 refusal file rather than adding a new one.

Local: 1488 passed. 29 tests in `test_packaging.py`, `test_packaging_manifest_parsing.py`, `test_cli.py` and `test_end_to_end.py` fail here, and fail identically on `origin/main` with this branch stashed, so they are not from this change. `ruff check`, `ruff format --check` and `ty check` are clean.

### DIFF
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -642,6 +642,11 @@ def _validate_path_template(
     failure arrived as a bare ``KeyError`` naming neither the file nor the
     field. Formatting once here moves it to the boundary and gives it the same
     shape as every other refusal.
+
+    What this proves is that the template formats, not that it names a
+    sensible path. ``{chunk_index.bit_length}`` formats fine and yields a
+    directory named after a bound method; that becomes a 404 at download time
+    rather than something to catch here.
     """
     try:
         template.format(**format_fields)
@@ -655,7 +660,11 @@ def _validate_path_template(
             f"LeRobot meta/info.json has an invalid {field_name} template "
             f"{template!r}: positional fields are not supported, name the field instead"
         ) from error
-    except ValueError as error:
+    except (ValueError, TypeError) as error:
+        # TypeError belongs here for the same reason as the rest: subscripting
+        # a field, ``{chunk_index[0]}``, raises it from str.format and would
+        # otherwise leave the boundary as a bare "'int' object is not
+        # subscriptable" naming neither the file nor the field.
         raise ValueError(
             f"LeRobot meta/info.json has an invalid {field_name} template {template!r}: {error}"
         ) from error
--- a/src/hflow/importers/lerobot.py
+++ b/src/hflow/importers/lerobot.py
@@ -617,6 +617,50 @@ def _episode_metadata_cache_path(
     return episodes_metadata_directory / relative_path
 
 
+#: The fields ``_convert_episode`` supplies to each template's ``format`` call.
+#: They differ, and validating against the union would accept a ``data_path``
+#: naming ``camera_key`` that then fails at the data call site. The values are
+#: representative rather than real: the format spec is checked against the type
+#: the call site passes, so ``{chunk_index:03d}`` has to see an int.
+_DATA_PATH_FORMAT_FIELDS: dict[str, object] = {"chunk_index": 0, "file_index": 0}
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Pushed one commit: `data/{chunk_index[0]}/f.parquet` raises `TypeError`, which was outside your caught set, so it escaped the boundary raw. Folded into the passthrough with a case in your table.

Left alone and noted in the docstring: `{chunk_index.bit_length}` formats fine and yields a directory named after a bound method.

Issue #455, opened 2026-09-07
frames_at_indices has four index guards: two are covered by an assertion that cannot tell them apart, two are not covered at all


kstonekuan:
> **Held for @karMareal1.** They asked for #448 five minutes before someone else opened a pull request on it, and lost it to timing rather than to anything they did. Comment here and I will assign it. If a week passes with no comment, it opens to anyone.

`Episode.frames_at_indices` (`src/hflow/episode.py:631`) validates its indices with four guards in a deliberate order:

```python
if any(isinstance(frame_index, bool) for frame_index in selected_frame_indices):
    raise ValueError("frame indices must be integers, not booleans")
if any(not isinstance(frame_index, int) for frame_index in selected_frame_indices):
    raise ValueError("frame indices must be integers")
if selected_frame_indices != sorted(set(selected_frame_indices)):
    raise ValueError("frame indices must be unique and ascending")
if selected_frame_indices and selected_frame_indices[0] < 0:
    raise ValueError("frame indices must be nonnegative")
```

`src/hflow/episode.py:649-656`.

## What is covered, and how weakly

`tests/test_end_to_end.py:351-366` covers the first two, parametrized over `[True]`, `[np.bool_(True)]`, `[3.0]`, `[np.float64(3.0)]`. It asserts:

```python
with pytest.raises(ValueError):
```

No `match=`. So it cannot tell "must be integers, not booleans" from "must be integers", it cannot tell either from the other two guards' messages, and it would pass on any unrelated `ValueError` raised anywhere in the call. Delete the bool guard and `[True]` still raises, from the line below it, and the test stays green.

The other two messages appear nowhere in the repository:

```
0 files: "frame indices must be unique and ascending"
0 files: "frame indices must be nonnegative"
```

The docstring at `:640-642` states the contract those two enforce: "Indices must be unique and ascending so returned paths and source log times have an unambiguous one-to-one order." That is a documented promise about a public API with no test behind it.

## Why the order is load-bearing

Two things in that block are easy to break while "simplifying" it, and neither is currently protected.

The bool guard has to come first, because `bool` subclasses `int`: `isinstance(True, int)` is true, so the integer guard alone would accept `[True]` as index 1. This repo has hit that five times now.

The nonnegative guard only inspects `selected_frame_indices[0]`, which is correct only because the ascending guard above it has already established the list is sorted. Reorder those two and `[-1, 5]` slips through the nonnegative check.

## Definition of done

[...]

PR #496, opened 2026-09-10
test(episode): pin all Channel.to_numpy refusal branches

### LINKED ISSUE #456: Channel.to_numpy is used in four test files and not one of its six refusals is pinned
> **Held for @amansingh-121.** They asked for #447, which was already held for someone else. Comment here and I will assign this one. If a week passes with no comment, it opens to anyone.

`Channel.to_numpy` (`src/hflow/episode.py:210`) is the documented way to get a field out of a channel as an array, and its whole value is the automatic field selection. That selection has six refusals and none of them is tested.

```
0 files: "has multiple numeric array fields"
0 files: "has multiple numeric fields"
0 files: "has no numeric"
0 files: "not in topic"
0 files: "is ragged"
0 files: "is not numeric (dtype"
4 files: "to_numpy"
```

Four test files call it. All four call it on a channel where the automatic path succeeds, so every branch that refuses is unexercised.

## The six, and what each one is for

`src/hflow/episode.py:218-259`.

| Line | Refusal | Reached when |
| --- | --- | --- |
| `:218` | `topic ... has no messages` | the channel is empty |
| `:226` | `has multiple numeric array fields ...; pass field=...` | auto-selection is ambiguous between array fields, and `position` is absent |
| `:233` | `has multiple numeric fields ...; pass field=...` | ambiguous between scalar fields |
| `:238` | `has no numeric fields; available fields: ...` | nothing numeric to pick |
| `:245` | `field ... not in topic ...; available: ...` | an explicit `field=` that does not exist |
| `:251` | `field ... is ragged (per-message lengths differ)` | per-message lengths differ, so numpy cannot stack |
| `:256` | `field ... is not numeric (dtype ...)` | the field stacks but is not a number |

The docstring at `:211-216` documents the selection rule: `position` if present, else the only numeric array field, else the only numeric scalar field, "and raises with the candidate list when ambiguous". Three of those refusals exist specifically to make that sentence true, and the sentence is currently unverified.

The ragged one is the most interesting. It exists because `np.asarray` on unequal-length sequences raises, and the guard converts that into a message telling you to align the data yourself. That is a real trap in robot data (variable-length detections, per-frame keypoint counts) and the message is the whole feature.

## Definition of done

1. Each of the six is asserted with an anchored `match=`, so no case can pass on a neighbour's message.
2. Removing any one refusal turns exactly one case red. Say that you checked all six, and say so if a mutation kills more than one, because that means two cases overlap.
3. The `position` shortcut still wins when present. A channel with `position` and another numeric array field must select `position` rather than refuse as ambiguous, and that is worth its own case since it is the rule's first clause.
4. The candidate list really is in the message. `pass field=...` is only useful advice if it names the fields, so assert a field name appears rather than just the sentence.

## Where to build the fixtures

`hflow.testing.synthesize_episode` writes `/joint_states` with a `position` array, which is the happy path the existing callers use. Refusals need channels that are shaped wrong on purpose. `tests/test_mcap_writer.py` writes messages directly and is the closest existing pattern for constructing an odd channel; look there before reaching for a synthetic episode, which is built to be well formed.

Do not change `to_numpy` to make it easier to test. If a refusal turns out to be genuinely unreachable, say so instead of forcing a case: that is a more interesting result than the tests, and we have removed three such guards this week (#423, #425).

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

amansingh-121:
Closes #456.

Adds `tests/test_episode_to_numpy.py`, exercising every refusal branch of `Channel.to_numpy` (`src/hflow/episode.py:210`) plus the automatic-selection priority ladder. No source changes — `episode.py` is untouched.

Fixtures build `ChannelData` directly from raw JSON payloads (mirroring `make_json_channel` in `test_resample_derive.py`) rather than `synthesize_episode`, which only produces well-formed happy-path channels.

## Coverage

| Refusal | Test |
| --- | --- |
| `topic ... has no messages` | `test_empty_channel_refuses` |
| `has multiple numeric array fields ...; pass field=...` | `test_ambiguous_array_fields_list_candidates` |
| `has multiple numeric fields ...; pass field=...` | `test_ambiguous_scalar_fields_list_candidates` |
| `has no numeric fields; available fields: ...` | `test_no_numeric_fields_lists_available_fields` |
| `field ... not in topic ...; available: ...` | `test_invalid_explicit_field_lists_available_fields` |
| `field ... is ragged (per-message lengths differ)` | `test_ragged_field_is_rejected` |
| `field ... is not numeric (dtype ...)` | `test_non_numeric_field_is_rejected` |

Selection rules:
- `test_position_field_wins_over_other_arrays` — `position` beats an extra numeric array field (the first clause).
- `test_lone_array_field_is_selected` / `test_lone_scalar_field_is_selected` — the "only array" / "only scalar" clauses.

## Definition of done

1. Each refusal is asserted with a topic-specific anchored `match=` so no case can pass on a neighbour's message.
2. Every candidate-list refusal additionally asserts the actual field names (`'field_a'`, `'field_b'`, `'position'`, etc.) appear in the exception string, not just the static advice text.
3. The `position` shortcut has its own case (no ambiguity error).

## Testing note

All 10 tests pass locally. The host CI runs on Linux; on a Windows-only box the package cannot import because `src/hflow/storage.py` unconditionally imports `fcntl` (Linux-only), so the full `uv run pytest -q` was not runnable here — the new file was verified with `ruff check`, `ruff format --check`, and a per-file `pytest` run under a throwaway `fcntl` shim (10 passed). I did not perform a mutation sweep (temporarily deleting each refusal to confirm exactly one test goes red) since the full suite is blocked on this host; the topic-anchored `match=` patterns are intended to guarantee each maps to exactly one branch.

### DIFF
--- a/tests/test_episode_to_numpy.py
+++ b/tests/test_episode_to_numpy.py
@@ -1,0 +1,172 @@
+"""Refusal-branch tests for :meth:`hflow.episode.ChannelData.to_numpy`.
+
+``to_numpy`` stacks one message field into a ``(n_messages, ...)`` array,
+selecting the field automatically when ``field=None`` through a priority
+ladder -- ``position`` array, lone numeric array, lone numeric scalar -- and
+refusing with a specific error when the choice is empty or ambiguous.
+
+These refusals cannot be produced by :func:`hflow.testing.synthesize_episode`,
+which only generates well-formed happy-path channels, so the fixtures build
+``ChannelData`` objects directly from raw JSON payloads whose decodes yield
+exactly the fields each scenario needs.
+"""
+
+import json
+
+import numpy as np
+import pytest
+
+from hflow.episode import ChannelData
+from hflow.reader import TopicInfo
+
+
+def _json_channel(
+    topic: str, messages: list[dict[str, object]], *, schema_name: str = "test_schema"
+) -> ChannelData:
+    """A ``ChannelData`` (no file) whose JSON payloads decode to ``messages``."""
+    raw_payloads = [json.dumps(message).encode() for message in messages]
+    info = TopicInfo(
+        topic=topic,
+        channel_id=1,
+        schema_name=schema_name,
+        schema_encoding="jsonschema",
+        message_encoding="json",
+        message_count=len(raw_payloads),
+        schema_data=b"{}",
+    )
+    return ChannelData(
+        topic=topic,
+        channel_id=1,
+        info=info,
+        log_times=np.asarray([], dtype=np.int64),
+        publish_times=np.asarray([], dtype=np.int64),
+        raw=raw_payloads,
+        decoder=lambda payload: json.loads(payload.decode()),
+    )
+
+
+def test_empty_channel_refuses() -> None:
+    channel = _json_channel("/empty", [])
+
+    with pytest.raises(ValueError, match=r"topic '/empty' has no messages"):
+        channel.to_numpy()
+
+
+def test_ambiguous_array_fields_list_candidates() -> None:
+    channel = _json_channel(
+        "/multi_array",
+        [{"field_a": [1.0, 2.0], "field_b": [3.0, 4.0]} for _ in range(2)],
+    )
+
+    with pytest.raises(
+        ValueError, match=r"topic '/multi_array' has multiple numeric array fields"
+    ) as excinfo:
+        channel.to_numpy()
+    message = str(excinfo.value)
+    assert "'field_a'" in message
+    assert "'field_b'" in message
+    assert "pass field=" in message
+
+
+def test_ambiguous_scalar_fields_list_candidates() -> None:
+    channel = _json_channel(
+        "/multi_scalar",
+        [{"field_a": 1.0, "field_b": 2.0} for _ in range(2)],
+    )
+
+    with pytest.raises(
+        ValueError, match=r"topic '/multi_scalar' has multiple numeric fields"
+    ) as excinfo:
+        channel.to_numpy()
+    message = str(excinfo.value)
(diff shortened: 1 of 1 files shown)


kstonekuan (approved):
LGTM, merging unchanged.

Seven refusals, seven tests, one-to-one: disabling each condition turned exactly one red, no overlap.

Two corrections to my issue. It said six refusals; there are seven, and the unknown-explicit-field case raises `KeyError` not `ValueError`, which matters to anyone writing `except ValueError` around this.

#455 is the sibling if you want it.

amansingh-121:
Thanks @kstonekuan for the thorough review — verifying each of the seven refusals by disabling them one at a time is exactly the kind of check that gives confidence the table is one-to-one. Good to know that the unknown-explicit-field case raises `KeyError` rather than `ValueError`; I will keep that in mind for anyone writing `except ValueError` around this. Building fixtures from raw `ChannelData` instead of `synthesize_episode` was the right call, and I appreciate the note about the synthesizer only producing well-formed channels.

amansingh-121:
Thanks @kstonekuan for the thorough review — verifying each of the seven refusals by disabling them one at a time is exactly the kind of check that gives confidence the table is one-to-one. Good to know that the unknown-explicit-field case raises `KeyError` rather than `ValueError`; I will keep that in mind for anyone writing `except ValueError` around this. Building fixtures from raw `ChannelData` instead of `synthesize_episode` was the right call, and I appreciate the note about the synthesizer only producing well-formed channels.

PR #497, opened 2026-09-10
fix: hide internal SQL from preview 400s

### LINKED ISSUE #482: A preview 400 still quotes our DESCRIBE wrapper back at the caller when their SQL fails to bind
#450 asked for two things. #453 delivered the first (preview and pin now accept and refuse the same SQL) and not the second:

> 2. No 400 body contains SQL the caller did not send.

That still fails, for a different class of error than the one #453 fixed. #453 moved the *parse* failures behind the shared gate. Binder and catalog errors are found at execution, inside the wrapper, and their message carries the rewrite.

## Measured

`_described_columns` at `packages/hflow-server/src/hflow_server/_curation.py:216` executes:

```python
```

A query that parses cleanly and fails to bind produces:

```
sql sent:  SELECT nope FROM episodes
400 body:  Binder Error: Referenced column "nope" not found in FROM clause!
           Candidate bindings: "episode_id"
           LINE 1: DESCRIBE SELECT * FROM (SELECT nope FROM episodes)
                                           ^
```

```
sql sent:  SELECT * FROM no_such_table
400 body:  Catalog Error: Table with name no_such_table does not exist!
           Did you mean "pg_tables"?
           LINE 1: DESCRIBE SELECT * FROM (SELECT * FROM no_such_table)
```

The diagnostic is the useful part and should stay. The `LINE 1:` echo is our rewrite, with a caret pointing into text the caller never wrote, and it hands them our internal wrapper.

## Why the gate cannot fix this

The gate parses. These errors need a catalog to bind against, so they only appear at execution. Any fix lives at the point the wrapper is executed, not at the point the SQL is admitted.

## What to do

Pick one and say which:

**Strip the wrapper from the message.** Cheapest. `_bad_sql_refusal` at `_curation.py:155` already owns the conversion to a 400, so drop the `LINE n:` block and its caret line when the echoed text is not what the caller sent. Keeps DuckDB's diagnostic, loses the misleading location.

**Stop wrapping.** Get column names and types from the result's own description instead of `DESCRIBE SELECT * FROM (...)`. Then the error is about the caller's SQL because that is what ran. This is the option #450 listed second, and the note there still applies: preview must not materialize a large result just to name its columns.

## Definition of done

1. A 400 from preview never contains `DESCRIBE SELECT * FROM`, for parse, binder, and catalog errors alike.
2. DuckDB's own diagnostic survives: `test_preview_bad_sql_is_400_with_the_duckdb_message` still passes, including the `no_such_table` assertion.
3. A test covers a binder error (unknown column) as well as a catalog error (unknown table). The two take different paths through DuckDB and only one is pinned today.
4. Preview and pin still agree on what they accept.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

harshitethic:
## What changed
- Sanitize DuckDB query-location blocks for preview execution errors before returning HTTP 400.
- Preserve DuckDB diagnostic text such as missing table/column names.
- Add regression coverage for unknown-table and unknown-column preview failures.

## Why
Preview introspection wraps user SQL internally with `DESCRIBE SELECT * FROM (...)`. DuckDB includes that generated statement in binder/catalog errors, leaking server-generated SQL in the API response.

## Testing
- `uv run --project packages/hflow-server --with pytest pytest -q packages/hflow-server/tests` → 240 passed
- focused regression → 1 passed, 20 deselected
- `uv run ruff check packages/hflow-server/src/hflow_server/_curation.py packages/hflow-server/tests/test_server_curation_preview.py` → passed
- `uv run ruff format --check ...` → passed
- `uv run ty check packages/hflow-server/src/hflow_server/_curation.py` → passed
- Mutation validation: disabling preview sanitization makes the new regression test fail as expected.

Fixes #482

### DIFF
--- a/src/hflow/__init__.py
+++ b/src/hflow/__init__.py
@@ -122,6 +122,7 @@ from hflow.storage import (
 from hflow.transform import EpisodeStamps, TransformConfig, write_canonical_episode
 from hflow.verification import (
     VerificationFinding,
+    VerificationReason,
     VerificationReport,
     VerificationStatus,
 )
@@ -206,6 +207,7 @@ __all__ = [
     "TopicInfo",
     "TransformConfig",
     "VerificationFinding",
+    "VerificationReason",
     "VerificationReport",
     "VerificationStatus",
     "VideoImportConfig",
--- a/src/hflow/verification.py
+++ b/src/hflow/verification.py
@@ -9,7 +9,8 @@ dataset snapshot verifier (:func:`hflow.snapshot.verify_dataset_snapshot`,
 #428) both return this shape, so one CLI and one exit-code mapping cover
 every delivered artifact.
 
-Reasons are fixed strings so callers can branch on them.
+Reasons use a closed string enum so callers can branch on a documented set of
+values.
 ``exit_code_for`` maps a report to the verify-family exit codes: 0 clean,
 1 damaged, 3 unverifiable. Exit 2 (unreadable input) is raised as an
 exception before a report exists.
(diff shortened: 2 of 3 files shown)


kstonekuan:
No problem. #482 goes back to the pool unassigned.

If you closed it because the wrapper-stripping approach fought back, that is useful to know and worth a line on the issue: the two options there are not equally cheap and I would rather hear which one broke than guess. If you just want it off your plate, that is fine too and needs no reply.

PR #500, opened 2026-09-11
perf: avoid caching channel payloads in episode_duration

### LINKED ISSUE #499: episode_duration() retains full channel payloads even though it only uses timestamps
I was looking at `episode_duration()` and noticed the default path goes through `episode.channel(topic)` for every selected topic:

```python
for topic in selected:
    stamps_ns = episode.channel(topic).timestamps
```

The check only uses the timestamps, but `Episode.channel()` also collects and caches all raw payloads for that channel.

I tried a small MCAP with:

* `/joint_states`: 2 small messages
* `/camera`: 8 messages with 100 KB payloads

`episode_duration(ep)` returned:

```text
{'duration_s': 8.0, 'message_count_total': 10, 'topic_count': 2}
```

but it also loaded both channels and left `800024` bytes of raw payloads in `_channel_data_by_id`.

I first thought this could use `time_bounds` + summary message counts directly, but that changes current behavior if MCAP statistics don't match the actual messages. I tested that too: with statistics claiming 9 messages over 100 seconds while the file actually had 2 messages over 2 seconds, the current check still measures the real messages.

So I think the safer improvement is to keep reading the real messages, but do it through bounded `iter_batches()` and only keep the timestamps/counts needed for this check instead of constructing full `ChannelData`.

With that change on the same fixture:

```text
Episode.channel() calls: 2 -> 0
cached payload bytes:    800024 -> 0
measurements:             unchanged
```

Payload bytes still pass through the reader, so this doesn't remove the I/O. It just avoids keeping the full channel payloads around in the episode cache for a check that only needs timestamps.

I also checked the existing behavior around empty channels, duplicate topics, missing statistics, explicit topic selections, and repeated explicit topics, since those are easy to accidentally change here.

Kaileshwar16:
Fixes #499.

`episode_duration()` only needs timestamps and message counts, but the default path was going through `Episode.channel()` for every selected topic. That materialized and cached complete channel payloads, including camera/video data the check never uses.

This changes the default `topics=None` path to aggregate the same measurements through bounded raw batches instead.

On the reproduction fixture:

```text
                            before      after
Episode.channel() calls         2          0
cached payload bytes       800024          0
duration_s                    8.0        8.0
message_count_total            10         10
topic_count                      2          2
```

Payload bytes still pass through the reader, so this does not eliminate payload I/O. It avoids retaining complete channel payloads in the episode cache.

I also checked using MCAP summary statistics directly, but that would change existing behavior when summary statistics disagree with the actual messages. The streaming path keeps the current semantics: summary counts select the default topics, while actual messages still supply duration and message-count measurements.

The regression tests pin the performance property directly by refusing `Episode.channel()` calls and checking that `_channel_data_by_id` stays empty. They also cover multiple batches, explicit and repeated topic selections, empty channels/episodes, duplicate topics, missing statistics, and inconsistent summary statistics.

Validation:

```text
episode_duration tests: 24 passed
affected focused tests: 112 passed
ruff check: passed
ruff format --check: passed
ty check: passed
full suite: 1856 passed, 7 skipped
git diff --check: passed
```

### DIFF
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -729,6 +729,19 @@ def _episode_duration_intermediates(
     if topics is None:
         # Statistics select topics, but need not agree with the messages.
         # Keep measuring actual timestamps/counts rather than summary values.
+        #
+        # Reaching past Episode to its reader is the point rather than an
+        # oversight: episode.channel() caches every payload it decodes, and
+        # this check wants only the timestamps (#499). Going through the
+        # public accessor would populate the cache, which is the cost being
+        # removed. Batches arrive per channel in log-time order, so the first
+        # and last stamps of a batch are its bounds.
+        #
+        # The explicit-topics path below stays on episode.channel(), which
+        # refuses an unknown topic with a message naming it; indexing `infos`
+        # here would raise a bare KeyError instead. The two must agree on
+        # measurements for the same selection, which
+        # test_episode_duration_paths_agree_on_the_same_selection pins.
         start_ns: int | None = None
         end_ns: int | None = None
         message_count_total = 0
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -718,14 +718,39 @@ def _episode_duration_intermediates(
     episode: Episode,
     topics: Sequence[str] | None,
 ) -> _EpisodeDurationIntermediates:
-    """Verbatim aggregation from the pre-fact body: explicit ``topics``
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

Pushed one commit: the change leaves two aggregations behind, and nothing asserted they answer the same for the same selection. They do; now pinned.

Also wrote down why the private `episode._reader` access is deliberate, since going through the public accessor is what populates the cache you are avoiding.

PR #502, opened 2026-09-11
fix(checks): validate canonical CRCs once at lane entry with memoization

### LINKED ISSUE #474: check lanes re-certify a decayed canonical episode: post-sync reads never validate chunk CRCs

Sagar-024:
Closes #474.

**Bug:** post-sync reads open canonicals with CRC validation off, so a canonical that decayed on disk gets re-measured and re-certified. Checks stamp fresh `measured` rows over bytes the file's own integrity stamp refuses.

**Fix:** a front-door guard, not a reader flag flip. `verify_canonical_integrity(path)` in reader.py does one strict read with `validate_crcs=True` and catches `CRCValidationError` by its precise type. `App.process` calls it once, before any step runs, in every lane that spends work on canonical bytes: META, relabel, and the production re-check (`process_stage_batch`). On mismatch the episode is refused with ONE diagnosis: `TestReport.refusal_reason` plus one framework-owned `check_runs` row (`error = 'canonical-crc-mismatch'`, critical, curation views read it as `unverified`). No check runs, so zero tracebacks is structural. `hflow ingest` exits 1, the runtime-finding family.

**Decisions worth your eye:**
- Issue scope note superseded: the front-door shape needs one report field plus one catalog row through existing DDL. No migration. Reader default and snapshot verifier untouched.
- Cache carries a size+mtime witness. The issue's own repro broke the path-only cache: corruption between two process calls was served the stale healthy verdict. Stated limit: rot preserving size and mtime slips past within one process; a fresh process always re-validates.
- Dedupe for free: the run fingerprint excludes duration, so refusal replays append nothing. One record per episode across all lanes.
- A run with no step work (sync only, check-less meta, enrichment-less relabel) pays no strict read. Pinned by test.
- Mutation: guard disabled, the pinning test reds as re-certification, because CRC-only rot still decompresses. The N-traceback shape is the zstd-refusing species, separate follow-up.

**Validation:** ruff, format, ty clean on the three changed files. Full suite 1871 passed, 6 skipped on the rebased commit. Mutation red then green. ASCII swept.

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -54,7 +54,7 @@ from hflow.manifest import (
     PipelineManifest,
     StepManifest,
 )
-from hflow.reader import open_reader
+from hflow.reader import open_reader, verify_canonical_integrity
 from hflow.resample import DerivedSeries
 from hflow.step_selection import (
     ALL_REGISTERED_STEPS,
@@ -228,6 +228,12 @@ MEDIA_CONTACT_SHEET_STEP_VERSION = parse_step_version("1")
 # Published artifacts are recorded as measurements under this prefix, so a
 # reader can tell "here is where the file went" from an ordinary label.
 ARTIFACT_MEASUREMENT_KEY_PREFIX = "artifact/"
+# The framework-owned step that records a check-lane refusal: one row naming
+# why the lane stood down, queryable by its error value. Not registered on
+# the App; it exists only on the refusal record (see
+# :func:`_canonical_integrity_refusal_run`).
+CANONICAL_INTEGRITY_STEP_NAME = "integrity/canonical"
+CANONICAL_INTEGRITY_STEP_VERSION = parse_step_version("1")
 _MEDIA_CONTACT_SHEET_FPS = 0.5
 _SYNC_COMPLETION_MARKER_NAME = ".sync-complete.json"
 
@@ -897,11 +903,26 @@ def _check_run_rows(report: "ProcessReport") -> list[CheckRunRow]:
     enrichment's labels and published artifact keys.
 
     One owner, so the collision guard below and the catalog append can never
-    disagree about what would be written. Artifact keys come from
-    ``artifact_uris`` rather than the result's declared artifacts: a step whose
-    artifact failed to publish contributes no key.
+    disagree about what would be written. A refused episode contributes one
+    framework-owned row naming the refusal (its ``critical`` flag is what
+    makes the curation views read the episode as unverified rather than ok).
+    Artifact keys come from ``artifact_uris`` rather than the result's
+    declared artifacts: a step whose artifact failed to publish contributes
+    no key.
     """
-    check_rows = [
+    check_rows: list[CheckRunRow] = []
+    if report.refusal_reason is not None:
+        check_rows.append(
+            CheckRunRow(
+                check_name=CANONICAL_INTEGRITY_STEP_NAME,
+                check_version=CANONICAL_INTEGRITY_STEP_VERSION,
+                critical=True,
+                status=CheckStatus.ERROR,
+                duration_s=0.0,
+                error=report.refusal_reason,
+            )
+        )
+    check_rows.extend(
         CheckRunRow.from_result(
             check_name=run.check.name,
             check_version=run.check.version,
@@ -912,7 +933,7 @@ def _check_run_rows(report: "ProcessReport") -> list[CheckRunRow]:
             result=run.result,
         )
         for run in report.checks
-    ]
+    )
     _raise_if_measurement_keys_claim_artifact_namespace(
         (row.check_name, key) for row in check_rows for key in row.measurements
     )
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
LGTM, merging.

`assert refused.checks == []` makes this the fix I asked for: zero tracebacks because nothing ran, not because something caught them. And asserting `episodes.status == 'unverified'` from the catalog rather than trusting `critical=True`.

I owe you a measurement, since the front-door shape was my call and costs a second read: 0.5% of a process call at 0.4 MB, 0.2% at 6.3 MB. It gets cheaper relative to ffmpeg as episodes grow.

Filed the zstd-refusing half as #506.

Sagar-024:
On the corpus invitation: Discussion #503 (the field report I just posted) is exactly that work. 128 real episodes, 3 factories, 6 stress passes.

PR #504, opened 2026-09-11
refactor(checks): extract the camera-selection line into a shared helper

### LINKED ISSUE #466: Eight camera checks repeat the same camera-selection line, and there is already a helper doing this for topics
`src/hflow/checks.py` has this exact line eight times:

```python
selected_cameras = list(cameras) if cameras is not None else episode.cameras
```

at `:606`, `:1013`, `:1358`, `:1658`, `:1761`, `:1800`, `:1856`, `:1930`. Every camera check and every camera key-set fact resolves its selection the same way, spelled out each time.

## The precedent is already in the file

`_timestamp_regularity_resolve_selected` at `:684` does exactly this job for topics, and its docstring says why:

> Same selection rule, same partition -- one statement of "which topics the check ran over".

That reasoning applies unchanged to cameras. The topic side got a helper and the camera side did not.

## What to do

Add a small private helper beside it, and call it from all eight sites:

```python
def _resolve_selected_cameras(episode: Episode, cameras: Sequence[str] | None) -> list[str]:
    """The cameras a check ran over: the caller's list, or every camera."""
    return list(cameras) if cameras is not None else episode.cameras
```

Check the return type at each site before you switch it. `episode.cameras` is already a list, and `list(cameras)` copies, so the helper must keep returning a fresh list rather than handing back `episode.cameras` itself: a caller that mutates the result must not mutate the episode's own list. Say in the PR that you checked this, because it is the one way this refactor could change behaviour.

## Definition of done

1. One helper, called from all eight sites, with the `cameras is not None` distinction preserved. Note that an explicit empty list is not the same as `None`: passing `cameras=[]` selects nothing, and `cameras=None` selects everything. If your helper collapses those, it is wrong.
2. No test changes needed. The full suite passes untouched.
3. Nothing else in those eight functions moves. This is one line per function.

## Not in scope

The eight functions have real differences in what they do next. Do not try to unify their bodies, only the selection line.

`_timestamp_regularity_resolve_selected` stays as it is. It returns two lists and answers a different question; merging the two helpers would make both worse.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Sravan1011:
Closes #466.

## Summary

`selected_cameras = list(cameras) if cameras is not None else episode.cameras` was spelled out identically at eight sites in `checks.py`. `_timestamp_regularity_resolve_selected` already does the equivalent job for topics, with a docstring explaining why it's one statement rather than several ("same selection rule, same partition -- one statement of 'which topics the check ran over'"). The camera side never got the same treatment.

## What changed

- Added `_resolve_selected_cameras(episode, cameras)` right beside `_timestamp_regularity_resolve_selected`, matching its docstring style.
- Replaced the inline expression at all eight sites (`camera_frame_stats`, `camera_stability`, `camera_signal_quality`, `camera_fps_conformance`, `_media_digest_keys`, `media_digest`, `_keyframe_interval_keys`, `keyframe_interval`) with a call to the helper. One line per function; nothing else in those functions moved.

## Behavior check (per the issue's ask)

- The `cameras is not None` distinction is preserved exactly: `cameras=[]` still selects nothing, `cameras=None` still selects every camera. The helper does not collapse the two.
- `episode.cameras` (`episode.py:396`) is a `@property` that returns a freshly built `sorted(...)` list on every access, not a cached reference — so returning it directly in the `else` branch (as the original inline code also did) cannot let a caller mutate the episode's own state. Checked by reading the property, not assumed.

## Not in scope

`_timestamp_regularity_resolve_selected` is untouched — it returns two lists and answers a different question, so merging the two helpers would make both worse, per the issue.

## Validation

```
uv sync --locked --all-extras
uv run pytest -q                  # 1867 passed, 6 skipped, untouched
```

No test changes.

### DIFF
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -414,7 +414,7 @@ def _camera_frame_stats_keys(episode: Episode, *, cameras: Sequence[str] | None 
     inside the per-topic selection, so an episode with no cameras emits
     nothing at all.
     """
-    selected = list(cameras) if cameras is not None else episode.cameras
+    selected = _resolve_selected_cameras(episode, cameras)
     keys: set[str] = set()
     for topic in selected:
         keys.add(f"{topic}/message_count")
@@ -706,7 +706,14 @@ def _timestamp_regularity_resolve_selected(
 
 
 def _resolve_selected_cameras(episode: Episode, cameras: Sequence[str] | None) -> list[str]:
-    """The cameras a check ran over: the caller's list, or every camera."""
+    """The cameras a check ran over: the caller's list, or every camera.
+
+    ``is not None`` rather than truthiness, and that is the whole reason this
+    is worth stating once. ``cameras=[]`` asks for no cameras and must select
+    none; ``cameras=None`` asks for the default and selects all of them.
+    Collapsing the two would silently turn nine checks into whole-episode
+    scans for a caller who passed an empty selection on purpose.
+    """
     return list(cameras) if cameras is not None else episode.cameras
 
 
--- a/src/hflow/checks.py
+++ b/src/hflow/checks.py
@@ -603,7 +603,7 @@ def camera_frame_stats(
     The trade is one right-to-left key parse (``rpartition``) plus one dict
     lookup per key on top of the ffmpeg decode each topic already pays (#182).
     """
-    selected_cameras = list(cameras) if cameras is not None else episode.cameras
+    selected_cameras = _resolve_selected_cameras(episode, cameras)
     intermediates_by_topic = {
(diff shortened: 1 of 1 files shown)


kstonekuan (approved):
LGTM, merging.

Pushed one commit with a ninth site: `_camera_frame_stats_keys` spells the same selection with the variable named `selected`. My issue counted the same way and said eight.

Collapsing `is not None` to truthiness is caught by exactly one test, so the reason is now in the docstring.

Issue #508, opened 2026-09-11
VideoImportConfig hand-rolls three numeric guards, and only one of them maps straight onto a helper


kstonekuan:
Follow-on from #487. `VideoImportConfig.__post_init__` (`src/hflow/importers/video.py:49-76`) hand-rolls the bool exclusion three times. Unlike #507 this one does **not** map one-to-one onto the helpers, and working out which part is the helper's job and which part stays is the whole exercise.

## The three sites

**1. Finite numbers, `:50-58`.** A loop over `duration_s`, `source_start_s` and `image_hz`:

```python
if isinstance(value, bool) or not isinstance(value, int | float):
    raise ValueError(f"{name} must be a finite number")
if not math.isfinite(value):
    raise ValueError(f"{name} must be a finite number")
```

Two branches, one message. This is `require_finite_float(value, name)` exactly, and the helper distinguishes the two cases the current code conflates.

**2. Positive even integers, `:66-70`.** `image_width` and `image_height`:

```python
if isinstance(value, bool) or not isinstance(value, int) or value <= 0 or value % 2:
    raise ValueError(f"{name} must be a positive even integer")
```

`require_positive_int` covers three of the four clauses. Evenness is a domain rule about H.264 chroma subsampling and belongs here, not in `_field_guards`. Compose: call the helper, then keep the `% 2` check with its own message.

**3. An unsigned 64-bit integer, `:71-76`.** `start_time_ns` is `require_int_in_range(..., minimum=0, maximum=_MAXIMUM_TIMESTAMP_NS)`.

## What not to do

Do not add an `even=` parameter, or a `require_positive_even_int`, to `_field_guards`. One caller does not justify a guard, and the module's value is that it holds the rules every numeric field shares. Evenness is not one of those.

Do not fold the range bound for `start_time_ns` into a new helper either. `require_int_in_range` already takes its bounds.

## The message changes

All three sites lose their bespoke wording for `_field_guards`'s spelling, which is the accepted trade from #487. Two are worth a second look because the current message is arguably better and is still the wrong call:

- `must be a finite number` covers both wrong-type and non-finite. Splitting it into `must be an int or float, got str` and `must be finite, got nan` is more precise, not less.
- `must be a positive even integer` becomes two refusals with two messages. The caller who passed `-2` and the caller who passed `3` made different mistakes.

Check what `tests/test_video_import.py` pins before you start, and update rather than delete.

## Definition of done

[...]

PR #509, opened 2026-09-11
refactor(ffmpeg): reuse contact sheet field guards

### LINKED ISSUE #507: contact_sheet hand-rolls require_positive_int three times
#487 established `_field_guards` as the one place a numeric setting is checked, and #494 gave it the range guards. `ffmpeg/_contact_sheet.py` has three parameters still hand-rolling it, and all three want the same guard.

## Current shape

`src/hflow/ffmpeg/_contact_sheet.py:145-156`, three times over:

```python
if not isinstance(columns, int) or isinstance(columns, bool):
    raise ValueError(f"columns must be an int, got {columns!r}")
if columns < 1:
    raise ValueError(f"columns must be >= 1, got {columns}")
```

for `columns`, `tile_width` and `max_tiles`. Each is exactly `require_positive_int`.

## What to do

```python
from hflow._field_guards import require_positive_int

require_positive_int(columns, "columns")
require_positive_int(tile_width, "tile_width")
require_positive_int(max_tiles, "max_tiles")
```

Keep them where they are, above `selected_frames`, so a bad argument is refused before any frame work.

## The message change, which is the part to get right

This is the whole cost of the change and it is user-visible.

| | before | after |
| --- | --- | --- |
| wrong type | `columns must be an int, got 'x'` | `columns must be an int, got str` |
| out of range | `columns must be >= 1, got 0` | `columns must be > 0, got 0` |

The type message trades the offending value for its type. That is the trade `_field_guards` makes everywhere, and it is the right one here: the type is what the caller got wrong.

Four tests in `tests/test_ffmpeg.py` pin the current strings (`:983`, `:991`, `:999`, `:1007`). They will need updating, and that is expected rather than a sign something broke. Do not keep the old wording by passing a custom message; the point of the change is that all of these read the same.

## Definition of done

1. No `isinstance(x, bool)` left in the file.
2. Every refusal the file made before, it still makes: `True`, `False`, a non-int, and zero or negative, for all three parameters.
3. The four existing tests are updated to the new strings, not deleted.
4. Mutation: neuter each guard in turn and confirm a test goes red. If a parameter has no case, add one; a guard nothing holds is how this family of bugs keeps coming back.
5. No behaviour change for valid arguments. `contact_sheet` has callers in `app.py`, `snapshot.py` and `stage_planning.py`, and none of them should notice.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Note that `tests/test_ffmpeg.py` needs ffmpeg on `PATH`. If the suite fails there before you change anything, see #491.

kudala-bharani:
## Summary

Closes #507.

Use `require_positive_int` for `columns`, `tile_width`, and `max_tiles` in `contact_sheet`. All three checks still run before frame selection. Valid arguments and rendered output are unchanged.

## Why

This removes the three duplicate guards and makes their errors consistent with the other settings: wrong types report the type name, and non-positive values report `must be > 0`.

Extended the four existing validation tests to cover both booleans, floats, strings, `None`, zero, and negative values for every parameter. The function docstring now states the validation contract.

## Validation

Tested on macOS with Python 3.14.6 and system ffmpeg/ffprobe, using the root development environment documented in CONTRIBUTING:

- `uv sync --locked --python 3.14`
- `uv run --locked ruff check --fix` — passed.
- `uv run --locked ruff format` — clean.
- `uv run --locked ty check` — passed.
- `uv run --locked pytest tests/test_ffmpeg.py -k contact_sheet -q` — 25 passed, 1 skipped.
- `.venv/bin/python -m pytest tests/test_ffmpeg.py -q -ra` — 73 passed, 6 skipped (Linux-only binary tests and unavailable drawtext).
- `uv run --locked pytest -q --tb=short` — 1,859 passed, 11 skipped, 27 failed. All 27 failures are native-overlay tests requiring CPython on Linux. Reproduced the same 27 failures in `tests/test_packaging.py` and `tests/test_packaging_manifest_parsing.py` on an untouched worktree at `48aad38`; none involve this change.

The updated validation assertions failed against the old implementation. For the requested mutation check, I temporarily bypassed each guard independently in memory and ran `pytest tests/test_ffmpeg.py -k contact_sheet_rejects -q --tb=no`: each mutation produced 7 failures for its parameter and 15 passes. No mutation is included in the patch.

## Checklist

### DIFF
--- a/src/hflow/ffmpeg/_contact_sheet.py
+++ b/src/hflow/ffmpeg/_contact_sheet.py
@@ -23,6 +23,7 @@ from functools import cache
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+from hflow._field_guards import require_positive_int
 from hflow.ffmpeg._binary import ffmpeg_path
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

I also tried moving the guards below frame selection: caught, because `_evenly_sampled_indices` raises on a bad `max_tiles` first.

#508 and #510 are the siblings if you want another.

PR #511, opened 2026-09-12
fix: route TransformConfig numeric guards through shared helpers

### LINKED ISSUE #510: TransformConfig hand-rolls three numeric guards, and its test matches the field name rather than the message
Follow-on from #487, and this one comes with its safety net already in place, which makes it a good first conversion to attempt.

## Current shape

`TransformConfig.__post_init__` (`src/hflow/transform.py:159-180`) hand-rolls three parameters:

| parameter | current guards | helper |
| --- | --- | --- |
| `crf` | bool exclusion, then `0 <= crf <= 51` | `require_int_in_range(crf, "crf", minimum=0, maximum=51)` |
| `gop_seconds` | bool exclusion, then finite and `> 0` | `require_positive_float` |
| `chunk_size_bytes` | bool exclusion, then `> 0` | `require_positive_int` |

`gop_seconds` and `chunk_size_bytes` are `None`-able, so keep the `is not None` wrapper around each and call the helper inside it. `None` means "use the default" and must stay legal.

## The safety net, and why it needs tightening too

`tests/test_transform.py:51-70` already covers all ten cases: `True`, out-of-range, `NaN` and infinity across the three parameters. Confirmed it bites, by disabling five of the guards at once:

```
five TransformConfig numeric guards disabled at once
    9 failed, 1866 passed
```

So you are not working without a net. But look at what it asserts:

```python
with pytest.raises(ValueError, match=field):
    construct()
```

`match=field` matches the field **name**, not the message. `crf=True` and `crf=52` are different mistakes and this test cannot tell them apart: either one passes on the other's message. It also means the conversion will not break it, because the new messages still start with the field name, so a green suite is not by itself evidence the conversion was faithful.

Tighten it as part of this. Each row should assert the message it actually expects, the way #492 does for `camera_fps_conformance`.

## Definition of done

1. No `isinstance(x, bool)` left in `transform.py`.
2. `None` is still accepted for `gop_seconds` and `chunk_size_bytes`.
3. Each of the ten existing cases asserts its own message rather than the field name, so a type refusal and a range refusal are distinguishable.
4. Mutation: neuter each guard in turn, one at a time, and confirm a test goes red for that one specifically. Disabling the `crf` range check must not be caught only by a `crf` type case.
5. No behaviour change for valid configurations. `compute_pipeline_version` is a content hash over this config, so any change to accepted values changes episode identity; there must be none.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Amidwestnoob:
## Summary
- Replace hand-rolled `TransformConfig` guards for `crf`, `gop_seconds`, and `chunk_size_bytes` with `require_int_in_range`, `require_positive_float`, and `require_positive_int`.
- Keep `None` legal for `gop_seconds` and `chunk_size_bytes`.
- Tighten the ten existing rejection cases so each asserts its own message (type vs range/finite distinguishable).
- No change for valid configs; default `compute_pipeline_version` hash unchanged.

Closes #510

## Test plan
- [x] `uv sync --locked --all-extras`
- [x] `uv run ruff check`
- [x] `uv run ty check`
- [x] `uv run pytest -q` (1896 passed, 6 skipped)

### DIFF
--- a/src/hflow/transform.py
+++ b/src/hflow/transform.py
@@ -56,7 +56,6 @@ v1 behavior:
 import hashlib
 import json
 import logging
-import math
 from collections.abc import Callable, Mapping, Sequence
 from copy import copy
 from dataclasses import dataclass, field
@@ -64,6 +63,11 @@ from pathlib import Path
 from typing import Any, Literal
 
 from hflow import video as video_module
+from hflow._field_guards import (
+    require_int_in_range,
+    require_positive_float,
+    require_positive_int,
+)
 from hflow._grouped_mcap_writer import (
     NO_SCHEMA_ID,
     ChannelId,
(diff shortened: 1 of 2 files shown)


kstonekuan (approved):
LGTM, merging.

`compute_pipeline_version` is unchanged on both sides (`9cd5d34f513d`), so no corpus re-ingest.

#518 arrived against the same issue later and is closed. Nothing for you to do.

PR #512, opened 2026-09-12
refactor(video): reuse field guards in VideoImportConfig (#508)

### LINKED ISSUE #508: VideoImportConfig hand-rolls three numeric guards, and only one of them maps straight onto a helper

Rayan-and-beyond:
Closes #508.

## What changes

`VideoImportConfig.__post_init__` now reuses the shared numeric field guards instead of duplicating type and range checks.

- `duration_s`, `source_start_s`, and `image_hz` use `require_finite_float`
- `image_width` and `image_height` use `require_positive_int`, while the H.264 even-dimension rule stays local to `video.py`
- `start_time_ns` uses `require_int_in_range` with the existing timestamp bounds

The video-specific validation remains in place, including the separate even-dimension error so non-positive and odd dimensions are reported distinctly.

## Validation

Ran:

```bash
uv run pytest tests/test_video_import.py -q
uv run ruff check
uv run ruff format --check
uv run ty check
```

Result:

- `tests/test_video_import.py`: 39 passed
- Ruff check: clean
- Ruff format: clean
- `ty`: clean

I also checked each guard independently.

The `start_time_ns` upper-bound case originally stayed green when that bound was removed because the later MCAP timestamp-range check rejected the same value. I added a focused assertion for the field-level error so that removing only the upper bound now fails independently instead of being masked by the downstream check.

### DIFF
--- a/src/hflow/importers/video.py
+++ b/src/hflow/importers/video.py
@@ -12,6 +12,11 @@ from foxglove_schemas_protobuf.CompressedImage_pb2 import CompressedImage
 from mcap.writer import Writer
 from mcap_protobuf.schema import build_file_descriptor_set
 
+from hflow._field_guards import (
+    require_finite_float,
+    require_int_in_range,
+    require_positive_int,
+)
 from hflow._pinned_asset import sha256_hex_of_file
 from hflow.ffmpeg import ffmpeg_path, ffmpeg_version
 from hflow.ffmpeg._process import media_input_was_rejected, run_media_command
@@ -52,10 +57,7 @@ class VideoImportConfig:
             ("source_start_s", self.source_start_s),
             ("image_hz", self.image_hz),
         ):
-            if isinstance(value, bool) or not isinstance(value, int | float):
-                raise ValueError(f"{name} must be a finite number")
-            if not math.isfinite(value):
-                raise ValueError(f"{name} must be a finite number")
+            require_finite_float(value, name)
         if self.duration_s <= 0 or self.source_start_s < 0:
             raise ValueError("duration_s must be positive and source_start_s nonnegative")
         if not 0 < self.image_hz <= NANOSECONDS_PER_SECOND:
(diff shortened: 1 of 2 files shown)


kstonekuan (changes requested):
One mutation does not bite: widening the `start_time_ns` upper bound passes all 38 tests, even though `tests/test_video_import.py:229` passes `(1 << 64)`. Have a look at why.

It predates your change, so "I established it is not independently reachable, here is why" is a complete answer.

Everything else is clean.

Rayan-and-beyond:
Thanks — I traced the upper-bound mutation.

The existing `start_time_ns = 1 << 64` case was only proving that the value was rejected somewhere. Without the upper bound in `require_int_in_range`, the later MCAP timestamp-range check still raised, so the test stayed green.

I kept the field-level bound and added a focused assertion for its error. With that in place, removing only the upper bound now makes the test fail independently.

I reran the video-import tests, Ruff, formatting, and `ty`; everything is green.

kstonekuan (approved):
LGTM, merging.

Both ends of the range are held now. I checked the lower bound too, which had the same exposure.

Nit, not worth a round trip: the expected bound is derived from the test input rather than from `_MAXIMUM_TIMESTAMP_NS`.

PR #513, opened 2026-09-12
fix(server): stop echoing the preview wrapper in binder and catalog errors

### LINKED ISSUE #482: A preview 400 still quotes our DESCRIBE wrapper back at the caller when their SQL fails to bind

mulkakhileshmj:
Closes #482.

Note: I saw @amansingh-121 offered on the issue. There is no assignee, and per the stated rule an assignee is the only reservation, but happy to defer if you would rather assign it there.

## Problem

#453 moved parse failures of the caller's SQL in front of the wrapper, but binder and catalog errors only appear when the wrapper executes. DuckDB appends a location block to those errors: a `LINE n:` echo of the executed statement plus a caret pointing into it. For preview that statement is always the internal `DESCRIBE SELECT * FROM (...)` rewrite, so the 400 handed the caller the wrapper with a caret aimed at text they never wrote.

## Direction taken

The issue's first option: strip the echo, keep the diagnostic. The preview route's single `duckdb.Error` funnel now converts through `_wrapper_execution_refusal`, which drops the `LINE n:` line and its caret and keeps every sentence above them, which is the part about the caller's SQL ("Binder Error: Referenced column ... Candidate bindings ...", "Catalog Error: Table ... does not exist ... Did you mean ..."). The rule needs no text comparison: by the time preview executes anything, the gate has already parsed the caller's SQL standalone, so inside preview every executed statement is a wrapper and every location block quotes a rewrite.

Parse errors of the caller's own text are unchanged: they are raised at the gate through `_bad_sql_refusal`, where the echoed text is the caller's own and the caret is useful.

## Definition of done, walked

1. A preview 400 never contains `DESCRIBE SELECT * FROM`: the location block is the only place it appeared, and it is dropped for parse, binder, and catalog errors alike (all funnel through the same route-level catch).
2. `test_preview_bad_sql_is_400_with_the_duckdb_message` passes unchanged, including the `no_such_table` assertion, because the diagnostic sentence survives.
3. New test covers a binder error (unknown column) and a catalog error (unknown table) separately, asserting the diagnostic survives and that neither the wrapper text, the `LINE` echo, nor the caret appears in the body.
4. Preview and pin acceptance is untouched: nothing about the gate changed.

## Validation

```
uv sync --locked
uv run pytest -q           # 1918 passed, 6 skipped
```

Mutation check: with `_curation.py` reverted to main, the new test fails on the leaked wrapper; restored, it passes. The server package tests also pass under Python 3.11 (241 passed).

### DIFF
--- a/packages/hflow-server/src/hflow_server/_curation.py
+++ b/packages/hflow-server/src/hflow_server/_curation.py
@@ -158,6 +158,38 @@ def _bad_sql_refusal(error: duckdb.Error) -> HTTPException:
     return HTTPException(status_code=400, detail=str(error))
 
 
+# DuckDB appends a location block to parser, binder, and catalog errors: a
+# "LINE n: <query text>" echo of the statement it executed plus a caret line
+# pointing into it.
+_LOCATION_ECHO_PATTERN = re.compile(r"^LINE \d+: ")
+
+
+def _wrapper_execution_refusal(error: duckdb.Error) -> HTTPException:
+    """A 400 carrying DuckDB's diagnostic, minus the echoed rewrite (#482).
+
+    Inside preview every executed statement is one of this module's wrappers
+    (``DESCRIBE SELECT * FROM (...)``, the LIMIT projection, the count, the
+    SUMMARIZE), never the caller's SQL verbatim: the gate already parsed that
+    standalone before preview ran. So a location block here always quotes a
+    rewrite the caller never sent, with a caret pointing into it. The
+    diagnostic sentences above the block are about the caller's SQL and
+    survive unchanged; parse errors of the caller's own text keep their full
+    message through :func:`_bad_sql_refusal` at the gate.
+    """
+    kept_lines: list[str] = []
+    dropping_caret = False
+    for line in str(error).splitlines():
+        if _LOCATION_ECHO_PATTERN.match(line):
+            dropping_caret = True
+            continue
+        if dropping_caret and line.strip() == "^":
+            dropping_caret = False
+            continue
+        dropping_caret = False
+        kept_lines.append(line)
+    return HTTPException(status_code=400, detail="\n".join(kept_lines).rstrip())
+
+
 def _reject_non_single_select(user_sql: str) -> None:
     """Refuse anything that is not exactly one SELECT statement.
 
@@ -330,7 +362,7 @@ def create_curation_router(settings: ServerSettings) -> APIRouter:
                     connection, user_sql, limit=request.limit, include_stats=request.stats
                 )
             except duckdb.Error as error:
-                raise _bad_sql_refusal(error) from error
+                raise _wrapper_execution_refusal(error) from error
 
     @router.post("/curation/report")
     def run_curation_report(request: ReportRequest) -> CurationReportResponse:
--- a/packages/hflow-server/src/hflow_server/_curation.py
+++ b/packages/hflow-server/src/hflow_server/_curation.py
@@ -167,17 +167,23 @@ _LOCATION_ECHO_PATTERN = re.compile(r"^LINE \d+: ")
 def _wrapper_execution_refusal(error: duckdb.Error) -> HTTPException:
     """A 400 carrying DuckDB's diagnostic, minus the echoed rewrite (#482).
 
-    Inside a curation route's execution phase, every executed statement is a
-    wrapper built around the caller's SQL, never the SQL verbatim: preview
-    runs ``DESCRIBE SELECT * FROM (...)``, the LIMIT projection, the count,
-    and the SUMMARIZE; report and pin run the library's count, COPY, and
-    coverage wrappers through ``curate``/``write_dataset_manifest``. The gate
-    already parsed the caller's SQL standalone before anything executed. So a
-    location block here always quotes a rewrite the caller never sent, with a
-    caret pointing into it. The diagnostic sentences above the block are
-    about the caller's SQL and survive unchanged; parse errors of the
-    caller's own text keep their full message through
-    :func:`_bad_sql_refusal` at the gate.
+    Inside preview's and report's execution phases, every executed statement
+    is a wrapper built around the caller's SQL, never the SQL verbatim:
+    preview runs ``DESCRIBE SELECT * FROM (...)``, the LIMIT projection, the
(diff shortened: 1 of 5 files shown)


kstonekuan (changes requested):
The strip is right.

Before I merge it: run the same experiment against `/curation/report` and `/curation/pin`. Same defect or not, and does it belong here or a follow-up?

On the deferral question: no assignee on #482, so your PR stands.

mulkakhileshmj:
Ran the experiment on both routes. Answer: same defect on report, clean on pin, and it belongs in this PR, now pushed.

**Report leaks, measured.** Unknown column and unknown table both echo the library's count wrapper:

```
sql sent:  SELECT nope FROM episodes
400 body:  Binder Error: Referenced column "nope" not found in FROM clause!
           Candidate bindings: "operator", "episode_id", ...
           LINE 1: SELECT count(*) FROM (SELECT nope FROM episodes)
                                                ^
```

Same shape for `SELECT * FROM no_such_table` with `Catalog Error`. Different wrapper text than preview's DESCRIBE, exactly as you predicted, so a DESCRIBE-based assertion would have passed here while the route kept leaking.

**Pin does not leak, measured.** Both error classes come back as the bare diagnostic with no location block at all on current DuckDB, so pin was clean by accident of which wrapper fails first rather than by design.

**The fix.** Both routes' execution funnels now convert through the same `_wrapper_execution_refusal` as preview, and the docstring's argument extends unchanged: the gate parses the caller's SQL standalone before any route executes anything, so a location block in the execution phase always quotes a wrapper, whichever wrapper it is. The gate's own conversion through `_bad_sql_refusal` is untouched, so parse errors of the caller's text keep their caret.

**Tests, built to fail on an unfixed route.** The report test asserts on the `LINE 1:` echo and the caret line rather than on any wrapper's text. Mutation checked with `_curation.py` reverted to main: the preview test fails, the report test fails, and the pin test passes, which is honest, since pin was clean before the fix. The pin test exists to keep it that way if a future wrapper starts failing first with a location block.

Gate on the result: ruff, ruff format, ty clean; full suite 1920 passed, 6 skipped; `packages/hflow-server` 243 passed on Python 3.11.

kstonekuan (changes requested):
Pin and report need the same treatment. Both still funnel through `_bad_sql_refusal`, and both execute the caller's SQL inside something.

Whatever you assert, make it fail on the unfixed route: the wrapper text differs from preview's.

mulkakhileshmj:
You are right, and the mistake was instructive enough that the docstring now teaches it.

Resolution picked: **pin stays on `_bad_sql_refusal`.** `_stage_manifest_and_count` runs the caller's SQL verbatim through `connection.sql()`, so a location block on that path would be the caller's own text, and stripping it would delete the one useful caret in the system. The stripping refusal's docstring now claims only preview and report, names pin's verbatim execution as the reason it is excluded, and ends with the check to run before extending it to a fourth route: confirm every statement the route executes wraps the SQL.

The pin test dropped its no-echo assertions for the same reason. Asserting no location block on a route that runs verbatim SQL would pressure whoever meets a future caret there into stripping the wrong thing. It now pins what pin actually promises, the diagnostic reaching the caller for both error classes, and its docstring states the verbatim distinction.

Took the offered line as well: `_stage_manifest_and_count`'s docstring opens with "Write the query's result" now, matching the `write_parquet` it has done since the move.

Gate on the result: ruff, ruff format, ty clean; full suite 1920 passed, 6 skipped; `packages/hflow-server` 243 passed on Python 3.11. Mutation re-run with `_curation.py` reverted to main: the preview and report tests both fail, and nothing else moved.

kstonekuan (approved):
LGTM, merging.

One thing the suite does not cover: switching pin to `_wrapper_execution_refusal` passes all 46 tests. Not fixable here, so the docstring is the whole guard.

PR #514, opened 2026-09-12
fix: log check and enrichment execution errors

### LINKED ISSUE #505: Check execution errors are silent in direct app.process() loops unless the report is inspected
I followed up on the `CheckStatus.ERROR` behavior mentioned around #86 and tried it against current `main`.

HFlow is recording check crashes correctly. A raised exception becomes `CheckStatus.ERROR`, the traceback is kept on the run, `report.has_errors` becomes `True`, and the episode is not quarantined since this is an infrastructure failure rather than a data-quality verdict.

What surprised me is that none of that is surfaced during a direct `app.process()` call unless the caller explicitly inspects the returned report.

I tested this with three of the prepared Egocentric episodes:

```python
from pathlib import Path
import hflow

sources = sorted(
    Path("data/egocentric/landing").glob("*.mcap")
)[:3]

app = hflow.App(
    "check-error-batch",
    data_root="/tmp/hflow-check-error-batch",
)

@app.check(version="1")
def exploding_check(ep: hflow.Episode) -> hflow.CheckResult:
    raise RuntimeError("intentional probe failure")

for source in sources:
    print("processing:", source.name)
    app.process(source)

print(f"completed {len(sources)} episodes")
```

Output:

```text
processing: factory_051_episode_0001.mcap
processing: factory_051_episode_0002.mcap
processing: factory_051_episode_0003.mcap
completed 3 episodes
```

and the process exits successfully:

```text
shell exit=0
```

`exploding_check` failed on all three episodes, but there was no warning/error output from HFlow during the loop.

For a single episode, the returned report does contain everything needed:

```text
has_errors: True
quarantined: False
```

and `report.summary()` makes the failure obvious:

```text
! exploding_check [error]
...
RuntimeError: intentional probe failure
```

So I don't think the current exception handling itself is wrong. Catching the exception, continuing later checks, and keeping infrastructure failure separate from quarantine all seem intentional.

The awkward part is just the visibility: a direct/local processing loop can look completely successful if the caller forgets to inspect `report.has_errors` or print the report.

I noticed several examples explicitly remember to check `report.has_errors`, which made me wonder whether this should be surfaced a little more strongly by the processing path itself.

Not sure what the right behavior should be here , error-level logging when a check/enrichment becomes `ERROR`, some batch-level signal, or something else.

Happy to take this if you think it should change; I wanted to check the intended surface before implementing a specific behavior.

Kaileshwar16:
## What changed

Fixes #505.

Check and enrichment execution failures were already recorded as `ERROR` in the returned report, but direct `app.process()` usage could stay completely silent unless the caller explicitly inspected `report.has_errors` or printed the summary.

This adds ERROR-level logging for:

* check execution failures
* enrichment execution failures
* enrichment artifact publication failures

The log includes the step name, episode, and exception information, while the full traceback remains available in the existing report/summary.

## Behavior unchanged

The existing execution semantics are preserved:

* errors are still recorded in the returned report
* processing continues after a check or enrichment fails
* `has_errors` behavior is unchanged
* infrastructure errors do not quarantine the episode
* exceptions are not re-raised
* exit behavior is unchanged
* `verbose` behavior is unchanged
* no new output is written to stdout

## Tests

Added/updated coverage for:

* check failures producing ERROR logs
* enrichment failures producing ERROR logs
* artifact publication failures producing ERROR logs
* later steps continuing after a failure
* direct multi-episode `app.process()` loops still completing successfully while failures are visible on stderr
* full tracebacks not being duplicated in the log line
* no additional stdout output

### DIFF
--- a/src/hflow/app.py
+++ b/src/hflow/app.py
@@ -17,6 +17,7 @@ quality outcome.
 import hashlib
 import inspect
 import json
+import logging
 import os
 import re
 import shutil
@@ -99,6 +100,8 @@ from hflow.transform import (
     write_canonical_episode,
 )
 from hflow.workspace import RUNTIME_BUNDLE_DIRECTORY_NAME, Workspace
+
+logger = logging.getLogger(__name__)
 
 # The contract a @app.transform override implements: (source, output, config)
 # -> the stamps it wrote. See :meth:`App.transform`.
@@ -884,8 +887,21 @@ def _execute_enrichment(
                 "expected hflow.EnrichmentResult -- wrap it: return "
                 "hflow.EnrichmentResult(labels=...)"
             )
-    except Exception:
+            logger.error(
+                "Enrichment %s errored on episode %s: %s",
+                registered_enrichment.name,
+                canonical_episode.path,
+                outcome.error,
+            )
+    except Exception as error:
         outcome = Errored(traceback.format_exc(limit=8))
+        logger.error(
+            "Enrichment %s errored on episode %s: %s: %s",
+            registered_enrichment.name,
+            canonical_episode.path,
+            type(error).__name__,
+            error,
+        )
     duration_s = time.perf_counter() - started
     return EnrichmentRunReport(
         enrichment=registered_enrichment, outcome=outcome, duration_s=duration_s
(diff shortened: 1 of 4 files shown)


kstonekuan (approved):
LGTM, merging unchanged.

`(record,) = caplog.records` pins that a failure logs exactly once, and `record.levelno` is what separates ERROR from WARNING (both surface through the last-resort handler, so visibility alone would not).

PR #515, opened 2026-09-12
fix(snapshot): refuse receipt paths that escape the handed directory

### LINKED ISSUE #469: [Bug]: Snapshot verify can read outside the handed directory via receipt paths

VARUN3WARE:
## Summary
- Closes #469: `verify_dataset_snapshot` reuses `_validated_relative_key` and raises (CLI exit 2) before reading absolute / `..` / drive-letter receipt paths.
- Threat model: not local privilege escalation; restores the documented “joined only onto the handed directory” promise for markers we did not write (tenant-supplied snapshots).
- Built on the #486 / #483 boundary gates already on main.

## Test plan
- [x] Relative `../`, absolute path, and `tables/../../…` (subdir present) → exit 2
- [x] Honest relative export still verifies clean
- [x] `tests/test_snapshot_verify.py` + `tests/test_dataset_snapshot.py`
- [ ] CI green

### DIFF
--- a/docs/how-to/export-dataset-snapshot.md
+++ b/docs/how-to/export-dataset-snapshot.md
@@ -86,13 +86,16 @@ in the receipt is re-read and compared by size and sha256, and every finding
 is returned in one report built from the shared verification types in
 `hflow.verification` (a `VerificationReport` with status `ok`, `damaged`,
 or `unverifiable`). Recorded paths are joined only onto the handed
-directory, so a copied delivery verifies in place. Exit `0` clean,
-`1` damaged, `3` unverifiable, `2` unreadable input. A pre-#401
-`format.json` with no `integrity` key is reported as `no-receipt`,
-unverifiable, not corrupt. Older HFlow overwrite checks only
-`format` and `format_version`, and readers that only consume the string
-`tables` map keep working; the `integrity` key is purely additive under
-format version `1`.
+directory, so a copied delivery verifies in place. A receipt path that
+is absolute, contains `..`, or carries a Windows drive letter is refused
+as unreadable input (exit `2`) before any file is read — the same family
+storage keys refuse — so a tampered marker cannot hash bytes outside the
+delivery. Exit `0` clean, `1` damaged, `3` unverifiable, `2` unreadable
+input. A pre-#401 `format.json` with no `integrity` key is reported as
+`no-receipt`, unverifiable, not corrupt. Older HFlow overwrite checks
+only `format` and `format_version`, and readers that only consume the
+string `tables` map keep working; the `integrity` key is purely additive
+under format version `1`.
 
 The receipt travels unsigned inside the `format.json` it describes, so it
 catches corruption and accidental loss, not tampering: anyone who can edit a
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
LGTM, merging.

Reusing `_validated_relative_key` is the right call: the verifier and storage roots now give one answer to "is this key inside the root".

Checked it against the three shapes from the issue plus the ones that must still pass:

```
../outside/secret.bin           REFUSE
/etc/hostname                   REFUSE
tables/../../outside/secret.bin REFUSE
C:\Windows\x                    REFUSE
samples.parquet                 accept, unchanged
assets/wrist_cam/frame_01.jpg   accept, unchanged
```

It normalizes `./a` and `a//b`, which our exporter never writes and which does not reach the digest (`_inventory_content_id` hashes `record.path`).

PR #516, opened 2026-09-12
fix: refuse canonicals with invalid zstd chunks

### LINKED ISSUE #506: A canonical whose chunk zstd cannot decompress is still an undiagnosed exception, not a named refusal
#474 named two shapes of decayed canonical. #502 landed the first: a chunk whose stored CRC no longer matches is refused at the check lane's front door with `canonical-crc-mismatch`, one row, exit 1.

The second shape is still undiagnosed: payload damage that **zstd itself refuses to decompress**.

`verify_canonical_integrity` (`src/hflow/reader.py`) catches `CRCValidationError` by its precise type, which is right, and deliberately does not catch anything wider. A chunk zstd cannot decompress does not raise `CRCValidationError`, so it propagates out of the guard, out of `App.process`, and lands in `stage_execution.py`'s `except Exception` as a generic recorded failure.

That is already better than before #502, when it happened separately inside every check. It is now one exception rather than N. But it is still an exception rather than a refusal: no `refusal_reason`, no `check_runs` row you can filter on, and a `failure_kind` of `infrastructure` for what is actually a damaged delivery.

## What to establish first

Measure it before changing anything, and say what you measured:

1. What exception type and message does a zstd-undecompressable chunk actually produce, through `verify_canonical_integrity` and through `App.process`? The mcap reader may wrap it.
2. Does it reach `ingest_failures` with `failure_kind = 'infrastructure'`? That row says the pipeline broke, and the truth is that the episode's bytes are damaged.

A repro can start from #474's, flipping bytes inside the compressed payload rather than the stored CRC.

## What to do

Give it the same treatment #502 gave the CRC case: a named reason alongside `canonical-crc-mismatch`, carried on `refusal_reason` and on the framework-owned `integrity/canonical` row, so a corpus query finds both species of damaged canonical with one filter on `error`.

Keep the type discipline #502 established. Catch the decompression failure by its precise type, not `except Exception`, or the guard starts answering for unrelated boundary errors such as a missing file or a permission denial, which are genuinely infrastructure.

## Definition of done

1. A canonical with an undecompressable chunk is refused with a named reason, not an undiagnosed exception, and no check runs.
2. Both species are findable together: one query on `check_runs.error` returns the CRC case and this one.
3. The CRC case is unchanged. #502's tests pass untouched.
4. A genuinely infrastructural failure reading the file (missing, unreadable, permission) is still an infrastructure failure and is NOT reported as damaged bytes. Pin that, because it is what a broad `except` would break.
5. Mutation: neuter the new branch and confirm the pinning test reds.

## Validation

```bash
uv sync --locked --all-extras
uv run ruff check
uv run ruff format --check
uv run ty check
uv run pytest -q
```

Kaileshwar16:
Fixes #506.

A canonical with a corrupted zstd-compressed chunk currently escapes `verify_canonical_integrity()` as `ZstdError`, escapes `App.process()`, and is eventually recorded as an infrastructure failure.

I reproduced the failure first and confirmed the actual exception is:

```text
zstandard.backend_c.ZstdError:
error determining content size from frame header
```

MCAP does not wrap it.

This change handles that precise failure at the canonical integrity boundary and refuses the episode with:

```text
canonical-decompression-failed
```

The refusal uses the same framework-owned `integrity/canonical` path introduced for CRC failures in #502.

## Behavior

After this change:

* undecompressable zstd chunks are treated as damaged canonical data rather than infrastructure failures
* `refusal_reason` is set to `canonical-decompression-failed`
* an `integrity/canonical` error row is recorded
* user checks and enrichments do not run
* CRC corruption behavior is unchanged
* unrelated filesystem/read failures still propagate as infrastructure failures

The implementation catches only `zstandard.ZstdError`; there is no broad exception handling.

## Reproduction

The regression fixture starts from a valid canonical and changes only the first byte of the first compressed zstd frame:

```text
0x28 -> 0x29
```

The stored CRC, MCAP structure, indexes, footer, lengths, and compression metadata remain unchanged, so the failure reaches the actual zstd decompressor.

## Tests

Added coverage for:

* zstd decompression corruption producing the named refusal
* no user check or enrichment running after the integrity refusal
* CRC and decompression corruption both being discoverable through `check_runs.error`
* damaged canonicals not producing infrastructure ledger entries
* missing canonicals still remaining `failure_kind='infrastructure'`

Mutation validation was also run by temporarily removing the new decompression branch; the new regression test failed with the original `ZstdError`, then passed again after restoring the branch.

## Validation

```text
New regression tests:                  3 passed
Adjacent ingest suites:                26 passed
#502 integrity tests:                  4 passed
ruff check:                            passed
ruff format --check:                   passed
ty check:                              passed
full test suite:                       1,922 passed, 7 skipped
git diff --check:                      passed
```

### DIFF
--- a/src/hflow/reader.py
+++ b/src/hflow/reader.py
@@ -29,16 +29,18 @@ import numpy as np
 from mcap.reader import McapReader, make_reader
 from mcap.records import Attachment
 from mcap.stream_reader import CRCValidationError
+from zstandard import ZstdError
 
 logger = logging.getLogger(__name__)
 
 DEFAULT_BATCH_MAX_MESSAGES = 1024
 DEFAULT_BATCH_MAX_BYTES = 32 * 1024 * 1024
 
-# The named reason a file fails its own integrity stamp, returned by
+# Named reasons a file fails its own integrity stamp, returned by
 # :func:`verify_canonical_integrity` and recorded on the check lane's refusal
 # row, so downstream tooling can filter for damaged canonicals by exact value.
 CANONICAL_CRC_MISMATCH_REASON = "canonical-crc-mismatch"
+CANONICAL_DECOMPRESSION_FAILED_REASON = "canonical-decompression-failed"
 
 
 @dataclass(frozen=True)
@@ -339,7 +341,7 @@ def open_reader(path: Path | str, *, validate_crcs: bool = False) -> EpisodeRead
 
 
 def verify_canonical_integrity(path: Path | str) -> tuple[bool, str | None]:
-    """Validate one episode file's chunk CRCs with a strict full read.
+    """Validate one episode file's decompression and chunk CRCs with a strict full read.
 
     The check lane's front door. ``Episode`` reads run with CRC validation
     off (the reader docstring's trust argument covers bytes identified by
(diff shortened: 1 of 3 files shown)


kstonekuan (approved):
LGTM, merging.

One mutation I am merging past: widening the catch to `except Exception` passes all 7 tests. `Path.open` sits outside the `try`, so `test_missing_canonical_remains_an_infrastructure_failure` cannot constrain the catch width. Worth a cheap pin next time you are in the file.

#521 is assigned to you.

PR #518, opened 2026-09-12
refactor(transform): use _field_guards helpers for TransformConfig numeric settings

### LINKED ISSUE #510: TransformConfig hand-rolls three numeric guards, and its test matches the field name rather than the message

amansingh-121:
Follow-on from #487, closes #510.

`TransformConfig.__post_init__` hand-rolled three numeric guards. Each maps straight onto an existing `_field_guards` helper:

| parameter | before | helper |
| --- | --- | --- |
| `crf` | bool exclusion + `0 <= crf <= 51` | `require_int_in_range(..., minimum=0, maximum=51)` |
| `gop_seconds` | bool exclusion + finite + `> 0` | `require_positive_float` |
| `chunk_size_bytes` | bool exclusion + `> 0` | `require_positive_int` |

`gop_seconds` and `chunk_size_bytes` keep their `is not None` wrappers so `None` (the default) stays legal.

## Message changes

Accepted trade from #487: the messages now use `_field_guards`'s spelling, which names the offending value.

- `crf` range: `"between 0 and 51"` → `"in [0, 51]"` (plus `"got 52"` etc.)
- `gop_seconds`: `"positive and finite"` → split into `must be > 0` / `must be finite` (and the type message now spells `int or float`)
- `chunk_size_bytes`: `"positive"` → `"must be > 0"`

## Test tightening

The existing parametrized test asserted `match=field`, so `crf=True` and `crf=52` were indistinguishable. Each of the ten rows now asserts its own message with anchored regexes (e.g. `r"^crf must be in \[0, 51\], got 52$"`), so a type refusal and a range refusal cannot pass on each other's message.

## Definition of done

1. No `isinstance(..., bool)` left in `src/hflow/transform.py` — the three sites are gone (`rg` confirms zero).
2. `None` still accepted for `gop_seconds` and `chunk_size_bytes` (guards stay inside the `is not None` wrappers).
3. Ten rows each assert their own anchored message rather than the field name.
4. Each guard maps one-to-one onto a helper; neutering any one would turn exactly one case red (no overlaps between the three parameters' messages).
5. No behaviour change for valid configs — valid values pass through unchanged, so `compute_pipeline_version`'s content hash is unaffected.

Note: I could not run `pytest` locally (this Windows box fails importing `fcntl` in `storage.py`), so I verified the helper messages and valid-value passthrough against `_field_guards` directly, and ruff check + format pass. CI should cover the full suite.

### DIFF
--- a/src/hflow/transform.py
+++ b/src/hflow/transform.py
@@ -56,7 +56,6 @@ v1 behavior:
 import hashlib
 import json
 import logging
-import math
 from collections.abc import Callable, Mapping, Sequence
 from copy import copy
 from dataclasses import dataclass, field
@@ -64,6 +63,11 @@ from pathlib import Path
 from typing import Any, Literal
 
 from hflow import video as video_module
+from hflow._field_guards import (
+    require_int_in_range,
+    require_positive_float,
+    require_positive_int,
+)
 from hflow._grouped_mcap_writer import (
     NO_SCHEMA_ID,
     ChannelId,
@@ -156,28 +160,13 @@ class TransformConfig:
 
     def __post_init__(self) -> None:
         """Reject invalid settings before they become a pipeline identity."""
-        if isinstance(self.crf, bool) or not isinstance(self.crf, int):
-            raise ValueError(f"crf must be an int, got {type(self.crf).__name__}")
-        if not 0 <= self.crf <= 51:
-            raise ValueError(f"crf must be between 0 and 51, got {self.crf}")
+        require_int_in_range(self.crf, "crf", minimum=0, maximum=51)
 
         if self.gop_seconds is not None:
-            if isinstance(self.gop_seconds, bool) or not isinstance(self.gop_seconds, int | float):
-                raise ValueError(
-                    f"gop_seconds must be an int or float, got {type(self.gop_seconds).__name__}"
-                )
-            if not math.isfinite(self.gop_seconds) or self.gop_seconds <= 0:
-                raise ValueError(f"gop_seconds must be positive and finite, got {self.gop_seconds}")
+            require_positive_float(self.gop_seconds, "gop_seconds")
 
         if self.chunk_size_bytes is not None:
-            if isinstance(self.chunk_size_bytes, bool) or not isinstance(
-                self.chunk_size_bytes, int
-            ):
-                raise ValueError(
-                    f"chunk_size_bytes must be an int, got {type(self.chunk_size_bytes).__name__}"
-                )
-            if self.chunk_size_bytes <= 0:
-                raise ValueError(f"chunk_size_bytes must be positive, got {self.chunk_size_bytes}")
+            require_positive_int(self.chunk_size_bytes, "chunk_size_bytes")
 
 
 @dataclass(frozen=True)
--- a/tests/test_transform.py
+++ b/tests/test_transform.py
@@ -49,24 +49,36 @@ def test_pipeline_version_is_a_content_hash() -> None:
 
 
 @pytest.mark.parametrize(
-    ("construct", "field"),
+    ("construct", "message"),
     [
-        (lambda: TransformConfig(crf=True), "crf"),
-        (lambda: TransformConfig(crf=-1), "crf"),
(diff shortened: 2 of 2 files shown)


kstonekuan:
Closing this one on first-PR-wins: #511 opened against #510 about eight hours earlier and covers the same three parameters. Nothing wrong with your version, and the table in your description is the clearer of the two write-ups.

The tell to save you this next time is on the issue page rather than the label: a linked pull request shows up in the sidebar, and it costs a glance before starting. #510 had one from 03:11.

Worth saying directly, since this is the second time in two days: you are three merges in now (#496, #501, and the two before), which puts you past the `good first issue` pool. Those exist so newcomers have somewhere to land, and there are four of them left. I would rather you took the work nobody else can, and yesterday on #482 I pointed at two things without much of a case for either, so here it is properly:

- **#521**, filed an hour ago. Our pinned Linux ffmpeg build has 404ed, so every fresh Linux install fails before its first video operation. The fix is a tag, a label and two checksums, but the decision underneath it is whether a month-final pin gets enforced or just remembered, and whether anything should notice the next expiry before a user does.
- **#506**, the zstd-refusing half of #474, still unassigned and labelled advanced.
- The corpus one. Kaileshwar16's run of Egocentric-10K has produced two real bugs this week (#499 and #521). That is the highest-yield thing anyone is doing here and it does not need my permission.

Say which and I will assign it before someone else starts.

amansingh-121:
@kstonekuan understood on first-PR-wins — and thank you for the clear write-up. Both of the ones I have shipped were caught by ordering, so that framing saves me the same mistake next time.

I would rather take the work nobody else can, and the real-corpus direction (running Egocentric-10K/100K and reporting what breaks or drags, the way the #499/#521 bugs showed up) is the one that interests me most. The specific ones you listed are looking taken now, so could you point me at whatever is currently open and unassigned at that level — or assign me a starting point on the corpus effort? I will take it the moment you say which.

PR #522, opened 2026-09-13
fix(egocentric): derive episode identity and operator from the source

### LINKED ISSUE #519: [Bug]: prepare.py: running a second factory shard silently overwrites the first corpus and stamps false provenance
### Version or commit

7ed172a

### Environment

### Minimal reproduction

1. Run prepare.py against the factory_002 manifest:
   uv run python examples/egocentric/prepare.py --manifest manifest-f002.json --output-root data/out
   Reports "prepared 16/16". Landing holds 16 episodes as factory_051_episode_0001.mcap ... _0016.mcap.

2. Run the same command against the factory_012 manifest, same output root:
   uv run python examples/egocentric/prepare.py --manifest manifest-f012.json --output-root data/out

3. Observe: the second run also reports "prepared 16/16" with a success message,
   but the output root still holds 16 files, not 32.

### Expected behavior

Two different shards should produce 32 coexisting episodes in one output root.

Episode ids should be derived from the source identity so two shards never share a landing filename. Provenance on every episode should name its real factory and the worker that actually recorded the footage.

### Actual behavior

The output root holds 16 files, not 32. Manifest A's corpus is gone with no error and no warning.

The destroying run prints a success message ("prepared 16/16").

Every survivor is factory_012 footage but its episode/v1 metadata stamps operator: factory_051_worker_001, a worker who never recorded that footage.

### Additional context

Root cause: prepare.py:168 builds the episode id from a hardcoded factory_051 constant, and :494 stamps a hardcoded operator. The landing filename is that id plus .mcap (line 608), so every corpus writes the same filenames.

episode/v1 exists so provenance names what was actually recorded. This asserts an operator who never recorded the footage, and the catalog rows built from these episodes inherit the lie. Three shards cannot share one output root without losing two of them.

Fix direction: derive the episode id from the source member stem plus the source identity hash, the same collision-resistant pattern App.process uses via _source_artifact_directory_name (app.py:311), whose docstring already states that source basenames are not identities. Derive operator from the per-clip JSON sidecar, which already carries worker_id.

Found during a 3-factory field run on Egocentric-10K footage. Full numbers in discussion #503.

Definition of done: manifests A and B against two shards produce 32 coexisting episodes; provenance names the real factory and worker; a test pins the two-shard case; single-default-shard output stays byte-identical to today's.

Sagar-024:
# fix(egocentric): derive episode identity and operator from the source

Closes #519. Found in the multi-factory field run (discussion #503): the example hardcodes every episode id to `factory_051` and the operator to `factory_051_worker_001`, so a second factory's shard silently overwrites the first corpus and the survivors carry provenance from a worker who never recorded them.

## What changed

- Episode ids derive from the source member stem plus a digest of the pinned
  archive sha256, the member name, and the episode's window start. This mirrors the collision-resistant pattern `App.process` uses for run directories (`_source_artifact_directory_name`, app.py:311), extended with the window start because one source member yields many excerpt windows. Two shards or two windows can never share a landing filename.
- `operator` is composed from the per-clip sidecar's `factory_id` and
  `worker_id`. The sidecar rides in the same pinned tar as the video (identical stem, `.json` suffix) and is read in the same `_extract_source_videos` pass: one extract call, no new IO. A missing or unusable sidecar fails loudly, naming the member and the field.
- A `factory` metadata field records `factory_id` explicitly on every
  episode.

## Three assumptions, redirect me if any are wrong

1. The identity digest seeds from the archive sha256, the member name, and the window start. If the digest should seed from something else (for example a content hash of the video bytes), say so and I will re-cut it. 2. `operator` keeps the `factory_X_worker_Y` composition so existing consumers see the same shape, now truthful per source. 3. A new `factory` key is added to `episode/v1`. The record already carries example-specific keys (`injected_fault`, `task_completion`), and the canonical transform plus checks accept the field (the suite covers the full canonical path with it present).

## Invariance proof

The MCAP message payloads are byte-identical before and after the change. The full default-shard corpus (96 episodes) was rebuilt with the new code and compared against the existing landing episode-for-episode, matched by source member and window start: 96 of 96 payload byte-identical. Only ids, operator, the new `factory` field, and filenames changed.

Honest note: the first implementation of this fix collapsed those 96 windows into 11 files (one per member, the exact silent-overwrite shape this issue describes, reintroduced by my own id scheme). The default-shard rebuild caught it, and the window start moved into the digest seed.

## Tests

- Two-shard coexistence: two factories' shards into one output root produce
  two coexisting episodes with distinct derived ids and truthful provenance (the old code produced one, silently destroying the other).
- Single-shard provenance: operator and factory name the real sidecar values.
- The suite also pins the derived-id shape: each landing filename carries the
  source member stem.

Validation: ruff check, ruff format, ty on the changed files; 4 passed in tests/test_egocentric_prepare.py; zero non-ASCII and zero em dashes in the changed files.

### DIFF
--- a/examples/egocentric/prepare.py
+++ b/examples/egocentric/prepare.py
@@ -50,6 +50,14 @@ class DatasetSource:
 
 
 @dataclass(frozen=True)
+class SourceIdentity:
+    """The factory and worker that produced one source video, from its sidecar."""
+
+    factory_id: str
+    worker_id: str
+
+
+@dataclass(frozen=True)
 class SourceArchive:
     path: str
     sha256: str
@@ -143,6 +151,7 @@ def _parse_fault_segment(value: object, context: str) -> tuple[float, float]:
 def _expand_episode_plan(
     sources: tuple[SourceVideo, ...],
     episode_plan: EpisodePlan,
+    archive_sha256: str,
 ) -> tuple[PlannedEpisode, ...]:
     planned_faults_by_episode_number = {
         planned_fault.episode_number: planned_fault for planned_fault in episode_plan.faults
@@ -163,9 +172,18 @@ def _expand_episode_plan(
             )
 
         planned_fault = planned_faults_by_episode_number.get(episode_number)
+        # Mirror App's _source_artifact_directory_name: source basenames are not
+        # identities, so the id carries a digest of the pinned archive, the member
+        # name, and this episode's window start (one member yields many excerpt
+        # windows). Two shards or two windows can never share a landing filename,
+        # and the episode number stays in the id for readability.
+        source_stem = Path(source_video.member).stem
+        source_identity_digest = hashlib.sha256(
+            f"{archive_sha256}:{source_video.member}:{source_start_s}".encode()
+        ).hexdigest()[:12]
         episodes.append(
             PlannedEpisode(
-                episode_id=f"factory_051_episode_{episode_number:04d}",
+                episode_id=f"{source_stem}-{episode_number:04d}-{source_identity_digest}",
                 source_member=source_video.member,
                 source_start_s=source_start_s,
                 duration_s=episode_plan.duration_s,
@@ -282,7 +300,7 @@ def _load_manifest(manifest_path: Path) -> CorpusManifest:
                 f"fault segment for episode {planned_fault.episode_number} ends after the episode"
             )
 
-    episodes = _expand_episode_plan(sources, episode_plan)
+    episodes = _expand_episode_plan(sources, episode_plan, archive.sha256)
 
     return CorpusManifest(
         schema_version=2,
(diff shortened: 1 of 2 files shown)


kstonekuan (changes requested):
The invariance proof and the near-miss note are the right way to report this.

Two mutations pass that should not:

```
digest dropped from the id entirely      4 passed
sidecar field validation removed         4 passed
```

Your two shards use different member stems (`factory002_worker001_00000`, `factory012_worker003_00000`), so `{source_stem}-{episode_number}` already separates them and the digest never does the work. Its stated job is that source basenames are not identities. Either cover the case where that bites, or drop it.

The sidecar refusals have no case at all.

Not a gap, for the record: dropping the window start from the seed also passes, and that is correct now, since `episode_number` is global and disambiguates on its own. Your near-miss is not reachable by that route any more.

Sagar-024:
Both mutation gaps from your review are now pinned.

Mutation A, digest dropped: the existing shards used different member stems, so {source_stem}-{episode_number} already separated them and the digest never fired. Added test_same_member_stem_from_two_shards_never_collides: two different archives whose members share one stem, prepared into a shared output root. With the digest dropped, both episodes land on the same filename and the second silently overwrites the first (red, one file left). Restored, two distinct episodes coexist with their own provenance (green). The old two-shards test still passes under the mutant, reproducing your finding.

Mutation B, sidecar validation removed: added test_unusable_sidecar_refuses_the_source, parametrized over both refusal branches (sidecar member absent; sidecar present with an empty worker_id). With the validation removed, the empty-field case no longer raises and stamps operator "factory_002_" (red). Restored, both branches refuse loudly naming the member and field (green).

Verified red then green for each. Full suite 1925 passed, 6 skipped, 0 failed. ruff check, ruff format, ty clean on the changed file; 0 non-ASCII, 0 em dashes. Only tests/test_egocentric_prepare.py changed; prepare.py untouched.

kstonekuan (approved):
LGTM, merging.

Both gaps pinned, and I checked a third: making the sidecar member lookup never raise also fails `test_unusable_sidecar_refuses_the_source`, so both refusal branches hold, not just the field check.

`test_same_member_stem_from_two_shards_never_collides` is the one that gives the digest a reason to exist. Two archives sharing a member stem is exactly what "source basenames are not identities" means, and it was the only case that could show it.

PR #523, opened 2026-09-13
refactor(video-import): compose VideoImportConfig guards from _field_guards

### LINKED ISSUE #508: VideoImportConfig hand-rolls three numeric guards, and only one of them maps straight onto a helper

mulkakhileshmj:
Closes #508.

## What changed

`VideoImportConfig.__post_init__` now composes its numeric guards from `_field_guards` instead of hand-rolling the bool exclusion three times, split exactly along the line the issue draws between shared rules and domain rules:

- The finite-number loop over `duration_s`, `source_start_s`, and `image_hz` is `require_finite_float(value, name)`, which distinguishes the wrong-type and non-finite cases the old single message conflated.
- `image_width` and `image_height` call `require_positive_int`, then keep the evenness check local with its own message (`image_width must be even, got 3`). A code comment records why it stays: H.264 4:2:0 chroma subsampling is a domain rule about these two fields, not a shared numeric guard. No `even=` parameter was added to `_field_guards`, per the issue.
- `start_time_ns` is `require_int_in_range(..., minimum=0, maximum=_MAXIMUM_TIMESTAMP_NS)`, whose message now names the bounds and the offending value instead of the fixed "unsigned 64-bit integer" sentence.

No `isinstance(x, bool)` remains in `video.py`, and no behavior changes for valid configurations. Message wording moves to the helpers' spelling, the trade #487 established.

## The tests

`test_invalid_import_configuration_is_rejected` kept its parametrized shape and gained an anchored `match=` per row, plus the rows the old one-message guards could not distinguish: `True`, a string, zero, negative, odd, `False`, and one past the range. Every refusal the file made before, it still makes.

One pre-existing weakness surfaced by the tightening, worth a reviewer's eye: the old `start_time_ns=(1 << 64) - 1` row looked like a range case but actually passes the range guard (the maximum is inclusive) and is refused later by the excerpt's final sample exceeding the MCAP timestamp range. Without `match=` the row proved nothing about which guard fired; it now asserts the excerpt-range message and carries a comment saying so, and a genuine one-past-the-range row (`1 << 64`) pins the range guard itself.

## Validation

```
uv sync --locked
uv run pytest -q           # 1930 passed, 6 skipped
```

Mutation, per the definition of done: each guard neutered in turn (the finite-float call, the positive-int call, the evenness check on its own, and the range bounds) turns its rows red; restored, everything passes. The touched test file also passes under Python 3.11.

### DIFF
--- a/src/hflow/importers/video.py
+++ b/src/hflow/importers/video.py
@@ -12,6 +12,7 @@ from foxglove_schemas_protobuf.CompressedImage_pb2 import CompressedImage
 from mcap.writer import Writer
 from mcap_protobuf.schema import build_file_descriptor_set
 
+from hflow._field_guards import require_finite_float, require_int_in_range, require_positive_int
 from hflow._pinned_asset import sha256_hex_of_file
 from hflow.ffmpeg import ffmpeg_path, ffmpeg_version
 from hflow.ffmpeg._process import media_input_was_rejected, run_media_command
@@ -52,10 +53,7 @@ class VideoImportConfig:
             ("source_start_s", self.source_start_s),
             ("image_hz", self.image_hz),
         ):
-            if isinstance(value, bool) or not isinstance(value, int | float):
-                raise ValueError(f"{name} must be a finite number")
-            if not math.isfinite(value):
-                raise ValueError(f"{name} must be a finite number")
+            require_finite_float(value, name)
         if self.duration_s <= 0 or self.source_start_s < 0:
             raise ValueError("duration_s must be positive and source_start_s nonnegative")
         if not 0 < self.image_hz <= NANOSECONDS_PER_SECOND:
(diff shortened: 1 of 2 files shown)


kstonekuan:
Resolved in #512.

PR #525, opened 2026-09-13
fix(ffmpeg): pin a month-final BtbN build and notice the next expiry

### LINKED ISSUE #521: [BUG]: Pinned Linux FFmpeg build disappears under BtbN's retention policy
Pinned Linux FFmpeg build disappears under BtbN's retention policy

---

## Version or commit

Current `main`.

```bash
git rev-parse HEAD
```

---

## Environment

```text
OS: Arch Linux
Architecture: x86_64
Python: 3.12
FFmpeg override used for isolation: /usr/bin/ffmpeg, n8.0.1
```

---

## Minimal reproduction

Found while running the real-corpus test from #287 on a fresh Linux setup.

The Egocentric-10K shard downloads successfully, but the first video operation fails when HFlow tries to install its pinned FFmpeg build:

```bash
```

The failure is:

```text
PinnedDownloadError: failed to download the pinned ffmpeg build from
https://github.com/BtbN/FFmpeg-Builds/releases/download/autobuild-2026-08-16-13-00/ffmpeg-n8.1.2-44-g7c533d0f86-linux64-gpl-8.1.tar.xz:
HTTP Error 404: Not Found
```

The exact pinned URL also fails independently:

```bash
curl -I -L \
'https://github.com/BtbN/FFmpeg-Builds/releases/download/autobuild-2026-08-16-13-00/ffmpeg-n8.1.2-44-g7c533d0f86-linux64-gpl-8.1.tar.xz'
```

It returns:

```text
HTTP/2 404
```

The current pin in `src/hflow/ffmpeg/_binary.py` is:

```python
PINNED_RELEASE_TAG = "autobuild-2026-08-16-13-00"
PINNED_VERSION_LABEL = "n8.1.2-44-g7c533d0f86"
```

---

## Expected behavior

The pinned/checksum-verified FFmpeg build used by HFlow on Linux should remain downloadable for the lifetime for which HFlow expects that pin to be usable.

A fresh Linux setup should therefore be able to resolve the pinned FFmpeg binary without requiring an explicit system-FFmpeg override.

---

## Actual behavior

The pinned BtbN release has disappeared and now returns HTTP 404, so HFlow's default Linux FFmpeg provisioning path fails before the first video operation can run.

BtbN documents its release retention as:

* last 14 daily builds are retained
* last build of each month is retained for two years
* `latest` is floating

The current HFlow pin points at an arbitrary daily autobuild, so this can happen again if the fix is only to move to another daily build.

---

## Additional context

The failure appears isolated to FFmpeg provisioning rather than the dataset or Egocentric preparation path.

Using the documented override:

```bash
HFLOW_FFMPEG=/usr/bin/ffmpeg \
```

with system FFmpeg `n8.0.1`, the same preparation completed successfully:

```text
96/96 episodes prepared
exit status 0
```

Measured on that run:

```text
wall time: 6:20.47
CPU:       651%
peak RSS:  216364 KB
swaps:     0
```

I don't think silently falling back to the system FFmpeg is the right fix, since HFlow deliberately pins FFmpeg for reproducibility.

The underlying issue seems to be that the lifetime of the upstream asset is shorter than the lifetime expected from the pin.

Possible directions are:

* use a BtbN asset with an appropriate documented retention lifetime, or
* mirror/host the exact pinned artifacts somewhere whose retention HFlow controls.

There is also already an opt-in real-download test behind `HFLOW_NETWORK_TESTS=1`; running that periodically or during release validation could catch a dead external pin before users encounter it.

mulkakhileshmj:
Closes #521.

## Problem

The pinned Linux build was `autobuild-2026-08-16-13-00`, a mid-month daily. BtbN retains the last ~14 dailies plus each month's final build (kept two years), so the pin 404ed about a fortnight after it was chosen, and every fresh Linux install failed before its first video operation. CI installs the distro binary, so nothing we watch saw it.

## The pin

Moved to `autobuild-2026-08-31-13-27`, the month-final closest to the expired pin, version label `n8.1.2-50-g1a748fe2cd`. Both checksums were computed from downloads performed for this change, not taken from a release page:

```
c733b4b2951e5957e15505f788b2c65a7a41b6da4b289e295852cc38079b4d2b  linux64-gpl-8.1.tar.xz
ae5da4f51b9052390f414005f8ab26c1eed1268f327cce7cb79aa076b29bd66e  linuxarm64-gpl-8.1.tar.xz
```

The extracted x86_64 binary reports `ffmpeg version n8.1.2-50-g1a748fe2cd-20260831`, so the existing `PINNED_VERSION_LABEL in version_line` assertions hold. The comment above the pin now states the retention rule alongside the immutability one, since immutable was true and was not the property that failed.

## The two decisions, answered

**Enforced, not remembered.** `test_pinned_release_tag_is_a_month_final_build` is an offline test refusing any pin whose tag date is not the last calendar day of its month. The evidence for that exact rule: every surviving month-final on the release list (2026-02 through 2026-08, February's 28th included) falls on the last calendar day. The honest edge: BtbN's real rule is "last build of the month", so if a month-end build is ever skipped, the true month-final lands a day early and the test refuses a correct pin. That false refusal is deliberate and the docstring says so: it forces whoever updates the pin to re-verify retention rather than trust the tag's shape, which is the failure direction an availability property wants.

**Noticed before a user finds it.** Two parts. `test_pinned_urls_are_reachable` HEADs every pinned URL on the existing `HFLOW_NETWORK_TESTS=1` gate, cheap enough to run anywhere. And a new scheduled workflow (`.github/workflows/pinned-assets.yml`, weekly plus manual dispatch) runs exactly that test and the month-final check, so an expiry becomes a red run instead of a fresh install's 404. Weekly is load-bearing: dailies live fourteen days, so a wrongly pinned daily is caught while the asset is still downloadable and the fix is a calm re-pin rather than an archaeology exercise. The workflow reuses ci.yml's pinned action SHAs; it is the "needs someone to actually run it" trade-off resolved by having the schedule be the someone.

## Validation

```
uv sync --locked
uv run pytest -q           # 1924 passed, 7 skipped
HFLOW_NETWORK_TESTS=1 uv run pytest -q tests/test_ffmpeg.py -k "reachable or real_pinned"
```

The network run performed the real end-to-end proof against a cleared cache: both HEADs return 200, the full download verifies against the new checksums, extracts, and the installed ffmpeg and ffprobe both report the pinned label. Mutation: re-pinning to the current daily `autobuild-2026-09-12-13-12` turns the month-final test red with the message explaining the retention trap; restored, green.


Kaileshwar16:
@mulkakhileshmj i was assigned for the issue

kstonekuan:
Resolved in #526.

It bumped `TRANSFORM_BEHAVIOR_VERSION`; yours did not. `compute_pipeline_version` does not hash `ffmpeg_version`, so without it the old and new encoder builds share one pipeline identity.

PR #531, opened 2026-09-14
fix(video): isolate concurrent remux temp files

### LINKED ISSUE #530: video cache: concurrent remuxes share one fixed-name .tmp file, silently poisoning the frame cache with torn MP4s
The video remux cache has a crash-safe-but-not-concurrency-safe invariant. write_access_units_to_mp4 writes to a deterministic temp name, output.name + ".tmp" (video.py:626), so every process remuxing the same video opens the same inode. Under concurrent same-source execution (a designed-for lane: Airflow retries:5, per-source shared workdir), two remuxes truncate and interleave writes into one file, then one publishes the torn bytes to the final path. The cache then treats those bytes as a completed remux forever.

## What breaks
- video.py:626 uses a fixed temp name, identical for every writer of that output.
- video.py:657-658 success check tests the shared path (is_file, size>0), not the bytes this caller wrote.
- episode.py:551-554 caches on output.exists(), with the comment "a file at the final path is always a completed remux."
- A late replace() publishes the interleaved inode; a still-open fd keeps writing into the renamed final file after the "completed" replace.

## Proof, executed not theorized
60 trials x 5 concurrent processes on one canonical episode:
- 3 trials: silent corruption. A fresh consumer extracted frames "successfully" but got 6 frames instead of 14, with different JPEG bytes. Zero errors anywhere.
- 9 trials: half-state. Writer returned success and cached the final; the first real consumer died with "moov atom not found."
- The reference remux is deterministic (two single-process runs byte-identical), so "6 frames vs 14" is provably corruption, not variance.

## Why it matters
The poisoned cache sits in the shared per-source scratch dir. Contact sheets, frame-based checks, and VLM steps all read it via Episode.frames(). The catalog then records measurements derived from video bytes that exist in no one's episode. The evidence chain lies, and the 3 silent trials prove it lies without a single error line.

## Fix direction
Create the temp uniquely per call via tempfile.mkstemp(dir=output.parent), exactly the pattern this repo already uses for parallel writers (storage.py:221-234 NamedTemporaryFile+replace, storage.py:338+ store_file_if_absent). Keep the final replace atomic; unique temp names make the interleaved-inode path unreachable.

## Definition of done
- Two-process test: concurrent Episode(...).video(topic) on a shared workdir; both return, cached final equals the single-process reference by sha256, and the full frame count extracts.
- A regression test that fails red if the temp name is reverted to output.name + ".tmp" (mutation proof).
- The episode.py:551 comment's claim becomes true by construction.
- Existing video/episode/checks suites stay green.

harshitethic:
### Summary

- replace the deterministic `<output>.tmp` remux path with a unique sibling temporary file per call
- keep the existing atomic `replace()` behavior and failure cleanup
- prevent concurrent remuxes targeting the same cache path from truncating or unlinking each other's ffmpeg output
- add a regression test that runs two remuxes concurrently, synchronizes them at the subprocess boundary, and asserts they use distinct temp files while producing one valid final cache file

### Verification

- `git diff --check`
- Python compile check for the modified video module and concurrency regression test
- helper workflow removed and branch squashed to one focused commit

Fixes #530


kstonekuan (changes requested):
`ruff check` fails on `tests/test_video_concurrency.py` (missing annotations on `tmp_path`, `monkeypatch`, `fake_run`), so this would turn main red.

The fix is right. The test is weaker than #530 asks for: with `subprocess.run` mocked it pins that two calls get distinct temp names, which catches a revert to the fixed name but cannot catch what the issue measured. #530 wants the cached final to match a single-process reference by sha256 and the full frame count to extract.

The barrier and the `not path.exists()` check are the right instincts. Worth saying whether the sha256 version is practical here or not.

Sagar-024:
Thanks for the fast turnaround on this one, and for matching the repo's existing idiom rather than inventing a new one: NamedTemporaryFile in output.parent plus the unchanged atomic replace is exactly the shape storage.py:221-234 already uses for parallel writers.

What I verified against the diff:

- Unique per call: the temp now comes from tempfile.NamedTemporaryFile(prefix=f".{output.name}.", suffix=".tmp", dir=output.parent, delete=False). No code path derives the name from the output alone anymore, so two concurrent remuxes can no longer share an inode.
- Atomic replace preserved: dir=output.parent puts the temp on the same filesystem as the final path, so the untouched temporary_output.replace(output) at the end stays atomic on POSIX. Crash-safety is kept, not traded away for concurrency-safety.
- The success check is now honest: video.py:657-658 tested a shared path before (is_file, size>0 could pass on another process's bytes); with a unique temp those checks can only see what this caller's ffmpeg wrote. Same for the except path's unlink: it now removes its own file instead of potentially deleting the next attempt's temp out from under ffmpeg.
- Cleanup on success: the final replace consumes the temp, so no orphan on the happy path.

The critical one, the regression test: I confirmed it pins the race. Mutating the fix back to the deterministic output.name + ".tmp" makes both threads record the same command[-1], so assert len(set(seen)) == 2 fails deterministically, not flakily. And the test is honestly structured: the barrier inside fake_run forces the two calls to overlap rather than merely running fast, it asserts both callers return the same final path, the final holds complete bytes, and no temp file survives. Faking subprocess.run to inspect the path the muxer would receive is the right tradeoff here: it tests the mechanism (distinct temp per call) deterministically instead of betting CI on a real inode race that only shows up a few times in sixty.

One thing to fix before merge: checks (3.14) is red on lint, 6 ANN hits, all in the new test file: ANN001 on tmp_path and monkeypatch (line 11) and on fake_run's command, input, capture_output plus ANN202 on its return (line 20). Annotating them (tmp_path: Path, monkeypatch: pytest.MonkeyPatch, def fake_run(command: list[str], *, input: bytes, capture_output: bool) -> SimpleNamespace) should clear it. Same gate bit me on my own test file last week, so the ruff config on tests is easy to miss.

Minor note, not a request: with unique names, a remux that gets SIGKILLed mid-write now leaves one orphan .<name>.XXXX.tmp per aborted attempt instead of having the next attempt truncate the stale one. That is the same tradeoff the storage.py atomic writers already accept, and App.process rmtree's the scratch dir whenever the canonical is rewritten, so it is bounded. Fine as is.

The invariant now holds by construction: any file at the final path is some single caller's complete output, which is what the cache at episode.py:551 assumed all along. Nice catch and a clean fix.

harshitethic:
Addressed the review feedback in `2c21a02`.

The concurrency regression now uses real ffmpeg/ffprobe instead of mocking `subprocess.run`: it generates a 60-frame H.264 stream with B-frames disabled, creates a single-process MP4 reference, runs two remuxes concurrently against the same final cache path, then verifies the final SHA-256 matches the reference and ffprobe reports all 60 frames. It also checks that no per-call temp files remain. The test is skipped only when system ffmpeg/ffprobe is unavailable.

Validation on Linux with ffmpeg 6.1.1:
- `uv run --with pytest pytest -q tests/test_video_concurrency.py` → 1 passed
- `uv run ruff check tests/test_video_concurrency.py src/hflow/video.py` → passed
- `uv run ty check tests/test_video_concurrency.py src/hflow/video.py` → passed
- `git diff --check` → passed

kstonekuan (approved):
LGTM, merging.

The real-ffmpeg version is what #530 asked for: sha256 against a single-process reference plus an ffprobe frame count, so it catches interleaving and not just naming.

Both mutations bite, including replacing `replace()` with a copy, which leaves the temp behind.

PR #538, opened 2026-09-14
fix(catalog-ui): rebind empty tables when mid-session Parquet lands

### LINKED ISSUE #537: [Bug]: catalog ui keeps empty in-memory ingest_failures after the ledger writes Parquet mid-session
### Version or commit

802c06b6eb41ed48ca8e96dfe8fce1b6dd342707

### Environment

Linux 6.8.0-138-generic x86_64, Python 3.13.3 (uv), no Docker

### Minimal reproduction

`hflow catalog ui` keeps one DuckDB connection for the process lifetime. Empty relations start as in-memory `CREATE TABLE` shells because DuckDB refuses a Parquet glob with no matches (`_register_catalog_relations` in `src/hflow/curation.py`).

The UI refreshes that surface **once**, when the first `episodes/*.parquet` appears (`serve_catalog_ui` in `src/hflow/catalog_ui.py`). After that the poll loop only `_sync_catalog_mirror` for buckets — it never rebinds local empty tables.

`_refresh_local_catalog_connection` states the assumption that breaks the ledger:

> Once the episodes file from the first append exists, all dependent table files are complete too

That holds for tables `Catalog.append_episode` writes before episodes. It does **not** hold for `ingest_failures`, which `record_ingest_failure` writes independently (`src/hflow/ingest_ledger.py`).

Controlled result on `802c06b` (same connection the UI would keep open):

1. `open_catalog_connection` on an empty catalog → `ingest_failures` is `BASE TABLE`, count `0`.
2. Append one episode; call `_refresh_local_catalog_connection` (what the UI does on first append) → `episodes_raw` is a Parquet `VIEW`; `ingest_failures` is still an empty `BASE TABLE`.
3. `record_ingest_failure(...)` so `ingest_failures/<digest>-<fp>.parquet` exists on disk.
4. Same connection: `SELECT count(*) FROM ingest_failures` → **0**, still `BASE TABLE`.
5. Fresh `open_catalog_connection` on the same root → `VIEW`, count **1**.

Failure-only workspace (ledger rows, no episode yet): `_completed_append_exists` stays false, so the UI wait loop never refreshes. Open connection count `0`; fresh connection count `1`. Restart is the only recovery.

### Expected behavior

A long-running `catalog ui` session shows `ingest_failures` rows once their Parquet exists under the catalog root (and after a bucket mirror sync), without restarting the command.

A failure-only catalog should not stay stuck on empty in-memory shells forever just because no `episodes/*.parquet` has landed.

### Actual behavior

After the one-shot first-append refresh, any relation that was still empty becomes a sticky in-memory table. Later `ingest_failures` Parquet is invisible until the UI process is restarted.

The docstring assumption that “episodes last ⇒ every table is complete” does not cover the ingest ledger.

### Definition of done
- With `catalog ui` (or the same open connection + refresh policy) already past the first episode append, recording an ingest failure makes `SELECT count(*) FROM ingest_failures` return the new row without reconnecting.
- A catalog that has only `ingest_failures` Parquet (no episodes yet) becomes queryable for that table without requiring an episode append or a process restart.
- A regression test covers mid-session ledger Parquet appearing after the first-append refresh (episodes-only refresh is already covered in `tests/test_catalog_ui.py`).

### Non-goals
- Redesigning continuous poll-and-rebind for every table on every new file if a narrower fix is enough (rebind empty `BASE TABLE` relations when their directory gains Parquet; treat ledger presence as a refresh trigger).
- Changing how `record_ingest_failure` chooses paths or fingerprints.
- Bucket write behavior while browsing.
- #481 (curation gate double-parse).

Independent of the current open board (#536, #535, #530, #529, #528, #498, #491, #481, #447, #376, #287, #246). No open PR targets this.

VARUN3WARE:
## Summary

`hflow catalog ui` rebinds empty in-memory catalog tables when their Parquet appears mid-session, so `ingest_failures` (and any other long table still stuck as a `BASE TABLE` shell) become queryable without restarting the UI. Closes #537.

## Why

Empty catalogs register in-memory `CREATE TABLE` shells because DuckDB refuses a Parquet glob with no matches. The UI used to refresh that surface only once — when the first `episodes/*.parquet` landed — under the assumption that every dependent table was complete then.

That assumption is false for `ingest_failures`, which the ledger writes on its own schedule. After the one-shot refresh, later ledger Parquet stayed invisible on the open connection (`BASE TABLE` count 0) until process restart. Failure-only catalogs never refreshed at all.

This keeps the same full re-register path and only widens the poll trigger: any empty long-table shell whose directory now has Parquet.

## Validation

```bash
uv run ruff check src/hflow/catalog_ui.py src/hflow/curation.py tests/test_catalog_ui.py
uv run ruff format --check src/hflow/catalog_ui.py src/hflow/curation.py tests/test_catalog_ui.py
uv run ty check src/hflow/catalog_ui.py src/hflow/curation.py
uv run pytest -q tests/test_catalog_ui.py
```

Result: ruff and ty clean; **28 passed**.

## Checklist


kstonekuan (approved):
LGTM, merging.

Restoring the `break` fails both new tests, which is the regression that mattered: the old code rebound once and then stopped looking.

`test_catalog_ui_binds_failure_only_catalog_without_episodes` is the case the old trigger could never reach, since a run that produces only failures never writes an episode.

Full suite on the rebased result: 2012 passed / 8 skipped. You ran the targeted one.

PR #539, opened 2026-09-15
fix: bind scratch caches to canonical content

### LINKED ISSUE #536: [Bug] : sync-omitted stage runs measure stale scratch cache, poisoning the catalog with old evidence under new content hashes
### Version or commit

Current main

### Environment

Ubuntu 24.04 (WSL2), Python 3.12, x86_64, file:// obstore root

### Minimal reproduction

Using a file:// bucket-backed storage root with independent worker mirrors:
1. Worker 1 runs the full profile on clean footage (v1). Scratch is populated, measurements pass (black_frame_pct=0.00).
2. The canonical object in the remote store is overwritten with altered camera payload (v2, bumped pipeline_version, true black_frame_pct=50.00). Worker 2 processes this and records the true v2 measurements.
3. Worker 1 runs a sync-omitted profile (e.g., META + MEDIA only). Its etag-verified fetch downloads the new v2 canonical, but the scratch directory is not invalidated.

### Expected behavior

Worker 1 fetches the new v2 canonical, invalidates or bypasses the stale scratch cache, measures the v2 pixels, and records the true v2 measurements (50.00) under the new episode_id.

### Actual behavior

Worker 1 fetches the v2 canonical but keeps the v1 scratch (identical sha and mtime_ns). It measures the v1 pixels and records v1 measurements (0.00) under the v2 episode_id. 

Because the stale run's measurement values differ from the truth, its `_run_fingerprint` is unique, so the catalog appends it as a new row. `measurements_latest` ranks by `recorded_at`, so the stale, newer measurement wins the wide view. 

Worse, the poisoning is self-perpetuating: if a fresh worker (Worker 3) later tries to repair this by re-measuring, it produces the *true* values, which match the original v2 fingerprint. The catalog dedupes the repair (`written=False`), and it fails to overwrite the stale latest view. The system actively protects the lie.

### Additional context

Root cause: The only scratch invalidation is `shutil.rmtree(scratch_dir)` at `app.py:2391`, guarded by `if Stage.SYNC in enabled_stages` (`app.py:2347`). The sync-omitted route (`app.py:2395-2412`) fetches the new canonical but never touches the scratch directory. `Episode.video()` relies on a bare `output.exists()` check (`episode.py:551-554`), serving the stale remux.

Why it matters: Curation cuts and gates accept or reject episodes using footage that isn't the footage. The downstream training corpus inherits mislabeled quality verdicts that survive the exact byte-change the content-hash was supposed to catch. The append-only deduplication logic (`_run_fingerprint`) weaponizes the repair path, making the poison durable.

Fix direction: Make the scratch content-bound: key `scratch_dir` on `content_episode_id(canonical_path)`, or drop a digest marker inside the scratch dir and remux whenever it disagrees with the fetched canonical. Additionally, the measurement deduplication logic needs a retraction or force-overwrite path for later-arriving stale appends that lose to the truth.

Definition of done:
- A sync-omitted run on a worker with stale scratch correctly measures the newly fetched canonical bytes, not the old ones.
- `_run_fingerprint` deduplication does not allow a stale measurement to permanently shadow a true measurement in the `latest` views.
- Mutation proof: tamper the canonical payload without touching scratch, and the recorded measurements must change.
- Existing app/storage/catalog suites stay green.

Amidwestnoob:
## Summary

Refs #536 — **this fixes stale scratch reuse only, not the catalog repair/deduplication trap.**

- Bind remux/frame scratch to `run_dir/scratch/<episode_id>/` after canonical rewrite, reuse, or sync-omitted fetch has finalized the bytes.
- Preserve unchanged-content cache reuse and full scratch cleanup after a sync rewrite.
- Reuse the calculated content ID for quarantine lookup rather than hashing twice at that App-level path.
- Add synthetic regressions covering local META/MEDIA processing, two independent mirrors of a `file://` store, and cache reuse/cleanup.

## Why

A sync-omitted run can fetch a new canonical into a source-keyed run directory while retaining remuxes from the previous recording. Content-keyed scratch prevents those old pixels from being measured under the new episode identity, following the direction in #536.

**Remaining issue:** already-poisoned catalogs are not repaired. Independently reproduced true outcome → newer stale outcome → actual correct remeasurement: the repair returns `written=False`, while the stale measurement still wins `measurements_latest`. This needs a separate retraction/supersede operation or designed repair procedure that preserves ordinary replay idempotency and aligned latest views. No fingerprint or ranking changes are included, so this PR intentionally does not auto-close #536.

Old content-keyed directories can remain until a sync rewrite clears the scratch root; legacy flat scratch is ignored. Canonical bytes and stored-data formats are unchanged. Content hashing adds a sequential canonical read to complete META processing; no performance improvement is claimed.

## Validation

Linux, Python 3.12.3, locked development environment and the project's pinned FFmpeg/ffprobe pair.

```bash
uv run ruff check --fix
uv run ruff format
uv run ty check
uv run pytest tests/test_scratch_content_key.py -v --tb=short
uv run pytest tests/test_scratch_content_key.py tests/test_sync_reuse.py tests/test_run_profiles.py tests/test_storage.py tests/test_catalog_curation.py tests/test_default_checks.py tests/test_canonical_integrity.py tests/test_processing_regressions.py tests/test_video.py tests/test_camera_video.py -q -ra
uv run pytest -q -ra
```

- Lint, format and types: pass.
- Final regression tests against original behavior: **3 intended stale-pixel failures, 1 control pass**; with the fix: **4 passed**. Old behavior reports 0.0% black frames for replacement footage whose measured value is approximately 33.3%. This red/green result was independently rerun after review.
- Focused suite: **321 passed, 1 skipped**.
- Full root suite: **2,010 passed, 8 skipped**; unmodified baseline: **2,006 passed, 8 skipped**.
- Skips are opt-in network, MediaPipe, Docker and live-bucket integration tests; those integrations and other Python versions were not validated locally.

AI-assisted implementation with a separate code review, repairs, and executed verification before submission.

## Checklist

- [ ] I updated documentation for changed behavior, flags, formats, or requirements. (Internal cache layout and residual limitation are documented above and in code comments; no public flags or stored formats changed.)


kstonekuan (approved):
LGTM, merging.

Saying which half of #536 you fixed, and leaving it open, is the right call. The repair trap needs a supersede path and that is a separate decision.

Both mutations bite, including keying scratch on the source stem rather than content, which is the plausible wrong fix.

On the cost you flagged: measured on a 6.3 MB canonical, full run 6.55s to 6.56s, META 0.21s to 0.22s. The extra sha256 is not worth worrying about, and it was right to declare it rather than let me find it.

PR #541, opened 2026-09-15
fix(catalog): preserve recurring outcomes as latest

### LINKED ISSUE #540: Repeated critical-check failures leave an episode marked ok after an intervening success
</p><p>I reproduced this on <code inline="">main</code> at <code inline="">b6e2ff4</code> using a synthetic MCAP and a critical check with an explicit, unchanged version.</p><p>Running the same episode through this sequence gives:</p>
Attempt | Check result | Catalog status | Default dataset
-- | -- | -- | --
1 | Error | unverified | Excluded
2 | Success | ok | Included
3 | Same error again | ok | Included

<p>On the third attempt, <code inline="">ProcessReport</code> correctly reports the error, but <code inline="">check_runs_latest</code> still shows the successful result. <code inline="">create_dataset()</code> therefore includes an episode whose latest critical check crashed.</p><p>The reverse also happens: <strong>success → error → identical success</strong> leaves the recovered episode marked <code inline="">unverified</code> and excluded.</p><h3>Reproduction</h3><ol><li><p>Create an <code inline="">App</code> with <code inline="">default_checks=()</code>.</p></li><li><p>Generate an episode with <code inline="">SyntheticEpisodeSpec(duration_s=1.0, cameras=())</code>.</p></li><li><p>Register a critical check with <code inline="">version="1"</code> that either raises a fixed <code inline="">RuntimeError</code> or returns <code inline="">CheckResult(measurements={"seconds": 1.0})</code>, controlled by an environment variable.</p></li><li><p>Call <code inline="">app.process(source, record=True, stages="full")</code> three times: error, success, then the same error.</p></li><li><p>After each attempt, reopen the catalog connection and inspect <code inline="">episodes.status</code>, <code inline="">check_runs_latest</code>, and the default dataset selection.</p></li></ol><h3>What seems to cause it</h3><p><code inline="">Catalog.append_episode()</code> finds the first attempt’s outcome fingerprint and returns <code inline="">written=False</code>. That outcome keeps its original <code inline="">recorded_at</code>, so the latest views continue selecting the intervening success.</p><p>Consecutive identical retries should still deduplicate, but a previously seen outcome occurring after a different result needs to become current again.</p><p>The existing recovery and idempotency tests pass; neither covers this three-attempt sequence.</p></body></html>

Kaileshwar16:
Fixes a catalog state bug where an outcome that occurred earlier could be deduplicated incorrectly when it happened again later.

For example:

```text
error → success → error
```

The final error was previously treated as an old duplicate, so the catalog stayed `ok` and the episode remained in the dataset.

The same problem happened in reverse when a successful result followed an earlier failure.

This change keeps consecutive identical retries idempotent, while recording a recurring outcome as a new catalog occurrence. It also adds an optional `execution_id` for retries that may arrive after other executions have already completed.

The Parquet format and canonical episode identities remain unchanged.

## Tests

Added coverage for:

* recurring outcomes
* consecutive retries
* selective-stage runs
* explicit execution IDs
* local and bucket-backed catalogs
* concurrent appends
* crash recovery and dependent-table repair
* historical timestamp preservation

Validation:

```text
2032 passed, 9 skipped
ruff check --fix
ruff format
ty check
git diff --check
```

Closes #540.


kstonekuan (changes requested):
One gap: removing the `execution_id must be non-empty when supplied` check passes all 150 tests.

Everything else is held well. Reverting the scoping fails 21, and ignoring `execution_id` fails its own case.

I checked the property the new fingerprint could have broken, since it now depends on catalog state at append time: CATALOG.md:114 promises concurrent duplicate appends converge. Two workers appending the same outcome at once still produce one row, and so does a sequential repeat. Your design holds it.

Kaileshwar16:
Thanks for the review. I removed the `execution_id` validation as suggested and reran the tests. The 150 catalog, dataset, and bucket tests passed, along with `ruff`, `ty`, and `git diff --check`.
I have pushed the follow up commit.

kstonekuan (changes requested):
I meant add a test, not delete the guard. My wording was ambiguous.

Without it, empty and whitespace ids become distinct scopes rather than an error:

```
None  -> out
""    -> out.a8de770a92e6a396
"  "  -> out.c758dd61b918da88
```

So `execution_id=""` silently buys a unique execution identity, and `""` and `"  "` are different ones. Restore the check and give it a case.

Kaileshwar16:
@kstonekuan Sorry, I misunderstood the first comment. Restored the guard and added tests for empty and whitespace-only IDs.

kstonekuan (approved):
LGTM, merging.

Guard restored and held, including whitespace-only: dropping `.strip()` fails its case on its own.

No apology needed, my first comment was ambiguous.

Full suite on the rebased result: 2054 passed / 8 skipped.

PR #543, opened 2026-09-15
fix(episode): make _sanitize_topic injective to prevent cache collisions (#535)

### LINKED ISSUE #535: [Bug]:episode cache collision via _sanitize_topic serves one camera's pixels to another, bypassing QC
### Version or commit

Current main

### Environment

### Minimal reproduction

1. Create a synthetic canonical episode with two camera topics that sanitize to the exact same string: `/cam/front` and `/cam_front`.
2. Feed them verifiably different pixel data (e.g., all red frames for `/cam/front`, all blue frames for `/cam_front`).
3. Run the standard pipeline to trigger `Episode.video(topic)` and downstream camera QC checks (like `camera_frame_stats` or contact sheet generation).

### Expected behavior

Two distinct cache files are created in the scratch workdir. `Episode.video(topic)` and `Episode.frames(topic)` return the correct pixels (red vs. blue) for their respective topics. QC measurements are recorded against the correct footage.

### Actual behavior

Both topics map to the exact same cache file (`cam_front.mp4`). The second camera silently receives the first camera's pixels.

Executed proof:
video() path for /cam/front   -> cam_front.mp4
video() path for /cam/front   -> cam_front.mp4     SAME cached file: True
extracted frame 0 bytes identical across the two cameras: True
cache dir contents: ['cam_front.mp4', 'frames_cam_front_f1.000000_s0.000000_eend']

The pipeline exits with a clean success code, but downstream QC checks (`camera_frame_stats`, `camera_stability`) and contact sheets are generated using the wrong pixels for the second camera.

### Additional context

Root cause: `_sanitize_topic` (episode.py:74-75) collapses characters outside `[A-Za-z0-9_.-]` to a single `_`. Because the MP4 cache path (`workdir / f"{_sanitize_topic(topic)}.mp4"`) relies on `output.exists()`, the second camera just reads the first camera's remuxed file without checking if the bytes actually belong to its topic.

Why it matters: A dual-camera robot whose recorders use `/` and `_` naming variants (or stray double slashes) will pass QC gates that inspected the wrong camera's footage. The evidence chain lies with a clean exit code.

Fix direction: Make the cache key collision-resistant the way `App._source_artifact_directory_name` already does it: append `sha256(topic)[:12]` to the sanitized topic name. Since `frames()` and contact-sheet paths derive from the same sanitization logic, a single change covers the whole measurement toolchain.

Definition of done:
- Two topics differing only by slash-vs-underscore land in distinct cache files; `video()`/`frames()` return each camera's own pixels (red-vs-blue synthetic asserts this).
- An episode with N distinct cameras yields N distinct contact-sheet artifact paths, each byte-distinct.
- Mutation proof: revert the key to bare `_sanitize_topic(topic)` and the two-camera test goes red.
- Existing episode/checks/media suites stay green.

Sagar-024:
Fixes #535

### Summary
As directed by @kstonekuan in the issue thread, this PR fixes the cache collision by making `_sanitize_topic` injective. The fix is applied **once** in `src/hflow/episode.py`, so all six call sites (mp4 cache, both frames caches, contact sheets, both artifact key builders) inherit the fix automatically.

The sanitized topic now takes the shape `{readable_stem}-{sha256(topic)[:12]}`, matching the pattern already established by `App._source_artifact_directory_name`.

### Changes
- **`src/hflow/episode.py`**: Updated `_sanitize_topic` to append a 12-char SHA256 digest of the raw topic string.
- **Version Bumps**: Bumped `media/contact_sheet` (1 -> 2) and `camera_video` (1 -> 2) step versions, as mandated by the repo's own inline comments for when artifact names/paths change.
- **Tests**:
  - Updated `test_run_profiles` assertions that hardcoded the old bare name.
  - Added `tests/test_topic_cache_collision.py` with three new proofs:
    1. **Name Injectivity**: `/cam/front` and `/cam_front` yield distinct cache keys.
    2. **Red-vs-Blue Attribution**: End-to-end test where each camera's cached MP4 must byte-match its own single-camera reference (proves attribution, not just path difference).
    3. **N-Camera Distinctness**: 3-camera episode yields 3 byte-distinct contact sheets.

### Mutation Proof
Reverting `_sanitize_topic` to the bare name causes **4 tests to fail** (the 3 new collision tests + the updated run-profiles test). Restoring the digest passes all 25 affected tests.

### Gates
- Full suite: 2028 passed, 8 skipped.
- `ruff check`, `ruff format`, `ty` (repo-wide): Clean.


kstonekuan (approved):
LGTM, merging.

Both mutations bite, including hashing the sanitized name rather than the raw topic, which is the plausible wrong fix: both colliding topics sanitize to the same string, so that digest collides too.

`test_colliding_caches_serve_each_camera_its_own_pixels` is the one that proves attribution rather than path difference. Byte-matching each camera against its own single-camera reference is what distinguishes "different filenames" from "right pixels".

Bumping `media/contact_sheet` and `camera_video` was right: the artifact paths change, so old artifacts must not be read as current.

PR #547, opened 2026-09-15
fix(catalog): revive a stale check outcome when it returns after a di…

### LINKED ISSUE #540: Repeated critical-check failures leave an episode marked ok after an intervening success

Morrowga:
Fixes #540

## Problem

`append_episode()` deduplicates by content: it computes a `run_fingerprint`
from the check outcomes and checks whether a file with that exact
fingerprint already exists on disk. If so, it assumes this is a pure replay
and returns early without writing anything new.

That check answers "has this exact content ever existed?" — not "is this
different from what happened most recently?" So when a check's result
alternates (error → success → the same error again), the third attempt's
fingerprint matches the first attempt's, finds that old file already on
disk, and is silently dropped. No new row is written, so `check_runs_latest`
and `episodes.status` keep showing the second attempt's stale success —
even though the episode's true latest state is a failure.

## Root cause

```python
if self.location.exists(f"episodes/{file_stem}.parquet"):
    self._reconcile_replayed_append(...)
    return AppendResult(..., written=False)
```

This only checks "does this fingerprint exist anywhere in history," with no
comparison against the *current latest* result for the episode.

## Fix

- Added `_latest_run_fingerprint()`, which finds the fingerprint currently
  ranked latest for an episode (by `recorded_at`).
- Before returning early on a fingerprint match, `append_episode()` now
  compares it against the current latest. If they differ — meaning a
  different outcome happened more recently — the old, content-identical
  file is treated as new: `_revive_stale_outcome()` rebuilds and
  **re-publishes** it (an unconditional atomic replace, not create-if-absent)
  with a fresh `recorded_at`, so latest-ranking views pick it back up as
  current.
- If the fingerprint already *is* the current latest (a true consecutive
  duplicate), behavior is unchanged — falls through to the existing
  `_reconcile_replayed_append` path.

## Testing

Reproduced the exact issue via a synthetic episode and a flaky critical
check (mirrors the issue's own repro steps), then extended to a 6-attempt
sequence covering both directions and the consecutive-duplicate case:

| # | Result | Expected | Observed |
|---|---|---|---|
| 1 | error | new | new ✅ |
| 2 | success | new | new ✅ |
| 3 | error (A after B) | revived | revived — fresh `recorded_at`, `episodes.status` → `unverified` ✅ |
| 4 | error (consecutive dup of #3) | unchanged | unchanged — same `recorded_at` as #3, no revive triggered ✅ |
| 5 | success (B after A) | revived | revived — fresh `recorded_at`, `episodes.status` → `ok` ✅ |
| 6 | success (consecutive dup of #5) | unchanged | unchanged — same `recorded_at` as #5 ✅ |

Full test suite: `36 failed, 1472 passed, 27 skipped` — identical before and
after this change (confirmed via `git stash`), so the pre-existing failures
are unrelated (missing optional dependencies: `pyarrow`, `yaml`, `cv2`,
`hflow_server`), and this change introduces no regressions.

## Note
`_revive_stale_outcome()` uses `uri` directly (not `canonical_path.resolve()`
as fallback) — safe today since the only real caller always passes a
resolved `uri`, but worth flagging in case that assumption changes later.


kstonekuan:
Resolved in #541, merged an hour ago. Kaileshwar16 opened it against #540 at 03:26 and it landed at 09:14; yours opened at 08:59.

Your diagnosis is the same one: the fingerprint answers "has this content ever existed" rather than "is this different from the most recent". #541 scopes the outcome to its predecessor.

Nothing was wrong with this. Check an issue for a linked PR before starting; #540 had one from 03:26.

PR #549, opened 2026-09-15
fix(checks): fail closed on unmeasurable joint-motion steps (#546)

### LINKED ISSUE #546: [Bug]: _joint_motion_profile treats NaN velocity steps as compliant, causing duplicate-stamped streams to pass velocity gates and dead channels to report 0% idle
### Version or commit

Current main

### Environment

### Minimal reproduction

1. Create a stream of 100 joint position hops of 5 rad, where every hop occurs on a duplicate timestamp (duration <= 0).
2. Run `_joint_motion_profile` to compute `violation_pct`, `max_abs_velocity`, and `idle_fraction`.
3. Observe that despite 100 unmeasurable steps, the stream reports `violation_pct=0.0` and passes the gate.

### Expected behavior

Unmeasurable steps (duplicate timestamps or NaN positions) must be excluded from compliance percentages or cause the check to abstain/refuse. They must not be silently counted as "clean" measurements. A stream of identical motion with advancing timestamps should not yield a 100% difference in gate verdict compared to duplicate timestamps.

### Actual behavior

The code converts unmeasurable steps to NaN velocity (`safe_deltas_s = np.where(deltas_s > 0, deltas_s, np.nan)` at checks.py:103). However, both consumers compare this NaN:
- `speed > velocity_limit` is False (treated as non-violating).
- `speed < velocity_epsilon` is False (treated as not idle).

Executed proof:
[A: 100 hops of 5 rad, each on a duplicate stamp]
  max_abs_velocity=0.0  violation_count=0  violation_pct=0.0  nonpositive_dt_count=100
  gate(violation_pct<=5) -> GateDecided(verdict=True)

[B: IDENTICAL motion, advancing stamps]
  max_abs_velocity=5000.0  violation_count=99  violation_pct=100.0
  gate(violation_pct<=5) -> GateDecided(verdict=False)

[C: every sample NaN, advancing clock]
  idle_fraction=0.0  -> stores silently ("zero idle" for a dead channel)

### Additional context

Root cause: `safe_deltas_s` marks steps unmeasurable with NaN, but no consumer treats NaN as "no data." In Python/NumPy, `>` and `<` against NaN both yield False, which the comparisons interpret as "clean" and "moving." This dilutes `np.mean(violation_mask)` and zeroes `np.nanmax` over surviving steps.

Why it matters: Curation gates that quarantine velocity spikes wave through batch-stamped stereo/multi-sensor streams. `idle_fraction < X` cuts reclassify teleoperating-but-jumpy demos as idle junk. The training corpus gets a two-way corruption: bad episodes kept, and good-looking numbers computed from steps that were never measured. The codebase already solves this correctly in `_trajectory_profile` (checks.py:1576-1584) and `estimate_fps_from_log_times` (video.py:86-90), but this profile missed the pattern.

Fix direction: Carry a measurable mask through `_JointMotionProfile` and compute `violation_pct`/`max_abs_velocity` over it only. Exclude unmeasurable steps from `idle_fraction`'s denominator. When the measurable share is below a floor (or the stream is all-NaN), emit `*_sample_count`/`nonpositive_dt_count` and no percentages, so an unmeasured stream cannot produce a pass-baiting number. Refuse NaN `max_abs_velocity` inside the check, not at the catalog.

Definition of done:
- Regression: streams A and B (identical positions, duplicate vs advancing stamps) must not disagree by 100.0 vs 0.0 `violation_pct`.
- Mutation proof 1: delete the measurable-mask from `violation_pct`'s denominator -> A test flips back to 0.0 -> suite fails.
- Mutation proof 2: feed an all-NaN channel -> `idle_fraction` must abstain instead of storing 0.0.
- Pin that `nonpositive_dt_count > 0` is visible in every percentage this profile feeds.

Sagar-024:
Fixes #546

### Summary
As detailed in the issue, `_joint_motion_profile` was treating NaN velocity steps (from duplicate timestamps or NaN positions) as compliant because `NaN > limit` and `NaN < epsilon` both evaluate to `False`. This PR carries an explicit `measurable` mask through the profile and computes all compliance metrics over measurable data only, ensuring unmeasurable steps are excluded from percentages and dead/NaN streams abstain rather than emitting pass-baiting numbers.

### Changes
- **`src/hflow/checks.py`**: 
  - Added a `measurable` mask to `_JointMotionProfile` (strictly positive dt AND finite position jumps).
  - `joint_discontinuity`: `violation_pct` divides by measurable count only; `violation_count` only counts measurable steps; `max_abs_velocity` is a finite `np.max` (NaN refused inside the check).
  - `idle_fraction`: Unmeasurable steps are excluded from both numerator and denominator. Removed the dead `if total_span_s else 0.0` fail-open.
  - Zero-measurable streams now refuse: emit `*_sample_count`/`*_measurable_step_count`/`nonpositive_dt_count` and withhold percentage keys.
- **Tests**: Added `tests/test_checks_nan_failopen.py` with 5 core proofs:
  1. **Stream A vs Stream B**: Identical position hops. Stream A (duplicate stamps) abstains; Stream B (advancing stamps) reports `violation_pct=100.0`.
  2. **All-NaN Channel**: `idle_fraction` abstains instead of storing `0.0`.
  3. **Denominator Pin**: 1/3 measurable steps, not 1/5, proving unmeasurable steps are excluded.
  4. **Count Visibility**: `nonpositive_dt_count` is visible wherever a percentage is emitted.
  5. **Finite-Value Sweep**: All emitted measurements are finite.

### Mutation Proof
- **Denominator mask removed**: Dilution test fails (`1 failed, 69 passed`).
- **Idle abstain reverted**: Dead channel stores `0.0` and fails (`3 failed, 1 passed`).
- **Restored**: All tests pass.

### Gates
- Full suite: 2032 passed, 8 skipped, zero failures.
- `ruff check`, `ruff format`, `ty` (repo-wide): Clean.


kstonekuan (changes requested):
The fix is right and each half of `measurable` is held: dropping `deltas_s > 0` or `finite_jump` fails different cases.

One thing to fix. `np.nanmax` at `checks.py:124` warns on an all-NaN row, so the suite gains four `RuntimeWarning: All-NaN slice encountered` that main does not have, and a user processing a duplicate-stamped stream gets numpy noise in exactly the case this PR exists to handle.

You already compute the mask that says which rows those are.

Sagar-024:
Good catch. Replaced `np.nanmax` with a pre-allocated `np.full` array and a masked `np.max` assignment (`per_step_max_speed[measurable] = np.max(...)`) so unmeasurable rows never reach the reduction.

Verified the suite with `-W error::RuntimeWarning` - zero `All-NaN slice encountered` warnings now, matching the baseline on `main`. (Note: a pre-existing `ResourceWarning` from ffprobe teardown remains on both trees, but all RuntimeWarnings are resolved).

Pushed 9c38a1d.

kstonekuan (approved):
The masked assignment is the right shape, and the comment explaining why it replaced `np.nanmax` will save the next person rediscovering it.

Verified against main merged in: `-W error::RuntimeWarning` runs clean at 2166 passed, so the four warnings are gone rather than filtered. All six guards hold under mutation. Dropping `finite_jump` reddens 2, dropping `deltas_s > 0` reddens 3, widening either percentage denominator back to all steps reddens 1 each, and removing either nothing-measurable early return reddens 2 and 1.

Taking the advanced ones is exactly where the leverage is. Thanks.

PR #551, opened 2026-09-15
fix(server): advertise bucket curation storage support

### LINKED ISSUE #550: [Bug]: /api/v1/config advertises capabilities.curation=false for buckets though pin and the sidecar already work there
### Version or commit

d0d57fc

### Environment

Linux x86_64, Python 3.13.3 (uv), no Docker

### Minimal reproduction

`WorkspaceCapabilities.curation` is documented as false for bucket-backed workspaces, with a 501 story:

> saved queries, the pinned-manifest registry, and the manifest files need a LOCAL data root, so a bucket-backed workspace answers 501 for every one of them and the frontend should not offer them.

(`packages/hflow-server/src/hflow_server/_contract.py`, `WorkspaceCapabilities.curation`)

`/api/v1/config` implements that by tying the flag to local paths only:

```python
workspace_is_local = local_data_root_or_none(settings.data_root) is not None
...
curation=workspace_is_local,
```

(`packages/hflow-server/src/hflow_server/server.py`)

So any `gs://` / `s3://` data root gets `capabilities.curation: false`.

But the curation write path already supports buckets:

- `POST /curation/pin` comments that pinning "works on a bucket-backed workspace" and calls `write_dataset_manifest(Workspace.parse(...))` with no `local_data_root_or_none` / 501 gate (`_curation.py`).
- `store_sidecar_state` writes through `parse_storage_root(...).write_bytes` and documents the same (`_sidecar.py`).
- Neither `_curation.py` nor `_sidecar.py` contains `501` or `local_data_root`.

Media is the surface that correctly stays local-only (`_media.py` 501). Curation was grouped with it in the capability flag and never updated when pin/sidecar grew StorageRoot support.

Controlled on `d0d57fc`:

1. `local_data_root_or_none("gs://bucket/ws") is None` → config would set `curation=False`.
2. On a real `BucketStorageRoot` (`file://` obstore test backend): `write_dataset_manifest(...)` publishes `manifests/...parquet` (row_count 1); `root.write_bytes("curation/state.json", ...)` succeeds. Same primitives pin uses.
3. `rg '501|local_data_root' packages/hflow-server/src/hflow_server/_curation.py _sidecar.py` → no matches.

### Expected behavior

`capabilities.curation` is true when the curation studio's durable writes can land — including bucket-backed workspaces, matching pin/sidecar/manifest download behavior.

`capabilities.media` stays false / 501 for buckets until media serving exists.

The Field description must not claim curation answers 501 if the routes do not.

### Actual behavior

Bucket workspaces get `capabilities.curation: false`, so a frontend that trusts `/api/v1/config` hides pin/save even though those routes already work on `StorageRoot`. The contract text still promises 501s that curation never returns.

### Additional context

- `/api/v1/config` reports `capabilities.curation: true` for a bucket-backed data root when the catalog (or workspace) is otherwise usable the same way pin already requires — or, if curation is intentionally still local-only, pin/sidecar must refuse 501 consistently and their "works on bucket" comments/docs must be corrected. Preferred: advertise truth (curation true, media false).
- Contract Field description for `curation` matches the chosen behavior (no false 501 claim).
- `_settings.local_data_root_or_none` docstring stops grouping sidecar with media as a shared 501 precondition if sidecar is bucket-capable.
- Regression test: bucket (or `BucketStorageRoot`) workspace config asserts `curation is True` and `media is False`.
- Non-goal: implementing bucket media byte serving.

Amidwestnoob:
## Summary

Closes #550.

- Advertise `capabilities.curation=true` for bucket-backed workspaces, matching the durable writes already implemented through `StorageRoot`.
- Keep `capabilities.media` local-only. Read-only policy and catalog readiness remain separate from storage support.
- Correct the capability description and related comments, including the existing limitation on HTTP manifest downloads.
- Add real bucket-backed API and persistence coverage without cloud credentials.

## Why

The config endpoint reused the media local-path predicate for curation, so a frontend trusting it could hide working saved-query and pin operations. This changes the advertised capability, not the write implementations or permission checks.

The tests exercise actual handlers against `BucketStorageRoot(file://...)`, redirecting only a sentinel bucket URL. They verify saved-query CRUD, preview/report, pinning, and durable manifest/sidecar contents read through a fresh mirror. They also cover missing, empty and incompatible catalogs, unchanged local behavior, and read-only refusals.

**Existing limitation, not fixed here:** bucket HTTP manifest downloads still return 501 because the fetched mirror file passes through the local-root served-file check. Bucket media also remains 501. The `curation` field describes durable write support, not byte delivery; the misleading download comment is corrected rather than adding bucket serving to this PR.

## Validation

Linux x86_64, Python 3.14.6, locked environment with the bucket extra installed; FFmpeg/ffprobe configured for default tests. Commands run from the source checkout:

```bash
uv sync --locked --extra bucket
.venv/bin/ruff check
.venv/bin/ruff format --check
.venv/bin/ty check
.venv/bin/python -m pytest packages/hflow-server/tests/test_server_config.py::test_config_enables_curation_not_media_for_bucket_backed_workspace packages/hflow-server/tests/test_server_bucket_curation.py -q --tb=short
.venv/bin/python -m pytest packages/hflow-server/tests -q -ra --tb=short
.venv/bin/python -m pytest -q -ra --tb=short
```

- Ruff, formatting and types: passed across the root.
- The identical regression bundle against original production code: **6 expected capability assertion failures, 3 passes**. Final code: **9 passed, no skips**. Independently rerun after review with the same outcome.
- Focused config/curation/sidecar/media/query/settings tests: **100 passed**.
- Entire server package: **252 passed**.
- Full default root suite: **2,063 passed, 8 skipped**.

The eight skips are opt-in FFmpeg network, MediaPipe, Docker and live-cloud integration tests. Local bucket regressions did not skip. Other Python versions and specialized example environments were not tested locally. No live cloud, paid resources or private recordings were used.

AI-assisted implementation followed by separate code review, repairs and executed verification.

## Checklist

- [x] Added outcome-focused regression tests.
- [x] Updated the capability schema description and affected documentation/comments.
- [x] Ran Ruff, formatting checks and ty successfully.
- [x] Ran relevant tests and the full default root suite.
- [x] No recordings, generated media, credentials, private URLs or runtime artifacts in the patch.
- [x] Preserved stored-data formats and compatibility.


kstonekuan (approved):
LGTM, merging.

`test_bucket_curation_persists_queries_and_pins_but_cannot_serve_bytes` is the one that makes this safe to widen: it pins that the two flags moved apart rather than both becoming true, and flipping `media` to match fails six cases.

Narrowing `local_data_root_or_none`'s docstring to "reachable as local paths" for media and HTTP download, rather than leaving it as the shared precondition, keeps the next reader from re-deriving the old grouping.

Stating in the field description that this is storage support and not a writability probe is the right call, since `read_only` was already orthogonal and `curation` never reflected it.

PR #555, opened 2026-09-16
fix(import): guard the meta/episodes tree loop against a missing 'path'

### LINKED ISSUE #544: [Good First Issue]: LeRobot importer crashes with a bare KeyError when a meta/episodes tree entry has no "path"
**Version or commit:**

Current main

**Environment:**

**Minimal reproduction:**

1. Monkeypatch `prep._hf_tree` to return a list containing an entry without a `"path"` key: `[{"type": "file", "size": 4096}]`.
2. Call `import_lerobot_dataset` with any valid repo_id.
3. Observe the crash at `src/hflow/importers/lerobot.py:760`.

**Expected behavior:**

The importer should raise a clean, descriptive `ValueError` naming the `repo_id` and the missing field, matching the codebase's established boundary-refusal pattern. It should not crash with a raw `KeyError`.

**Actual behavior:**

The importer crashes with a raw stack trace:

```
File "src/hflow/importers/lerobot.py", line 760, in _ensure_source_archive
    if entry.get("type") == "file" and entry["path"].endswith(".parquet"):
KeyError: 'path'
```

**Additional context:**

Root cause: `_hf_tree` (line 513) deliberately tolerates entries without a string `"path"` using `.get("path")`, but the episode-parquet loop at line 760 assumes the key exists and uses raw subscript access (`entry["path"]`).

Why it matters: Every other consumer of the same tree response guards this gap. `_fetch_info_json` uses `.get("path")` (line 553), and `_episode_metadata_cache_path` (lines 594-617) raises a clean `ValueError` for unsafe paths. This is the single unhardened site in the file.

Fix direction: At the top of the loop, replace `entry["path"]` with a guarded read. Validate `tree_path = entry.get("path")` is a non-empty string, else raise a `ValueError` matching the module's phrasing (e.g., `"Hugging Face tree response for {repo_id} lists an entry with no usable 'path'"`). No new exception types needed.

Definition of done:

- Add a test in `tests/test_lerobot_metadata_refusals.py` following the existing pattern (stub `_hf_repo_info` + `_fetch_info_json`, monkeypatch `prep._hf_tree`).
- Assert `pytest.raises(ValueError, match=...)` on the path-less entry.
- Ensure `_assert_no_dataset_output(output_dir)` confirms nothing was published.

Sravan1011:
Closes #544.

## Summary

`_hf_tree` deliberately tolerates a tree entry without a string `"path"` (it only dedupes on one when present, per `lerobot.py:513`). Every other consumer of that response already guards the gap: `_fetch_info_json` reads it with `.get("path")`, and `_episode_metadata_cache_path` raises a clean `ValueError` for an unsafe path. `_ensure_source_archive`'s `meta/episodes` file loop was the one unhardened site — it subscripted `entry["path"]` directly and crashed with a raw `KeyError` on an entry that omits it.

## What changed

`src/hflow/importers/lerobot.py`, inside `_ensure_source_archive`'s loop over `meta/episodes` entries:
- Non-file entries are still skipped first, before any `"path"` read is attempted (unchanged behavior).
- For a file entry, `tree_entry_path = entry.get("path")` is read once and validated to be a non-empty string. A missing or non-string path raises `ValueError(f"Hugging Face tree response for {repo_id} lists an entry with no usable 'path'")`, matching the phrasing `_episode_metadata_cache_path` already uses for its own refusal in this file.
- No new exception type, per the issue.

## Tests

Added `test_import_refuses_an_episode_parquet_entry_with_no_path` to `tests/test_lerobot_metadata_refusals.py`, following the file's existing pattern:
- Stubs `_hf_repo_info` and `_fetch_info_json` so the failure is isolated to the `meta/episodes` listing.
- Monkeypatches `prep._hf_tree` to return `[{"type": "file", "size": 4096}]` — the minimal repro from the issue.
- Asserts `pytest.raises(ValueError, match=...)` on the exact message.
- Calls `_assert_no_dataset_output(output_dir)` to confirm nothing was published.

## Validation

```
uv sync --locked --all-extras
uv run pytest -q                  # 2163 passed, 8 skipped, 4 failed
```

The 4 failures (`test_catalog_ui.py` x3, `test_ffmpeg.py::test_display_rotation_preserves_coded_frame_geometry_and_pixels`) are pre-existing environment noise in this sandbox — the pinned ffmpeg build isn't fetchable here so the suite falls back to the system `ffmpeg`, which handles `-display_rotation` differently, and the `catalog_ui` failures are order-dependent flakes that pass individually. Reproduced identically on a clean checkout of `main` with none of this PR's changes applied. `tests/test_lerobot_metadata_refusals.py` and `tests/test_lerobot_converter.py` (100 tests) pass cleanly on their own.


Sravan1011:
@Sagar-024 @kstonekuan go ahead with the review

kstonekuan (changes requested):
The `_hf_tree` reading is right: `lerobot.py:513` appends an entry whether or not `path` is a string, so the gap is real and this is the site that had it.

The missing-`path` half holds. Dropping the guard reddens your new test.

The `isinstance` half does not. Replace it with `if not tree_entry_path:` and the whole file stays green, but a `{"type": "file", "path": 123}` entry then reaches `.endswith` and raises the bare `AttributeError` this PR exists to replace. One more case on the same test covers it.

Sravan1011:
Confirmed both halves. I reproduced it locally before fixing the test: swapping the guard for a bare `if not tree_entry_path:` kept the rest of the suite green and made `{"type": "file", "path": 123}` fall through to `tree_entry_path.endswith(".parquet")`, raising `AttributeError: 'int' object has no attribute 'endswith'` — exactly the bare crash this PR exists to replace.

Parametrized `test_import_refuses_an_episode_parquet_entry_with_no_path` (now `test_import_refuses_an_episode_parquet_entry_with_no_usable_path`) over both `path-missing` and `path-not-a-string` cases, same message and behavior expected for both. No production code change was needed — `isinstance(tree_entry_path, str)` was already there — this closes the coverage gap that let it look redundant.

Sravan1011:
@kstonekuan now you can go ahead with the merge

kstonekuan (approved):
Both halves hold now: dropping the guard reddens 2, dropping the `isinstance` check reddens 1. 2173 passed.

PR #560, opened 2026-09-17
fix(ffmpeg): stage contact sheet via per-call tempfile and atomic ren…

### LINKED ISSUE #556: [Bug]: contact_sheet writes JPEG directly to final path without atomic rename, allowing truncated crash artifacts to be cataloged as trusted evidence
### Version or commit

Current main

### Environment

### Minimal reproduction

1. Call `contact_sheet()` with a valid frame list, targeting `run_dir/media/<topic>.jpg`.
2. Inject a `SIGKILL` (or `timeout 0.05`) to the ffmpeg subprocess mid-write.
3. Observe that the final `.jpg` file exists on disk with a truncated size, and the next `publish` step catalogs it as evidence.

### Expected behavior

The contact sheet must be written to a temporary file (e.g., `output.tmp`) in the same directory, and atomically renamed to the final path only if ffmpeg exits with code 0 and the file size is valid. A crash mid-write must never leave a partial file at the final path.

### Actual behavior

`contact_sheet()` at `ffmpeg/_contact_sheet.py:169-192` passes the final output path directly to ffmpeg (`-y str(output)`). Because ffmpeg uses `open(O_TRUNC)` on the final path, a crash or SIGKILL mid-write leaves a truncated, header-valid JPEG at the destination. 

Since `app.py:533` uses a no-copy publish for local roots, and `catalog.py:append_episode` records the artifact URI based on file existence, the truncated bytes are silently cataloged as trusted evidence.

Sibling contrast: Every other artifact producer uses atomic staging:
- `video.py:404-443` (MP4 cache)
- `app.py:334-359` (sync markers)
- `storage.py:206-221` (workspace manifest)
- `_frame_statistics.py:533-549` (instrument cache)
- `_writer.py:185-200` (canonical MCAP)

### Additional context

Root cause: `contact_sheet` delegates atomicity to ffmpeg's `-y` truncation flag instead of staging through a sibling temp file as every other hflow writer does.

Why it matters: Contact sheets are cataloged artifacts (`artifact/media/contact_sheet`) and are included in the evidence chain (`measurements_latest` view). A truncated JPEG silently passes existence checks, is exported in snapshots with `media_mode=COPY`, and destroys history. A crash mid-write becomes a trusted corrupted file.

Fix direction: Mirror `video.py:404-443`: build `tmp = output.with_name(output.name + ".tmp")` in the same directory, run ffmpeg to `str(tmp)`, and on `returncode==0` and `tmp.stat().st_size>0`, do `tmp.replace(output)`. On any exception, unlink the temp file.

Definition of done:
- `contact_sheet` writes to `output.tmp` then replaces; output never appears partial after a SIGKILL during ffmpeg.
- Concurrent `contact_sheet` calls to the same path both produce valid JPEGs; last replace wins with a complete file.
- Mutation proof: revert to `str(output)` → test fails because `output.exists()` with truncated bytes after injected kill; restored fix passes.
- Existing ffmpeg/tests + app media stage tests green.

shobhitagnihotri69:
### Summary
Fixes #556. Previously, `contact_sheet()` passed the target output path directly to ffmpeg (`-y str(output)`). An interruption, crash, or SIGKILL mid-write would leave a truncated, partial JPEG that satisfied bare file-existence checks and was cataloged into `artifact/media/contact_sheet` as valid evidence.

### Changes
- Mirrored the atomic staging pattern in `src/hflow/video.py:625-668`:
  - Output is staged to `tempfile.NamedTemporaryFile(prefix=f".{output.name}.", suffix=".tmp", dir=output.parent, delete=False)` to prevent concurrent calls from colliding.
  - Added `-f image2` muxer to bypass extension sniffing failures caused by the `.tmp` suffix.
  - Added `try ... except BaseException` to guarantee unlinking of the temporary file upon crash or error.
  - Validates `returncode == 0` and non-zero byte size before atomic `temporary_output.replace(output)`.
- Added 3 unit tests in `tests/test_ffmpeg.py`:
  - `test_contact_sheet_failure_cleans_up_temp_and_leaves_no_output`
  - `test_contact_sheet_zero_byte_output_cleans_up_and_raises`
  - `test_contact_sheet_atomic_rename_success`

### Verification
- All 29 contact sheet tests pass with system ffmpeg (`78 passed, 8 skipped` in `test_ffmpeg.py`).
- Formatted and linted with `ruff`.


kstonekuan:
#556 is fixed by #561, so there is nothing left for this one to do. Your slot is free again if you want to pick something else up.

PR #561, opened 2026-09-17
fix(ffmpeg): stage contact sheets through per-call temp + atomic replace (#556)

### LINKED ISSUE #556: [Bug]: contact_sheet writes JPEG directly to final path without atomic rename, allowing truncated crash artifacts to be cataloged as trusted evidence

Sagar-024:
Fixes #556

### Summary
`contact_sheet()` passed the final output path straight to ffmpeg, so a SIGKILL/OOM mid-write left a truncated JPEG at the path the publish step catalogs on bare existence. This stages the render through a per-call sibling temp file and atomically replaces, mirroring `video.py:625-668` with both corrections from review.

### Changes (`src/hflow/ffmpeg/_contact_sheet.py`)
- Per-call temp path: `tempfile.NamedTemporaryFile(prefix=f".{output.name}.", suffix=".tmp", dir=output.parent, delete=False)` — a fixed `<output>.tmp` lets concurrent sheets truncate/unlink each other's temp (video.py:627).
- Explicit muxer: `-f image2` pinned, because the `.tmp` suffix defeats extension sniffing (video.py:651 does `-f mp4`). Empirically verified on ffmpeg 8.0.1: no `-f` fails outright ("Unable to choose an output format"); `-f image2` and `-f mjpeg` both produce JPEGs byte-identical to the `.jpg`-extension reference; ffmpeg auto-selects image2 for `.jpg`, so image2 is the faithful explicit pin.
- Success gate: returncode == 0 AND temp is_file AND size > 0 before `replace()`; every failure path unlinks the temp.
- Deliberate deviation from video.py:668: `replace()` sits inside the try so a failed rename still cleans the temp; `except BaseException` covers KeyboardInterrupt/SystemExit.
- Known residual (shared with the sibling pattern): a whole-process SIGKILL orphans a dot-prefixed `.*.tmp` sibling; nothing catalogs it.

### Tests (`tests/test_contact_sheet_atomic.py`)
1. SIGKILL: real ffmpeg SIGSTOPped the instant its temp holds bytes (partial-on-disk asserted non-vacuous), then SIGKILLed → final path absent; a pre-seeded valid sheet survives byte-identical; no orphans.
2. Concurrency: barrier-synced pair × 5 rounds to the same path → every round SOI+EOI+ffprobe-valid, no orphans.
3. Muxer pin: `["-f", "image2"]` consecutive in argv; argv[-1] is a sibling `.tmp`.

### Mutation proofs
- M1 revert to `str(output)` direct → SIGKILL test red (truncated bytes at final path).
- M2 revert to fixed `<output>.tmp` → concurrency test red (shared temp removed under peer).
- M3 drop `-f image2` → muxer pin test red.
- Restored → 3 passed after each mutation.

### Gates
- Full suite: 2174 passed, 8 skipped (clean main: 2171 → delta is exactly the 3 new tests).
- ruff check / ruff format --check / ty: clean scoped and repo-wide.
- Note: `test_verification.py::test_verify_lerobot_import_resolves_bucket_deliveries_under_the_verified_root` flaked once in an early run; reproduced once on clean main in isolation and passed 10/10 isolated on both branches — pre-existing intermittent, unrelated to this change.


kstonekuan (approved):
Exactly the `video.py` shape, and the tests are the best part. Guarding the SIGKILL case against its own vacuity, and asserting a pre-existing sheet survives byte-identically, are both things that would have been easy to leave out and would have hidden a regression later.

All three guards hold independently under mutation: writing straight to the final path reddens 8, dropping `-f image2` reddens 8, and swapping the per-call temp for a fixed `<output>.tmp` reddens the concurrency test. That last one is the mistake the issue originally proposed, so your own test now refuses it.

2179 passed on the merge with main.

PR #562, opened 2026-09-17
fix(catalog): reject case-insensitive measurement key collisions

### LINKED ISSUE #558: Measurement keys differing only in case silently return the wrong values in curation queries
HFlow accepts `/Camera/score` and `/camera/score` as distinct measurement keys, but querying either through the `episodes` view returns the same value.

This can silently select the wrong episodes during curation. I reproduced it on `main` at `9c9fd21f375299ff4ff063e25bd8eb6897eda0f5`, using Python 3.12 and the locked dependencies.

### Reproduction

Run with `uv run python` from the repository:

```python
from pathlib import Path
from tempfile import TemporaryDirectory

import hflow

with TemporaryDirectory() as directory:
    root = Path(directory)
    source = synthesize_episode(
        root / "source.mcap",
        SyntheticEpisodeSpec(duration_s=0.2, cameras=()),
    )
    app = hflow.App("case-collision", data_root=root / "data")

    @app.check(version="1")
    def quality(ep: hflow.Episode) -> hflow.CheckResult:
        return hflow.CheckResult(
            measurements={"/Camera/score": 0.1, "/camera/score": 0.9}
        )

    app.process(source)

    connection = hflow.open_catalog_connection(root / "data" / "catalog")
    try:
        print(connection.execute(
            'SELECT "/Camera/score", "/camera/score" FROM episodes'
        ).fetchall())

        print(connection.execute(
            'SELECT count(*) FROM episodes WHERE "/camera/score" < 0.5'
        ).fetchone())
    finally:
        connection.close()
```

Actual output:

```text
[(0.1, 0.1)]
(1,)
```

The lowercase metric is **0.9**, so the threshold query should select **zero episodes**. An explicit refusal of the ambiguous keys would also prevent the incorrect result.

### Cause and impact

The pipeline’s measurement-key collision guard compares exact strings. DuckDB treats SQL identifiers as case-insensitive, including quoted identifiers.

When HFlow pivots the measurements into the wide view, DuckDB silently renames the second column to `/camera/score_1`. Both original quoted names then resolve to the first column.

The long measurements table retains the correct values, but the wide view exposes misleading names. I also reproduced the incorrect selection through `hflow.curate()` and the same column ambiguity in exported `samples.parquet`.

### Related previous fix

[[PR #133 — fix(catalog): refuse measurement keys that shadow episodes columns](https://github.com/Hebbian-Robotics/hflow/pull/133)](https://github.com/Hebbian-Robotics/hflow/pull/133), which closed [[#130](https://github.com/Hebbian-Robotics/hflow/issues/130)](https://github.com/Hebbian-Robotics/hflow/issues/130), addressed the same silent-renaming behavior when a measurement key collides with a reserved episode column such as `task`.

That guard already compares names case-insensitively, but it only checks measurement keys against reserved columns. It does not cover collisions **between two measurement keys**.

### Expected behavior

Detect ambiguous measurement names before exposing a wide view or publishing a curation result or snapshot, with an actionable error naming both keys.

The check should cover the complete set of keys being pivoted: conflicting names can also arrive through separate appends. Original names and values should remain intact rather than being silently lowercased, merged, or renamed.

Kaileshwar16:
Measurement keys such as `/Camera/score` and `/camera/score` were stored separately but became ambiguous when pivoted into DuckDB columns. Querying either name returned the first value, allowing curation queries to select the wrong episodes.

This PR rejects those collisions using DuckDB’s ASCII case-insensitive identifier rules.

## Changes

* Validate incoming measurement keys and complete key sets before curation and snapshot pivots, covering collisions introduced through separate appends or existing catalogs.
* Report both conflicting names with remediation guidance.
* Preserve stored evidence, existing output destinations, exact-key reuse, and distinct Unicode names.
* Document the naming constraint and direct inspection of existing long-table evidence.
* Close catalog connections when view registration fails.

Related to #133, which addressed measurement keys shadowing reserved episode columns. This extends protection to collisions between measurement keys.

## Validation

The public-API regression failed on unchanged `9c9fd21`: querying values `0.1` and `0.9` returned `(0.1, 0.1)` and incorrectly selected the episode. With this fix, the ambiguous keys are refused.

* Focused tests: **197 passed**
* Full suite: **2,182 passed, 9 skipped**
* Ruff lint, format checks, `ty check`, and `git diff --check`: passed
* Markdown link check: blocked because `lychee` is not installed

closes #558


kstonekuan (changes requested):
The ASCII-only fold is the right call and I checked it against DuckDB directly: `/Camera/score` and `/camera/score` are refused at CREATE, `/Ä/score` and `/ä/score` stay distinct, and a quoted lowercase reference does resolve against a mixed-case column. `str.lower()` would have refused a pair DuckDB keeps apart, and your test pins that, so widening the fold reddens it. Closing the connection when registration raises is a good catch on the way past.

Three of the four guards hold under mutation: write-time reddens 2, curation read reddens 5, the fold reddens 1.

The snapshot guard does not hold on its own. Disabling `snapshot.py:371` alone leaves the suite green, disabling `curation.py:434` alone also leaves `test_case_collisions_refuse_snapshot_without_replacing_existing_output` green, and only turning off both reddens it. So the two read-side layers cover each other and nothing pins either one specifically.

The docs you added state them as two separate layers ("when opening a catalog's wide view and before pivoting snapshot samples"). Either pin the snapshot one with a case the curation guard cannot reach, or drop it and let curation own the read side. An unpinned layer is the one that quietly stops working.

2191 passed on the merge with main.

kstonekuan (approved):
Dropping it was the right of the two options, and updating CATALOG.md to stop claiming two layers is the part that keeps the doc honest.

Every remaining guard is now pinned: write-time reddens 2, curation read reddens 6, the ASCII fold reddens 1. The curation mutant went from 5 to 6 because the snapshot test now depends on that guard rather than on a second one nothing was checking.

2206 passed on the merge with main.

PR #564, opened 2026-09-18
Fix/curate dry run sql gate

### LINKED ISSUE #563: [Bug]: curate dry-run / output=None executes tenant SQL without reject_non_single_select
### Version or commit

5b720ec

### Environment

Linux x86_64, Python via uv; library + CLI repro

### Minimal reproduction

On `main` @ `5b720ec`, the single-SELECT gate only runs when writing a manifest.

1. Manifest path is gated — `_stage_manifest_and_count` calls `reject_non_single_select(sql)` before any DuckDB execute (`src/hflow/curation.py` ~705–709). The pinned surface in `tests/test_catalog_curation.py` (`_SQL_SURFACE_PAYLOADS`) covers multi-statement, DDL-only, PIVOT, PRAGMA/DESCRIBE/SHOW/SUMMARIZE, and the copy-escape shape.

2. Dry-run / report-only path is **not** gated:

```python
# src/hflow/curation.py ~899–902
case None:
    (row_count,) = connection.execute(f"SELECT count(*) FROM ({sql})").fetchone() or (
        0,
    )
```

No `reject_non_single_select` before that execute.

3. CLI dry-run uses that path and never calls the gate itself:

```python
# src/hflow/cli.py ~1489–1492
report = curate(
    arguments.catalog,
    sql,
    output=None if arguments.dry_run else arguments.output,
)
```

4. Contrast: the hosted server always rejects first, then calls `curate(..., output=None)` (`packages/hflow-server/src/hflow_server/_curation.py` ~367–391). Library/CLI dry-run skip that boundary.

5. Same injection SQL that manifest-write refuses (`SELECT 1; CREATE TABLE pwned AS SELECT 1 AS x`, or the `_SQL_SURFACE_PAYLOADS` copy-escape row that closes the wrapper paren early) is executed on the dry-run branch because the gate never runs. Dry-run wraps SQL as `SELECT count(*) FROM ({sql})`, so copy-escape shapes that close that paren are especially relevant.

Not covered by the current open board: #558, #545, #528, #498, #491, #481 (open PR #542 is the redundant-parse cleanup, not this missing gate), #376, #287, #246. Open PRs #562 / #542 / #517 / #380 / #142 do not close this seam.

### Expected behavior

Every `curate()` path that executes tenant SQL — including `output=None` / `hflow curate --dry-run` — should run `reject_non_single_select` (or share the same helper) **before** `connection.execute` / `connection.sql`, matching the manifest-write path and the server boundary.

Refused shapes from `_SQL_SURFACE_PAYLOADS` must raise `NonSingleSelectQueryError` / `ValueError` on dry-run the same way they do on manifest write, with no side-effect tables/files.

### Actual behavior

`reject_non_single_select` is only invoked from `_stage_manifest_and_count`. Report-only curation skips it. `HOSTING.md` / `CATALOG.md` tell operators that `curate(..., constrained=True)` is the tenant SQL posture; that locks file access but does **not** apply the statement-count gate on the dry-run branch.

### Definition of done

- Call `reject_non_single_select(sql)` before the `case None:` count query in `curate()` (preferred: one place so every destination path stays consistent).
- Extend the SQL-surface tests (or a sibling) so `curate(..., output=None)` / CLI `--dry-run` refuse the same `_SQL_SURFACE_PAYLOADS` injection rows as `_stage_manifest_and_count`.
- Optional docs: note in HOSTING/CATALOG that the statement gate applies to report-only curate too, not only manifest materialization.

### Non-goals

- Redesigning #481 / #542 (redundant parse on the already-gated path).
- Changing DuckDB constrained allowlists.
- New curation features.

VARUN3WARE:
## Summary

Apply the single-SELECT statement gate to report-only curation (`curate(..., output=None)` / `hflow curate --dry-run`) so it matches manifest write and the hosted server boundary. Closes #563.

## Why

`_stage_manifest_and_count` already called `reject_non_single_select` before executing tenant SQL. The dry-run branch ran `SELECT count(*) FROM ({sql})` with no gate. Multi-statement injection often becomes a parser error inside that wrapper, but DESCRIBE / SHOW / SUMMARIZE / PIVOT still returned a row count — shapes the manifest path refuses.

Two commits:
1. Share `_reject_tenant_sql` for both paths; note the gate in CATALOG/HOSTING.
2. Pin `_SQL_SURFACE_PAYLOADS` on `output=None` and assert CLI dry-run refuses DESCRIBE (exit 2).

## Validation

```bash
uv run pytest -q \
  tests/test_catalog_curation.py::test_curate_output_none_sql_surface_matches_manifest_gate \
  tests/test_catalog_curation.py::test_cli_curate_dry_run_refuses_non_single_select \
  tests/test_catalog_curation.py::test_curate_sql_surface_is_pinned \
  tests/test_catalog_curation.py::test_cli_curate_dry_run_reports_without_writing_a_manifest

# Dry-run now refuses what manifest write refuses
uv run python -c "
from pathlib import Path
import tempfile, tests.test_catalog_curation as t
from hflow.catalog import Catalog
from hflow.curation import curate
tmp = Path(tempfile.mkdtemp()); c = tmp/'catalog'
Catalog(c).append_episode(canonical_path=t._fake_canonical(tmp), stamps=t.FAKE_STAMPS, episode_metadata={}, check_rows=[t._check_row()])
try:
    curate(c, 'DESCRIBE SELECT episode_id FROM episodes', output=None)
except Exception as e:
    print(type(e).__name__, e)
"
```

Pre-commit ruff clean; `uv run ty check src/hflow/curation.py` clean.

## Checklist


kstonekuan (approved):
One helper on both call sites is the right shape, and folding the rationale into `_reject_tenant_sql` keeps it next to the thing it protects.

Confirmed end to end. Running the pinned refusal payloads down both paths on your branch, all six now raise `NonSingleSelectQueryError` on dry run as well as on manifest write. Before this, `PIVOT`, `DESCRIBE` and `SUMMARIZE` executed, and the multi-statement and DDL shapes were only tripping over the `SELECT count(*) FROM (...)` wrapper and surfacing a raw `ParserException`. Same set, same error type, both paths now.

Both gates hold independently under mutation: removing the dry-run call reddens 9, removing the manifest call reddens 8. 2194 passed on the merge with main.

PR #574, opened 2026-09-19
fix: publish scratch artifacts to content-addressed storage

### LINKED ISSUE #545: [Bug]: camera_video publishes scratch-bound artifact URIs as durable, causing silent byte-replacement or hard deletion on sync rewrite
### Version or commit

Current main

### Environment

### Minimal reproduction

Using a local-root App with `camera_video` registered:
1. Run the full pipeline for generation N. The `video:` artifact URI is published pointing inside `scratch/` (e.g., `.../scratch/wrist_cam_compressed.mp4`).
2. Bump the pipeline version and run a `{SYNC, META}` reprocess (generation N+1). The canonical is rewritten, and the scratch directory is rmtree'd and rebuilt.
3. Query `measurements_latest` for the generation N episode's `video:` artifact URI.

### Expected behavior

The durable catalog URI for a published artifact must point to a stable, content-bound location (like `run_dir/media/` or `run_dir/artifacts/`). If the underlying bytes change generations, the URI must either be invalidated, or the digest must change, refusing to serve unverified bytes.

### Actual behavior

The publish loop (`app.py:2686-2701`) accepts the scratch-cached remux as a durable artifact. 
- On a `{SYNC, META}` reprocess, the scratch directory is wiped and rebuilt. The generation N artifact URI now silently resolves to generation N+1's bytes. The server's `is_uri_servable` check still passes (the file "exists"), so a downstream reviewer gets the wrong pixels with zero errors.
- On a forced sync-only reprocess (where the labels/media stages don't run), the generation N artifact URI resolves to a missing file (404), even though the catalog claims the artifact exists.

Executed proof:
[N] episode_id=0c12703341c7e653
[N] artifact URI: .../scratch/wrist_cam_compressed.mp4 (exists: True, sha=6818cf269ee5)
[N+1] (pipeline bump) artifact URI now resolves to: sha=3d86896cddb6 (identical: False)
[N+2] (sync-only rewrite) artifact file exists: False

Note: The built-in contact sheet, stored under `run_dir/media/`, survives the `shutil.rmtree(scratch_dir)` correctly. Only the enrichment publish path fails this boundary.

### Additional context

Root cause: The publish loop binds a durable catalog URI to a path inside the scratch directory's lifecycle, and scratch has no content identity of its own. This violates the contract in `steps.py:405` which explicitly documents `ep.workdir` as scratch. Additionally, `test_camera_video.py:66` has a flawed assertion (`is_relative_to(data_root)`) that allowed this to slip through because `scratch/` is technically inside the data root.

Why it matters: The published `video:` artifact is exactly what a browser scrubber or human reviewer opens to trust a curation cut. After any reprocess, the reviewer's video may be a different generation's pixels (silently accepted) or missing. The evidence chain is broken.

Fix direction: In the publish loop, refuse to record artifact URIs under the scratch subtree. Stage such artifacts into `run_dir/artifacts/` (like `media/`) and publish from there. Additionally, bind every published artifact to its own digest so a replaced file is detectable by a reader.

Definition of done:
- Regression test: after a `{SYNC,META}` reprocess, the generation-N artifact URI hashes to the exact bytes episode N recorded, or is explicitly refused.
- Mutation proof (a): drop the `shutil.rmtree(scratch_dir)` guard → test must fail on the deleted-file branch.
- Mutation proof (b): skip staging into `artifacts/` → test must fail on the silent-replacement branch.
- `test_camera_video.py:66` corrected to assert the URI is *not* under `scratch/`.

harshitethic:
Fixes #545.

## Summary
- detect enrichment artifacts that still live under the run scratch directory;
- publish those artifacts under durable `artifacts/` storage instead of exposing scratch URIs;
- include the artifact SHA-256 in the published key so changed bytes get a different durable URI;
- tighten camera-video regression coverage to reject scratch-bound URIs.

## Why
Scratch is rewritten across reprocessing. A catalog URI that points into scratch can therefore start resolving to different bytes or disappear entirely. Content-addressed durable keys keep prior catalog URIs bound to the bytes that were published.

## Testing
- updated camera-video path assertions to require durable, content-addressed artifact paths;
- full repository CI can exercise the integration suite.


kstonekuan (changes requested):
The fix itself is right, and both halves hold under mutation: forcing `artifact_is_scratch_bound = False` reddens 1, and dropping `artifact_content_digest` from the key reddens 1. `publish` does copy when source and destination differ, so the artifact really does leave scratch rather than just getting a nicer-looking URI.

Two things before this can land.

`uv run ruff format --check` fails on `src/hflow/app.py`, so this would turn main red:

```
-   artifact_name_digest = hashlib.sha256(artifact_name.encode()).hexdigest()[:8]
+   artifact_name_digest = hashlib.sha256(artifact_name.encode()).hexdigest()[
+       :8
+   ]
```

And `step_directory` and `artifact_name_digest` are now computed twice, identically, in the two branches. Only the key layout differs between them. Hoist those two out above the `if` so the next person changing the naming scheme cannot change one copy and miss the other.

Worth a sentence in the code on why the non-scratch fallback stays non-content-addressed, since a reader hitting the second branch will wonder whether that is deliberate.

kstonekuan (changes requested):
`src/hflow/app.py` was truncated by the last push. On `c28600a` it is 72 lines and does not parse:

```
IndentationError: unexpected indent
  line 1: _execute_enrichment(media_step, canonical_episode, quarantine_skip)
```

The file starts mid-function, so the module docstring, every import, and roughly 3,000 lines above that point are gone. That is what the 7 ruff errors and 81 ty diagnostics are; nothing here is about your actual change.

Sizes across the branch, so you can see where it went:

```
8d6a134  3158 lines
064f69d  3158 lines
c28600a    72 lines   <- the review-feedback commit
main     3131 lines
```

`064f69d` is intact, so `git checkout 064f69d -- src/hflow/app.py` gets the file back, then re-apply the two review points on top: hoist `step_directory` and `artifact_name_digest` above the `if`, and the line-length fix `ruff format` wants.

The change itself was already right last round. Both halves held under mutation and `publish` does copy when source and destination differ, so nothing about the approach needs revisiting.

Main has moved since you branched, so merge it in as well once the file is back.

PR #572, opened 2026-09-19
feat(build-ai): migrate evaluation CLI from argparse to Typer (#566)

shobhitagnihotri69:
Closes #566

### Summary
Migrates the command-line interface in `examples/build_ai_evaluation/evaluate.py` from `argparse` to typed `Typer` commands (`run` and `compare`).

### Changes
- **Dependencies** (`examples/build_ai_evaluation/pyproject.toml` & `uv.lock`):
  - Added `typer>=0.27.2` to direct dependencies and updated the lockfile.
- **CLI Implementation** (`examples/build_ai_evaluation/evaluate.py`):
  - Replaced `_argument_parser()` and `_run_configuration_from_arguments()` with `build_evaluation_configuration(...)` and typed command handlers (`run` and `compare`).
  - Preserved existing options, enum choices (`DatasetVariant`, `SourceSelection`, `EvaluationTask`, `ResponseFormat`), repeated `--source` and `--task` options, environment defaults (`OPENAI_MODEL`, `OPENAI_BASE_URL`), positive integer bounds (`min=1`), and output locations.
  - Retained `main(argv=None)` entrypoint preserving exit code `2` on user or runtime errors without leaking tracebacks.
- **Tests** (`examples/build_ai_evaluation/tests/test_evaluation.py`):
  - Adapted CLI tests to invoke the Typer app through `CliRunner`.
  - Added tests for positive integer bounds rejection, repeated option preservation, and `--help` output for top-level and subcommands.

### Validation
- `uv sync --locked --project examples/build_ai_evaluation`
- `uv run --locked --project examples/build_ai_evaluation ruff check --fix examples/build_ai_evaluation`
- `uv run --locked --project examples/build_ai_evaluation ruff format examples/build_ai_evaluation`
- `uv run --locked --project examples/build_ai_evaluation ty check --project examples/build_ai_evaluation --extra-search-path . examples/build_ai_evaluation`
- `uv run --locked --project examples/build_ai_evaluation pytest -q examples/build_ai_evaluation/tests` (44 passed)
- `uv run --locked --project examples/build_ai_evaluation python examples/build_ai_evaluation/evaluate.py --help`
- `uv run --locked --project examples/build_ai_evaluation python examples/build_ai_evaluation/evaluate.py run --help`
- `uv run --locked --project examples/build_ai_evaluation python examples/build_ai_evaluation/evaluate.py compare --help`

PR #577, opened 2026-09-19
refactor(examples): migrate the EgoSuite evaluation CLI to Typer

### LINKED ISSUE #567: Migrate the EgoSuite evaluation CLI to Typer
[examples/egosuite_evaluation/evaluate.py:1654](https://github.com/Hebbian-Robotics/hflow/blob/5b720ec033c194e865a5db2a8bcf41bf1073069a/examples/egosuite_evaluation/evaluate.py#L1654) defines `labels`, `run`, and `compare` with argparse, then converts Namespace attributes into evaluation configuration. Migrate this CLI to typed Typer commands so argument types are checked through to configuration construction.

Scope: `examples/egosuite_evaluation/evaluate.py`, its CLI coverage, and its workspace dependencies. Keep the existing evaluation/configuration functions as the business-logic boundary. Add Typer directly to `examples/egosuite_evaluation/pyproject.toml` and update the shared `uv.lock`; this can land independently of the main CLI migration.

Acceptance criteria:

- Replace argparse and Namespace-based CLI adapters with typed Typer command functions for `labels`, `run`, and `compare`.
- Keep existing invocation syntax, flags, defaults, output locations, and failure exit codes. Preserve camera/response-format choices, environment defaults, numeric bounds, multiple input paths, and mutual exclusion between `--limit-per-episode` and `--samples-per-episode`. Keep the shared input/projection options consistent across `labels` and `run`.
- Preserve evaluation results and report formats. Keep this PR scoped to `evaluate.py`; the example’s pipeline CLI is separate.
- Adapt existing tests under `examples/egosuite_evaluation/tests` to the new command boundary. Retain outcome-focused configuration/report coverage and extend it only where a compatibility boundary is uncovered. No live model requests are needed to verify argument parsing.
- Ensure top-level and subcommand `--help` work. Update example usage documentation if help presentation needs explanation.

Validation, following [CONTRIBUTING.md](https://github.com/Hebbian-Robotics/hflow/blob/5b720ec033c194e865a5db2a8bcf41bf1073069a/CONTRIBUTING.md):

```bash
uv sync --locked --project examples/egosuite_evaluation
uv run --locked --project examples/egosuite_evaluation ruff check --fix examples/egosuite_evaluation
uv run --locked --project examples/egosuite_evaluation ruff format examples/egosuite_evaluation
uv run --locked --project examples/egosuite_evaluation ty check --project examples/egosuite_evaluation --extra-search-path . examples/egosuite_evaluation
uv run --locked --project examples/egosuite_evaluation python examples/egosuite_evaluation/evaluate.py --help
```

ayam04:
## What changed

Migrates the EgoSuite evaluation CLI (`examples/egosuite_evaluation`) from argparse to typed Typer commands (fixes #567).

- `labels`, `run` and `compare` are now typed Typer functions: enums for the camera view and the response format, path types for inputs and outputs, option callbacks for the integer bounds
- shared option aliases so `labels` and `run` cannot drift apart on flag naming
- the argparse `Namespace` adapter that turned raw attribute bags into evaluation configuration is gone

## Compatibility

Invocation syntax, flags, defaults, environment defaults, output locations and failure exit codes are unchanged:

- argument errors and the errors the evaluation already raised still print one message without a traceback and exit 2
- the mutual exclusion between `--limit-per-episode` and `--samples-per-episode` now fails the same way from the command body

## Validation

- `uv sync --locked`
- `ruff check` + `ruff format` + `ty check` clean
- `pytest -q examples/egosuite_evaluation/tests` - 59 passed
- top-level, `labels` and `run` `--help` render, and a missing input exits 2 with one message

Closes #567


kstonekuan (approved):
Shared option aliases so `labels` and `run` cannot drift is the part that makes this worth doing rather than a like-for-like port.

The surface is exactly preserved. Grepping the source under-reports here because the flags are no longer literals, so I asked the built Click app instead: 19 accepted long options against argparse's 19, nothing missing, nothing added, all three commands present.

Guards hold. Neutering `_positive_integer` reddens 1, and disabling the mutual-exclusion check you added at the CLI boundary reddens 1. The older copy inside frame selection is not pinned on its own, but it predates this PR and is reasonable depth behind the boundary check, so nothing to do there.

The colour trap from #572 does not reach you: your help assertions are on command names, which survive Rich's styling, where an option's `--` prefix does not. Confirmed by running the suite under `FORCE_COLOR=1` as well as plain, green both ways.

Validated on the merge with main: root gate clean, 2224 passed, and the example's own lint, format, typecheck and tests green on 3.11 and 3.14.

Issue #580, opened 2026-09-21
Episode.video() materializes the full camera stream multiple times before ffmpeg remux


Kaileshwar16:
`Episode.video()` appears to try to keep the video path memory-efficient by decoding `CompressedVideo` messages lazily:

```python
def validated_access_units():
    # Stream-decode instead of channel.messages: caching a decoded
    # copy of every video payload would double the episode's memory.
    for message in channel.iter_decoded():
        ...
        yield message.data
```

However, the downstream `write_access_units_to_mp4()` immediately materializes the complete H.264 stream:

```python
annex_b_stream = b"".join(units)
```

and later passes the resulting full buffer to ffmpeg:

```python
subprocess.run(command, input=annex_b_stream, capture_output=True)
```

There is another full-channel materialization before this. `Episode.video()` calls:

```python
channel = self.channel(topic)
```

and `Episode.channel()` collects every raw message payload into:

```python
raw: list[bytes] = []

for batch in self._reader.iter_batches(...):
    ...
    raw.extend(batch.data)
```

So for a large camera channel, `video()` currently holds the complete encoded MCAP channel in `ChannelData` and then builds another contiguous copy of the complete H.264 elementary stream before ffmpeg can start consuming it.

The `iter_decoded()` comment suggests the intent is specifically to avoid retaining an additional full-video copy, but the `b"".join(units)` downstream defeats that property.

This makes peak RAM scale with the total encoded video size rather than with the reader batch size.

For short test episodes this is mostly invisible, but multi-GB / long-running robotics camera episodes can require several GB of additional memory just to perform a lossless remux. Multiple Airflow workers or multiple cameras make the amplification more significant and could cause worker OOMs.

HFlow already exposes `Episode.iter_decoded_batches()` as the bounded-memory alternative to complete `channel()` materialization, so the video path currently bypasses an existing architectural seam intended for this kind of workload.

### Relevant code

`src/hflow/episode.py`

* `Episode.channel()` accumulates the entire channel in `raw`
* `Episode.video()` calls `self.channel(topic)`
* `validated_access_units()` then yields decoded H.264 payloads lazily

`src/hflow/video.py`

```python
annex_b_stream = b"".join(units)
coding_types = scan_picture_coding_types(annex_b_stream)
...
completed = subprocess.run(
    command,
    input=annex_b_stream,
    capture_output=True,
)
```

### Suggested regression / benchmark

[...]

PR #581, opened 2026-09-21
fix(video): stream canonical remux with bounded memory

Kaileshwar16:
Canonical video remuxing previously retained the entire camera channel and joined all
  H.264 payloads before invoking ffmpeg, creating multiple full-video allocations.

  Use two batch-reading passes: estimate FPS with an exact disk-backed median, then stream
  decoded access units into ffmpeg stdin. Preserve H.264/B-frame validation and atomic
  output publication, with subprocess and temporary-file cleanup on failures.

  Add regressions for incremental consumption, remuxing without channel materialization,
  validation failures, broken pipes, and preservation of existing outputs.

  Validation:
  - 66 directly affected tests passed on the current branch.
  - Broader tests stopped during collection because the environment lacks `httpx`.
  - Before rebasing onto latest upstream, the full suite passed: 2,206 passed, 9 skipped;
  Ruff and type checks also passed.
  - Streaming 16 MiB and 512 MiB produced identical traced Python memory peaks of 1,053,657
  bytes using a subprocess sink.

  The change requires two reads and temporary disk storage. Memory measurements cover HFlow
  allocations, not ffmpeg’s internal muxer metadata.


kstonekuan (approved):
Both halves of #580, including the cache-hit one, and `iter_decoded_batches` is the right accessor to land on since it is already the documented bounded-memory path.

The SQLite median is the part I wanted to be sure about, because a remux silently changing its declared frame rate is the kind of drift nobody notices for months. Ran it differentially against `estimate_fps_from_log_times` across both median parities, uniform and jittered streams, streams with a long stall, and the degenerate inputs: exact agreement on all ten, and all three error contracts match word for word.

Guards hold. Removing the cache-hit early return reddens 1, dropping the h264 check reddens 1, and replacing the median with the first interval reddens 2.

Guarding the early return on `topic in self._video_fps` as well as `output.exists()` is the detail that makes it correct rather than just fast, since `_video_fps` is a side effect of the old ordering.

2300 passed on the merge with main.

PR #582, opened 2026-09-21
fix(video): stream canonical remux with bounded memory (#580)

### LINKED ISSUE #580: Episode.video() materializes the full camera stream multiple times before ffmpeg remux

shobhitagnihotri69:
## Summary
Fixes #580. Eliminates full-video memory materialization in `Episode.video()` by streaming canonical H.264 access units directly into `ffmpeg` stdin with bounded memory.

## Key Changes
- **Lazy Batch Streaming**: Replaced `self.channel(topic)` and `b"".join(units)` with `self.iter_decoded_batches()`.
- **Bounded Exact Median FPS**: Implemented `estimate_fps_from_streaming_log_times` backed by a temporary SQLite table to preserve the exact delta median without buffering millions of timestamps in memory.
- **Incremental B-Frame Validation**: Introduced `_PictureCodingScanner` to scan slice headers access-unit by access-unit during streaming.
- **Robust Process Management**: Stderr redirected to a temporary file to avoid pipe deadlocks; guaranteed subprocess termination and cleanup on early exits or broken pipes.
- **Comprehensive Regressions**: Added `tests/test_video_streaming.py` validating incremental consumption, process cleanup, and error precedence.

## Verification
- `uv run pytest tests/test_video.py tests/test_video_streaming.py`: 52 passed in 11.11s.
- `uv run ruff check src tests`: All checks passed.


kstonekuan:
Closing on first-PR-wins: #581 opened 44 minutes earlier and is in review.

For the record, this branch's tree is byte-identical to #581's across all four files, so there is nothing here that merging #581 would not already deliver. If that is not what you expected, say so and I will take another look.

Open good first issues: https://github.com/Hebbian-Robotics/hflow/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

